PR #12, opened 2026-04-26
[Gautam's Dev bot] Standalone MockProviderHost + Claude Agent SDK env-var override (issue #11 Phase 2)

### LINKED ISSUE #11: Standalone mock provider HTTP service wrapping ScriptedSseResponder for OpenAI + Anthropic E2E
## Summary

Wrap the existing instruction-chain message handlers
(`TestSseMessageHandler`, `AnthropicTestSseMessageHandler`,
`ScriptedSseResponder`) into a standalone, runnable HTTP service that
speaks the OpenAI and Anthropic on-the-wire conventions, so external CLI
tools — Claude Agent SDK, GitHub Copilot CLI, OpenAI Codex CLI — can be
pointed at it (`OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, etc.) and exercised
end-to-end against deterministic, scripted responses.

This is a thin transport wrapper. **No new response logic.** The service
must delegate every request to the existing in-process handlers / the
instruction-chain parser already used by the unit tests.

Follow-up to #7 — that issue wired the same handlers into the sample app
via `WebApplicationFactory<Program>` for in-process xUnit tests. This issue
takes the next step: a real listening port reachable from external
processes.

## Value

- **Unblocks E2E coverage of external agent CLIs** (Claude SDK, Copilot
  CLI, Codex) against the same scripted scenarios already used in unit
  tests — today these tools cannot be redirected through an in-process
  `HttpMessageHandler`.
- **Single source of truth for mock behavior.** Reusing
  `ScriptedSseResponder` / `InstructionChainParser` keeps unit, in-proc
  E2E (#7), and external-CLI E2E coverage all driven by the same scripts.
- **CI-stable.** Deterministic, offline, no upstream provider rate
  limits or flakiness.

## Proposed Approach

1. **New project:** `samples/MockProviderHost/` (or
   `tools/MockProviderHost/`) — an ASP.NET Core minimal-API host on
   Kestrel, runnable as `dotnet run -- --port 5099 --script <path>`.

2. **Endpoints (initial scope):**
   - `POST /v1/chat/completions` — streaming + non-streaming. Delegates
     to `TestSseMessageHandler` / `ScriptedSseResponder`.
   - `POST /v1/messages` — streaming + non-streaming. Delegates to
     `AnthropicTestSseMessageHandler` / `AnthropicSseStreamHttpContent`.
   - `GET /healthz` — readiness probe for CI orchestration.

3. **Wrapper-only constraint.** The endpoints reconstruct the inbound
   `HttpRequestMessage` and hand it to the existing handlers; the SSE
   bytes the handlers produce are streamed back as the response body
   unchanged. No bespoke response synthesis in the host.

4. **Scripting model.** Reuse the existing `ScriptedSseResponder` /
   `InstructionChainParser` script format. The script source is
   configured at launch (`--script ./scenario.yaml` or env var) and
   loaded once on startup. Optional: per-request `X-Mock-Scenario` header
   to select among multiple loaded scripts for parallel-test isolation.

5. **Auth passthrough.** Accept any `Authorization: Bearer …` /
   `x-api-key` header and ignore — but echo back the conventional
   provider headers (`openai-version`, `anthropic-version`, request id)
   so client SDKs that assert on them do not break.

## Acceptance Criteria

- [ ] New host project builds as part of `LmDotnetTools.sln` and runs
      with `dotnet run --project samples/MockProviderHost -- --port N`.
- [ ] OpenAI chat/completions endpoint streams an existing scripted
      scenario byte-identically to `TestSseMessageHandler` output (test
      asserts equality of SSE frames).
- [ ] Anthropic messages endpoint streams `message_start` …
      `content_block_delta` (`thinking_delta`, `input_json_delta`,
      `text_delta`) … `message_stop` byte-identically to
      `AnthropicTestSseMessageHandler` output.
- [ ] One smoke test points the OpenAI .NET SDK (or LmCore's OpenAI
      provider) at `http://localhost:N/v1` via `OPENAI_BASE_URL` and
      consumes a multi-turn scripted conversation end-to-end.
- [ ] One smoke test points an Anthropic-shaped client at the host and
      consumes a streaming conversation with thinking + tool-use blocks.
- [ ] No duplication of response-generation logic — host code only does
      transport adaptation; all SSE bytes originate from the existing
      handlers.
- [ ] README under the new project documenting how to run and how to
      author scripts (link to existing instruction-chain docs).

## Out of Scope

- Embeddings, rerank, OpenAI `/v1/responses`, OpenAI image, OpenAI
  audio — file as follow-ups if needed.
- Admin API for hot-swapping scripts at runtime.
- Docker image — can be a follow-up once endpoints stabilize.
- Recording / replay against real providers.

## Priority

P2 — planned, not blocking current work.

## Related

- #7 — In-process scripted SSE harness via `WebApplicationFactory`
  (shared handlers / scripts).
- #8 — Playwright UI E2E (parallel test surface).

gautam-achieveai:
[Gautam's Dev bot]

Closes #11.

## Summary

Phase 2 of issue #11: a standalone, in-process **mock provider HTTP service** that wraps `ScriptedSseResponder` behind OpenAI- and Anthropic-compatible endpoints, plus an additive extension to `ClaudeAgentSdkOptions` so the spawned Claude Agent SDK CLI can be redirected at the host via env vars.

### `samples/MockProviderHost/` — minimal-API ASP.NET Core host
- `POST /v1/chat/completions` → forwards to ScriptedSseResponder's OpenAI handler.
- `POST /v1/messages` → forwards to ScriptedSseResponder's Anthropic handler.
- `GET /healthz` → liveness probe.
- **Wrapper-only invariant:** every body byte streamed back originates from the inner `ScriptedSseResponder`. The host does no SSE re-framing — it `CopyToAsync`s the upstream body into the response stream untouched. Status code, content-type, and SSE `data:` frames are all sourced from the inner handler. The host adds only conventional response *headers* (e.g. `openai-version`, `anthropic-version`) which are out-of-band of the streamed body.

### `tests/MockProviderHost.Tests/` — in-process coverage (9 tests)
- **ByteIdentityTests** — `TeeHandler` captures the inner handler's exact body bytes and asserts they equal what the host serves.
- **HealthAndRoutingTests** — health endpoint, route mapping, header pass-through (`Authorization` + `x-api-key` simultaneously).
- **OpenAiProviderConsumptionTests** — LmCore `OpenClient` streams scripted text through the host on a real ephemeral port.
- **AnthropicProviderConsumptionTests** — LmCore `AnthropicClient` streams scripted text through the host on a real ephemeral port.

### `tests/MockProviderHost.E2E.Tests/` — real-CLI scaffolding (1 test, skipped by default)
- `RealPortHostFixture` boots the host on `127.0.0.1:0` and exposes the bound URL for external processes.
- `ClaudeCliPrerequisites.Detect()` probes for Node.js + the `claude` CLI on PATH and standard install locations; tests skip with a clear reason when missing.
- `ClaudeAgentSdkAgainstMockTests` is **double-gated**: requires both `LMDOTNET_RUN_CLAUDE_E2E=1` env var and CLI prerequisites. Default CI runs see it as `Skipped`. The CLI's protocol-level handshake is the subject of follow-up work.

### `src/ClaudeAgentSdkProvider/` — additive options + env var mapping
- `ClaudeAgentSdkOptions` adds three optional properties: `BaseUrl`, `AuthToken`, `DisableExperimentalBetas`.
- `ClaudeAgentSdkClient.ApplyMockHostOverrides(IDictionary<string, string?>, ClaudeAgentSdkOptions)` maps them onto `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, and `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` on the spawned CLI process. Helper is `internal static` and unit-tested directly without spawning a process.
- All-default options leave the spawned environment untouched — this change is strictly additive for existing callers.

## Test results

| Project | Result |
|---|---|
| `MockProviderHost.Tests` | 9 passed |
| `ClaudeAgentSdkProvider.Tests` | 25 passed (incl. 3 new `MockHostOverrides` tests) |
| `MockProviderHost.E2E.Tests` | 1 skipped (gated by `LMDOTNET_RUN_CLAUDE_E2E=1`) |

## Self-review

- **Iteration 1** flagged one defensive-correctness issue: `ClaudeCliPrerequisites.TryRun` called `StandardOutput.ReadToEnd()` *before* `WaitForExit(5000)`, which could deadlock the test host if a probed CLI binary hung without ever closing stdout. Fixed in commit 2: `WaitForExit` first, kill the process tree on timeout, then read stdout.
- **Iteration 2** confirmed the fix and verified the wrapper-only byte-identity invariant, additive-only behavior, default-skipped E2E gating, and resource cleanup. No blocking issues found.

## Follow-ups (separate issues)

The plan's four follow-up issues will be filed against this PR's outcome:
1. Codex E2E using the same mock host + scripted scenarios.
2. GitHub Copilot SDK provider investigation against the mock host.
3. JSON scenario loader for `ScriptedSseResponder`.
4. Header-based scenario routing (`X-Mock-Scenario`) for multi-tenant E2E suites.

## Test plan

- [x] `dotnet test tests/MockProviderHost.Tests/MockProviderHost.Tests.csproj` — green.
- [x] `dotnet test tests/MockProviderHost.E2E.Tests/MockProviderHost.E2E.Tests.csproj` — 1 skipped (expected).
- [x] `dotnet test tests/ClaudeAgentSdkProvider.Tests/ClaudeAgentSdkProvider.Tests.csproj` — green (25/25, including new env-var helper unit tests).
- [x] Local self-review iteration 1 (with fix) and iteration 2 (clean).


gautam-achieveai (commented):
## **[Gautam's review bot]** PR #12 — Deep Review Summary

### Context
**User Story #11**: Standalone mock provider HTTP service wrapping `ScriptedSseResponder` for OpenAI + Anthropic E2E.

Phase 2 adds a standalone `MockProviderHost` (ASP.NET Core minimal API) wrapping `ScriptedSseResponder` behind `/v1/chat/completions` and `/v1/messages`, plus env-var overrides for `ClaudeAgentSdkClient` to redirect the Claude CLI at the mock host.

### Review Agents Dispatched
| Agent | Key Finding |
|---|---|
| Architecture Review | Clean dependency flow; `samples/` as test infra is consistent with existing convention |
| Duplicate Code Detector | `RealPortHostFixture` ≈ `MockProviderHostFixture` (~37 lines verbatim duplication) |
| Exception Handling Review | 4 MEDIUM issues — bare catches, missing logging in `ForwardAsync`, broad OCE catch |
| Performance Review | Unbounded request body buffering (low risk for test infra) |
| Test Coverage Review | Missing error-path and non-streaming-path tests |

### Build & Compilation
✅ Build succeeds with **0 warnings, 0 errors** (after NuGet restore for 3 new projects).

### Strengths
- **Byte-identity test pattern is excellent** — `TeeHandler` + `ByteIdentityTests` prove the wrapper-only invariant by capturing inner handler bytes and comparing against wire bytes.
- **Provider consumption tests** use real `OpenClient`/`AnthropicClient` through real TCP on ephemeral ports — genuine end-to-end validation.
- **`ApplyMockHostOverrides` tests** are thorough with full branch coverage (all-set, all-default, blank-string).
- **Self-review iteration** correctly fixed a real deadlock risk in `TryRun` (commit 2).
- **E2E test is properly double-gated** (`LMDOTNET_RUN_CLAUDE_E2E=1` + CLI prerequisites) — CI without the SDK installed passes cleanly.
- **Architecture is clean**: dependency flow is `tests/` → `samples/MockProviderHost` → `src/LmTestUtils`, production code has no back-dependency on the mock host.

### Findings Summary

| # | Severity | Finding | File |
|---|---|---|---|
| 1 | MEDIUM | Duplicate test fixtures — `RealPortHostFixture` is strict subset of `MockProviderHostFixture` | `RealPortHostFixture.cs` |
| 2 | MEDIUM | `ForwardAsync` has no error logging — failures invisible in test output | `MockProviderHostBuilder.cs` |
| 3 | MEDIUM | `catch (OperationCanceledException)` doesn't filter by CTS | `ClaudeAgentSdkAgainstMockTests.cs` |
| 4 | MEDIUM | `TryRun` bare `catch` swallows all exceptions without logging | `ClaudeCliPrerequisites.cs` |
| 5 | MEDIUM | No tests for error status code propagation through the host | `MockProviderHost.Tests/` |

See inline comments for details on each finding.

### Verdict
**No blocking issues.** The code is well-structured, the test design is strong, and the architecture is clean. The 5 MEDIUM findings are all improvement opportunities — not correctness bugs. Recommend addressing items 1–4 before merge for maintainability; item 5 (error-path tests) can be a follow-up.

tests/MockProviderHost.E2E.Tests/Infrastructure/RealPortHostFixture.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.AspNetCore.Builder;
+
+namespace AchieveAi.LmDotnetTools.MockProviderHost.E2E.Tests.Infrastructure;
+
+/// <summary>
+/// Boots <see cref="MockProviderHostBuilder"/> on a real ephemeral 127.0.0.1 port and exposes
+/// the bound URL. External processes (the Claude Agent SDK CLI) need a real TCP socket; an
+/// in-process <c>HttpMessageHandler</c> seam is not enough.
+/// </summary>
+internal sealed class RealPortHostFixture : IAsyncDisposable
```
gautam-achieveai: **[Gautam's review bot]** MEDIUM — **Duplicate test fixture**

`RealPortHostFixture` is a strict subset of `MockProviderHostFixture` — every line here has a verbatim counterpart in the unit-test fixture. The only difference is `MockProviderHostFixture` adds a second `StartAsync` overload taking raw `HttpMessageHandler`s.

| Aspect | `MockProviderHostFixture` | `RealPortHostFixture` | Match? |
|---|---|---|---|
| Class shape | `sealed : IAsyncDisposable` | `sealed : IAsyncDisposable` | ✅ |
| Field | `WebApplication _app` | `WebApplication _app` | ✅ |
| Private ctor | `(WebApplication, string)` | `(WebApplication, string)` | ✅ |
| `StartAsync(responder)` | ✅ | ✅ | ✅ verbatim |
| URL extraction | `app.Urls.FirstOrDefault() ?? throw …` | Same | ✅ verbatim |
| Error message | `"Mock host failed to bind to a URL."` | Same | ✅ verbatim |
| `DisposeAsync` | `StopAsync + DisposeAsync` | Same | ✅ verbatim |

Both test projects already reference `samples/MockProviderHost/`. Consider consolidating into a single `EphemeralHostFixture` in the `MockProviderHost` project (which already has `InternalsVisibleTo` for both test assemblies), and deleting both per-project copies.

samples/MockProviderHost/MockProviderHostBuilder.cs
```diff
@@ -0,0 +1,162 @@
...
+        app.MapPost("/v1/chat/completions", ctx =>
+            ForwardAsync(ctx, openAiClient, "/v1/chat/completions", openAiHeaders: true));
+
+        app.MapPost("/v1/messages", ctx =>
+            ForwardAsync(ctx, anthropicClient, "/v1/messages", openAiHeaders: false));
+
+        return app;
+    }
+
+    private static async Task ForwardAsync(
```
gautam-achieveai, line 104: **[Gautam's review bot]** MEDIUM — **ForwardAsync has no error logging**

`ForwardAsync` is the endpoint handler for both `/v1/chat/completions` and `/v1/messages`. Multiple operations can throw (`CopyToAsync` → `OperationCanceledException`, `SendAsync` → `HttpRequestException`/`TaskCanceledException`, stream copy → `IOException`). When any throw, ASP.NET Core converts the unhandled exception into a generic 500, but with the host's log level at `Warning` by default and no exception-handler middleware, these failures are invisible in test output.

For a mock host used in E2E tests where the external CLI is a black box, diagnosing "why did the test fail?" often means looking at the mock host's logs. A 500 with no logged cause is a debugging dead end.

**Suggested fix** — lightweight log-and-rethrow:
```csharp
try
{
    // ... existing body ...
}
catch (OperationCanceledException) when (ctx.RequestAborted.IsCancellationRequested)
{
    // Client disconnected — not an error.
}
catch (Exception ex)
{
    var logger = ctx.RequestServices.GetService<ILoggerFactory>()
        ?.CreateLogger("MockProviderHost.Forward");
    logger?.LogError(ex, "ForwardAsync failed for {Path}", upstreamPath);
    throw; // let ASP.NET Core generate the 500
}
```

tests/MockProviderHost.E2E.Tests/Scenarios/ClaudeAgentSdkAgainstMockTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            await client.StartAsync(request, cts.Token);
+            await foreach (var _ in client.SendMessagesAsync([], cts.Token))
+            {
+                if (responder.RemainingTurns["parent"] == 0)
+                {
+                    break;
+                }
+            }
+        }
+        catch (OperationCanceledException)
```
gautam-achieveai: **[Gautam's review bot]** MEDIUM — **Broad `catch (OperationCanceledException)` doesn't filter by CTS**

This catches `OperationCanceledException` (and its subclass `TaskCanceledException`) without inspecting which cancellation token fired. The 20-second timeout CTS is the intended source, but `client.StartAsync` could cancel internally (CLI fails to start), and `SendMessagesAsync` could throw `TaskCanceledException` from an HTTP timeout — neither should be silently swallowed.

If `StartAsync` throws OCE *before any request reaches the mock host*, the catch swallows it and the assertion at line 82 catches it with a confusing message ("expected 0 but found 1") instead of the real stack trace.

**Suggested fix:**
```csharp
catch (OperationCanceledException) when (cts.IsCancellationRequested)
{
    // 20-second timeout expired — acceptable if the CLI reached the host at least once.
}
```
This way unexpected internal cancellations propagate as proper test failures with stack traces.

tests/MockProviderHost.E2E.Tests/Infrastructure/ClaudeCliPrerequisites.cs
```diff
@@ -0,0 +1,86 @@
...
+            // child, which a hung CLI may never do.
+            if (!p.WaitForExit(5000))
+            {
+                try { p.Kill(entireProcessTree: true); } catch { /* best-effort */ }
+                return false;
+            }
+            output = p.StandardOutput.ReadToEnd();
+            return p.ExitCode == 0;
+        }
+        catch
```
gautam-achieveai: **[Gautam's review bot]** MEDIUM — **Bare `catch` swallows all exceptions without logging**

This bare `catch` swallows every exception type — including `OutOfMemoryException` and other critical CLR failures. More importantly, there's **zero diagnostic signal**: when troubleshooting why a prerequisite check returned `false` on a machine where the tool *is* installed (permission issue, PATH misconfiguration), there's nothing in the logs.

The `/* best-effort */` catch on line 75 for `Kill()` is fine (race condition with process exit), but this outer catch should at minimum:
1. Use `catch (Exception ex)` instead of bare `catch`
2. Add a trace-level log

**Suggested fix** — if you want a never-throw probe:
```csharp
catch (Exception ex)
{
    System.Diagnostics.Debug.WriteLine(
        $"Prerequisite probe failed for {fileName}: {ex.GetType().Name}: {ex.Message}");
    return false;
}
```

Or catch only the expected types:
```csharp
catch (SystemException ex) when (ex is System.ComponentModel.Win32Exception
                                     or InvalidOperationException
                                     or PlatformNotSupportedException)
{
    return false;
}
```

tests/MockProviderHost.Tests/ByteIdentityTests.cs
```diff
@@ -0,0 +1,82 @@
...
+          "stream": true,
+          "max_tokens": 1024,
+          "system": "You are a helpful assistant.",
+          "messages": [
+            {"role": "user", "content": "say hello"}
+          ]
+        }
+        """;
+
+    [Fact]
```
gautam-achieveai, line 40: **[Gautam's review bot]** MEDIUM — **Missing error-path tests for ForwardAsync**

The test suite covers happy paths thoroughly (streaming success, byte identity, headers, routing). However, there are no tests verifying that `ForwardAsync` faithfully propagates error responses from the inner handler.

Two specific gaps:

1. **Error status code propagation** — When `ScriptedSseResponder` returns a non-200 (e.g., 400 for empty body), does the host propagate the status code and error body? The code at `MockProviderHostBuilder.cs:131` copies it, but no test locks this behavior.

2. **Non-streaming request rejection** — `ScriptedSseResponder` rejects `stream:false` with HTTP 400. The code comment at `MockProviderHostBuilder.cs:134` references "application/json for non-streaming errors" but no test exercises this path.

These could be a follow-up, but given this is a transport wrapper whose core contract is "every byte passes through unchanged," testing the error path would strengthen confidence in that invariant.

*Example test:*
```csharp
[Fact]
public async Task OpenAi_endpoint_returns_400_for_non_streaming_request()
{
    var responder = ScriptedSseResponder.New()
        .ForRole("parent", _ => true).Turn(t => t.Text("unused"))
        .Build();
    await using var fixture = await MockProviderHostFixture.StartAsync(responder);
    using var client = new HttpClient { BaseAddress = new Uri(fixture.BaseUrl) };
    var body = """{"model":"gpt-test","stream":false,"messages":[{"role":"user","content":"hi"}]}""";
    using var content = new StringContent(body, Encoding.UTF8, "application/json");
    using var response = await client.PostAsync("/v1/chat/completions", content);
    response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
}
```

tests/MockProviderHost.E2E.Tests/Infrastructure/RealPortHostFixture.cs
gautam-achieveai: **[Gautam's Dev bot]**

Done. Both fixtures consolidated into a single `internal sealed class EphemeralHostFixture` in `samples/MockProviderHost/` (InternalsVisibleTo wires both test projects). Both copies deleted; all 11 unit tests + the E2E scenario use the new fixture. Fixed in 3993699dc095f67700aebe953b5ceea2d7b9b058.

samples/MockProviderHost/MockProviderHostBuilder.cs
gautam-achieveai: **[Gautam's Dev bot]**

Done. `ForwardAsync` now wraps its body in try/catch:
- `OperationCanceledException when ctx.RequestAborted.IsCancellationRequested` — client disconnect, rethrow without logging
- General `Exception` — `ILoggerFactory.CreateLogger("MockProviderHost.Forward").LogError(ex, …)` then rethrow so ASP.NET Core still emits the 500.

Fixed in 3993699dc095f67700aebe953b5ceea2d7b9b058.

tests/MockProviderHost.E2E.Tests/Scenarios/ClaudeAgentSdkAgainstMockTests.cs
gautam-achieveai: **[Gautam's Dev bot]**

Done. Catch now reads `catch (OperationCanceledException) when (cts.IsCancellationRequested)`, so internal cancellations from `StartAsync`/`SendMessagesAsync` propagate as proper test failures with stack traces. Fixed in 3993699dc095f67700aebe953b5ceea2d7b9b058.

tests/MockProviderHost.E2E.Tests/Infrastructure/ClaudeCliPrerequisites.cs
gautam-achieveai: **[Gautam's Dev bot]**

Done. Replaced the bare catch with a typed filter (`Win32Exception`, `InvalidOperationException`, `PlatformNotSupportedException`, `IOException`) and emit a `Debug.WriteLine` with the exception type/message so a missing-tool diagnosis isn't silent on machines where the CLI is supposedly installed. Fixed in 3993699dc095f67700aebe953b5ceea2d7b9b058.

tests/MockProviderHost.Tests/ByteIdentityTests.cs
gautam-achieveai: **[Gautam's Dev bot]**

Done. Added two tests in `ByteIdentityTests`:
- `OpenAi_endpoint_propagates_400_for_non_streaming_request`
- `Anthropic_endpoint_propagates_400_for_non_streaming_request`

Both send `stream:false`, assert `StatusCode == 400`, and verify the wire bytes match the inner handler's `TeeHandler` capture byte-for-byte (the wrapper-only invariant still holds on the error path). Fixed in 3993699dc095f67700aebe953b5ceea2d7b9b058.

gautam-achieveai:
**[Gautam's Dev bot]**

Addressed all 5 MEDIUM findings from the deep-review summary in 3993699dc095f67700aebe953b5ceea2d7b9b058, plus merged `origin/main` to resolve the `LmDotnetTools.sln` conflict (combined both branches' new projects).

| # | Finding | Resolution |
|---|---|---|
| 1 | Duplicate fixture | Consolidated into `samples/MockProviderHost/EphemeralHostFixture.cs`; both copies deleted |
| 2 | `ForwardAsync` no error logging | try/catch with `ILoggerFactory.LogError` + rethrow; client-disconnect (`RequestAborted`) bypasses logging |
| 3 | Broad OCE catch | Now `when (cts.IsCancellationRequested)`; internal cancellations propagate as test failures |
| 4 | Bare catch in `TryRun` | Typed filter (`Win32Exception`/`InvalidOperationException`/`PlatformNotSupportedException`/`IOException`) + `Debug.WriteLine` trace |
| 5 | Missing error-path tests | Added 2 tests covering OpenAI + Anthropic 400 propagation with byte-identity assertion |

**Build & tests:** 0 errors, MockProviderHost.Tests 11/11 passing (was 9, +2 new), ClaudeAgentSdkProvider MockHostOverrides 3/3 passing, E2E test correctly skips without env gate.

gautam-achieveai (commented):
## **[Gautam's review bot]** PR #12 — Incremental Review (Iterations 3-4)

### Context
**User Story #11**: Standalone mock provider HTTP service wrapping `ScriptedSseResponder` for OpenAI + Anthropic E2E.

Reviewing changes since commit `65f3e77` (last reviewed) through `3993699` (PR head). This iteration addresses all 5 MEDIUM findings from the initial deep review, plus merges `origin/main`.

### Previous Findings Verification

All 5 MEDIUM findings properly addressed:

| # | Finding | Status | Assessment |
|---|---|---|---|
| 1 | Duplicate fixture | ✅ Fixed | Consolidated into `EphemeralHostFixture` in `samples/MockProviderHost/`. Both copies deleted. Clean factoring with `StartCoreAsync` helper and two overloads (responder vs. raw handlers). `InternalsVisibleTo` wires both test projects. |
| 2 | ForwardAsync no error logging | ✅ Fixed | try/catch with `ILoggerFactory.LogError` + rethrow. Client disconnect bypasses logging correctly via `ctx.RequestAborted.IsCancellationRequested`. |
| 3 | Broad OCE catch | ✅ Fixed | `when (cts.IsCancellationRequested)` filter added — internal cancellations from `StartAsync`/`SendMessagesAsync` now propagate as proper test failures. |
| 4 | Bare catch in TryRun | ✅ Fixed | Typed filter (`Win32Exception`, `InvalidOperationException`, `PlatformNotSupportedException`, `IOException`) + `Debug.WriteLine` trace. |
| 5 | Missing error-path tests | ✅ Fixed | 2 new tests (`OpenAi_endpoint_propagates_400_for_non_streaming_request`, `Anthropic_endpoint_propagates_400_for_non_streaming_request`) covering status code + byte-identity assertion on the error path. |

### New Finding (LOW)

**Unused using directives** — The fixture refactoring left behind `using AchieveAi.LmDotnetTools.MockProviderHost.Tests.Infrastructure` in 3 files that no longer reference types from that namespace (`HealthAndRoutingTests.cs`, `AnthropicProviderConsumptionTests.cs`, `OpenAiProviderConsumptionTests.cs`), and `using Microsoft.AspNetCore.Builder` in `EphemeralHostFixture.cs` (covered by SDK implicit usings). Build emits IDE0005 warnings. Non-blocking — clean up at convenience.

### Test Results
✅ **MockProviderHost.Tests: 11/11 passed** (was 9, +2 new error-path tests)

### Verdict
**APPROVE.** All previous MEDIUM findings cleanly addressed. No new blocking or MEDIUM-severity issues introduced. The only new finding is LOW (unused usings from the refactoring). Code quality is solid, the fixture consolidation is well-executed, and the error handling improvements make the mock host significantly more debuggable for E2E scenarios.

gautam-achieveai:
[Gautam's Dev bot] Addressed the LOW finding from review 4176593029 in c005757:

- Removed `using AchieveAi.LmDotnetTools.MockProviderHost.Tests.Infrastructure;` from `HealthAndRoutingTests.cs`, `AnthropicProviderConsumptionTests.cs`, and `OpenAiProviderConsumptionTests.cs` — none reference `TeeHandler` after the fixture consolidation. `ByteIdentityTests.cs` still uses `TeeHandler`, so the using stays there.
- Removed `using Microsoft.AspNetCore.Builder;` from `samples/MockProviderHost/EphemeralHostFixture.cs` — covered by the Web SDK's implicit usings.

Build: 0 errors, no IDE0005 in changed files. Tests: 11/11 MockProviderHost.Tests pass.

samples/MockProviderHost/MockProviderHostBuilder.cs
```diff
@@ -0,0 +1,179 @@
...
+                ctx.Response.Headers["anthropic-version"] = "2023-06-01";
+                ctx.Response.Headers["anthropic-request-id"] = requestId;
+                ctx.Response.Headers["request-id"] = requestId;
+            }
+
+            // Copy the (possibly streaming) response body straight through. CopyToAsync flushes
+            // periodically, which is sufficient for SSE since each `data:` frame from the inner
+            // SseStreamHttpContent crosses a write boundary.
+            await using var upstreamStream = await upstreamResponse.Content.ReadAsStreamAsync(ctx.RequestAborted).ConfigureAwait(false);
+            await upstreamStream.CopyToAsync(ctx.Response.Body, ctx.RequestAborted).ConfigureAwait(false);
```
gautam-achieveai, line 162: **[Gautam's review bot]** MEDIUM — **`CopyToAsync` does not flush SSE frames to the network**

`Stream.CopyToAsync` calls `WriteAsync` in a loop but **never** calls `FlushAsync()`. Kestrel's `HttpResponsePipeWriter`-backed stream accumulates writes in its output buffer and flushes only when: (a) `FlushAsync()` is called explicitly, (b) the ~64KB internal buffer threshold is exceeded, or (c) the response completes.

For SSE, this means individual `data:` frames (~50–500 bytes each) will **not** be flushed to the TCP client as they arrive. They accumulate and are delivered in a batch when the stream completes. The comment at line 158 states "CopyToAsync flushes periodically" — this is incorrect.

Current tests pass because they consume the full response after stream completion. But for the host's stated purpose (E2E testing of external CLIs expecting real-time SSE), frames would arrive batched/delayed.

**Suggested fix** — manual copy loop with per-frame flush:
```csharp
await using var upstreamStream = await upstreamResponse.Content
    .ReadAsStreamAsync(ctx.RequestAborted).ConfigureAwait(false);
var buffer = new byte[4096];
int bytesRead;
while ((bytesRead = await upstreamStream.ReadAsync(
    buffer, ctx.RequestAborted).ConfigureAwait(false)) > 0)
{
    await ctx.Response.Body.WriteAsync(
        buffer.AsMemory(0, bytesRead), ctx.RequestAborted).ConfigureAwait(false);
    await ctx.Response.Body.FlushAsync(ctx.RequestAborted).ConfigureAwait(false);
}
```

samples/MockProviderHost/MockProviderHostBuilder.cs
```diff
@@ -0,0 +1,179 @@
...
+        HttpContext ctx,
+        HttpClient client,
+        string upstreamPath,
+        bool openAiHeaders)
+    {
+        try
+        {
+            // Buffer the body up-front so the inner ScriptedHandler can read it via
+            // ReadAsStringAsync without racing the original request stream's cancellation.
+            ctx.Request.EnableBuffering();
```
gautam-achieveai, line 114: **[Gautam's review bot]** LOW — **Redundant `EnableBuffering()` call**

`EnableBuffering()` wraps the body in a `FileBufferingReadStream` to allow re-reading. But the body is immediately copied to a `MemoryStream` (line 115-117) and never re-read or seeked — `EnableBuffering` adds an unnecessary buffer layer for zero benefit. Safe to remove this line.

gautam-achieveai (commented):
**[Gautam's review bot]** Re-review (iteration 5) — **APPROVE** ✅

All 5 MEDIUM and 1 LOW findings from the previous review have been thoroughly addressed. The incremental diff is clean — 4 unused `using` removals. Build passes with 0 errors, all 14 tests pass (11 MockProviderHost + 3 MockHostOverrides).

---

### 🔍 Deep Review Summary (6 agents dispatched)

| Agent | Result |
|---|---|
| Architecture | ✅ Clean — acyclic dependencies, no SOLID violations |
| Exception Handling | ✅ Clean — all patterns correct, proper re-throw, good filters |
| Performance | 1 MEDIUM (SSE flush), 1 LOW (EnableBuffering) |
| Test Coverage | 14 tests, 1 minor gap (500 exception path) |
| Design Simplification | Well-designed, 2 INFO-level cleanups |
| Review Grader | APPROVE WITH COMMENTS |

---

### New Finding — MEDIUM: `CopyToAsync` doesn't flush SSE frames (non-blocking)

**File:** `MockProviderHostBuilder.cs:158-162`

`Stream.CopyToAsync` never calls `FlushAsync()`. Kestrel's output pipe writer buffers writes (~64KB threshold) and only flushes when the stream completes or the buffer fills. This means SSE frames from the inner `ScriptedSseResponder` accumulate and are delivered to the client as a batch when the response completes — not in real-time as they're produced.

The comment at line 158 states *"CopyToAsync flushes periodically"* — this is incorrect. `CopyToAsync` calls `WriteAsync` in a loop but never `FlushAsync`.

**Current tests pass** because they consume full responses (`ReadAsByteArrayAsync`, `await foreach` to completion). But for the stated E2E use case — external CLIs expecting real-time streaming SSE — frames would arrive batched/delayed.

**Suggested follow-up fix** — manual copy loop with per-frame flush:
```csharp
await using var upstreamStream = await upstreamResponse.Content
    .ReadAsStreamAsync(ctx.RequestAborted).ConfigureAwait(false);
var buffer = new byte[4096];
int bytesRead;
while ((bytesRead = await upstreamStream.ReadAsync(
    buffer, ctx.RequestAborted).ConfigureAwait(false)) > 0)
{
    await ctx.Response.Body.WriteAsync(
        buffer.AsMemory(0, bytesRead), ctx.RequestAborted).ConfigureAwait(false);
    await ctx.Response.Body.FlushAsync(ctx.RequestAborted).ConfigureAwait(false);
}
```

---

### Lower-priority observations (all non-blocking)

| Sev | Finding | File |
|---|---|---|
| LOW | `EnableBuffering()` is redundant — body is never re-read after `MemoryStream` copy. Safe to remove. | `MockProviderHostBuilder.cs:114` |
| LOW | No test for the 500 exception-propagation path (the catch-log-rethrow added in review 1). Consider a `FaultingHandler` test in a follow-up. | `MockProviderHostBuilder.cs:169-177` |
| LOW | `MockProviderHost` has `InternalsVisibleTo` + `internal` classes — atypical for `samples/`. Consider moving to `src/` alongside `LmTestUtils` if more test hosts are added. | `MockProviderHost.csproj` |
| INFO | `TryRun` has an unused `out string output` parameter — both callers pass `out _`. | `ClaudeCliPrerequisites.cs:51` |

---

### Work Item Context
- **User Story #11**: Standalone mock provider HTTP service wrapping ScriptedSseResponder for OpenAI + Anthropic E2E

**Overall: Well-implemented feature with responsive iteration on review feedback. The SSE flush issue is the only remaining concern — recommend addressing in a follow-up before wiring up real external CLI E2E tests.**

PR #4, opened 2026-04-14
feat: Sub-agent orchestration via tool calls

gautam-achieveai:
## Summary

- Adds sub-agent orchestration enabling parent `MultiTurnAgentLoop` to spawn, monitor, and resume child agents through LLM tool calls (`Agent` / `CheckAgent`)
- Sub-agents inherit filtered parent tools with recursive delegation support, concurrency gating via `SemaphoreSlim`, and conversation persistence for agent resurrection
- Includes `SubAgentManager` (lifecycle orchestration), `SubAgentToolProvider` (`IFunctionProvider` integration), plus 5 supporting types and 9 new tests (7 unit + 2 integration)

### New files (`src/LmMultiTurn/SubAgents/`)
- **SubAgentTemplate.cs** — Named template configuration (system prompt, agent factory, tool filters, max turns)
- **SubAgentOptions.cs** — Top-level config: template dictionary, max concurrency, default store factory
- **SubAgentState.cs** — Runtime state: status enum, turn summary record, per-agent tracking
- **SubAgentManager.cs** — Core orchestrator: spawn, resume, peek, monitor, dispose
- **SubAgentToolProvider.cs** — `IFunctionProvider` exposing `Agent` and `CheckAgent` tools

### Modified
- **MultiTurnAgentLoop.cs** — New `subAgentOptions` constructor parameter; snapshots tools before registering sub-agent provider; disposes manager on shutdown

### Tests
- **SubAgentManagerTests.cs** — 7 unit tests covering spawn, concurrency limits, peek, completion relay, disposal
- **SubAgentIntegrationTests.cs** — 2 integration tests: full parent→sub-agent→parent flow, tool registration verification

## Test plan
- [x] All 9 new tests pass
- [x] Full solution builds with 0 errors
- [x] 137/138 existing tests pass (1 pre-existing flaky `CodexAgentLoopTests` unrelated to this PR)
- [ ] Verify CI passes on GitHub

Fixes #3

Issue #3, opened 2026-04-14
Sub-Agent Orchestration: Enable LLM to spawn and monitor child agents via tool calls


gautam-achieveai:
## Summary

Add sub-agent orchestration to `MultiTurnAgentLoop`, allowing the primary LLM to delegate tasks to independent child agents via tool calls. Sub-agents run their own `MultiTurnAgentLoop` with independent configuration (system prompt, tools, model) and communicate results back to the parent through a direct reference.

This follows a similar pattern to Claude Code's Task tool for sub-agent spawning, but with **bidirectional communication** — the parent can peek into sub-agent progress, and the sub-agent pushes completion messages back to the parent.

## Value

- **Developers** building complex AI workflows get first-class multi-agent orchestration
- Enables **parallel task delegation** — an LLM can farm out research, analysis, or sub-tasks while continuing its main work
- **Context isolation** — each sub-agent has its own conversation history, avoiding token bloat in the parent
- **Observability** — the parent can monitor sub-agent progress without waiting for completion

#### 1. `SubAgentManager`
Manages the lifecycle of active sub-agents within a parent `MultiTurnAgentLoop`.
- Tracks active sub-agents in a `ConcurrentDictionary<string, SubAgentEntry>`
- Enforces configurable concurrency limit (default: 5)
- Handles cleanup when parent stops
- Buffers recent turns per sub-agent for the peek mechanism

#### 2. `spawn_subagent` Tool
Registered in the parent's `FunctionRegistry`. Called by the LLM to create a sub-agent.
- **Parameters:**
  - `agentName` — selects a pre-defined named sub-agent template (system prompt + default tools)
  - `task` — user prompt / task description for the sub-agent
  - `addTools` _(optional)_ — additional tools to grant beyond the template defaults
  - `removeTools` _(optional)_ — tools to exclude from the template defaults
- **Returns:** `{ agentId, status: "running" }` — non-blocking
- **Behavior:** Creates a new `MultiTurnAgentLoop` with the template's system prompt, resolved tool set, and the task as initial input. Passes a parent reference (`IMultiTurnAgent`) so the sub-agent can send completion messages.

#### 3. `peek_subagent` Tool
Registered in the parent's `FunctionRegistry`. Called by the LLM to check sub-agent progress.
- **Parameters:** `agentId`
- **Returns:** Compact summary:
  - Sub-agent status (`running`, `completed`, `failed`)
[...]

PR #27, opened 2026-05-03
Feat/deferred tool execution

rinz13r:
## Problem

`MultiTurnAgentLoop` executes tool calls synchronously: the run loop awaits every handler via `Task.WhenAll(pendingToolCalls.Values)` before starting the next turn. This works for fast tools (filesystem reads, in-process computation, quick HTTP calls) but breaks down whenever a tool's result depends on something the loop can't drive forward on its own.

Concrete scenarios that don't fit the current model:

- **Human-in-the-loop approval.** LLM asks to read a directory; a human must approve before it proceeds. Approval may take minutes to hours.
- **External async APIs with callbacks.** A bank verification API accepts a request and fires a webhook 10 minutes later. Nothing useful to await locally.
- **Long-running batch jobs.** Submit work to a queue or scheduler; result comes back via a separate channel.
- **Sub-agents with their own tool loops.** Already long-lived; modeling them as deferred calls is a natural fit.

In all of these the handler can't reasonably block. Today, doing so would:

- Pin a turn (and the entire run loop) for the full duration.
- Prevent the loop from draining new inputs from `SendAsync` until the slow tool returns (`TryDrainInputs` runs at turn boundaries).
- Provide no cooperative cancellation — handler signatures are `Func<string, Task<string>>` with no `CancellationToken`.
- Hide partial progress from the UI and the LLM.

## The generalizing insight

Approval is not a special case. The unifying primitive is **deferred tool execution**: a handler that returns a placeholder synchronously and is resolved later by *something* outside the loop. The loop should not care whether that "something" is a human, a webhook, a worker, or a timer.

Specializations collapse into thin integration layers:

| Use case | What "resolves" the call |
|---|---|
| HITL approval | UI calls `ResolveToolCallAsync` when the human acts |
| Bank webhook | Webhook receiver calls `ResolveToolCallAsync` |
| Long batch job | Worker calls `ResolveToolCallAsync` on completion |
| Timeout/cancellation | Host timer calls `ResolveToolCallAsync` with an error |

## Why a placeholder is required, not optional

Both Anthropic and OpenAI enforce that every `tool_use` (or `tool_call`) ID in an assistant message must have a matching `tool_result` before the next assistant inference. There is no "partially answered tool calls" state in either contract.

So while the loop is waiting on a deferred resolution, it must still place *some* `tool_result` in history for protocol validity. A placeholder string (`"PENDING ..."`) plus a flag like `IsDeferred = true` on the message satisfies both the contract and the loop's own state-tracking needs.

When resolution arrives, the placeholder is replaced in-place by the real result (mutating history by `ToolCallId`). Appending a second `tool_result` for the same ID is not a defined operation and would produce an invalid request payload.

## Design implications worth noting

These are flagged here as motivation context; the actual design lives in `design.md` (TBD).

- **History as the single source of truth.** Pending state lives on `ToolCallResultMessage` (e.g., `IsDeferred`), not in a parallel registry. Persistence via `IConversationStore` is then automatic.
- **Resolution is one API.** `ResolveToolCallAsync(toolCallId, result, isError)` — used identically by every integrator. Approvals, webhooks, and timeouts all funnel through it.
- **Idempotency matters once async is in scope.** Webhooks retry; resolution must be safe against double-delivery in a way approvals never required.
- **Restart durability is the hardest open question.** When the process dies mid-deferral, the in-flight outbound work the handler initiated (HTTP request, queue publish) is *not* persisted. Handlers must be responsible for their own outbound idempotency / recovery; the loop should expose enough metadata (deferred call inspection on startup) to support that, but cannot solve it generically.
- **Progress updates use the existing subscriber stream.** Tool results are write-once; partial progress flows as separate published events to UI subscribers without resolving the placeholder.

## Out of scope (for now)

- Cross-process resolution and durable workflow integration (Temporal, Durable Functions). The first cut targets in-process async; durable hand-off is an explicit future concern.
- Built-in timeout policy. Timeouts are host-driven (host calls `ResolveToolCallAsync` with an error after its own deadline) rather than enforced by the loop. Helpers may follow.
- Streaming partial tool results into LLM context. The LLM only sees a deferred call's result on final resolution.

### DIFF
--- a/src/LmCore/Messages/ToolCallResultBuilder.cs
+++ b/src/LmCore/Messages/ToolCallResultBuilder.cs
@@ -1,0 +1,60 @@
+namespace AchieveAi.LmDotnetTools.LmCore.Messages;
+
+/// <summary>
+/// Single mapping point from a handler-side <see cref="ToolHandlerResult"/> to the wire-shape
+/// <see cref="ToolCallResult"/>. Stamps in framework-controlled fields (tool call id, tool
+/// name, deferral marker, timestamps) that handlers should not — and now cannot — set
+/// themselves.
+/// </summary>
+/// <remarks>
+/// Both <c>FunctionCallMiddleware</c> and <c>MultiTurnAgentLoop</c> route through here so
+/// the two paths can't drift. Handlers signal "deferred" with an empty
+/// <see cref="ToolHandlerResult.Deferred"/> record; the builder produces a
+/// <see cref="ToolCallResult"/> with empty <c>Result</c> text and <c>IsDeferred=true</c> —
+/// host loops are responsible for refusing to send any provider request while such a
+/// result is unresolved in history (see <c>MultiTurnAgentLoop.ExecuteTurnAsync</c>).
+/// </remarks>
+public static class ToolCallResultBuilder
+{
+    /// <summary>
+    /// Maps a handler return into a wire-shape <see cref="ToolCallResult"/>.
+    /// </summary>
+    /// <param name="result">The handler's return value.</param>
+    /// <param name="toolCallId">Tool call id from <c>ToolCallContext.ToolCallId</c> (or the
+    /// originating <c>ToolCallMessage</c>). May be null on call paths that don't carry one.</param>
+    /// <param name="toolName">Function name. Optional but recommended — used by providers
+    /// that surface <c>tool_name</c> separately from <c>tool_call_id</c>.</param>
+    /// <param name="executionTarget">Distinguishes local function tools from provider-executed
+    /// server tools. Defaults to <see cref="ExecutionTarget.LocalFunction"/>.</param>
+    public static ToolCallResult FromHandlerResult(
+        ToolHandlerResult result,
+        string? toolCallId,
+        string? toolName = null,
+        ExecutionTarget executionTarget = ExecutionTarget.LocalFunction)
+    {
+        return result switch
+        {
+            ToolHandlerResult.Resolved r => new ToolCallResult(
+                toolCallId,
+                r.Payload.Text,
+                r.Payload.ContentBlocks,
+                executionTarget)
+            {
+                ToolName = toolName,
+                IsError = r.Payload.IsError,
+                ErrorCode = r.Payload.ErrorCode,
+            },
+            ToolHandlerResult.Deferred => new ToolCallResult(
+                toolCallId,
+                string.Empty,
+                executionTarget)
+            {
+                ToolName = toolName,
+                IsDeferred = true,
+                DeferredAt = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds(),
+            },
+            _ => throw new InvalidOperationException(
+                $"Unknown ToolHandlerResult variant '{result.GetType().Name}'."),
+        };
+    }
+}
--- a/src/LmCore/Middleware/LegacyHandlerAdapter.cs
+++ b/src/LmCore/Middleware/LegacyHandlerAdapter.cs
@@ -1,0 +1,118 @@
+using AchieveAi.LmDotnetTools.LmCore.Messages;
+
+namespace AchieveAi.LmDotnetTools.LmCore.Middleware;
+
+/// <summary>
+/// Adapters that bridge the new <see cref="ToolHandlerResult"/> handler shape to/from the
+/// legacy <c>Func&lt;string, Task&lt;string&gt;&gt;</c> and
+/// <c>Func&lt;string, Task&lt;ToolCallResult&gt;&gt;</c> shapes used by external integrations
+/// (MCP server adapters, external SDK tool bridges) that cannot support deferred tool execution.
+/// </summary>
+/// <remarks>
+/// Deferred tool execution requires an out-of-band resolution channel (see
+/// <c>MultiTurnAgentLoop.ResolveToolCallAsync</c>). Components that lack one — external SDK
+/// bridges, <c>FunctionCallMiddleware</c>-driven flows — should adapt new-shape handlers via
+/// these helpers. Deferred returns surface as <see cref="NotSupportedException"/> at invocation
+/// time.
+/// </remarks>
+public static class LegacyHandlerAdapter
+{
+    /// <summary>
+    /// Wraps a single new-shape <see cref="ToolHandlerResult"/> handler into the legacy
+    /// <c>Func&lt;string, Task&lt;string&gt;&gt;</c> shape. Throws
+    /// <see cref="NotSupportedException"/> at invocation time on
+    /// <see cref="ToolHandlerResult.Deferred"/>.
+    /// </summary>
+    public static Func<string, Task<string>> ToLegacyHandler(
+        Func<string, Task<ToolHandlerResult>> handler,
+        string toolKey = "<unknown>")
+    {
+        ArgumentNullException.ThrowIfNull(handler);
+        return async args =>
+        {
+            var result = await handler(args);
+            return result switch
+            {
+                ToolHandlerResult.Resolved r => r.Result.Result,
+                ToolHandlerResult.Deferred => throw new NotSupportedException(
+                    $"Tool '{toolKey}' returned a deferred result. Deferred tool execution is "
+                    + "only supported when handlers are dispatched by MultiTurnAgentLoop."
+                ),
+                _ => throw new InvalidOperationException(
+                    $"Unknown ToolHandlerResult variant '{result.GetType().Name}' for tool '{toolKey}'."
+                ),
+            };
+        };
+    }
+
+    /// <summary>
+    /// Wraps a handler dictionary of the new <see cref="ToolHandlerResult"/> shape into the
+    /// legacy <c>Func&lt;string, Task&lt;string&gt;&gt;</c> shape. Each wrapped handler unwraps
+    /// <see cref="ToolHandlerResult.Resolved"/> and throws <see cref="NotSupportedException"/>
+    /// on <see cref="ToolHandlerResult.Deferred"/>.
+    /// </summary>
+    public static IDictionary<string, Func<string, Task<string>>> WrapToLegacyHandlers(
+        IDictionary<string, Func<string, Task<ToolHandlerResult>>> source,
+        IEqualityComparer<string>? keyComparer = null)
+    {
+        ArgumentNullException.ThrowIfNull(source);
+        var wrapped = new Dictionary<string, Func<string, Task<string>>>(
+            source.Count,
+            keyComparer ?? StringComparer.Ordinal);
+
+        foreach (var kvp in source)
+        {
+            wrapped[kvp.Key] = ToLegacyHandler(kvp.Value, kvp.Key);
+        }
+
+        return wrapped;
+    }
+
+    /// <summary>
+    /// Wraps a single legacy <c>Func&lt;string, Task&lt;string&gt;&gt;</c> handler into the new
+    /// <see cref="ToolHandlerResult"/> shape. The wrapped handler always returns
+    /// <see cref="ToolHandlerResult.Resolved"/> with a text-only <see cref="ToolCallResult"/>;
+    /// legacy handlers cannot signal deferral.
+    /// </summary>
+    public static Func<string, Task<ToolHandlerResult>> ToNewHandler(
+        Func<string, Task<string>> legacy)
+    {
+        ArgumentNullException.ThrowIfNull(legacy);
+        return async args => new ToolHandlerResult.Resolved(new ToolCallResult(null, await legacy(args)));
+    }
+
+    /// <summary>
+    /// Wraps a single legacy <c>Func&lt;string, Task&lt;ToolCallResult&gt;&gt;</c> handler into
+    /// the new <see cref="ToolHandlerResult"/> shape. Useful for multi-modal handlers where the
+    /// legacy callsite produces a <see cref="ToolCallResult"/> with content blocks; the wrapped
+    /// handler forwards the entire <see cref="ToolCallResult"/> verbatim.
+    /// </summary>
+    public static Func<string, Task<ToolHandlerResult>> ToNewHandler(
+        Func<string, Task<ToolCallResult>> legacy)
+    {
+        ArgumentNullException.ThrowIfNull(legacy);
+        return async args => new ToolHandlerResult.Resolved(await legacy(args));
+    }
+
+    /// <summary>
+    /// Wraps a handler dictionary of the legacy <c>Func&lt;string, Task&lt;string&gt;&gt;</c>
+    /// shape into the new <see cref="ToolHandlerResult"/> shape. Each wrapped handler always
+    /// resolves synchronously with a text-only <see cref="ToolCallResult"/>.
+    /// </summary>
+    public static IDictionary<string, Func<string, Task<ToolHandlerResult>>> WrapToNewHandlers(
+        IDictionary<string, Func<string, Task<string>>> source,
+        IEqualityComparer<string>? keyComparer = null)
+    {
+        ArgumentNullException.ThrowIfNull(source);
+        var wrapped = new Dictionary<string, Func<string, Task<ToolHandlerResult>>>(
+            source.Count,
+            keyComparer ?? StringComparer.Ordinal);
+
+        foreach (var kvp in source)
+        {
+            wrapped[kvp.Key] = ToNewHandler(kvp.Value);
+        }
+
+        return wrapped;
+    }
+}
--- a/src/LmCore/Middleware/LegacyHandlerAdapter.cs
+++ b/src/LmCore/Middleware/LegacyHandlerAdapter.cs
@@ -18,19 +18,20 @@ namespace AchieveAi.LmDotnetTools.LmCore.Middleware;
 public static class LegacyHandlerAdapter
 {
     /// <summary>
-    /// Wraps a single new-shape <see cref="ToolHandlerResult"/> handler into the legacy
-    /// <c>Func&lt;string, Task&lt;string&gt;&gt;</c> shape. Throws
-    /// <see cref="NotSupportedException"/> at invocation time on
-    /// <see cref="ToolHandlerResult.Deferred"/>.
+    /// Wraps a single new-shape <see cref="ToolHandler"/> into the legacy
+    /// <c>Func&lt;string, Task&lt;string&gt;&gt;</c> shape used by external SDKs that
+    /// can't supply a <see cref="ToolCallContext"/>. The wrapper invokes the handler with
+    /// a default <see cref="ToolCallContext"/>. Throws <see cref="NotSupportedException"/>
+    /// at invocation time on <see cref="ToolHandlerResult.Deferred"/>.
     /// </summary>
     public static Func<string, Task<string>> ToLegacyHandler(
-        Func<string, Task<ToolHandlerResult>> handler,
+        ToolHandler handler,
         string toolKey = "<unknown>")
     {
         ArgumentNullException.ThrowIfNull(handler);
         return async args =>
         {
-            var result = await handler(args);
+            var result = await handler(args, new ToolCallContext());
             return result switch
             {
                 ToolHandlerResult.Resolved r => r.Result.Result,
@@ -52,7 +53,7 @@ public static class LegacyHandlerAdapter
     /// on <see cref="ToolHandlerResult.Deferred"/>.
     /// </summary>
     public static IDictionary<string, Func<string, Task<string>>> WrapToLegacyHandlers(
-        IDictionary<string, Func<string, Task<ToolHandlerResult>>> source,
+        IDictionary<string, ToolHandler> source,
         IEqualityComparer<string>? keyComparer = null)
     {
         ArgumentNullException.ThrowIfNull(source);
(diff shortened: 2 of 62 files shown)


src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,905 @@
...
+        string result,
+        bool isError = false,
+        IList<ToolResultContentBlock>? contentBlocks = null,
+        CancellationToken ct = default)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(toolCallId);
+        ArgumentNullException.ThrowIfNull(result);
+
+        return ResolveToolCallInternalAsync(toolCallId, result, isError, contentBlocks, ct);
+    }
```
gautam-achieveai, line 652: **[Gautam's review bot]** Bug: Missing `ObjectDisposedException` guard. Every other public API method checks this. `ResolveToolCallAsync` does not. Add `ObjectDisposedException.ThrowIf(_isDisposed, this);` at the top.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,905 @@
...
+        // empty by design — resumes contribute no new messages, only a wake-up.
+        var sentinel = new ResumeSentinel(runId, generationId);
+        var emptyInput = new UserInput([], InputId: null, ParentRunId: runId);
+        var queuedInput = new QueuedInput(
+            emptyInput,
+            ReceiptId: $"resume:{Guid.NewGuid():N}",
+            QueuedAt: DateTimeOffset.UtcNow,
+            Resume: sentinel);
+
+        _ = WriteResumeSentinelAsync(queuedInput, ct);
```
gautam-achieveai, line 890: **[Gautam's review bot]** Bug: Fire-and-forget can permanently stall auto-resume. If `WriteResumeSentinelAsync` fails, `_resumeScheduled` is already `true` and the deferring markers are cleared. No future resolution can re-trigger. On failure, restore scheduling state so next resolution can retry.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,905 @@
...
+            }
+
+            var ctx = new ToolCallContext
+            {
+                ToolCallId = toolCall.ToolCallId,
+            };
+            var result = await handler(functionArgs, ctx, ct);
+            return BuildResultMessage(toolCall, result);
+        }
+        catch (Exception ex)
```
gautam-achieveai, line 599: **[Gautam's review bot]** `OperationCanceledException` swallowed by broad catch. Add `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` before the general catch to propagate user abort.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,905 @@
...
+                _lastDeferringGenerationId = mostRecentDeferringGen;
+                _resumeScheduled = false;
+            }
+
+            Logger.LogInformation(
+                "Restored {Count} deferred tool call(s) from persisted history",
+                _deferred.Count);
+        }
+
+        return Task.CompletedTask;
```
gautam-achieveai, line 818: **[Gautam's review bot]** No tests for `OnHistoryRestoredAsync`. This 65-line method is the process-restart recovery path. Pre-populate a store with `IsDeferred=true` messages, call `RecoverAsync`, verify `GetDeferredToolCallsAsync` returns entries and `ResolveToolCallAsync` triggers auto-resume.

src/LmMultiTurn/MultiTurnAgentBase.cs
```diff
@@ -247,6 +247,142 @@ protected void RestoreHistory(IReadOnlyList<IMessage> messages)
...
+
+        try
+        {
+            var persisted = MessagePersistenceConverter.ToPersistedMessage(message, ThreadId, runId);
+            await Store.AppendMessagesAsync(ThreadId, [persisted], ct);
+        }
+        catch (Exception ex)
+        {
+            Logger.LogWarning(ex, "Failed to persist deferred-tool placeholder");
+        }
```
gautam-achieveai, line 291: **[Gautam's review bot]** Persistence failure contradicts documented guarantee. Doc says placeholder must be durable before subscribers see it. Exception is swallowed and `PublishToAllAsync` continues. Either propagate error or update doc to say best-effort.

src/LmCore/Middleware/LegacyHandlerAdapter.cs
```diff
@@ -0,0 +1,119 @@
...
+        string toolKey = "<unknown>")
+    {
+        ArgumentNullException.ThrowIfNull(handler);
+        return async args =>
+        {
+            var result = await handler(args, new ToolCallContext(), CancellationToken.None);
+            return result switch
+            {
+                ToolHandlerResult.Resolved r => r.Result.Result,
+                ToolHandlerResult.Deferred => throw new NotSupportedException(
```
gautam-achieveai, line 38: **[Gautam's review bot]** Missing test coverage for Deferred -> `NotSupportedException` path. MCP server adapter calls `ToLegacyHandler` for every tool. This is the only guardrail for deferred handlers exposed via MCP. Add `LegacyHandlerAdapterTests`.

gautam-achieveai (changes requested):
**[Gautam's review bot]** Deep review of PR #27 — Feat/deferred tool execution

## Overview

Well-architected feature. The discriminated union pattern (`ToolHandlerResult`), handler signature unification (`ToolHandler` delegate), and layer separation between `LmCore` and `LmMultiTurn` are clean. The 3-case idempotency in `ResolveToolCallAsync` is correctly designed for webhook retries.

Found **2 bugs** and **significant test coverage gaps** to address before merge.

---

## Bugs

### 1. Missing `ObjectDisposedException` guard on `ResolveToolCallAsync`
Other public API methods check `ObjectDisposedException.ThrowIf(_isDisposed, this)`. `ResolveToolCallAsync` does not. Resolutions arrive externally on arbitrary threads and can hit after disposal.

### 2. Fire-and-forget resume sentinel can permanently stall auto-resume
`EnqueueResumeSentinel` fires-and-forgets the sentinel write. If it fails, `_resumeScheduled` is already `true` and `_lastDeferringRunId`/`_lastDeferringGenerationId` are cleared. No future resolution can re-trigger. Loop permanently stalled.

---

## Test Coverage Gaps

### 3. `OnHistoryRestoredAsync` has zero tests (process restart recovery)
65 lines rebuilding `_deferred` from persisted history. None of the 6 deferred tests exercise this.

### 4. `ReplaceMessageAsync` untested for File and SQLite stores
InMemory has 3 tests. File (full-file rewrite) and SQLite (hand-written UPDATE) have zero.

### 5. No `LegacyHandlerAdapter` tests for Deferred path
MCP server adapter calls `ToLegacyHandler` for every tool. The `NotSupportedException` guardrail is untested.

### 6. No streaming-path deferred test for `FunctionCallMiddleware`
3 non-streaming deferred tests exist. Zero streaming tests.

### 7. No multi-deferred-per-turn partial resolution test at loop level

---

## Code Quality

### 8. `OperationCanceledException` swallowed in tool execution (line 551)
Converts cancellation into LLM error messages instead of propagating abort.

### 9. Deferred-specific methods in `MultiTurnAgentBase` only used by one subclass
Consider keeping in `MultiTurnAgentLoop` or extracting to a helper.

### 10. `AddDeferredToHistoryAsync` persistence failure contradicts documented guarantee
Doc says placeholder must be durable before publish. Exception is swallowed and publish continues.

---

## Minor
- `_deferred` dictionary has no eviction
- LmCore XML docs reference `MultiTurnAgentLoop` by name (doc coupling)
- `HasUnresolvedDeferralsForGeneration` iterates ConcurrentDictionary inside lock

---

## Positive Observations
- Layer boundary discipline excellent
- Discriminated union pattern well-executed
- Idempotency handling correct for webhook retries
- Tool calls correctly executed in parallel
- Handler signature unification is clean
- PR description is excellent
- Commit messages are detailed

**Verdict: REQUEST CHANGES** — Fix bugs #1, #2 and add critical tests #3, #4 before merge.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental re-review of iteration 5 (f7662cab -> 56588f18).

## Summary

The decoupling of `ToolHandlerResult` from `ToolCallResult` via `ToolCallResultBuilder` is architecturally clean -- the Conformist pattern prevents the two code paths (FunctionCallMiddleware / MultiTurnAgentLoop) from drifting. The sealed discriminated union with `ToolHandlerResultPayload` correctly separates handler-controlled state from framework-controlled state. The empty `Deferred()` decision is the right YAGNI call -- extensible later via optional fields if needed.

**Build**: 0 errors (146 pre-existing warnings)
**Tests**: LmCore 702/702, LmMultiTurn 167/167

## New Findings (iteration 5)

Two new MEDIUM architecture findings posted inline below. The structural changes themselves are well-designed.

## Previous Comment Threads

**All 6 active threads from the initial review remain unaddressed in this iteration.** The iteration focused exclusively on type decoupling. Specifically:

| Thread | Finding | Status |
|--------|---------|--------|
| ObjectDisposedException guard on ResolveToolCallAsync | Bug | Still open |
| Fire-and-forget WriteResumeSentinelAsync stall | Bug | Still open |
| OperationCanceledException swallowed by broad catch | Bug | Still open |
| No tests for OnHistoryRestoredAsync | Coverage gap | Still open |
| Persistence failure contradicts doc guarantee | Design gap | Still open |
| Missing LegacyHandlerAdapter Deferred test | Coverage gap | Still open |

## Verdict

**COMMENT** -- withholding approval. The iteration 5 decoupling is correct and well-tested for the happy paths. However, the 3 bug-class issues from the initial review (ObjectDisposedException, OperationCanceledException, fire-and-forget stall) should be addressed before merge -- they represent real failure modes in production shutdown/cancellation scenarios.

_Review agents used: architecture-review, test-coverage-review, exception-handling-review_

src/LmCore/Messages/ToolCallResultBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+                ErrorCode = r.Payload.ErrorCode,
+            },
+            ToolHandlerResult.Deferred => new ToolCallResult(
+                toolCallId,
+                string.Empty,
+                executionTarget)
+            {
+                ToolName = toolName,
+                IsDeferred = true,
+                DeferredAt = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds(),
```
gautam-achieveai, line 54: **[Gautam's review bot]** [MEDIUM] Ambient time dependency. `DateTimeOffset.UtcNow` hardcoded in a static utility method makes timestamp assertions non-deterministic for SDK consumers testing deferral flows. Tests must use fuzzy checks (`.Should().BeGreaterThan(0)`) instead of asserting exact values.

Consider accepting an optional `TimeProvider` parameter:
```csharp
public static ToolCallResult FromHandlerResult(
    ToolHandlerResult result,
    string? toolCallId,
    string? toolName = null,
    ExecutionTarget executionTarget = ExecutionTarget.LocalFunction,
    TimeProvider? timeProvider = null)
{
    var now = (timeProvider ?? TimeProvider.System).GetUtcNow().ToUnixTimeMilliseconds();
    // ...
}
```
Keeps the default zero-ceremony while enabling test determinism. Non-blocking.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,897 @@
...
+            ThreadId = ThreadId,
+        };
+
+        // Build messages list with system prompt prepended (if configured)
+        var messagesToSend = GetMessagesWithSystemPrompt();
+
+        // Precondition: never send a request while any deferred tool result is unresolved.
+        // The provider would reject an empty tool_result.content, but the real bug is sending
+        // at all — host code must call ResolveToolCallAsync before resuming.
+        foreach (var m in messagesToSend)
```
gautam-achieveai: **[Gautam's review bot]** [MEDIUM] O(n) history scan on every turn. This precondition guard iterates the full `messagesToSend` list on every turn, even when `_deferred` is empty (the common case). For long-running agent conversations (200+ messages), this accumulates.

Short-circuit with the in-memory index first:
```csharp
// Fast path: if no known deferrals, skip the expensive scan.
if (_deferred.Count > 0)
{
    foreach (var m in messagesToSend)
    {
        // ... existing scan
    }
}
```
Preserves the defense-in-depth property (scan when deferrals exist) while making the common case O(1). Non-blocking.

src/LmCore/Messages/ToolCallResultBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+        string? toolCallId,
+        string? toolName = null,
+        ExecutionTarget executionTarget = ExecutionTarget.LocalFunction)
+    {
+        return result switch
+        {
+            ToolHandlerResult.Resolved r => new ToolCallResult(
+                toolCallId,
+                r.Payload.Text,
+                r.Payload.ContentBlocks,
```
gautam-achieveai, line 40: **[Gautam's review bot]** [MEDIUM] Missing unit test: `ContentBlocks` propagation through builder. The `Resolved` branch maps `r.Payload.ContentBlocks` into the wire-shape `ToolCallResult`, but no direct unit test verifies this. The existing `_StampsToolCallIdAndToolName` covers text-only, `_PropagatesErrorPayload` covers error. Multi-modal is only tested indirectly via `MultiModalHandlerIntegrationTests`.

Suggested test:
```csharp
[Fact]
public void ToolCallResultBuilder_FromHandlerResult_PropagatesContentBlocks()
{
    var blocks = new List<ToolResultContentBlock>
    {
        new ImageToolResultBlock { Data = ""abc"", MimeType = ""image/png"" },
    };
    var result = ToolHandlerResult.FromMultiModal(""text"", blocks);
    var tcr = ToolCallResultBuilder.FromHandlerResult(result, ""tc_mm"", ""image_tool"");
    tcr.ContentBlocks.Should().BeSameAs(blocks);
    tcr.Result.Should().Be(""text"");
}
```

gautam-achieveai (approved):
**[Gautam's review bot]** Re-review of iteration 6 (diff `56588f18..de69d851`).

## ✅ Verdict: APPROVE

All 6 bug-class findings from the initial review are addressed with correct implementations. 8 of 9 previous comment threads resolved. Exception handling patterns verified clean by deep review (all 7 patterns correct).

## Previous Thread Status (8/9 addressed)

| Thread | Finding | Status |
|--------|---------|--------|
| ObjectDisposedException guard | `ThrowIfDisposed()` added to `ResolveToolCallAsync` + `GetDeferredToolCallsAsync` | ✅ |
| Fire-and-forget stall | State restoration on `WriteResumeSentinelAsync` failure with concurrent guard | ✅ |
| OperationCanceledException swallowed | `when (ct.IsCancellationRequested)` guard before broad catch | ✅ |
| No tests for OnHistoryRestoredAsync | Full process-restart recovery test added | ✅ |
| Persistence failure contradicts guarantee | Persist-before-append + exception propagation + _deferred unwind | ✅ |
| LegacyHandlerAdapter test coverage | 5 tests in new file | ✅ |
| O(n) history scan | `!_deferred.IsEmpty` fast path | ✅ |
| ContentBlocks propagation test | Direct unit test added | ✅ |
| Ambient time dependency (ToolCallResultBuilder) | Still open — non-blocking MEDIUM | ⏳ |

## New Findings (non-blocking)

**[MEDIUM] No test for `_deferred` unwind on persistence failure** (`MultiTurnAgentLoop.cs:495-503`). The `catch { TryRemove; throw; }` block that prevents split-brain between `_deferred` and history has no test. A regression (removing the `TryRemove`) would leave a phantom deferred entry that `ResolveToolCallAsync` finds but `UpdateToolResultByCallId` can't match — a data-consistency corruption that's hard to diagnose. Consider adding a test with a mock `IConversationStore` that throws on `AppendMessagesAsync`.

**[LOW]** `WriteResumeSentinelAsync` state restoration and `ToolsCallResultMessage` aggregate guard path are untested — acceptable given complexity of exercising them and correctness verified by exception-handling review.

## Build & Test
- Build: ✅ 0 errors (159 pre-existing warnings)
- LmCore.Tests: ✅ 709/709
- LmMultiTurn.Tests: ✅ 172/172

## What's Done Well
- Persist-before-append ordering in `AddDeferredToHistoryAsync` correctly prevents split-brain
- `when (ct.IsCancellationRequested)` guard correctly distinguishes caller vs tool-internal cancellation
- `WriteResumeSentinelAsync` concurrent state check is sound (only restores when markers match the failed sentinel's state)
- 12 new tests provide strong coverage of all production fixes
- `MultipleDeferred_PartialResolution_AutoResumeOnlyAfterAllResolved` is an excellent contract test

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -1,481 +1,946 @@
...
+
+            // Persist synchronously so a webhook-triggered ReplaceMessageAsync cannot race
+            // the placeholder's append. If persistence fails, unwind the deferred-entry
+            // registration so a future resolution doesn't find a half-applied state — the
+            // exception propagates up to the run loop and surfaces as a RunFailed.
+            try
+            {
+                await AddDeferredToHistoryAsync(result, ct);
+            }
+            catch
```
gautam-achieveai, line 499: **[Gautam's review bot]** [MEDIUM] No test for this unwind path. If a future refactor removes the `TryRemove`, `_deferred` retains an entry whose placeholder never made it to history or store — `ResolveToolCallAsync` would find the entry but `UpdateToolResultByCallId` would throw `InvalidOperationException`. Consider a test with a mock `IConversationStore` that throws on `AppendMessagesAsync`, then assert `GetDeferredToolCallsAsync()` returns empty.

Non-blocking — the implementation is correct.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in de69d851. `ThrowIfDisposed()` added as a protected helper on `MultiTurnAgentBase`, called at the top of both `ResolveToolCallAsync` and `GetDeferredToolCallsAsync`. Tests added for both paths. Correct implementation.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. On `EnqueueRawAsync` failure, the catch block now restores `_lastDeferringRunId`/`_lastDeferringGenerationId` and clears `_resumeScheduled` under `_resumeLock`, with a concurrent-state guard that avoids clobbering fresh markers from a new deferral wave. Verified correct by exception-handling review.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` now precedes the broad catch. The `when` guard correctly distinguishes caller-initiated cancellation (propagates) from tool-internal cancellation (becomes LLM-visible error). Test added: `OperationCanceledException_FromHandler_PropagatesAsRunCanceled`.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. New test `OnHistoryRestoredAsync_RebuildsDeferredRegistry_FromPersistedHistory` pre-populates an `InMemoryConversationStore` with a deferred placeholder, recovers via `RecoverAsync()`, and verifies `GetDeferredToolCallsAsync` surfaces the entry with correct tool name, args, and timestamps.

src/LmMultiTurn/MultiTurnAgentBase.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. `AddDeferredToHistoryAsync` now persists BEFORE the in-memory append. If `AppendMessagesAsync` throws, no in-memory state is mutated. `ExecuteAndPublishToolCallAsync` wraps the call with `catch { TryRemove; throw; }` to unwind the `_deferred` entry on failure. Doc and code are now consistent.

src/LmCore/Middleware/LegacyHandlerAdapter.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. New `LegacyHandlerAdapterTests.cs` with 5 tests: `ToLegacyHandler` resolved + deferred paths, `WrapToLegacyHandlers` lazy throw, and both `ToNewHandler` overloads (string and ToolCallResult).

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. `if (!_deferred.IsEmpty)` now gates the O(n) scan. Comment explains the belt-and-suspenders rationale for keeping the scan when deferrals exist.

src/LmCore/Messages/ToolCallResultBuilder.cs
gautam-achieveai: **[Gautam's review bot]** ✅ Addressed in `de69d851`. `ToolCallResultBuilder_FromHandlerResult_PropagatesContentBlocks` added with `BeSameAs` reference equality check — confirms no copy/rewrap.

Issue #28, opened 2026-05-03
MockProviderHost: codex-mock hangs because /v1/responses route is not implemented


gautam-achieveai:
[Gautam's Dev bot]

## Symptom

In `samples/LmStreaming.Sample`, selecting the `codex-mock` provider and sending a user message causes the chat UI to hang on the "Stop" state. The WebSocket never receives a `done` event; only the user turn is persisted.

Reproduction: PR #26 preview env, choose `codex-mock`, type "hello", send.

## Root cause

Codex CLI 0.128 (and current versions) defaults to the OpenAI **Responses API** (`POST /v1/responses`). `samples/MockProviderHost/MockProviderHostBuilder.cs:95-99` only routes:

```csharp
app.MapPost("/v1/chat/completions", …);
app.MapPost("/v1/messages", …);
```

Codex's app-server hits the unrouted `/v1/responses` path; the request 404s, the JSON-RPC turn never resolves, and the agent loop hangs. See `src/CodexSdkProvider/Transport/CodexAppServerTransport.cs:84-106` for how `OPENAI_BASE_URL` is plumbed (no `wire_api` override today).

## Already acknowledged in tree

`samples/MockProviderHost/README.md:175-178`:

> Newer Copilot model paths (GPT-5 family) use `/v1/responses`. If the probe finds the CLI defaults to that route on the version under test, the follow-up E2E issue must add a `/v1/responses` route to `MockProviderHostBuilder` before E2E coverage is possible.

Codex hits the same gap as Copilot's GPT-5 path.

## Proposed fixes (pick one)

**(a) Add `/v1/responses` route to `MockProviderHostBuilder`** that adapts `ScriptedSseResponder` output to OpenAI Responses-API SSE shape. Larger surface (new wire format in `src/LmTestUtils/TestMode/`), but future-proofs for the Copilot GPT-5 path called out in the README.

**(b) Override Codex wire format to chat-completions when a mock base URL is set** by passing `-c model_providers.openai.wire_api="chat"` to `codex app-server` in `CodexAppServerTransport.cs:87` when `baseUrl` is non-null. ~10 LoC, targeted, but couples the Codex transport to mock-mode behavior and skips Responses-API-only Codex features.

## Definition of done

- `codex-mock` provider produces a streaming reply end-to-end against the in-process mock host on the LmStreaming.Sample preview env.
- A regression test (E2E or unit) covers the chosen wire path so the failure mode can't silently return.

## Linked PR

[...]

PR #30, opened 2026-05-03
[Gautam's Dev bot] fix(claude-mock): surface mock CLI's silent renders (#29)

### LINKED ISSUE #29: MockProviderHost: claude-mock completes silently with no assistant content rendered
[Gautam's Dev bot]

## Symptom

In `samples/LmStreaming.Sample`, selecting `claude-mock` and sending a user message causes the WebSocket to emit `done` with **zero** assistant `TextMessage` content. `GET /api/conversations/{thread}/messages` shows only the user turn was persisted. No console errors; no failed network requests.

Reproduction: PR #26 preview env, choose `claude-mock`, type "hello", send.

## Root cause (suspected)

The Claude Agent SDK CLI reaches `POST /v1/messages` on the mock and the run completes successfully, but the streaming-event shape produced by `AnthropicSseStreamHttpContent` (in `src/LmTestUtils/TestMode/`) does not match what the CLI version under test parses into a `TextContentBlock`. Either the CLI emits no JSONL output blocks or our SDK event mapping in `src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs` filters them out.

Likely sensitivities (need to confirm with `ANTHROPIC_LOG=debug` capture):
1. `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` toggle (set when mock base URL is configured — see `samples/LmStreaming.Sample/Program.cs:996`).
2. Whether `ANTHROPIC_BASE_URL` carries the `/v1` suffix (the Claude Agent SDK E2E test uses `BaseUrl = fixture.BaseUrl + "/v1"`, but `Program.cs:270` passes the raw `mockBaseUrl` without `/v1`).
3. The exact `message_start` / `content_block_start` / `content_block_delta` / `message_stop` event shape the CLI requires.

## Already acknowledged in tree

`tests/MockProviderHost.E2E.Tests/Scenarios/ClaudeAgentSdkAgainstMockTests.cs:23-28`:

> Real-CLI E2E is opt-in: the protocol-level handshake the CLI performs against `/v1/messages` depends on CLI version and is the subject of a follow-up issue. Run with `LMDOTNET_RUN_CLAUDE_E2E=1` once a stable scenario contract is recorded.

This bug **is** that follow-up issue.

## Proposed approach

1. Capture an `ANTHROPIC_LOG=debug` trace of `claude` CLI against a real Anthropic API and diff the streaming event shape against what `AnthropicSseStreamHttpContent` emits.
2. Pin a known-good `claude` CLI version range and document the contract (versions newer than that may need probe updates).
3. Reconcile the `BaseUrl` shape: `Program.cs:270` should pass `$"{mockBaseUrl}/v1"` (matching `codex-mock` and `copilot-mock`), or `MockProviderHostBuilder` should accept the bare base URL — pick one and document.
4. Add a probe test that fails fast if the CLI's actual request shape diverges (e.g., missing `anthropic-version` header, unexpected `?beta=` query string).

## Definition of done

- `claude-mock` provider produces a streaming reply end-to-end against the mock host with at least one rendered `TextMessage` on the LmStreaming.Sample preview env.
- The `ClaudeAgentSdkAgainstMockTests` skip-fact is upgraded to assert content rendering (not just "the host got a request").

## Linked PR

#26 — where the limitation was reported and a UX banner was added pointing here.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Issue #29 (`MockProviderHost: claude-mock completes silently with no assistant content rendered`) was three independent bugs masking each other. This PR untangles them, ships a unified BaseUrl helper to keep the two conflicting conventions straight at every wiring site, and lands two layers of regression coverage.

### Root causes

1. **BaseUrl convention collision.** The `LmStreaming.Sample` passed `ANTHROPIC_BASE_URL` with a trailing `/v1`. The Claude Agent SDK CLI (and the underlying Anthropic SDK) appends `/v1/messages` itself, so requests landed on `/v1/v1/messages`, which the mock host 404'd silently. The in-house `AnthropicClient` uses the opposite convention (it appends `/messages`, so the configured URL **must** include `/v1`) — easy to confuse at a wiring site.
2. **Unsupported CLI flags.** The sample set `DisableCheckpoints = true` / `DisableSessionPersistence = true`, which emit `--no-checkpoints` / `--no-session-persistence`. CLI v0.1.55 does not accept those flags and exits with "unknown option" before any `/v1/messages` call.
3. **No diagnostic trail.** `MockProviderHost` returned a bare 404 for unmatched paths with no log entry. A BaseUrl misconfiguration produced no surfaced symptom anywhere.

### Fixes

- **`LmCore.Utils.BaseUrlNormalizer`** — new helper with `EnsureV1Suffix` (for the in-house client) and `StripV1Suffix` (for the SDK CLI). XML docs spell out both conventions and how they failed in #29. 14 unit tests cover null/whitespace, with/without trailing slash, with/without `/v1`, and the `/api/v1service` substring edge case.
- **Sample wiring** — `Program.cs` now uses `BaseUrlNormalizer.StripV1Suffix(mockBaseUrl)` for `claude-mock`, and the unsupported `DisableCheckpoints` / `DisableSessionPersistence` options are removed.
- **`MockProviderHostBuilder.MapFallback`** — unmatched requests now log a warning and return a body that names the path, so the next BaseUrl drift fails loudly instead of silently.
- **Docs** — `.env.example` and `samples/MockProviderHost/README.md` document the convention split explicitly.

### Hardening

- **`AnthropicSseRoundTripProbeTests`** (new) — pins the mock-emitter ↔ production-parser contract by feeding `AnthropicSseStreamHttpContent` output through `AnthropicStreamParser` and asserting the text comes out as a `TextMessage`. Plus a spec-shape pin on SSE event ordering. Fast, no I/O, runs in CI.
- **`ClaudeAgentSdkAgainstMockTests`** (rewritten) — was previously skipped with an opt-in env var **and** had the wrong BaseUrl, so it was effectively dead code. Now: drops the bogus `/v1`, enqueues a small pool of scripted turns to absorb the CLI's pre-flight requests, and asserts on the **rendered text content** (not just "queue drained"). Passes against real CLI v0.1.55 when run with `LMDOTNET_RUN_CLAUDE_E2E=1`. Still skipped by default in CI.

## Test plan

- [x] `dotnet build LmDotnetTools.sln` — 0 errors, no new warnings on touched files.
- [x] `dotnet test tests/LmCore.Tests --filter BaseUrlNormalizer` — 14/14 passed.
- [x] `dotnet test tests/AnthropicProvider.Tests --filter AnthropicSseRoundTrip` — 2/2 passed.
- [x] `LMDOTNET_RUN_CLAUDE_E2E=1 dotnet test tests/MockProviderHost.E2E.Tests` — passes against real CLI v0.1.55, marker text round-trips through the CLI as a `TextMessage`.
- [x] Full non-browser test suite passes; pre-existing failures (`McpIntegrationTests`, `LmStreaming.Sample.Browser.E2E.Tests`) are environment-bound and not touched by this PR.

Closes #29

### DIFF
--- a/src/LmCore/Utils/BaseUrlNormalizer.cs
+++ b/src/LmCore/Utils/BaseUrlNormalizer.cs
@@ -1,0 +1,74 @@
+namespace AchieveAi.LmDotnetTools.LmCore.Utils;
+
+/// <summary>
+///     Normalizes provider <c>BaseUrl</c> values to the convention each consumer expects.
+/// </summary>
+/// <remarks>
+///     Two conventions exist in this codebase and they are easy to confuse:
+///     <list type="bullet">
+///         <item>
+///             <description>
+///                 The in-house <c>AnthropicClient</c> appends <c>/messages</c> to <c>BaseUrl</c>
+///                 and therefore expects the configured URL to <em>include</em> a trailing
+///                 <c>/v1</c> segment (e.g. <c>https://api.anthropic.com/v1</c>).
+///             </description>
+///         </item>
+///         <item>
+///             <description>
+///                 The official <c>@anthropic-ai/claude-agent-sdk</c> CLI (and the underlying
+///                 Anthropic SDK) appends <c>/v1/messages</c> to whatever <c>ANTHROPIC_BASE_URL</c>
+///                 it reads, so that env var must <em>exclude</em> the <c>/v1</c> suffix
+///                 (e.g. <c>https://api.anthropic.com</c>). Passing a URL ending in <c>/v1</c>
+///                 produces requests against <c>/v1/v1/messages</c>, which 404 silently and surface
+///                 as "the agent completes with no assistant content rendered" (issue #29).
+///             </description>
+///         </item>
+///     </list>
+///     This helper exposes both transformations so callers can pick the convention they need
+///     instead of open-coding string surgery at every wiring site.
+/// </remarks>
+public static class BaseUrlNormalizer
+{
+    private const string V1Segment = "/v1";
+
+    /// <summary>
+    ///     Returns <paramref name="baseUrl"/> with a single trailing <c>/v1</c> segment, suitable
+    ///     for the in-house <c>AnthropicClient</c> which appends <c>/messages</c> directly.
+    /// </summary>
+    /// <param name="baseUrl">The configured base URL. May or may not already end in <c>/v1</c>.</param>
+    /// <returns>The URL guaranteed to end in <c>/v1</c> (no trailing slash). Returns <c>null</c>
+    /// when <paramref name="baseUrl"/> is null or whitespace.</returns>
+    public static string? EnsureV1Suffix(string? baseUrl)
+    {
+        if (string.IsNullOrWhiteSpace(baseUrl))
+        {
+            return baseUrl;
+        }
+
+        var trimmed = baseUrl.TrimEnd('/');
+        return trimmed.EndsWith(V1Segment, StringComparison.Ordinal)
+            ? trimmed
+            : trimmed + V1Segment;
+    }
+
+    /// <summary>
+    ///     Returns <paramref name="baseUrl"/> with any trailing <c>/v1</c> segment removed,
+    ///     suitable for <c>ANTHROPIC_BASE_URL</c> as consumed by the Anthropic SDK / Claude
+    ///     Agent SDK CLI which append <c>/v1/messages</c> themselves.
+    /// </summary>
+    /// <param name="baseUrl">The configured base URL. May or may not already end in <c>/v1</c>.</param>
+    /// <returns>The URL guaranteed to <em>not</em> end in <c>/v1</c> (no trailing slash). Returns
+    /// <c>null</c> when <paramref name="baseUrl"/> is null or whitespace.</returns>
+    public static string? StripV1Suffix(string? baseUrl)
+    {
+        if (string.IsNullOrWhiteSpace(baseUrl))
+        {
+            return baseUrl;
+        }
+
+        var trimmed = baseUrl.TrimEnd('/');
+        return trimmed.EndsWith(V1Segment, StringComparison.Ordinal)
+            ? trimmed[..^V1Segment.Length]
+            : trimmed;
+    }
+}
--- a/tests/LmStreaming.Sample.Browser.E2E.Tests/Infrastructure/BrowserWebAppFactory.cs
+++ b/tests/LmStreaming.Sample.Browser.E2E.Tests/Infrastructure/BrowserWebAppFactory.cs
@@ -35,25 +35,41 @@ namespace LmStreaming.Sample.Browser.E2E.Tests.Infrastructure;
 public sealed class BrowserWebAppFactory : WebApplicationFactory<Program>
 {
     private readonly string _providerMode;
-    private readonly ITestAgentBuilder _builder;
+    private readonly ITestAgentBuilder? _builder;
     private IHost? _kestrelHost;
     private string? _serverAddress;
 
-    public BrowserWebAppFactory(string providerMode, ITestAgentBuilder builder)
+    public BrowserWebAppFactory(string providerMode, ITestAgentBuilder? builder)
     {
-        if (
-            !string.Equals(providerMode, "test", StringComparison.OrdinalIgnoreCase)
-            && !string.Equals(providerMode, "test-anthropic", StringComparison.OrdinalIgnoreCase)
-        )
+        // Scripted SSE modes ('test' / 'test-anthropic') drive a fake handler via ITestAgentBuilder.
+        // 'claude-mock' (and other *-mock providers) drive the real CLI against the in-process
+        // MockProviderHostLifetime that boots automatically at app startup; for those modes the
+        // builder is unused and may be null.
+        var isScriptedMode =
+            string.Equals(providerMode, "test", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "test-anthropic", StringComparison.OrdinalIgnoreCase);
+        var isMockHostMode =
+            string.Equals(providerMode, "claude-mock", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "codex-mock", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "copilot-mock", StringComparison.OrdinalIgnoreCase);
+
+        if (!isScriptedMode && !isMockHostMode)
         {
             throw new ArgumentException(
-                $"providerMode must be 'test' or 'test-anthropic'; got '{providerMode}'",
+                $"providerMode must be 'test', 'test-anthropic', or a *-mock variant; got '{providerMode}'",
                 nameof(providerMode)
             );
         }
 
+        if (isScriptedMode && builder is null)
+        {
+            throw new ArgumentNullException(
+                nameof(builder),
+                "Scripted provider modes require an ITestAgentBuilder.");
+        }
+
         _providerMode = providerMode;
-        _builder = builder ?? throw new ArgumentNullException(nameof(builder));
+        _builder = builder;
 
         // Program.cs reads LM_PROVIDER_MODE at the top-level, well before any host-builder
         // callback fires. Set it here so the sample picks the right test-mode agent factory.
@@ -80,11 +96,16 @@ public sealed class BrowserWebAppFactory : WebApplicationFactory<Program>
 
         // ConfigureTestServices runs after Program.cs's default registrations, so
         // AddSingleton + RemoveAll last-wins semantics guarantee our builder is resolved.
-        builder.ConfigureTestServices(services =>
+        // For *-mock modes the builder is unused (real CLI drives the in-process mock host),
+        // so leave the default registration in place when none is supplied.
+        if (_builder is not null)
         {
-            services.RemoveAll<ITestAgentBuilder>();
-            services.AddSingleton(_builder);
-        });
+            builder.ConfigureTestServices(services =>
+            {
+                services.RemoveAll<ITestAgentBuilder>();
+                services.AddSingleton(_builder);
+            });
+        }
     }
 
     protected override IHost CreateHost(IHostBuilder builder)
--- a/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ClaudeMockProviderTests.cs
+++ b/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ClaudeMockProviderTests.cs
@@ -1,0 +1,90 @@
+using FluentAssertions;
+using LmStreaming.Sample.Browser.E2E.Tests.Infrastructure;
+
+namespace LmStreaming.Sample.Browser.E2E.Tests.Scenarios;
+
+/// <summary>
+/// Browser-level regression for issue #29: drives the real Claude CLI against the
+/// in-process MockProviderHost via the <c>claude-mock</c> provider mode and asserts the
+/// chat client renders non-empty assistant text. This is the Playwright counterpart to
+/// the SDK-level <c>ClaudeAgentSdkAgainstMockTests</c> in
+/// <c>tests/MockProviderHost.E2E.Tests/</c>: the SDK test pins the protocol contract,
+/// this test pins the UI rendering of the same flow so the silent-completion failure
+/// mode is caught at both layers.
+/// </summary>
+/// <remarks>
+/// Skipped unless <c>LMDOTNET_RUN_CLAUDE_E2E=1</c> AND a <c>claude</c> CLI binary is
+/// reachable on PATH (or via <c>CLAUDE_CLI_PATH</c>). Mirrors the gating of the SDK-level
+/// test so machines without the CLI installed (most CI runners today) skip cleanly
+/// rather than fail.
+/// </remarks>
+[Collection(PlaywrightCollection.Name)]
+public sealed class ClaudeMockProviderTests
+{
+    private readonly PlaywrightFixture _fixture;
+
+    public ClaudeMockProviderTests(PlaywrightFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [SkippableFact]
+    public async Task Claude_mock_renders_assistant_text_in_chat_ui()
+    {
+        Skip.IfNot(
+            string.Equals(
+                Environment.GetEnvironmentVariable("LMDOTNET_RUN_CLAUDE_E2E"),
+                "1",
+                StringComparison.Ordinal),
+            "Set LMDOTNET_RUN_CLAUDE_E2E=1 to run claude-mock browser E2E tests.");
+        Skip.IfNot(ClaudeCliIsAvailable(), "Claude CLI not found on PATH or via CLAUDE_CLI_PATH.");
+
+        await using var factory = new BrowserWebAppFactory("claude-mock", builder: null);
+        await using var context = await _fixture.Browser.NewContextAsync();
+        var page = await context.NewPageAsync();
+
+        await page.GotoAsync(factory.ServerAddress);
+        await page.Textarea().WaitForAsync();
+
+        await page.SendMessageAsync("say hello");
+        await page.WaitForStreamIdleAsync(timeoutMs: 60_000);
+
+        await page.AssistantText().WaitForCountAtLeastAsync(1);
+        var assistantTexts = await page.AssistantText().AllInnerTextsAsync();
+        var combined = string.Join(" ", assistantTexts).Trim();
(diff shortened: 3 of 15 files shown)


gautam-achieveai:

| Service | URL | Status |
|---|---|---|
| lmdotnettools | http://pr-30.lmdotnettools.localhost/ | ✅ Healthy |

gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #30 — fix(claude-mock): surface mock CLI's silent renders (#29)

## Context
Issue #29: Three independent bugs combined to make `claude-mock` complete with zero assistant content:
1. `ANTHROPIC_BASE_URL` with trailing `/v1` → CLI re-appended → `/v1/v1/messages` → silent 404
2. `--no-checkpoints` / `--no-session-persistence` unsupported in CLI v0.1.55 → "unknown option" exit
3. Unmatched mock-host requests returned bare 404 with no log entry → zero diagnostic trail

## Verdict: ✅ APPROVE with suggestions

Well-scoped fix that addresses all three root causes with good hardening. The commit message is excellent — documents each root cause and its fix. The round-trip probe tests are a particularly strong addition.

## Findings

### [MEDIUM] Normalization burden falls entirely on callers — provider doesn't guard its own invariant

`ClaudeAgentSdkClient` writes `options.BaseUrl` directly to `ANTHROPIC_BASE_URL` without verifying the invariant that the CLI requires no trailing `/v1`. The `BaseUrlNormalizer` utility exists but nothing enforces its use — future callers who forget `StripV1Suffix` will reproduce the exact silent-404 bug this PR fixes.

**Suggestion:** Consider one of:
- Add defensive `StripV1Suffix` inside `ClaudeAgentSdkClient` itself (enforce at the boundary)
- Add a runtime `LogWarning` if `BaseUrl` ends in `/v1` when wiring `ANTHROPIC_BASE_URL`
- Add an XML doc warning on `ClaudeAgentSdkOptions.BaseUrl` noting the convention

This isn't blocking — the documentation is thorough — but the root cause of #29 could recur at a different call site.

### [LOW] `EnsureV1Suffix` has zero production callers

Only referenced in tests and README. The symmetric API design is reasonable for future use, but currently no production code calls it.

### [LOW] E2E test `StartAsync` called without `CancellationToken`

`ClaudeAgentSdkAgainstMockTests:77` — The previous version passed `cts.Token` to `StartAsync`; the new version omits it. If CLI startup hangs, the test blocks indefinitely. `ProcessTimeoutMs` provides a process-level safeguard, but passing `cts.Token` would be more defensive.

### [LOW] Missing `EnsureV1Suffix` substring-guard test

`StripV1Suffix` has an explicit test for the `/v1service` substring case but `EnsureV1Suffix` doesn't have the symmetric test. The code handles it correctly, but parity communicates intent.

## Positive Observations
- 🏆 Excellent commit message — documents all three root causes and each fix
- ✅ `BaseUrlNormalizer` correctly placed in `LmCore.Utils` (shared innermost layer, no new dependency arrows)
- ✅ `MapFallback` diagnostic anchor — actionable log output for future BaseUrl drift
- ✅ Round-trip probe tests catch emitter↔parser protocol drift at unit-test speed
- ✅ E2E test now asserts on rendered text content, directly testing the #29 failure mode
- ✅ Exception handling is clean throughout — proper catch ordering, filtered OCE, structured logging

## Review Agents Used
| Agent | Result |
|---|---|
| architecture-review | Clean. BaseUrlNormalizer placement correct, no layer violations, no circular deps |
| test-coverage-review | Good coverage. All three root causes have regression tests |
| exception-handling-review | Clean. No issues — proper patterns throughout |

tests/MockProviderHost.E2E.Tests/Scenarios/ClaudeAgentSdkAgainstMockTests.cs
```diff
@@ -63,27 +74,39 @@ public async Task Cli_routes_through_mock_host_via_BaseUrl_and_AuthToken_overrid
             ],
         };
 
+        await client.StartAsync(request);
```
gautam-achieveai: **[Gautam's review bot]** [LOW] `StartAsync` is called without the `CancellationToken` that the previous version passed (`cts.Token`). The `cts` with its 20-second timeout is scoped to the `SendMessagesAsync` loop below. If CLI startup hangs for any reason, this call blocks indefinitely until the test runner kills it.

`ProcessTimeoutMs = 30000` provides a process-level safeguard, so this is unlikely to be a real problem in practice. But passing `cts.Token` here would be more defensive:
```csharp
await client.StartAsync(request, cts.Token);
```

src/LmCore/Utils/BaseUrlNormalizer.cs
```diff
@@ -0,0 +1,74 @@
...
+///                 (e.g. <c>https://api.anthropic.com</c>). Passing a URL ending in <c>/v1</c>
+///                 produces requests against <c>/v1/v1/messages</c>, which 404 silently and surface
+///                 as "the agent completes with no assistant content rendered" (issue #29).
+///             </description>
+///         </item>
+///     </list>
+///     This helper exposes both transformations so callers can pick the convention they need
+///     instead of open-coding string surgery at every wiring site.
+/// </remarks>
+public static class BaseUrlNormalizer
```
gautam-achieveai, line 30: **[Gautam's review bot]** [MEDIUM — design consideration] This utility is well-designed, but the normalization burden falls entirely on callers. `ClaudeAgentSdkClient` writes `options.BaseUrl` directly to `ANTHROPIC_BASE_URL` without validating the convention. A future caller who forgets `StripV1Suffix` would silently reproduce issue #29.

Consider adding a defensive guard at the provider boundary — e.g., a `LogWarning` inside `ClaudeAgentSdkClient` if `BaseUrl` ends in `/v1`, or applying `StripV1Suffix` internally. The documentation is thorough, but documentation doesn't prevent bugs the way code does.

gautam-achieveai: [Gautam's Dev bot]

Addressed in 46e18ff. `ApplyMockHostOverrides` now applies `BaseUrlNormalizer.StripV1Suffix` to `options.BaseUrl` defensively before writing `ANTHROPIC_BASE_URL`, and emits a `LogWarning` when stripping actually changed the value so future callers see a clear diagnostic and can correct their configuration. New unit test `Defensively_strips_trailing_v1_from_BaseUrl` pins the behavior; the existing happy-path test was updated to the canonical (no `/v1`) form.

tests/MockProviderHost.E2E.Tests/Scenarios/ClaudeAgentSdkAgainstMockTests.cs
gautam-achieveai: [Gautam's Dev bot]

Addressed in 46e18ff. The `CancellationTokenSource` is now created before `StartAsync` and `cts.Token` is passed in, so a hung CLI startup is bounded by the same 20s deadline that already covered `SendMessagesAsync`. The `using` declaration ensures the source is still disposed even if startup throws.

gautam-achieveai:
[Gautam's Dev bot]

Pushed 46e18ff addressing the review:

- **[MEDIUM] Provider-boundary defensive guard** — `ApplyMockHostOverrides` now strips a trailing `/v1` from `options.BaseUrl` via `BaseUrlNormalizer.StripV1Suffix` before writing `ANTHROPIC_BASE_URL`, and emits a `LogWarning` when stripping actually changed the value. New `Defensively_strips_trailing_v1_from_BaseUrl` unit test pins the behavior. The future-caller-forgets path can no longer reproduce #29 silently.
- **[LOW] StartAsync without `cts.Token`** — `cts` is now created before `StartAsync` and passed in, so the 20s deadline also bounds CLI startup.
- **[LOW] Symmetric substring-guard for `EnsureV1Suffix`** — added `EnsureV1Suffix_only_treats_a_trailing_v1_segment_not_substrings_as_already_present` for parity with the `StripV1Suffix` substring test.
- **[LOW] `EnsureV1Suffix` has zero production callers** — leaving the symmetric API in place; it documents the OpenAI-style convention next to its inverse and is exercised by tests/README. Happy to remove it in a follow-up if you'd prefer to gate on actual production use.

Targeted tests: `BaseUrlNormalizerTests` 15/15 ✅, `ClaudeAgentSdkClientMockHostOverridesTests` 4/4 ✅, E2E project builds clean.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review of iteration 2 (dc007ca -> 46e18ff)

## Verdict: APPROVE

All four findings from iteration 1 have been addressed:

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | **MEDIUM** | Provider boundary doesn't guard its own /v1 invariant | ApplyMockHostOverrides now defensively strips /v1 and emits LogWarning |
| 2 | LOW | EnsureV1Suffix has zero production callers | Acknowledged - keeping for API symmetry |
| 3 | LOW | StartAsync called without CancellationToken | cts now created before StartAsync, token passed in |
| 4 | LOW | Missing EnsureV1Suffix substring-guard test | Symmetric test added |

## Incremental Changes Quality
The iteration 2 commit is clean and precisely scoped - each change maps directly to a review finding with no scope creep. The `ILogger? logger = null` default parameter is a nice touch for backward compatibility.

## Context
- **Issue #29**: Three independent bugs (BaseUrl /v1 collision, unsupported CLI flags, silent 404s) combined to make claude-mock complete with no assistant content.
- This PR fixes all three root causes, adds BaseUrlNormalizer utility, MapFallback diagnostic anchor, round-trip probe tests, and upgraded E2E coverage.

No new issues found. Ship it!

gautam-achieveai:
In Streaming.SampleApp, I see that ClaudeMock is marked as may not work, though it's working. Also can you write/test the streamingApp with playwright tests to make sure this provider is also tested for basic functionality?

gautam-achieveai:
[Gautam's Dev bot]

Both addressed in ef85917:

1. **Removed the `claude-mock` "may not work" caveat** in `ProviderRegistry` — issue #29 is fixed and the provider works end-to-end, so the banner shouldn't surface anymore. The `ProviderRegistryTests.KnownLimitation_TaggedOnBrokenMocks_UnsetOnHealthyOnes` assertion was flipped to require `claude-mock.KnownLimitation` to be `null`, matching `copilot-mock`. Only `codex-mock` (issue #28) still carries a caveat.

2. **Added a Playwright UI test for `claude-mock`** at `tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ClaudeMockProviderTests.cs`. It drives the real Claude CLI through the chat client (which auto-starts `MockProviderHostLifetime` in `claude-mock` mode), sends "say hello", waits for stream-idle, and asserts non-empty assistant text rendered — the exact failure mode from #29. To make this work, `BrowserWebAppFactory` was relaxed to accept `*-mock` provider modes and treat the `ITestAgentBuilder` injection as optional (the real CLI doesn't use it). Skipped unless `LMDOTNET_RUN_CLAUDE_E2E=1` AND a `claude` CLI is reachable on PATH/`CLAUDE_CLI_PATH`, mirroring the gating of the SDK-level test so machines without the CLI installed (most CI runners) skip cleanly. Verified locally: `Skipped! - Failed: 0, Passed: 0, Skipped: 1`.

Targeted tests still green: `ProviderRegistryTests` 27/27 ✅, `LmStreaming.Sample.Browser.E2E.Tests` build clean (0 warnings).

tests/LmStreaming.Sample.Browser.E2E.Tests/Infrastructure/BrowserWebAppFactory.cs
```diff
@@ -35,25 +35,41 @@ namespace LmStreaming.Sample.Browser.E2E.Tests.Infrastructure;
...
+        // 'claude-mock' (and other *-mock providers) drive the real CLI against the in-process
+        // MockProviderHostLifetime that boots automatically at app startup; for those modes the
+        // builder is unused and may be null.
+        var isScriptedMode =
+            string.Equals(providerMode, "test", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "test-anthropic", StringComparison.OrdinalIgnoreCase);
+        var isMockHostMode =
+            string.Equals(providerMode, "claude-mock", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "codex-mock", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(providerMode, "copilot-mock", StringComparison.OrdinalIgnoreCase);
```
gautam-achieveai, line 54: **[Gautam's review bot]** [MEDIUM] This explicit allowlist duplicates knowledge from `ProviderRegistry.IsMockProvider()` which uses `EndsWith("-mock")`. Adding a fourth `*-mock` provider would require updating both places.

Consider:
```csharp
var isMockHostMode = providerMode.EndsWith("-mock", StringComparison.OrdinalIgnoreCase);
```

This eliminates the synchronization point and stays forward-compatible. The `isScriptedMode` check can remain explicit since `test`/`test-anthropic` are a closed set with different semantics.

tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ClaudeMockProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        await page.WaitForStreamIdleAsync(timeoutMs: 60_000);
+
+        await page.AssistantText().WaitForCountAtLeastAsync(1);
+        var assistantTexts = await page.AssistantText().AllInnerTextsAsync();
+        var combined = string.Join(" ", assistantTexts).Trim();
+        combined.Should().NotBeEmpty(
+            "issue #29: claude-mock previously completed silently with no rendered text");
+    }
+
+    private static bool ClaudeCliIsAvailable()
```
gautam-achieveai, line 59: **[Gautam's review bot]** [LOW] This reimplements `FileSystemProbe.IsExecutableOnPath()` with subtle differences (missing `.bat` suffix, hardcoded separator). The Browser.E2E.Tests project already references the project containing `FileSystemProbe`. Consider delegating:

```csharp
private static bool ClaudeCliIsAvailable()
{
    var probe = new FileSystemProbe();
    var explicitPath = Environment.GetEnvironmentVariable("CLAUDE_CLI_PATH");
    if (!string.IsNullOrWhiteSpace(explicitPath) && probe.FileExists(explicitPath))
        return true;
    return probe.IsExecutableOnPath("claude");
}
```

4 lines instead of 30, and stays in sync with the production PATH-scanning logic.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review of iteration 3 (46e18ff -> ef85917).

## Summary

Iteration 3 addresses the follow-up request: removes the `claude-mock` "may not work" caveat from `ProviderRegistry` (since #29 is now fixed), relaxes `BrowserWebAppFactory` to accept `*-mock` provider modes, and adds a Playwright browser-level regression test for the silent-rendering bug.

**Review agents used:** architecture-review, duplicate-code-detector, test-coverage-review.

## Findings

### [MEDIUM] Hardcoded mock-provider allowlist in BrowserWebAppFactory

`BrowserWebAppFactory` explicitly enumerates `claude-mock`, `codex-mock`, `copilot-mock` (lines 51-54), while `ProviderRegistry.IsMockProvider()` uses the convention `EndsWith("-mock")`. Adding a new `*-mock` provider (e.g. `gemini-mock`) would require updating both places. Consider:

```csharp
var isMockHostMode = providerMode.EndsWith("-mock", StringComparison.OrdinalIgnoreCase);
```

### [LOW] `ClaudeCliIsAvailable()` reimplements PATH scanning logic

The private method in `ClaudeMockProviderTests` reimplements `FileSystemProbe.IsExecutableOnPath()` with subtle gaps (missing `.bat` on Windows, hardcoded separator). The project already references `FileSystemProbe` - delegate to it instead (4 lines vs 30).

### [LOW] Guard clause test coverage gap

The two new validation paths in `BrowserWebAppFactory` (unknown mode, null builder for scripted mode) have no direct tests. Low priority since it's test infrastructure.

## Verdict: APPROVE

All prior findings addressed. No blockers. One MEDIUM suggestion for maintainability, two LOWs as nice-to-haves.

gautam-achieveai:
**[Gautam's review bot]** Both requests addressed in ef85917. The caveat removal is clean and properly tested — `KnownLimitation_TaggedOnBrokenMocks_UnsetOnHealthyOnes` now asserts `claude-mock.KnownLimitation` is `null`. The Playwright test is well-gated with `[SkippableFact]` and exercises the exact failure mode from #29. 

One suggestion: the `ClaudeCliIsAvailable()` helper in the new test reimplements PATH scanning that `FileSystemProbe` already handles (and it's on the classpath). See inline comment for a 4-line alternative.

PR #36, opened 2026-05-09
Fix Codex mock provider routing

### LINKED ISSUE #28: MockProviderHost: codex-mock hangs because /v1/responses route is not implemented

gautam-achieveai:
## Summary
Fixes Codex(Mock) routing so a mock `BaseUrl` is passed to Codex app-server as an explicit OpenAI Responses-compatible custom model provider instead of relying only on `OPENAI_BASE_URL`.

This allows Codex app-server to use `MockProviderHost` for Responses/WebSocket traffic and removes the stale `codex-mock` known-limitation banner from the sample provider registry.

## Changes
- Add explicit `modelProvider` and `config.model_provider/model_providers` wiring when `CodexBridgeInitOptions.BaseUrl` is set.
- Configure the custom provider with `wire_api = responses`, `requires_openai_auth = false`, and `supports_websockets = true`.
- Add regression tests for start/resume app-server config generation.
- Remove the outdated Codex(Mock) limitation message from `LmStreaming.Sample`.

## Testing
- `dotnet format whitespace LmDotnetTools.sln --verify-no-changes`
- `dotnet test tests\CodexSdkProvider.Tests\CodexSdkProvider.Tests.csproj --no-restore`
- `dotnet test tests\LmStreaming.Sample.Tests\LmStreaming.Sample.Tests.csproj --no-restore --filter "FullyQualifiedName~ProviderRegistryTests|FullyQualifiedName~CodexAgentLoopTests"`
- `dotnet test tests\MockProviderHost.Tests\MockProviderHost.Tests.csproj --no-restore --filter "FullyQualifiedName~Responses"`
- `dotnet build LmDotnetTools.sln --no-restore`
- Live smoke: `CodexSdkClient` + `MockProviderHost` consumed the scripted mock turn (`remaining_turns=0`) and emitted visible mock text.

## Related Issues
Fixes #28

### DIFF
--- a/tests/CodexSdkProvider.Tests/Agents/CodexAppServerRequestConfigTests.cs
+++ b/tests/CodexSdkProvider.Tests/Agents/CodexAppServerRequestConfigTests.cs
@@ -1,0 +1,93 @@
+using System.Reflection;
+using AchieveAi.LmDotnetTools.CodexSdkProvider.Agents;
+using AchieveAi.LmDotnetTools.CodexSdkProvider.Configuration;
+using AchieveAi.LmDotnetTools.CodexSdkProvider.Models;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace AchieveAi.LmDotnetTools.CodexSdkProvider.Tests.Agents;
+
+public sealed class CodexAppServerRequestConfigTests
+{
+    private const string ProviderId = "lm-dotnet-tools-openai-responses";
+
+    [Theory]
+    [InlineData("BuildThreadStartParams")]
+    [InlineData("BuildThreadResumeParams")]
+    public void BuildThreadParams_BaseUrlSet_RoutesThroughExplicitResponsesModelProvider(string methodName)
+    {
+        var options = new CodexBridgeInitOptions
+        {
+            Model = "gpt-5.3-codex",
+            ApprovalPolicy = "never",
+            SandboxMode = "read-only",
+            WorkingDirectory = @"B:\sources\LmDotnetTools",
+            BaseUrl = "http://127.0.0.1:5099/v1",
+            ApiKey = "mock-token",
+            ThreadId = "thread-1",
+        };
+
+        var parameters = InvokeThreadParams(methodName, options);
+
+        parameters["modelProvider"].Should().Be(ProviderId);
+        var config = parameters["config"].Should().BeAssignableTo<Dictionary<string, object?>>().Subject;
+        config["model_provider"].Should().Be(ProviderId);
+
+        var modelProviders = config["model_providers"]
+            .Should()
+            .BeAssignableTo<Dictionary<string, object?>>()
+            .Subject;
+        var provider = modelProviders[ProviderId]
+            .Should()
+            .BeAssignableTo<Dictionary<string, object?>>()
+            .Subject;
+
+        provider["name"].Should().Be("LmDotnetTools OpenAI Responses");
+        provider["base_url"].Should().Be("http://127.0.0.1:5099/v1");
+        provider["wire_api"].Should().Be("responses");
+        provider["env_key"].Should().Be("OPENAI_API_KEY");
+        provider["requires_openai_auth"].Should().Be(false);
+        provider["supports_websockets"].Should().Be(true);
+        provider.Should().NotContainKey("experimental_bearer_token");
+    }
+
+    [Fact]
+    public void BuildThreadParams_BaseUrlWithoutApiKey_DoesNotRequireEnvKey()
+    {
+        var options = new CodexBridgeInitOptions
+        {
+            Model = "gpt-5.3-codex",
+            BaseUrl = "http://127.0.0.1:5099/v1",
+        };
+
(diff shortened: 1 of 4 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** — Deep review of PR #36

## Context
Fixes #28 — Codex app-server hangs when routed through MockProviderHost because the mock host doesn't implement `/v1/responses`. This PR configures an explicit custom model provider in the Codex app-server config when `BaseUrl` is set, with `wire_api = "responses"`, `supports_websockets = true`, and `requires_openai_auth = false`.

## Review Summary

**Approach is correct and well-executed.** The fix adds two clean helper methods (`ApplyModelProviderOverride` and `AddResponsesModelProviderConfig`) that inject explicit model provider config only when `BaseUrl` is present. Both thread-start and thread-resume paths are handled consistently. The limitation banner cleanup in `ProviderRegistry` is the right follow-through.

### What's good
- Clean early-return guard pattern in both helpers
- `env_key` logic correctly conditional on `ApiKey` presence
- Both start/resume paths tested via `[Theory]`
- Separate test for the no-API-key edge case

### Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | MEDIUM | Missing negative test — no assertion that `modelProvider`/`model_providers` are absent when `BaseUrl` is null. Protects against accidental config leakage into non-mock scenarios. |
| 2 | LOW | Reflection-based `InvokeThreadParams` helper is fragile — signature changes break at runtime, not compile time. Acceptable tradeoff for now. |
| 3 | INFO | `OPENAI_BASE_URL` env var (set in `CodexAppServerTransport`) is now redundant when the explicit model provider config is present. Not a bug, just an observation. |

### Tests
- `CodexAppServerRequestConfigTests`: 3/3 pass
- `ProviderRegistryTests.KnownLimitation_*`: 1/1 pass

**Verdict: APPROVE** — The fix is clean, well-scoped, and correctly resolves the hang. The missing negative test is a suggestion, not a blocker.

tests/CodexSdkProvider.Tests/Agents/CodexAppServerRequestConfigTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            methodName,
+            BindingFlags.Instance | BindingFlags.NonPublic)
+            ?? throw new MissingMethodException(nameof(CodexSdkClient), methodName);
+
+        return method.Invoke(client, [options])
+                .Should()
+                .BeAssignableTo<Dictionary<string, object?>>()
+                .Subject;
+    }
+}
```
gautam-achieveai, line 93: **[Gautam's review bot]** MEDIUM — Consider adding a negative test that verifies when `BaseUrl` is null/empty, neither `modelProvider` is set on the params dictionary nor `model_provider`/`model_providers` appear in the config. This guards against regression where the custom provider config accidentally leaks into production (non-mock) scenarios.

PR #41, opened 2026-05-12
Bump version to 1.0.27 and migrate ModelContextProtocol to 1.3.0 stable

gautam-achieveai:
## Summary

- Bump release version to **1.0.27** (Directory.Build.props). Published all 10 packages to NuGet.org from this branch.
- Migrate `ModelContextProtocol` and `ModelContextProtocol.AspNetCore` from `0.4.1-preview.1` to **1.3.0 stable** across all consuming projects.

## Why

`dotnet pack` was failing for `McpMiddleware` and `McpSampleServer` with **NU5104** (`Warning As Error: A stable release of a package should not have a prerelease dependency`). That blocked the AgentRuntimeProfile release from publishing those two packages. Bumping MCP to a stable line unblocks packing.

## Changes

**Version bump** (`d3b28ad` → cherry-picked as `1a48360`):
- `Directory.Build.props`: PatchVersion 26 → 27.

**MCP migration** (`1e311a4` → cherry-picked as `6ec7b7b`):
- Updated 6 csproj files referencing MCP to `1.3.0`:
  - `src/McpMiddleware`, `src/McpSampleServer`, `src/McpServer.AspNetCore`
  - `tests/McpIntegrationTests`, `tests/McpServer.AspNetCore.Tests`
  - `example/LmConfigUsageExample`
- Bumped transitive dependency floors required by MCP 1.3.0:
  - `Microsoft.Extensions.Logging.Abstractions` 10.0.0 → 10.0.7
  - `Microsoft.Extensions.DependencyInjection.Abstractions` 10.0.0 → 10.0.7
  - `System.Text.Json` 10.0.0 → 10.0.6
- Adapted `src/McpMiddleware/McpClientFunctionProvider.cs` to MCP 1.x breaking change: `ImageContentBlock.Data` is now `ReadOnlyMemory<byte>` (raw decoded bytes) instead of `string` (base64). Decode is now a no-op; re-encode to base64 when producing `ImageToolResultBlock` for the LmCore tool-result contract.

## Verification

- `dotnet build LmDotnetTools.sln -c Release` — **0 errors, 0 warnings**
- `dotnet test` — all publishable-package tests pass:
  - `LmCore.Tests`: 730 passed
  - `ClaudeAgentSdkProvider.Tests`: 64 passed
  - `LmMultiTurn.Tests`: 210 passed
  - `McpIntegrationTests`: 21 passed
- All 10 packages **already published** to NuGet.org as `1.0.27` from this branch.

## Out of scope

`samples/McpServer.AspNetCore.Sample/Tools/ExampleTools.cs` has 4 pre-existing build errors against the `ToolHandler` delegate signature. Verified pre-existing via `git stash`. Not packaged. Should be tracked separately.

## Test plan

- [x] Solution builds clean
- [x] All publishable-package test suites pass
- [x] `dotnet pack` succeeds for all 10 packages at 1.0.27
- [x] `dotnet nuget push` succeeds for all 10 packages

### DIFF
--- a/src/McpMiddleware/McpClientFunctionProvider.cs
+++ b/src/McpMiddleware/McpClientFunctionProvider.cs
@@ -689,11 +689,11 @@ public partial class McpClientFunctionProvider : IFunctionProvider
         var imageIndex = 0;
         foreach (var content in response.Content.Where(c => c?.Type == "image"))
         {
-            if (content is ImageContentBlock imgBlock && !string.IsNullOrEmpty(imgBlock.Data))
+            if (content is ImageContentBlock imgBlock && !imgBlock.Data.IsEmpty)
             {
                 try
                 {
-                    var bytes = Convert.FromBase64String(imgBlock.Data);
+                    var bytes = imgBlock.Data.ToArray();
                     var detectedMimeType = DetectImageMimeType(bytes, imgBlock.MimeType, logger);
 
                     if (detectedMimeType != imgBlock.MimeType)
@@ -719,7 +719,7 @@ public partial class McpClientFunctionProvider : IFunctionProvider
                     }
 
                     imageBlocks.Add(
-                        new ImageToolResultBlock { Data = imgBlock.Data, MimeType = detectedMimeType }
+                        new ImageToolResultBlock { Data = Convert.ToBase64String(bytes), MimeType = detectedMimeType }
                     );
                 }
                 catch (FormatException ex)
@@ -729,7 +729,7 @@ public partial class McpClientFunctionProvider : IFunctionProvider
                         "Invalid base64 data in MCP image response: ToolName={ToolName}, ImageIndex={ImageIndex}, DataLength={DataLength}",
                         toolName,
                         imageIndex,
-                        imgBlock.Data?.Length ?? 0
+                        imgBlock.Data.Length
                     );
                 }
                 catch (Exception ex)
--- a/src/McpMiddleware/McpClientFunctionProvider.cs
+++ b/src/McpMiddleware/McpClientFunctionProvider.cs
@@ -693,8 +693,8 @@ public partial class McpClientFunctionProvider : IFunctionProvider
             {
                 try
                 {
-                    var bytes = imgBlock.Data.ToArray();
-                    var detectedMimeType = DetectImageMimeType(bytes, imgBlock.MimeType, logger);
+                    var dataSpan = imgBlock.Data.Span;
+                    var detectedMimeType = DetectImageMimeType(dataSpan, imgBlock.MimeType, logger);
 
                     if (detectedMimeType != imgBlock.MimeType)
                     {
@@ -704,7 +704,7 @@ public partial class McpClientFunctionProvider : IFunctionProvider
                             imageIndex,
                             imgBlock.MimeType,
                             detectedMimeType,
-                            bytes.Length
+                            dataSpan.Length
                         );
                     }
                     else
@@ -714,12 +714,12 @@ public partial class McpClientFunctionProvider : IFunctionProvider
                             toolName,
                             imageIndex,
                             detectedMimeType,
-                            bytes.Length
+                            dataSpan.Length
                         );
                     }
 
                     imageBlocks.Add(
-                        new ImageToolResultBlock { Data = Convert.ToBase64String(bytes), MimeType = detectedMimeType }
+                        new ImageToolResultBlock { Data = Convert.ToBase64String(dataSpan), MimeType = detectedMimeType }
                     );
                 }
                 catch (Exception ex)
@@ -778,7 +778,7 @@ public partial class McpClientFunctionProvider : IFunctionProvider
     ///     Detects the MIME type of an image from its byte content.
     ///     Uses magic bytes to identify common image formats.
     /// </summary>
-    private static string DetectImageMimeType(byte[] bytes, string fallbackMimeType, ILogger logger)
+    private static string DetectImageMimeType(ReadOnlySpan<byte> bytes, string fallbackMimeType, ILogger logger)
     {
         if (bytes.Length >= 8)
         {
--- a/src/McpMiddleware/McpClientFunctionProvider.cs
+++ b/src/McpMiddleware/McpClientFunctionProvider.cs
@@ -722,16 +722,6 @@ public partial class McpClientFunctionProvider : IFunctionProvider
                         new ImageToolResultBlock { Data = Convert.ToBase64String(bytes), MimeType = detectedMimeType }
                     );
                 }
-                catch (FormatException ex)
-                {
-                    logger.LogWarning(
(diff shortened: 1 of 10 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #41 — MCP 0.4.1-preview.1 → 1.3.0 migration + version bump to 1.0.27.

## Review Summary

**Verdict: ✅ APPROVE with comments**

The migration is functionally correct and complete. All 6 direct MCP consumers updated, no stale `0.4.1-preview` references remain, and the `ImageContentBlock.Data` type change (`string` → `ReadOnlyMemory<byte>`) is handled correctly. The PR description is thorough and the "out of scope" note about `ExampleTools.cs` is accurate (confirmed those errors are unrelated to MCP).

### Review Agents Used
| Agent | Key Finding |
|-------|-------------|
| exception-handling-review | Dead `catch (FormatException)` block + misleading log message |
| performance-review | `Span`-based optimization to eliminate `ToArray()` copy |
| architecture-review | Dependency version fragmentation (10.x vs 9.x split) |
| test-coverage-review | Zero test coverage for `ExtractImageBlocks` data flow |

### Findings (3 inline comments below + 2 informational)

**Informational (not inline):**

1. **CHANGELOG.md** — No entry for 1.0.27 despite packages already published to NuGet. Consider adding a changelog entry documenting the MCP migration and version bump.

2. **Dependency version fragmentation** — MCP-touching projects now use `Microsoft.Extensions.*.Abstractions` 10.0.x while 11 other `src/` projects stay on 9.0.5. NuGet will unify at build time, but the declared versions are misleading. Long-term recommendation: adopt Central Package Management (`Directory.Packages.props`) to prevent this class of inconsistency. Not blocking for this PR.

src/McpMiddleware/McpClientFunctionProvider.cs
```diff
@@ -719,7 +719,7 @@ ILogger<McpClientFunctionProvider> logger
                     }
 
                     imageBlocks.Add(
-                        new ImageToolResultBlock { Data = imgBlock.Data, MimeType = detectedMimeType }
+                        new ImageToolResultBlock { Data = Convert.ToBase64String(bytes), MimeType = detectedMimeType }
                     );
                 }
                 catch (FormatException ex)
```
gautam-achieveai: **[Gautam's review bot]** [MEDIUM] **Dead code: `catch (FormatException)` is now unreachable**

This catch block was meaningful when the try block contained `Convert.FromBase64String(imgBlock.Data)`. After the migration to MCP 1.3.0:

| Operation in try block | Can throw `FormatException`? |
|---|---|
| `imgBlock.Data.ToArray()` | No — copies `ReadOnlyMemory<byte>` to `byte[]` |
| `DetectImageMimeType(bytes, ...)` | No — pure byte-header inspection |
| `Convert.ToBase64String(bytes)` | No — **encoding** never fails (only decoding does) |

This catch is dead code. The log message also says "Invalid base64 data" but the data is now raw bytes, which is misleading for debugging.

**Suggestion:** Remove this `catch (FormatException)` block entirely — the generic `catch (Exception ex)` on line 735 already handles any unexpected failure.

src/McpMiddleware/McpClientFunctionProvider.cs
```diff
@@ -689,11 +689,11 @@ ILogger<McpClientFunctionProvider> logger
         var imageIndex = 0;
         foreach (var content in response.Content.Where(c => c?.Type == "image"))
         {
-            if (content is ImageContentBlock imgBlock && !string.IsNullOrEmpty(imgBlock.Data))
+            if (content is ImageContentBlock imgBlock && !imgBlock.Data.IsEmpty)
             {
                 try
                 {
-                    var bytes = Convert.FromBase64String(imgBlock.Data);
+                    var bytes = imgBlock.Data.ToArray();
```
gautam-achieveai: **[Gautam's review bot]** [LOW] **Optimization opportunity: avoid `ToArray()` copy using `Span`**

`imgBlock.Data.ToArray()` allocates a full copy of the image bytes. Since the project targets .NET 9, both `Convert.ToBase64String(ReadOnlySpan<byte>)` and indexer-based `DetectImageMimeType` work with `ReadOnlySpan<byte>` directly:

`csharp
var detectedMimeType = DetectImageMimeType(imgBlock.Data.Span, imgBlock.MimeType, logger);
// ...
new ImageToolResultBlock { Data = Convert.ToBase64String(imgBlock.Data.Span), MimeType = detectedMimeType }
`

This eliminates one full image-sized heap allocation per tool call (5MB saved for a 5MB image). Not blocking — just a nice cleanup if you're already touching this code.

*(Would also require changing `DetectImageMimeType` to accept `ReadOnlySpan<byte>` instead of `byte[]` — the method body already uses only indexer access, so it's a signature-only change.)*

src/McpMiddleware/McpClientFunctionProvider.cs
```diff
@@ -689,11 +689,11 @@ ILogger<McpClientFunctionProvider> logger
         var imageIndex = 0;
         foreach (var content in response.Content.Where(c => c?.Type == "image"))
         {
-            if (content is ImageContentBlock imgBlock && !string.IsNullOrEmpty(imgBlock.Data))
+            if (content is ImageContentBlock imgBlock && !imgBlock.Data.IsEmpty)
```
gautam-achieveai, line 692: **[Gautam's review bot]** [INFO] **No test coverage for `ExtractImageBlocks` data flow**

The image extraction pipeline was fundamentally changed (base64 string → raw bytes → base64 string) with zero test coverage for `McpClientFunctionProvider` anywhere in the codebase. This is pre-existing but worth calling out since the data flow semantics changed.

If the conversion is wrong, all image tool results from MCP servers would silently produce corrupt data. Consider adding at least a round-trip test (raw bytes in → base64 out → decode matches original) as a follow-up.

gautam-achieveai:
**[Gautam's review bot]** ✅ **APPROVE** — Incremental re-review (iteration 4)

Both findings from the previous review have been cleanly addressed:

| Finding | Status | Commit |
|---------|--------|--------|
| F1: Dead `catch (FormatException)` | ✅ Removed | `9be0a28` |
| F2: `ToArray()` → `Span` optimization | ✅ Applied | `ddd617b` |

**F2 implementation notes:**
- `imgBlock.Data.Span` used directly — no heap allocation per image
- `DetectImageMimeType` signature correctly changed to `ReadOnlySpan<byte>` (method body only uses indexer access, so this is a clean fit)
- `Convert.ToBase64String(ReadOnlySpan<byte>)` is available on the .NET 9 target
- Span doesn't escape the method scope — safe usage

**Remaining non-blocking items** (tracked, not gating):
- F3: Test coverage for `ExtractImageBlocks` (pre-existing gap)
- F6: CHANGELOG entry for 1.0.27

PR #45, opened 2026-05-12
Harden SystemInitEvent skills/agents against schema drift (#44)

### LINKED ISSUE #44: JsonlStreamParser: harden Skills and Agents against same schema-drift pattern as Plugins (#42)
[Gautam's Dev bot]

## Context

#42 / PR #43 fixed `JsonlStreamParser` after `claude-agent-sdk` CLI v2.0.55+ changed `system.init.plugins` from `string[]` to `[{name, path}]`. The mis-typed field caused the entire `system.init` frame to be silently dropped (including `SessionId`/`Model`), breaking the `--resume` flow.

## Risk

In `src/ClaudeAgentSdkProvider/Models/JsonlEvents/SystemInitEvent.cs`, two sibling fields are still typed as `List<string>?`:

- `Agents` (line 45)
- `Skills` (line 48)

If upstream evolves either of these to an object array (mirroring the Plugins change), `JsonlStreamParser` will exhibit the same silent-drop failure mode.

## Options

1. **Reactive** (current strategy for Plugins): wait until the CLI changes, then re-type as a record. Low effort, but each drift causes a silent production outage until detected.
2. **Proactive defensive typing**: introduce a polymorphic `JsonConverter` (prior art: `src/LmCore/Utils/Union.cs`) that accepts both `string` and `{name, ...}` shapes, applied to `Plugins`, `Skills`, `Agents`.
3. **Monitor**: add an integration test that pins against the latest CLI binary and surfaces drift in CI before it hits production.

## Recommendation

Option 2 (polymorphic converter) is the smallest defensive change that eliminates the silent-drop failure mode across all three fields. Option 3 is complementary if a real CLI binary is available in CI.

## Tracking

Follow-up to #42 review feedback on PR #43.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Hardens `SystemInitEvent.skills` / `.agents` against the same `system.init` schema-drift class as `.plugins` (#42 / PR #43). With the prior PR, only `plugins` was protected; the sibling fields stayed `List<string>?` and would silently drop the entire `system.init` frame (including `SessionId`/`Model`, breaking `--resume`) if the CLI ever evolved them to `[{name, ...}]`.

This PR applies the defensive typing strategy across all three fields:

- Replace `PluginRef` with a single tolerant `NamedRef` record that captures unknown fields in `Extra` via `[JsonExtensionData]`, future-proofing against further drift (`version`, `source`, …).
- Add `NamedRefJsonConverter` that accepts either a JSON string or `{name, path, ...}` object **per list element**, so a single drifted item no longer drops the whole frame. Tokens outside the tolerated envelope (numbers, arrays) still surface as `JsonException`, which `ParseLine` catches and logs.
- Re-type `SystemInitEvent.Agents` / `.Skills` / `.Plugins` to `List<NamedRef>?`.

No public API of `ClaudeAgentSdkClient` changes; `PluginRef` had no in-repo consumers outside `JsonlStreamParser.Tests`, which only dereferenced `.Name` / `.Path` (both still present).

## Test plan

- [x] New `JsonlStreamParser.Tests` cases mirror the PR #43 plugin matrix for both Skills and Agents: string-only, object-only, mixed string/object, empty.
- [x] `ParseLine_SystemInitWithObjectSkills_CapturesUnknownFieldsInExtra` pins extension-data behaviour against `{name, version, source}`.
- [x] `ParseLine_SystemInitWithMalformedSkillItem_FailsLoudly` asserts drift outside the tolerated envelope (`[123]`) still surfaces as a parse failure rather than silently passing.
- [x] Existing PR #43 plugin tests continue to pass (rename only).
- [x] Full solution build succeeds with `TreatWarningsAsErrors`.
- [x] `dotnet format --verify-no-changes` is clean.

Closes #44

### DIFF
--- a/src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
+++ b/src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
@@ -16,20 +16,30 @@ namespace AchieveAi.LmDotnetTools.ClaudeAgentSdkProvider.Models.JsonlEvents;
 ///         <item><see cref="JsonTokenType.String"/> → <c>new NamedRef { Name = value }</c>.</item>
 ///         <item><see cref="JsonTokenType.StartObject"/> → deserialize as <see cref="NamedRef"/>
 ///             (including <see cref="NamedRef.Extra"/> overflow).</item>
-///         <item>Anything else → throw <see cref="JsonException"/>. <c>ParseLine</c>
-///             catches this and logs a warning, surfacing drift outside the tolerated
-///             envelope rather than silently swallowing it.</item>
+///         <item>Anything else, including a literal JSON <c>null</c> element, → throw
+///             <see cref="JsonException"/>. <c>ParseLine</c> catches this and logs a
+///             warning, surfacing drift outside the tolerated envelope rather than
+///             silently passing an element into a non-nullable list slot.</item>
 ///     </list>
 /// </summary>
 public sealed class NamedRefJsonConverter : JsonConverter<NamedRef>
 {
+    // Force STJ to route literal JSON `null` array elements through Read() instead
+    // of short-circuiting and inserting a raw null into List<NamedRef>; the Read()
+    // null-token branch then throws JsonException, matching the rest of the
+    // out-of-envelope handling.
+    public override bool HandleNull => true;
+
     public override NamedRef Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
         return reader.TokenType switch
         {
             JsonTokenType.String => new NamedRef { Name = reader.GetString() },
             JsonTokenType.StartObject => ReadObject(ref reader),
-            JsonTokenType.Null => null!,
+            // Explicitly reject null array elements: List<NamedRef> declares a
+            // non-nullable element type, and silently letting null sneak into
+            // the list would reintroduce the same NRE-shaped failure mode this
+            // converter exists to prevent (consumers iterating `.Name` on null).
             _ => throw new JsonException(
                 $"Unexpected token '{reader.TokenType}' when reading NamedRef; expected String or StartObject."),
         };
@@ -87,7 +97,11 @@ public sealed class NamedRefJsonConverter : JsonConverter<NamedRef>
             }
 
             var propertyName = reader.GetString();
-            _ = reader.Read();
+            if (!reader.Read())
+            {
+                throw new JsonException(
+                    $"Unexpected end of JSON after property '{propertyName}' inside NamedRef object.");
+            }
 
             switch (propertyName)
             {
--- a/src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
+++ b/src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
@@ -1,0 +1,123 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace AchieveAi.LmDotnetTools.ClaudeAgentSdkProvider.Models.JsonlEvents;
+
+/// <summary>
+///     Tolerant per-item converter for <see cref="NamedRef"/>.
+///     <para>
+///         The claude-agent-sdk CLI has shipped <c>system.init.plugins</c> as both
+///         <c>string[]</c> and <c>[{name, path}]</c> (see #42 / PR #43). The sibling
+///         <c>skills</c>/<c>agents</c> fields are at risk of the same drift.
+///         This converter accepts either shape per-item so that a single drifted
+///         element no longer drops the entire <c>system.init</c> frame.
+///     </para>
+///     <list type="bullet">
+///         <item><see cref="JsonTokenType.String"/> → <c>new NamedRef { Name = value }</c>.</item>
+///         <item><see cref="JsonTokenType.StartObject"/> → deserialize as <see cref="NamedRef"/>
+///             (including <see cref="NamedRef.Extra"/> overflow).</item>
+///         <item>Anything else → throw <see cref="JsonException"/>. <c>ParseLine</c>
+///             catches this and logs a warning, surfacing drift outside the tolerated
+///             envelope rather than silently swallowing it.</item>
+///     </list>
+/// </summary>
+public sealed class NamedRefJsonConverter : JsonConverter<NamedRef>
+{
+    public override NamedRef Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return reader.TokenType switch
+        {
+            JsonTokenType.String => new NamedRef { Name = reader.GetString() },
+            JsonTokenType.StartObject => ReadObject(ref reader),
+            JsonTokenType.Null => null!,
+            _ => throw new JsonException(
+                $"Unexpected token '{reader.TokenType}' when reading NamedRef; expected String or StartObject."),
+        };
+    }
+
+    public override void Write(Utf8JsonWriter writer, NamedRef value, JsonSerializerOptions options)
+    {
+        ArgumentNullException.ThrowIfNull(writer);
+        ArgumentNullException.ThrowIfNull(value);
+
+        writer.WriteStartObject();
+        if (value.Name != null)
+        {
+            writer.WriteString("name", value.Name);
+        }
+
+        if (value.Path != null)
+        {
+            writer.WriteString("path", value.Path);
+        }
+
+        if (value.Extra != null)
+        {
+            foreach (var kvp in value.Extra)
+            {
+                writer.WritePropertyName(kvp.Key);
+                kvp.Value.WriteTo(writer);
+            }
+        }
+
+        writer.WriteEndObject();
+    }
+
+    private static NamedRef ReadObject(ref Utf8JsonReader reader)
+    {
+        string? name = null;
+        string? path = null;
+        Dictionary<string, JsonElement>? extra = null;
+
+        while (reader.Read())
+        {
+            if (reader.TokenType == JsonTokenType.EndObject)
+            {
+                return new NamedRef
+                {
+                    Name = name,
+                    Path = path,
+                    Extra = extra,
+                };
+            }
+
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException($"Unexpected token '{reader.TokenType}' inside NamedRef object.");
+            }
+
+            var propertyName = reader.GetString();
+            _ = reader.Read();
+
+            switch (propertyName)
+            {
+                case "name":
+                    name = ReadOptionalString(ref reader, propertyName);
+                    break;
+                case "path":
+                    path = ReadOptionalString(ref reader, propertyName);
+                    break;
+                default:
+                    extra ??= [];
+                    using (var doc = JsonDocument.ParseValue(ref reader))
+                    {
+                        extra[propertyName!] = doc.RootElement.Clone();
+                    }
+                    break;
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #45 — Harden SystemInitEvent skills/agents against schema drift (#44)

## Context
Follow-up to #42 / PR #43. That fix addressed `system.init.plugins` changing from `string[]` to `[{name, path}]`, which silently dropped the entire `system.init` frame (including `SessionId`). This PR proactively applies the same defensive typing to the sibling `skills` and `agents` fields.

## What the PR delivers
- Unified `NamedRef` record replacing `PluginRef`, with `[JsonExtensionData]` overflow for forward-compat
- `NamedRefJsonConverter` accepting both `string` and `{name, ...}` shapes per-item
- Re-typed `Agents`/`Skills`/`Plugins` as `List<NamedRef>?`
- 11 new tests covering string/object/mixed/empty/extra/malformed shapes

## Build & Tests
✅ Build: 0 warnings, 0 errors (changed project)
✅ Tests: 39 passed, 0 failed

## Review Agents Used
- architecture-review ✅ — No issues
- exception-handling-review — 2 findings
- test-coverage-review — 3 findings
- class-design-simplifier — 2 info items
- review-grader — Escalated 1 finding to HIGH

## Findings Summary

| # | Severity | Finding | Blocker |
|---|----------|---------|---------|
| 1 | **HIGH** | `null!` return inserts null into non-nullable `List<NamedRef>`, bypasses ParseLine safety net | ✅ |
| 2 | MEDIUM | Unchecked `reader.Read()` return value — misleading error on truncated JSON | |
| 3 | MEDIUM | Missing string-form plugins test (asymmetric with skills/agents coverage) | |
| 4 | MEDIUM | Write/serialization path of `NamedRefJsonConverter` completely untested | |
| 5 | MEDIUM | null-in-array behavior not tested (companion to Finding 1) | |
| 6 | Info | `[JsonExtensionData]` attribute inert when custom converter active | |
| 7 | Info | `ReadObject` could be simplified with `JsonDocument.ParseValue` | |

## Verdict: REQUEST CHANGES
The blocker (Finding 1) is a one-line fix: change `null!` to `throw new JsonException(...)`. The PR's own documentation says unexpected tokens should throw — the null case should follow that contract. Everything else is solid work.

## Positive Observations
- Architecture is clean — correct layer placement, complete `PluginRef` removal, no stale references
- Exception propagation design (converter throws → ParseLine catches/logs → returns null) is well-structured
- Test quality is high with thorough shape coverage
- XML documentation with cross-references to #42/PR #43 is excellent
- `Extra` field for forward-compatibility is a smart design choice

src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
```diff
@@ -0,0 +1,123 @@
...
+/// </summary>
+public sealed class NamedRefJsonConverter : JsonConverter<NamedRef>
+{
+    public override NamedRef Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return reader.TokenType switch
+        {
+            JsonTokenType.String => new NamedRef { Name = reader.GetString() },
+            JsonTokenType.StartObject => ReadObject(ref reader),
+            JsonTokenType.Null => null!,
```
gautam-achieveai: **[Gautam's review bot]** 🔴 **HIGH (Blocker)** — `null!` silently inserts null into non-nullable `List<NamedRef>`

This creates exactly the kind of silent failure the PR aims to prevent:

1. If the CLI emits `"skills": [null, "code-review"]`, a `null` sneaks into `List<NamedRef>` (non-nullable element type)
2. Any consumer doing `foreach (var s in skills) { Log(s.Name); }` gets `NullReferenceException`
3. **NRE is NOT a `JsonException`** — it bypasses the `ParseLine` safety net (line 44 of `JsonlStreamParser.cs` only catches `JsonException`)
4. The converter's own XML doc (lines 19-21) says: *"Anything else → throw JsonException, surfacing drift rather than silently swallowing it."* The null case directly violates this contract.

**Fix (one-liner):**
```csharp
JsonTokenType.Null => throw new JsonException(
    "Unexpected null element in NamedRef array; expected String or StartObject."),
```

This makes the null case consistent with the default arm two lines below.

src/ClaudeAgentSdkProvider/Models/JsonlEvents/NamedRefJsonConverter.cs
```diff
@@ -0,0 +1,123 @@
...
+                };
+            }
+
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException($"Unexpected token '{reader.TokenType}' inside NamedRef object.");
+            }
+
+            var propertyName = reader.GetString();
+            _ = reader.Read();
```
gautam-achieveai: **[Gautam's review bot]** 🟡 **MEDIUM** — Unchecked `reader.Read()` return value

`Utf8JsonReader.Read()` returns `false` when the reader reaches end of payload. By discarding the result, if JSON is truncated after a property name, `reader.TokenType` remains `PropertyName` and the switch falls through with a misleading error message (e.g. "got PropertyName") instead of the intended "Unexpected end of JSON" on line 110.

**Fix:**
```csharp
if (!reader.Read())
{
    throw new JsonException("Unexpected end of JSON while reading NamedRef.");
}
```

tests/ClaudeAgentSdkProvider.Tests/Parsers/JsonlStreamParser.Tests.cs
```diff
@@ -552,6 +552,253 @@ public void ParseLine_SystemInitWithPluginObjects_PreservesSessionIdForResume()
         Assert.Equal(expectedSessionId, initEvent.SessionId);
     }
 
+    // ---- #44 schema-drift tolerance for skills/agents (mirrors PR #43 plugin coverage) ----
```
gautam-achieveai, line 555: **[Gautam's review bot]** 🟡 **MEDIUM** — Missing string-form plugins test

Skills and agents each have 4 shape tests (string/object/mixed/empty). Plugins — the original drift case from #42 — only has object and empty tests. No test for `"plugins": ["code-search", "shell-runner"]` (string form).

Since `Plugins` was re-typed from `List<PluginRef>` to `List<NamedRef>` in this PR, a string-form test proves the original drift case is tolerated bidirectionally. Without it, you have a coverage gap at the field that motivated the entire feature.

**Suggested:** Add `ParseLine_SystemInitWithStringPlugins_DeserializesPluginsAndSessionId` mirroring the skills/agents string tests.

Also, the `Write()` method of `NamedRefJsonConverter` (25 lines, 3 conditional branches) has no test at all — consider adding a round-trip test.

src/ClaudeAgentSdkProvider/Models/JsonlEvents/SystemInitEvent.cs
```diff
@@ -67,13 +68,23 @@ public record McpServerStatus
...
+public record NamedRef
 {
     [JsonPropertyName("name")]
     public string? Name { get; init; }
 
     [JsonPropertyName("path")]
     public string? Path { get; init; }
+
+    [JsonExtensionData]
+    public Dictionary<string, JsonElement>? Extra { get; init; }
```
gautam-achieveai, line 93: **[Gautam's review bot]** ℹ️ **INFO** — `[JsonExtensionData]` attribute is inert here

`[JsonExtensionData]` is never consulted at runtime because `NamedRefJsonConverter` (via `[JsonConverter]` on line 79) takes full control of serialization. The overflow capture is actually handled by the converter's `ReadObject` default branch (lines 100-105) and the `Write` loop (lines 56-61).

The attribute serves as useful documentation of intent and would become functional if the converter were ever removed — so keeping it is reasonable. Consider a brief comment to prevent confusion:
```csharp
// Note: Overflow is handled by NamedRefJsonConverter at runtime.
// Attribute retained as documentation of the contract.
[JsonExtensionData]
public Dictionary<string, JsonElement>? Extra { get; init; }
```

gautam-achieveai:
[Gautam's Dev bot] Iteration 2 — addressed review-bot findings on PR #45.

**HIGH**
- `NamedRefJsonConverter.Read` no longer returns `null!` for `JsonTokenType.Null`. The null-token branch is removed so it falls through to the default arm and throws `JsonException`. `HandleNull => true` is overridden so STJ routes literal-null array elements through `Read()` instead of short-circuiting and inserting raw `null` into `List<NamedRef>`. Covered by new test `ParseLine_SystemInitWithNullSkillItem_FailsLoudly`.

**MEDIUM**
- `ReadObject` now checks the return value of `reader.Read()` after consuming the `PropertyName` token and throws `JsonException` on premature end-of-payload, instead of silently advancing into an invalid token.
- Added `ParseLine_SystemInitWithStringPlugins_DeserializesPluginsAndSessionId` to cover the pre-existing `plugins: ["string", ...]` shape end-to-end through `ParseLine`, not just via the converter directly.
- Added Write-side tests: `NamedRefJsonConverter_WritesObjectShape_WithNameAndPath`, `NamedRefJsonConverter_WritesExtraFieldsAlongsideNameAndPath`, and `NamedRefJsonConverter_RoundTrip_PreservesNamePathAndExtra`.

**INFO**
- Removed the inert `[JsonExtensionData]` attribute on `NamedRef.Extra`. STJ never consults extension-data attributes once a custom `JsonConverter` is registered, so the attribute was misleading. Added a comment on the property documenting that `Extra` is populated/emitted by the converter's overflow path.
- Deferred: simplifying `ReadObject` to use `JsonSerializer.Deserialize` for the object case. The hand-written loop preserves precise error messages per property and the path is small enough that the indirection isn't worth the churn.

**Verification**
- `dotnet build LmDotnetTools.sln` — 0 warnings, 0 errors.
- `dotnet test ... --filter "FullyQualifiedName~Parsers"` — 31 passed, 0 failed.
- `dotnet format ... --verify-no-changes` — clean.

Pushed as `250846e` to `wi-44-skills-agents-tolerant`.

gautam-achieveai:

| Service | URL | Status |
|---|---|---|
| lmdotnettools | http://pr-45.lmdotnettools.localhost/ | ✅ Healthy |

gautam-achieveai (commented):
**[Gautam's review bot]** ✅ **APPROVE** — Iteration 2 resolves all findings from the initial review.

### Findings Status

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 HIGH | `null!` inserts null into non-nullable list | ✅ Fixed — `HandleNull => true` + removed Null case |
| 2 | 🟡 MEDIUM | Unchecked `reader.Read()` return | ✅ Fixed — throws `JsonException` with property name |
| 3 | 🟡 MEDIUM | Missing string-form plugins test | ✅ Added |
| 4 | 🟡 MEDIUM | Write/serialization path untested | ✅ 3 tests added (write, write+extra, round-trip) |
| 5 | 🟡 MEDIUM | Null-in-array behavior untested | ✅ `FailsLoudly` test added |
| 6 | ℹ️ INFO | Inert `[JsonExtensionData]` | ✅ Removed, documented |
| 7 | ℹ️ INFO | `ReadObject` simplification | ⏭️ Deferred (reasonable) |

### Notes

- The `HandleNull => true` override is the correct STJ mechanism — without it, STJ bypasses the converter entirely for null array elements and silently inserts `default(NamedRef)` into the list. Good catch and good fix.
- The `reader.Read()` check now includes the property name in the error message, which aids debugging truncated payloads.
- Comment on `Extra` explaining the deliberate `[JsonExtensionData]` omission is clear and prevents future confusion.

### Verification
- Build: 0 warnings, 0 errors
- Tests: 26 parser tests passed, 0 failed

**Linked:** #44

PR #53, opened 2026-05-13
fix(LmMultiTurn): surface unmapped Copilot sessionUpdate kinds at Information

### LINKED ISSUE #52: CopilotEventTranslator drops post-tool answer — silent loss of model output when sessionUpdate.kind is unmapped
## What's broken

When `CopilotAgentLoop` runs a turn that includes tool calls, the agent's final answer (everything the model produces after the last tool result) **never reaches the IMessage stream**. The host sees only the agent's pre-tool opening narration as `TextUpdateMessage` chunks + a consolidated `TextMessage`, then tool-call / tool-result, then nothing of substance until `RunCompletedMessage`. The model's actual answer is silently dropped somewhere between the Copilot ACP server and the host SDK.

## Concrete repro

[revobot](https://github.com/gautam-achieveai/revobot)'s `probe-run-review --backend copilot` against `achieveai/LmDotnetTools#46`:

**Prompt:**
> Review the diff of PR achieveai/LmDotnetTools#46 using gh CLI. Write a one-paragraph summary and a verdict line. Output just the markdown, no fences.

**Observed message-kind sequence (per the orchestrator's `transcript.jsonl`):**
1. `RunAssignmentMessage`
2 – 28. `TextUpdateMessage` (~27 chunks) — the agent's *opening narration*: "I'll fetch the PR diff with `gh`, capture review notes... then return only the requested markdown verdict."
29 – 30. `ReasoningUpdateMessage`
31. `ToolCallMessage`
32. `ToolCallResultMessage`
33. `TextMessage` — **consolidated form of the *opening*, not a new burst**
34. `ReasoningMessage`
35. `UsageMessage`
36. `RunCompletedMessage` (IsError=false)

In a second run with the same prompt + a more permissive system prompt: 42 messages, 10 tool calls, **still 119-byte artifact** containing just the opening narration. `RunCompletedMessage.IsError == false` and the bridge reports `copilot.bridge.run.completed completed` and `copilot.turn.latency completed`. From the SDK's perspective everything succeeded.

Run-level telemetry:
```
runId=5ef991fdce19 turns=1 messages=42 toolCalls=10
promptTokens=0 completionTokens=0
kindsObserved=RunAssignmentMessage,TextUpdateMessage,ReasoningUpdateMessage,
              ToolCallMessage,ToolCallResultMessage,TextMessage,
              ReasoningMessage,UsageMessage,RunCompletedMessage
IsError=false  durationMs=31764
```

After the `ToolCallResultMessage` (event 32), **zero `TextUpdateMessage` events arrive** — even though `completionTokens=0` rules out "the model emitted text we missed" and the agent's stated intent ("then return only the requested markdown verdict") indicates a post-tool answer was planned.

## Suspected root cause

The translator at [src/LmMultiTurn/CopilotEventTranslator.cs:45-110](src/LmMultiTurn/CopilotEventTranslator.cs#L45-L110) switches on a small set of `sessionUpdate` kinds:

```csharp
switch (kind)
{
    case "agent_message_chunk":  return ConvertAgentMessageChunk(...);
    case "agent_thought_chunk":  return ConvertAgentThoughtChunk(...);
    case "tool_call":            return ConvertToolCall(...);
    case "tool_call_update":     return ConvertToolCallUpdate(...);
    case "plan":                 return ConvertPlan(...);
    default:
        _logger?.LogDebug(  // ← unmapped kinds are silently dropped at Debug only
            "{event_type} {event_status} {provider} {provider_mode} {session_update_kind}",
            "copilot.session_update.unmapped", "ignored", ..., kind);
        return [];
}
```

If Copilot CLI 1.0.45 (current GA) emits the post-tool answer under a `sessionUpdate.kind` value not in the switch (e.g. `agent_message`, `final_response`, `message_complete`, `agent_response`, or similar), it falls into the `default` arm and gets **silently dropped at Debug level**. Downstream subscribers (and `BuildTerminalMessages`, which only reads from `_agentMessageAccumulator`) never see those tokens.

The post-tool consolidated `TextMessage` we DO observe is the SAME opening text — that's the `BuildTerminalMessages` flush at `session/prompt/completed` ([line 342-362](src/LmMultiTurn/CopilotEventTranslator.cs#L342-L362)) emitting the accumulator's *remaining* contents from before the tool call. Line 361 then `_agentMessageAccumulator.Remove("agent_message")`, so if the same accumulator were reused for a post-tool burst, that data would also be lost on the next terminal — but with the current switch, no second burst ever lands in the accumulator anyway.

## Why I'm not certain it's just the default-arm drop

Two alternative hypotheses I can't rule out without more visibility:

1. **Copilot agent self-terminates after tool calls.** Maybe Copilot CLI 1.0.45's planner decides "I've seen the diff, my job is done" and emits nothing. But `completionTokens=0` is also consistent with "we never counted them" — the `UsageMessage` shape may not surface model-side completion tokens.

2. **ACP `session/prompt/completed` fires before the model finishes streaming.** A timing race where the bridge ships the terminal envelope before the post-tool token stream catches up.

Both would have the same external symptom. The fastest way to disambiguate is to **promote the default-arm log line to Information** so operators see every unmapped kind in production.

### Tier 1 (5-line diagnostic; should ship regardless)

Promote the default-arm Debug log to Information and include the *event_status* `dropped` (not `ignored`) so downstream operators can spot this immediately:

```csharp
default:
    _logger?.LogInformation(   // ← was LogDebug
        "{event_type} {event_status} {provider} {provider_mode} {session_update_kind}",
        "copilot.session_update.unmapped",
        "dropped",
        _options.Provider, _options.ProviderMode, kind);
    return [];
```

Even if it turns out Copilot is sending well-known kinds we just don't handle, the operator-facing signal becomes immediate.

### Tier 2 (when Tier 1 reveals the actual unmapped kind)

Add a case for whatever `kind` value the diagnostic surfaces. Likely candidates from looking at the Copilot ACP protocol docs / source:

- `agent_message` (non-chunk, complete message)
- `final_response`
- `message_complete`

The fix is mechanical — same shape as `ConvertAgentMessageChunk` but with `accumulatedAgentText = update.Text` (full text, not delta) and immediate consolidation.

### Tier 3 (medium effort)

If it turns out the post-tool answer arrives via the existing `agent_message_chunk` kind but AFTER `_agentMessageAccumulator.Remove("agent_message")` ran on a prior terminal flush, the accumulator needs to handle "burst re-open" — re-creating the entry when a chunk arrives after a clear, with the same `runId`/`generationId`.

## Concrete consumer impact

revobot Phase 5 + the Copilot backend (revobot commit `924c6bd`) shipped a silent-failure guard (`HostModeReviewProbe.HasSuspectThinTurn`) specifically to catch this — `outcome=Succeeded` runs now log a Warning when `ToolInvocationCount>0 && OutputLength<256`. That's a downstream Band-Aid; the upstream fix unblocks **every consumer of `CopilotAgentLoop` that uses tool-augmented reviews**, not just revobot.

## Related upstream issues

- #47 — pre-seeding `CurrentSessionId` for `--resume` (re-review)
- #48 — `WasForked` always false on Claude (forked-run pattern)
- #49 — pluggable process transport for Docker isolation

All four are independent but together unblock revobot Phase 6 (event-driven re-review) and Phase 4d (containerization).

## What I'd need to confirm root cause

A run with `REVOBOT_LOG_LEVEL=Verbose` against the same prompt, with the `copilot.session_update.unmapped` log lines visible. Happy to capture and paste here — just let me know if Tier 1 lands as a quick patch and I'll re-run and post the actual `kind` values.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Closes #52.

Tier 1 fix for the silent post-tool answer loss reported in #52: `CopilotEventTranslator`'s default switch arm logged unmapped `sessionUpdate.kind` values at `Debug` level with `event_status="ignored"`, hiding the signal in default-filtered logs and making the silent-drop hypothesis hard to disambiguate from a Copilot-side termination or an ACP race.

This PR converts that branch to `Information` level with `event_status="dropped"` so operators see the line under default filters when a new ACP `sessionUpdate.kind` shows up. No behavioural change beyond log level and status token — return shape (`[]`) and structured field names are identical, so downstream log consumers keying off `event_type=copilot.session_update.unmapped` keep working.

Tier 2 (adding case arms for `agent_message` / `final_response` / `message_complete`) is deferred to a follow-up issue once a verbose Information-level run captures the actual kind value(s) emitted by Copilot CLI 1.0.45+. Speculatively adding case arms risks miscategorising a full `agent_message` as a chunk delta or stomping accumulator state — a louder regression than the silent drop we are surfacing.

## Changes

- `src/LmMultiTurn/CopilotEventTranslator.cs` — switch default arm: `LogDebug` → `LogInformation`, `"ignored"` → `"dropped"`.
- `tests/LmMultiTurn.Tests/CopilotEventTranslatorUnmappedKindTests.cs` — new regression test that constructs the (internal) translator with a captured `ILogger`, feeds a `session/update` envelope carrying `sessionUpdate: "some_unknown_kind"`, and asserts:
  - Translator returns `[]` (no messages emitted).
  - Exactly one `Information`-level log entry fires for `event_type=copilot.session_update.unmapped`.
  - Structured properties carry `event_status=dropped`, `provider=copilot`, `provider_mode=copilot`, `session_update_kind=some_unknown_kind`.

`InternalsVisibleTo("LmMultiTurn.Tests")` was already present, so no csproj changes were needed.

## Out of scope

- No new `sessionUpdate.kind` case arms — Tier 2 follow-up.
- No changes to `BuildTerminalMessages` accumulator clear logic — Tier 3 only applies if Tier 2 shows post-tool tokens arrive after the existing accumulator window.
- No changes to `CopilotAgentLoop` — it remains a pass-through.

## Test plan

- [x] `dotnet build src/LmMultiTurn/AchieveAi.LmDotnetTools.LmMultiTurn.csproj` — clean.
- [x] `dotnet test tests/LmMultiTurn.Tests/LmMultiTurn.Tests.csproj --filter "FullyQualifiedName~CopilotEventTranslatorUnmappedKindTests"` — new test green.
- [x] `dotnet test tests/LmMultiTurn.Tests/LmMultiTurn.Tests.csproj --filter "FullyQualifiedName~Copilot"` — 13/13 existing Copilot translator/loop tests stay green.
- [x] `dotnet test tests/LmMultiTurn.Tests/LmMultiTurn.Tests.csproj` — 233/233 full LmMultiTurn test suite green.

## Follow-up

After this lands, please re-run `revobot probe-run-review --backend copilot` with `REVOBOT_LOG_LEVEL=Information` and post any captured `session_update_kind` values back on #52. Those values will drive the Tier 2 follow-up: `WI-52-followup: handle Copilot ACP <kind> for post-tool answer`. If no unmapped kinds are logged but the answer is still missing, the follow-up pivots to investigating `session/prompt/completed` timing.

### DIFF
--- a/src/LmMultiTurn/CopilotEventTranslator.cs
+++ b/src/LmMultiTurn/CopilotEventTranslator.cs
@@ -98,10 +98,10 @@ internal sealed class CopilotEventTranslator
                 // Plans are informational; surface as a summary reasoning message.
                 return ConvertPlan(update, runId, generationId);
             default:
-                _logger?.LogDebug(
+                _logger?.LogInformation(
                     "{event_type} {event_status} {provider} {provider_mode} {session_update_kind}",
                     "copilot.session_update.unmapped",
-                    "ignored",
+                    "dropped",
                     _options.Provider,
                     _options.ProviderMode,
                     kind);
--- a/tests/LmMultiTurn.Tests/CopilotEventTranslatorUnmappedKindTests.cs
+++ b/tests/LmMultiTurn.Tests/CopilotEventTranslatorUnmappedKindTests.cs
@@ -1,0 +1,89 @@
+using System.Text.Json;
+using AchieveAi.LmDotnetTools.CopilotSdkProvider.Configuration;
+using AchieveAi.LmDotnetTools.LmMultiTurn;
+using FluentAssertions;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace LmMultiTurn.Tests;
+
+public class CopilotEventTranslatorUnmappedKindTests
+{
+    [Fact]
+    public void ConvertEventToMessages_UnmappedSessionUpdateKind_LogsInformationDroppedWithStructuredProps()
+    {
+        var logger = new ListLogger();
+        var options = new CopilotSdkOptions { Provider = "copilot", ProviderMode = "copilot" };
+        var translator = new CopilotEventTranslator(options, logger);
+
+        const string unmappedKind = "some_unknown_kind";
+        var envelope = JsonDocument.Parse($$"""
+            {
+              "type": "session/update",
+              "sessionId": "sess_x",
+              "update": { "sessionUpdate": "{{unmappedKind}}" }
+            }
+            """).RootElement;
+
+        var result = translator.ConvertEventToMessages(envelope, runId: "run_1", generationId: "gen_1");
+
+        result.Should().BeEmpty("unmapped kinds must not produce messages");
+
+        var matching = logger.Entries
+            .Where(e =>
+                e.Level == LogLevel.Information
+                && string.Equals(GetProp(e.State, "event_type"), "copilot.session_update.unmapped"))
+            .ToList();
+
+        matching.Should().HaveCount(1, "exactly one Information-level diagnostic must fire for an unmapped kind");
+
+        var entry = matching[0];
+        GetProp(entry.State, "event_status").Should().Be("dropped");
+        GetProp(entry.State, "provider").Should().Be("copilot");
+        GetProp(entry.State, "provider_mode").Should().Be("copilot");
+        GetProp(entry.State, "session_update_kind").Should().Be(unmappedKind);
+    }
+
+    private static string? GetProp(IReadOnlyList<KeyValuePair<string, object?>> state, string key)
+    {
+        foreach (var kv in state)
+        {
+            if (string.Equals(kv.Key, key, StringComparison.Ordinal))
+            {
+                return kv.Value?.ToString();
+            }
+        }
+
+        return null;
+    }
+
+    private sealed class ListLogger : ILogger
+    {
+        public List<LogEntry> Entries { get; } = [];
+
+        public IDisposable? BeginScope<TState>(TState state) where TState : notnull => null;
+
+        public bool IsEnabled(LogLevel logLevel) => true;
+
+        public void Log<TState>(
+            LogLevel logLevel,
+            EventId eventId,
+            TState state,
+            Exception? exception,
+            Func<TState, Exception?, string> formatter)
+        {
+            var props = new List<KeyValuePair<string, object?>>();
+            if (state is IEnumerable<KeyValuePair<string, object?>> structured)
+            {
+                props.AddRange(structured);
+            }
+
+            Entries.Add(new LogEntry(logLevel, formatter(state, exception), props));
+        }
+    }
+
+    private sealed record LogEntry(
+        LogLevel Level,
+        string Message,
+        IReadOnlyList<KeyValuePair<string, object?>> State);
+}


gautam-achieveai (commented):
**[Gautam's review bot]** ✅ APPROVE — Deep review of iteration 1 (commit b3389f17)

## Summary

Clean, minimal Tier 1 fix for #52. Promotes the default-arm log from `Debug`→`Information` and changes status from `"ignored"`→`"dropped"` so operators can see unmapped `sessionUpdate` kinds under default log filters. Regression test pins the level, status token, and all structured properties.

**Build:** ✅ 0 warnings, 0 errors
**Tests:** ✅ 1/1 passed

## Agents Dispatched
- **code-simplifier** — 3 Info-level suggestions, none blocking
- **test-coverage-review** — excellent coverage, no gaps
- **over-engineering-review** — no over-engineering detected

## Minor Suggestions (non-blocking)

1. **`ContainSingle` pattern** (`CopilotEventTranslatorUnmappedKindTests.cs:32-40`): The `.Where(...).ToList()` + `.Should().HaveCount(1)` + `matching[0]` chain could be replaced with FluentAssertions' `.Should().ContainSingle(predicate).Which` — combines filter, count assertion, and element extraction in one step.

2. **Shared structured logger** (`CopilotEventTranslatorUnmappedKindTests.cs:60-83`): The `ListLogger` is the first capturing logger in the codebase that preserves structured properties (existing ones only capture formatted text). As more diagnostics get promoted, consider extracting to `LmTestUtils` as a shared `StructuredCapturingLogger`. Not needed now — the private sealed pattern matches `ResolveBatchParentTests.CapturingLogger`.

No blocking issues. Ship it. 🚀

PR #54, opened 2026-05-13
[Gautam's Dev bot] Pluggable process transport for CLI agent providers

### LINKED ISSUE #49: Feature: Pluggable process transport for ClaudeAgentSdkClient — enable Docker-isolated agent runs
## What's missing

`ClaudeAgentSdkClient` hard-codes the agent CLI launch as a direct host-process spawn at [src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs:149-197](src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs#L149-L197):

```csharp
var startInfo = new ProcessStartInfo
{
    FileName = nodePath,
    Arguments = $"\"{cliPath}\" {args}",
    RedirectStandardInput = true,
    RedirectStandardOutput = true,
    ...
};
_process = Process.Start(startInfo) ?? throw ...;
```

There's no extension point for a host to redirect the spawn into a containerized environment (`docker run` / `docker exec` / `kubectl exec` / `ssh` / a remote runner). Today the only way to containerize is to either (a) re-implement the entire `ClaudeAgentSdkClient` plus `ClaudeAgentLoop` outside the SDK, or (b) run the entire host orchestrator inside the container.

## Why it matters

**Per-PR isolated execution** is the canonical multi-tenant pattern for an agent runner:
- One container per `(repo, prNumber, agent)` so parallel review/dev runs can't see each other's worktree state
- Network isolation (no Docker socket, no broad home mount)
- Tooling pinned per repo (Node version, language SDKs)
- Credentials scoped via env, never via shared host config

revobot's design doc §3-9 makes this its core invariant. We already shipped Phase 3-5 against the host-mode flow (your SDK works great there), but Phase 4d — putting the agent in a container — has no clean upstream seam to plug into.

### 1. Minimum-viable: `IProcessLauncher` injection on `ClaudeAgentSdkOptions`

```csharp
public interface IProcessLauncher
{
    /// <summary>Launch the underlying CLI process and return its handle.
    /// Default implementation: Process.Start(startInfo). Custom impls
    /// can transform argv/env (e.g. prepend docker run --rm -v ... -e ...),
    /// or attach to an existing container with docker exec.</summary>
    Process Launch(ProcessStartInfo startInfo);
}

public record ClaudeAgentSdkOptions
{
    // ... existing fields ...

    /// <summary>Custom process launcher. Null = default host-mode spawn.</summary>
    public IProcessLauncher? ProcessLauncher { get; init; }
}
```

Then in the client:

```csharp
_process = (_options.ProcessLauncher ?? DefaultProcessLauncher.Instance)
    .Launch(startInfo)
    ?? throw new InvalidOperationException("Failed to start claude-agent-sdk process");
```

This is a single new interface + one-line change at the spawn site. Existing callers see zero behavior change.

### 2. Optional: ship a `DockerProcessLauncher` helper

Even more useful: a built-in `DockerProcessLauncher` that takes the standard mount/env/image configuration and produces a working `docker run` spawn. Then downstream consumers (revobot, others) don't have to rediscover the operational tricks we list below.

```csharp
public class DockerProcessLauncher : IProcessLauncher
{
    public DockerProcessLauncher(DockerLaunchOptions opts) { ... }

    public Process Launch(ProcessStartInfo startInfo)
    {
        // Replace FileName/Arguments with `docker run --rm -v ... -e ... <image> <inner-argv>`
        // Keep stdin/stdout/stderr redirection so the JSONL stream works unchanged.
    }
}
```

## Operational tricks the SDK could expose as helpers

These are things we (revobot) had to figure out by reading the CLI source and shipping bug reports against your repo. Centralizing them upstream means every downstream gets them for free:

### A. The default mount layout for sharing host config with the container

For Claude to load the host's plugin skills, sub-agents, MCP servers, and user settings inside a container, the mounts must be:

| Host path                                    | Container path                       | Mode  | Why |
|----------------------------------------------|--------------------------------------|-------|-----|
| `~/.claude` (settings, agents, skills)       | `/root/.claude`                      | ro    | settings/skills are read-only; bad data here breaks discovery silently |
| `~/.claude/projects` (sessions overlay)      | `/root/.claude/projects`             | rw    | session state writes here; mount AFTER the parent ro mount so last-mount-wins gives write access |
| `~/.claude.json`                             | `/root/.claude.json`                 | ro    | API config; ro to prevent the container from overwriting on host |
| host plugin/skills dir                       | `/root/.claude/plugins`              | ro    | plugin skills + sub-agents (`code-reviewer:*`, etc.) |
| `~/.config/github-copilot`                   | `/root/.config/github-copilot`       | rw    | Copilot SDK auth + session state |
| `~/.codex`                                   | `/root/.codex`                       | rw    | Codex SDK home for parity |
| worktree path                                | `/repo`                              | rw    | checked-out code |
| run artifact dir                             | `/revobot-artifacts`                 | rw    | session.json, manifest.json, etc. |

This table could ship as `ClaudeContainerMounts.Default(homePath, worktreePath, artifactPath)` returning an `IReadOnlyList<MountSpec>`.

### B. Don't pass `--setting-sources ""` (issue #42's sister bug)

Empirically: when `IncludeUserSettings`/`IncludeProjectSettings`/`IncludeLocalSettings` are all true (the default), the CLI must run with `--setting-sources` **omitted** so its own default discovery (user + project + local) loads:
- Plugin skills (`code-reviewer:pr-review`, etc.)
- Plugin sub-agents (`code-reviewer:architecture-review`, etc.)
- Worktree-level `.mcp.json`

Passing `--setting-sources ""` (empty string) silently disables ALL discovery. Your `BuildSettingSources` at [src/LmMultiTurn/ClaudeAgentLoop.cs:1055-1085](src/LmMultiTurn/ClaudeAgentLoop.cs#L1055-L1085) does this correctly today — but it's invisible behavior that's easy to break in a Docker rewrite, and ought to be documented as a contract.

### C. Don't inject "no-op" MCP servers

We had a phase in revobot where we tried to inject a placeholder MCP server to suppress auto-discovery — it actively broke skill loading. The right approach is: pass an empty MCP dict and let the CLI auto-discover from `.mcp.json` in the worktree (which is mounted at `/repo`).

### D. Auth tokens via env, never via mounted files

`GH_TOKEN`, `ADO_TOKEN`, `ANTHROPIC_API_KEY` all go in env vars on the container. Mounting host token files creates a permanent backdoor that survives container exit.

### E. The `claude_code_version: 2.0.55+` `plugins` shape

Already covered by closed issue #42, but worth noting in a containerization doc: when targeting `claude-agent-sdk` ≥ 2.0.55 from inside a container, the host SDK version must be ≥ 1.0.28 to parse the new `plugins: [{name, path}]` shape correctly.

### F. The `host.docker.internal` requirement for credential helpers

Git operations from inside the container need a host-side credential helper (otherwise the container needs PAT files mounted, which we don't want — see D). The pattern: host runs an HTTP service on port `N`, container reaches it via `host.docker.internal:N`, configured via `GIT_CREDENTIAL_HELPER=revobot` + `REVOBOT_CREDHELPER_URL=http://host.docker.internal:<port>/`. This is revobot-specific tooling but the *pattern* could be documented.

## Concrete consumer

[revobot](https://github.com/gautam-achieveai/revobot)'s Phase 4d (containerization) is blocked on the missing extension point. We already shipped:
- `IWorkbenchRuntime` seam ([dotnet/src/RevoBot.Agents/Workbench/IWorkbenchRuntime.cs](https://github.com/gautam-achieveai/revobot/blob/master/dotnet/src/RevoBot.Agents/Workbench/IWorkbenchRuntime.cs))
- `DockerSpawnPlanner` for the host-side spawn argv ([dotnet/src/RevoBot.Agents/Workbench/DockerSpawnPlanner.cs](https://github.com/gautam-achieveai/revobot/blob/master/dotnet/src/RevoBot.Agents/Workbench/DockerSpawnPlanner.cs))
- `MountTranslator` for Windows-host → Linux-container paths
- `WorkbenchKey` for the `(repo, pr, agent)` triple

What's missing is the connection between `WorkbenchKey` → live agent process. With an `IProcessLauncher` hook, revobot would implement `DockerProcessLauncher` and inject it via `ClaudeAgentSdkOptions.ProcessLauncher`. Total downstream change: ~100 lines + a credhelper integration test.

## Related upstream issues

- #47 — pre-seeding `CurrentSessionId` for `--resume` (also needed before Phase 6 / re-review)
- #48 — `WasForked` always false (needed for the forked-extractor pattern)

#47 and #48 together with this feature unblock the full revobot review + dev loop architecture in design doc §3-12. None are urgent — host-mode works today — but they all gate the next round of cleanup.

## Open design question

Should the hook be `Process Launch(ProcessStartInfo)` or `IAsyncDisposable LaunchAsync(LaunchPlan) → IProcessHandle`? The async/abstract shape is better long-term (works for non-Process backends like a remote runner), but requires more upstream churn. For now the sync `Process` return type matches the existing field type in `ClaudeAgentSdkClient`.

gautam-achieveai:
[Gautam's Dev bot] Implements the approved v3 plan from issue #49 (developer-approved as comment 4442997209).

## Summary
- New `AchieveAi.LmDotnetTools.ProcessLauncher` package owns process spawning for the Claude / Codex / Copilot SDK providers. The seam is `IProcessLauncher` returning `IProcessHandle`. The default implementation (`DefaultProcessLauncher`) wraps `System.Diagnostics.Process` and absorbs the agent-specific discovery (Node + cli.js probe for Claude, Windows shebang shim probe for Copilot) that previously lived in each provider.
- Each provider now hands the launcher a `ProcessLaunchRequest` containing tokenized arguments (no more manual quoting), environment overrides, and a typed list of `HostPathReference` entries (working directory, MCP config file, system-prompt temp file, staging directory). A non-local launcher (Docker, SSH, remote VM) can consume those without re-parsing CLI flags.
- Each provider exposes a `ProcessLauncher` option defaulting to `DefaultProcessLauncher.Instance` so existing in-process behavior is unchanged.

## Why
Issue #49 wants Docker-isolated agent runs without forking the providers. The seam lets a host swap the launcher in `*SdkOptions` and route spawn through a container/SSH/remote target. The default path remains identical to today's `Process.Start` flow.

## Test plan
- [x] `dotnet build LmDotnetTools.sln` — 0 warnings, 0 errors (`TreatWarningsAsErrors` + `EnforceCodeStyleInBuild`).
- [x] New `tests/ProcessLauncher.Tests` — request defaulting / record equality / host-path immutability; `SystemProcessHandle` exit observation, `Kill`, idempotent dispose; `DefaultProcessLauncher` end-to-end against benign child processes (`dotnet --info`, `pwd`, env-echo, missing-executable, cancelled-token). 15 tests, all pass.
- [x] One contract test per provider asserts the exact `Agent` / `ExecutableHint` / `Arguments` / `EnvironmentOverrides` / `HostPaths` the provider hands the launcher seam, using a `RecordingProcessLauncher` test double so no real CLI is invoked. All pass.
- [x] Full provider test suites: `ClaudeAgentSdkProvider.Tests` 83/83, `CodexSdkProvider.Tests` 143/143, `CopilotSdkProvider.Tests` 83/83 pass.
- [x] Full solution test pass — the only failures (`LmStreaming.Sample.Browser.E2E.Tests`, `McpIntegrationTests`) are pre-existing environment issues unrelated to this change (missing Playwright browser binaries, hard-coded `.exe` extension in test fixtures running on Linux).

Closes #49


gautam-achieveai:

⚠️ Preview environment failed to start.

```
git: git -C b:\review_repos\LmDotnetTools\.dev-worktrees\wi-49 fetch origin feat/49-pluggable-process-launcher exited 128: fatal: not a git repository: /repo/.git/worktrees/wi-49
```

src/ProcessLauncher/SystemProcessHandle.cs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public void Kill(bool entireProcessTree = true)
+    {
+        if (_process.HasExited)
+        {
+            return;
+        }
+
+        _process.Kill(entireProcessTree);
```
gautam-achieveai: **[Gautam's review bot] 🔴 HIGH — TOCTOU race in Kill()**

Classic time-of-check-to-time-of-use: the process can exit between the `HasExited` check (line 75) and the `Kill()` call (line 80), causing `InvalidOperationException`. This is the exact path called from `ClaudeAgentSdkClient.Dispose()` (line 712) and both Codex/Copilot `StopAsync` methods.

**Fix:** Catch the specific exception:
```csharp
public void Kill(bool entireProcessTree = true)
{
    try
    {
        if (!_process.HasExited)
        {
            _process.Kill(entireProcessTree);
        }
    }
    catch (InvalidOperationException)
    {
        // Process exited between check and kill — safe to ignore.
    }
}
```
The `HasExited` guard remains as an optimization for the common case, but the `catch` is necessary for correctness.

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
```diff
@@ -691,13 +698,20 @@ public void Dispose()
...
-            // Give process time to exit gracefully
-            var exited = _process?.WaitForExit(5000) ?? true;
+            // Give process time to exit gracefully (best-effort; brief poll).
+            if (_process != null)
+            {
+                var deadline = DateTime.UtcNow.AddMilliseconds(5000);
+                while (!_process.HasExited && DateTime.UtcNow < deadline)
+                {
+                    Thread.Sleep(50);
+                }
```
gautam-achieveai: **[Gautam's review bot] 🔴 HIGH — Thread.Sleep busy-wait in sync Dispose called from async context**

This polls with `Thread.Sleep(50)` for up to 5 seconds (100 iterations), blocking a thread pool thread. The caller is `ClaudeAgentLoop.DisposeClientResourcesAsync()` at line 274:
```csharp
(_client as IDisposable)?.Dispose(); // calls this busy-wait from async context
```

With concurrent agent sessions, multiple thread pool threads block simultaneously.

**Options:**
1. **Best:** Fix the caller to use `IAsyncDisposable` — `await (_client as IAsyncDisposable)?.DisposeAsync()` which already uses proper async shutdown via `ShutdownAsync()`.
2. **If sync Dispose must stay:** Add `bool WaitForExit(TimeSpan timeout)` to `IProcessHandle` wrapping `Process.WaitForExit(TimeSpan)` (available since .NET 5). No busy-wait needed.
3. **Quick fix:** At minimum use `_process.WaitForExitAsync()` with a timeout CTS and `.GetAwaiter().GetResult()` — still blocks but doesn't spin.

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
```diff
@@ -997,84 +1004,100 @@ internal string BuildCliArguments(ClaudeAgentSdkRequest request)
...
+    ///     Builds the CLI arguments as a pre-tokenized list. Each flag and value
+    ///     is a separate element so the launcher hands them straight to the
+    ///     OS process argument list — no string-level quoting for the consumer
+    ///     to worry about. Tri-state semantics for
+    ///     <c>--setting-sources</c> match <see cref="BuildCliArguments"/>:
+    ///     <c>null</c> omits the flag, <c>""</c> emits the flag with an empty
+    ///     value (isolated agent), any other value emits the explicit subset.
     /// </summary>
-    private static string FindNodeJs()
+    internal IReadOnlyList<string> BuildCliArgumentTokens(ClaudeAgentSdkRequest request)
```
gautam-achieveai, line 1025: **[Gautam's review bot] 🟡 MEDIUM — BuildCliArgumentTokens has zero test coverage**

This 70-line method is now the **sole runtime code path** for constructing CLI arguments. The old `BuildCliArguments` (string-concatenation, line 954) has 7 dedicated tests, but `BuildCliArgumentTokens` has none — the contract test only exercises one specific request shape.

Untested critical behaviors:
- Empty/whitespace `ModelId` should omit `--model` token
- `SettingSources` tri-state: `null` omits flag, `""` emits `--setting-sources ""`, `"user"` emits explicit value
- `--mcp-config` JSON serialization (no shell-escaping needed in tokenized form)
- All the boolean flags: `--verbose`, `--include-partial-messages`, `--no-checkpoints`, `--no-session-persistence`

A subtle divergence from `BuildCliArguments` would cause silent CLI argument misparses at runtime.

**Suggestion:** Mirror the existing `BuildCliArguments_*` tests for `BuildCliArgumentTokens`. The method is `internal` so it is already accessible.

src/ProcessLauncher/DefaultProcessLauncher.cs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private static readonly string[] WindowsCopilotExtensions = [".cmd", ".exe", ".ps1", ".bat"];
+
+    /// <summary>
+    /// Windows-only probe for npm shebang shims (e.g., <c>node_modules/.bin/copilot</c>)
+    /// that have no extension. Returns the resolved wrapper path, or the original
+    /// input when no probe matches.
+    /// </summary>
+    private static string ResolveCopilotCliPath(string cliPath)
```
gautam-achieveai, line 174: **[Gautam's review bot] 🟡 MEDIUM — Code duplication with CopilotCliPathResolver**

This `ResolveCopilotCliPath` method (lines 167-201) is a near-exact copy of `CopilotSdkProvider/Agents/CopilotCliPathResolver.Resolve`. The original is still used by `CopilotVersionChecker` (line 19) but no longer by the transport.

Now the same logic lives in two places — if the probe order or extension list changes, both need updating.

**Options:**
1. Have `DefaultProcessLauncher` call `CopilotCliPathResolver.Resolve` (but this creates an upward dependency — wrong direction).
2. Move `CopilotCliPathResolver` into the ProcessLauncher package and have both consumers reference it there.
3. Keep discovery in the provider (see architecture comment on `CliAgentKind`) and have the provider pass a fully-resolved path.

src/ProcessLauncher/IProcessLauncher.cs
```diff
@@ -0,0 +1,22 @@
...
+/// </remarks>
+public interface IProcessLauncher
+{
+    /// <summary>
+    /// Launch the process described by <paramref name="request"/>. Returns a
+    /// handle the caller uses for stdio, liveness probes, and termination.
+    /// Implementations should throw <see cref="ProcessLauncherException"/>
+    /// (wrapping the underlying cause) when the launch fails.
+    /// </summary>
+    IProcessHandle Launch(ProcessLaunchRequest request, CancellationToken cancellationToken = default);
```
gautam-achieveai: **[Gautam's review bot] 🟡 MEDIUM — Synchronous Launch() constrains the stated Docker/K8s/SSH extensibility goals**

#49 explicitly lists Docker, Kubernetes, and SSH as target scenarios — all require async operations (image pull, pod scheduling, SSH handshake). A sync interface forces these implementations to block with `.GetAwaiter().GetResult()`.

Since this is a published NuGet interface, changing from sync to async later is a **breaking change** for all `IProcessLauncher` implementations.

**Suggestion:** Ship `Task<IProcessHandle> LaunchAsync(...)` as the primary API. `DefaultProcessLauncher` wraps with `Task.FromResult(...)`. If sync convenience is needed, add it as a default interface method:
```csharp
public interface IProcessLauncher
{
    Task<IProcessHandle> LaunchAsync(ProcessLaunchRequest request, CancellationToken ct = default);

    IProcessHandle Launch(ProcessLaunchRequest request, CancellationToken ct = default)
        => LaunchAsync(request, ct).GetAwaiter().GetResult();
}
```

This is the one design decision that is hardest to change later, so worth getting right in v1.

src/ProcessLauncher/CliAgentKind.cs
```diff
@@ -0,0 +1,18 @@
+namespace AchieveAi.LmDotnetTools.ProcessLauncher;
+
+/// <summary>
+/// Identifies the CLI agent being launched. Custom launchers can use this to
+/// apply agent-specific behavior (e.g., a Docker launcher may select different
+/// base images per agent kind).
+/// </summary>
+public enum CliAgentKind
```
gautam-achieveai, line 8: **[Gautam's review bot] 🟡 MEDIUM — Design question: Should the shared package know about agent kinds?**

The `CliAgentKind` enum + `switch` dispatch in `DefaultProcessLauncher.ResolveExecutable` (line 76-83) forces the shared package to be modified every time a new agent type is added. A 4th provider (e.g., Gemini CLI) requires: (1) new enum member, (2) new `Resolve*` method in `DefaultProcessLauncher`, (3) potentially new properties on `ProcessLaunchRequest` (like `NodeJsPath` is Claude-specific today).

**Alternative:** Each provider resolves its own executable *before* calling `Launch`, and `ProcessLaunchRequest` carries a fully-resolved `FileName` + `Arguments`. `DefaultProcessLauncher` becomes a simple process spawner with no agent-specific logic. The ~130 lines of file-system discovery (lines 86-201) move back to each provider where they lived before and can evolve independently.

I realize this is a bigger design discussion — happy to talk through tradeoffs. The current approach works correctly; this is about long-term extensibility of the seam.

src/ProcessLauncher/ProcessLaunchRequest.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>Caller-supplied executable path. Bypasses the default
+    /// launcher's discovery logic. Maps to today's <c>CliPath</c> /
+    /// <c>CodexCliPath</c> / <c>CopilotCliPath</c> option fields.</summary>
+    public string? ExecutableOverride { get; init; }
+
+    /// <summary>Path to Node.js. Only consulted when
+    /// <see cref="Agent"/> is <see cref="CliAgentKind.Claude"/>; null delegates
+    /// to the launcher's auto-discovery.</summary>
+    public string? NodeJsPath { get; init; }
```
gautam-achieveai, line 36: **[Gautam's review bot] 🟡 MEDIUM — Claude-specific property on shared contract**

The XML doc explicitly says this is "only consulted when Agent is Claude." As new agents arrive, this pattern invites more provider-specific fields (`PythonPath`? `JavaPath`?), gradually bloating the shared record.

If the design moves toward agent-agnostic requests (see comment on `CliAgentKind`), this property would be consumed internally by the Claude provider when building its launch request.

src/ProcessLauncher/SystemProcessHandle.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            return;
+        }
+
+        try
+        {
+            _process.Exited -= OnProcessExited;
+        }
+        catch
+        {
```
gautam-achieveai, line 118: **[Gautam's review bot] 🟢 LOW — Bare catch swallows all exception types**

This catches everything including `OutOfMemoryException`, `StackOverflowException`, etc. The comment says "Process may already be disposed" — the expected types are `ObjectDisposedException` and `InvalidOperationException`. Narrowing is cheap:
```csharp
catch (ObjectDisposedException) { }
catch (InvalidOperationException) { }
```
Minor since this is a Dispose path, but good hygiene.

gautam-achieveai (commented):
## [Gautam's review bot] PR #54 — Deep Review

**Feature:** Pluggable process transport for CLI agent providers (closes #49)

### Context
Issue #49 requests a pluggable process launch seam so revobot's Phase 4d (Docker-isolated agent runs) can redirect spawn without forking the providers. This PR extracts a new `ProcessLauncher` package with `IProcessLauncher` / `IProcessHandle` interfaces, `DefaultProcessLauncher` as the host-mode implementation, and integrates all three providers (Claude, Codex, Copilot).

### Build & Test
✅ `dotnet build` — 0 warnings, 0 errors
✅ `ProcessLauncher.Tests` — 15/15 passed
✅ Contract tests (Claude, Codex, Copilot) — 3/3 passed

### Overall Assessment
The core idea is **sound and well-executed** — the seam is clean, dependency direction is correct (ProcessLauncher is a leaf package), contract tests are a solid pattern, and existing behavior is preserved. However, there are several issues ranging from design concerns that will compound over time to a concrete race condition and a performance anti-pattern.

### Review Agents Used
| Agent | Duration | Key Finding |
|-------|----------|-------------|
| architecture-review | 260s | `CliAgentKind` enum couples shared package to all agents (OCP) |
| exception-handling-review | 146s | TOCTOU race in `Kill()`, bare catches |
| test-coverage-review | 170s | `BuildCliArgumentTokens` (the runtime path) has zero test coverage |
| performance-review | 185s | `Thread.Sleep(50)` busy-wait in sync Dispose from async context |

### Findings Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 HIGH | `SystemProcessHandle.cs` | TOCTOU race in `Kill()` — process can exit between `HasExited` check and `Kill()` call |
| 2 | 🔴 HIGH | `ClaudeAgentSdkClient.cs` | `Thread.Sleep(50)` busy-wait loop in sync `Dispose()` called from async context |
| 3 | 🟡 MEDIUM | `ClaudeAgentSdkClient.cs` | `BuildCliArgumentTokens` — the actual runtime code path — has zero dedicated tests |
| 4 | 🟡 MEDIUM | `DefaultProcessLauncher.cs` | `ResolveCopilotCliPath` duplicates `CopilotCliPathResolver.Resolve` exactly |
| 5 | 🟡 MEDIUM | `IProcessLauncher.cs` | Sync `Launch()` method constrains the stated Docker/K8s/SSH goals |
| 6 | 🟡 MEDIUM | `CliAgentKind.cs` / `DefaultProcessLauncher.cs` | Agent-specific discovery in shared package violates OCP |
| 7 | 🟡 MEDIUM | `IProcessHandle.cs` | No synchronous `WaitForExit(TimeSpan)` — forces busy-wait in sync Dispose |
| 8 | 🟢 LOW | Contract tests | Three identical `RecordingLauncher` inner classes |

### Verdict: REQUEST CHANGES
Items 1-2 should be fixed before merge. Items 3-7 are strong suggestions — especially item 3 (test coverage for the runtime code path) and item 5 (sync vs async interface for the stated extensibility goals).

src/ProcessLauncher/SystemProcessHandle.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 8baa441 — wrapped the `Kill(entireProcessTree)` call in `try { ... } catch (InvalidOperationException) { }`. The `HasExited` guard remains as an optimization for the common case; the catch closes the race window where the process exits between check and kill.

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 8baa441 by going with the second option in your suggestion list. Added `bool WaitForExit(TimeSpan timeout)` to `IProcessHandle` (wraps `Process.WaitForExit(int)` — OS-managed wait, no busy-spin) and the Dispose path is now:

```csharp
if (_process != null && !_process.HasExited)
{
    _process.WaitForExit(TimeSpan.FromMilliseconds(5000));
}
```

Kept sync `Dispose` for now since the caller chain into `IDisposable.Dispose()` is wider than just `ClaudeAgentLoop.DisposeClientResourcesAsync`. `DisposeAsync` already takes the proper async path via `ShutdownAsync` — the sync path is the fallback.

src/ProcessLauncher/SystemProcessHandle.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 8baa441 — narrowed to `catch (ObjectDisposedException) { }` and `catch (InvalidOperationException) { }` (process never started). The pathological out-of-memory / stack-overflow cases will now propagate as they should.

src/ProcessLauncher/IProcessLauncher.cs
gautam-achieveai: [Gautam's Dev bot]

Agreed — applied your exact suggestion in 8baa441. `LaunchAsync` is now the primary contract on `IProcessLauncher`; sync `Launch` is a default interface method that wraps the async call via `.GetAwaiter().GetResult()`. `DefaultProcessLauncher` keeps sync work but exposes `LaunchAsync` as `Task.FromResult(Launch(...))` for callers that want the async path uniformly.

All three provider call sites (`ClaudeAgentSdkClient.StartAsync`, `CodexAppServerTransport.StartAsync`, `CopilotAcpTransport.StartAsync`) now `await` `LaunchAsync`. Test `RecordingLauncher` stubs and `RecordingProcessLauncher` implement `LaunchAsync`.

A future Docker/K8s/SSH launcher can perform image pull / pod schedule / handshake without blocking the caller thread.

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
gautam-achieveai: [Gautam's Dev bot]

Added in 8baa441. `tests/ClaudeAgentSdkProvider.Tests/Agents/ClaudeAgentSdkClient.CliArguments.Tests.cs` now has 11 `BuildCliArgumentTokens_*` tests that mirror the existing `BuildCliArguments_*` coverage:

- ModelId omit when empty / whitespace; emit when set
- SettingSources tri-state: null → omit, "" → emit empty value (no shell-quoting), "user" → emit explicit value
- `--include-partial-messages` when OutputFormat is stream-json
- `--no-checkpoints` when `DisableCheckpoints=true`
- `--no-session-persistence` when `DisableSessionPersistence=true`
- `--mcp-config` JSON token contains raw JSON without surrounding shell quotes
- `--resume {sessionId}` when SessionId is set

All 94 tests pass in ClaudeAgentSdkProvider.Tests.

src/ProcessLauncher/CliAgentKind.cs
gautam-achieveai: [Gautam's Dev bot]

Thanks for laying out the trade-off — this one I want to push back on, but with full receipts.

**The current shape is intentional** for the docker/k8s/SSH scenarios called out in #49. Here's why the alternative (provider resolves its own `FileName`/`Arguments` ahead of time) doesn't work well for the target use case:

A Docker launcher implementation needs to translate **host paths** (mcp config, system-prompt temp file, working dir, staging dir) into **container paths** via `-v` bind mounts. To do that translation correctly it needs to know:

1. Which CLI is being launched (different agents stage MCP differently — Claude inlines a JSON file, Codex/Copilot configure MCP over RPC after start).
2. Which arguments reference host paths (so the launcher can rewrite them or refuse a path it can't mount).
3. Which agents need a Node.js binary path *inside* the container vs the host's path.

If the provider pre-resolves `FileName` to e.g. `/Users/foo/.nvm/.../node` and bakes that into the request, the docker launcher can't know that token used to be Node — it just sees an opaque host path. Same with `/Users/foo/AppData/Roaming/npm/.../cli.js`.

Keeping `CliAgentKind` + `ExecutableHint` on the request lets the launcher decide "this is Claude → I need node inside the container, cli.js at `/opt/claude-agent-sdk/cli.js`, and the host's `NodeJsPath` is irrelevant in this environment".

**On the bloat concern:** `NodeJsPath` is the only agent-specific property today. If a 4th provider needs a similar property, I'd add it as another nullable string with clear XML docs; the request record stays small and the discriminator stays explicit. If we hit 4+ such properties, that's the signal to introduce per-agent sub-records — but I'd rather pay that cost when we have a concrete case than design for it preemptively.

**On the enum-grows-with-providers concern:** adding a 4th CLI agent already requires a new provider package, new options, new transport — adding one enum member and one `Resolve*` method in `DefaultProcessLauncher` is the smallest part of that work, and it lives in one file. Custom launchers don't need to switch on `Agent`; they can treat it as metadata and route by their own policy.

Happy to revisit if/when we actually start writing the Docker launcher and discover the shape doesn't fit — but for v1 I'd like to keep the discriminator in. WDYT?

src/ProcessLauncher/DefaultProcessLauncher.cs
gautam-achieveai: [Gautam's Dev bot]

You're right that two copies of Windows shebang-shim probing exist. The reason I duplicated rather than moved `CopilotCliPathResolver` is that the file lives next to other provider-internal types (`CopilotVersionChecker` consumes it), and pulling it into the shared `ProcessLauncher` package would force the package to carry the Copilot-shaped probe order (`.cmd`, `.exe`, `.ps1`, `.bat`) — that's actually agent-specific knowledge.

The longer-term answer (tied to the discussion on `CliAgentKind`) is one of:

1. Keep current shape, and dedupe the two copies by having `DefaultProcessLauncher.ResolveCopilotCliPath` call `CopilotCliPathResolver.Resolve` via an `InternalsVisibleTo` (option 1 in your list — the dependency is downward from launcher → provider in this case, so it's the wrong direction for an architectural seam, but it's pragmatic).
2. Move `CopilotCliPathResolver` into the ProcessLauncher package (option 2) and have `CopilotVersionChecker` reference it there. This is the cleanest dedupe but commits the shared package to caring about Copilot's filesystem layout.

I'd lean toward option 2 if we keep the current design after the `CliAgentKind` discussion settles. Will do the cleanup in a follow-up commit if you concur — flagged as a TODO in my head but not blocking on it right now since both call sites currently agree.

src/ProcessLauncher/ProcessLaunchRequest.cs
gautam-achieveai: [Gautam's Dev bot]

This is the same trade-off as the `CliAgentKind` discussion — replied there in full. Short version: `NodeJsPath` lives on the shared contract because a Docker / SSH launcher needs to know that this particular request needs Node.js (vs a bare binary) so it can pull the right base image / install Node in the container / locate Node on the remote host. If the Claude provider pre-resolves to e.g. `/Users/x/.nvm/.../node` and bakes it into `FileName`, the launcher loses the semantic information.

If we hit 4+ similar agent-specific properties, that's the signal to introduce per-agent sub-records (`ClaudeLaunchInfo`, etc.). For one nullable string with clear XML docs, the explicit-discriminator shape is the right call IMO.

gautam-achieveai:
Can you check build failure?

gautam-achieveai:
[Gautam's Dev bot]

Yes — diagnosed and pushed a fix in `5a97214`.

**Two issues showed up in CI run [25818773727](https://github.com/achieveai/LmDotnetTools/actions/runs/25818773727):**

**1. The actual blocker — silent hang (introduced by this PR):** After `LmEmbeddings.Tests` finished at 18:41:57, the run went completely silent for 22 minutes and then hit the workflow's 30-min `timeout-minutes` and got cancelled. No test result line, no error — just dead air. The test projects that would have run next (`LmMultiTurn.Tests`, `LmStreaming.Sample.Tests`, `MockProviderHost.E2E.Tests`, and the new `ProcessLauncher.Tests`) never reported. The most likely culprit is `SystemProcessHandleTests.Kill_TerminatesRunningProcess`, whose Windows path used `cmd /c pause` — `pause` reads from the console and behaves unpredictably when stdin is a redirected pipe on Windows Server 2025. Replaced with `cmd /c timeout /t 30 /nobreak` (stable 30 s sleep, no stdin read).

**2. Pre-existing on main — 17 Playwright failures in `LmStreaming.Sample.Browser.E2E.Tests`:** Browser binaries aren't installed (`Executable doesn't exist at ...\chromium_headless_shell-1148\chrome-win\headless_shell.exe`). These same 17 tests fail on main's most recent green build ([25816709692](https://github.com/achieveai/LmDotnetTools/actions/runs/25816709692)) — the `Build and Test` workflow's `dotnet test` exit code isn't being checked by `scripts/ci-test.ps1` (PowerShell doesn't auto-bubble native exe non-zero exits with `$ErrorActionPreference="Stop"`), so 17 failures silently pass on main. Out of scope for this PR; happy to file a follow-up if you want.

**Belt-and-braces:** Added `--blame-hang --blame-hang-timeout 4m` to both `dotnet test` invocations in `scripts/ci-test.ps1`. If anything hangs longer than 4 min the runner now dumps and names the offending test instead of silently eating the outer 30-min budget. Verified locally: `ProcessLauncher.Tests` runs in 679 ms with `--blame-hang --blame-hang-timeout 1m`, all 15 tests pass.

gautam-achieveai (commented):
## [Gautam's review bot] PR #54 — Incremental Review (Iteration 2)

**Diff reviewed:** `9dd7ddf...8baa441` (12 files, +216 / -19)

### First Review Items — Status

| # | First Review Finding | Status |
|---|---|---|
| 1 | :red_circle: TOCTOU race in `Kill()` | :white_check_mark: **Fixed** — `try/catch(InvalidOperationException)` is correct pattern. |
| 2 | :red_circle: `Thread.Sleep` busy-wait | :white_check_mark: **Fixed** — `WaitForExit(TimeSpan)` delegates to OS primitives. No CPU spin. |
| 3 | :yellow_circle: `BuildCliArgumentTokens` untested | :white_check_mark: **Fixed** — 11 new tests mirror existing coverage. |
| 4 | :yellow_circle: Code duplication `CopilotCliPathResolver` | :left_right_arrow: **Acknowledged** — pushback accepted. |
| 5 | :yellow_circle: Sync `Launch()` constrains extensibility | :white_check_mark: **Fixed** — `LaunchAsync` is now primary contract. |
| 6 | :yellow_circle: `CliAgentKind`/OCP | :left_right_arrow: **Acknowledged** — reasoning about Docker/SSH is coherent. Accepted. |
| 7 | :yellow_circle: `NodeJsPath` Claude-specific | :left_right_arrow: **Acknowledged** — same reasoning as #6. |
| 8 | :green_circle: Bare catch in Dispose | :white_check_mark: **Fixed** — narrowed to specific types. |

### New Findings (Iteration 2) — see inline comments

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | :yellow_circle: MEDIUM | `SystemProcessHandle.cs` | `WaitForExit` comment contradicts implementation (negative timeout) |
| 2 | :yellow_circle: MEDIUM | `CliArguments.Tests.cs` | `--tools` (AllowedTools) token emission has no dedicated test |
| 3 | :green_circle: LOW | `IProcessLauncher.cs` | DIM `Launch` `.GetAwaiter().GetResult()` — add doc warning |

### Build & Test
:white_check_mark: `dotnet build` — 0 warnings, 0 errors
:white_check_mark: `ProcessLauncher.Tests` — 15/15 passed
:white_check_mark: `ClaudeAgentSdkClient.CliArguments.Tests` — 21/21 passed

### Agents Used
| Agent | Duration | Key Finding |
|-------|----------|-------------|
| exception-handling-review | 178s | WaitForExit comment/code mismatch |
| performance-review | 203s | Confirmed busy-wait fix correct |
| test-coverage-review | 221s | `--tools` gap, `WaitForExit(TimeSpan)` untested |

### Verdict: **LOOKS GOOD** :+1:
All HIGH items properly addressed. Remaining findings are non-blocking suggestions. Developer pushbacks are reasonable. Nice iteration.

src/ProcessLauncher/SystemProcessHandle.cs
```diff
@@ -0,0 +1,137 @@
...
+        await _process.WaitForExitAsync(cancellationToken).ConfigureAwait(false);
+        return _process.ExitCode;
+    }
+
+    public bool WaitForExit(TimeSpan timeout)
+    {
+        // Process.WaitForExit(TimeSpan) blocks the calling thread but defers
+        // to the OS — no busy-wait. Negative timeouts mean "wait indefinitely"
+        // per the BCL contract.
+        return _process.WaitForExit((int)Math.Min(int.MaxValue, Math.Max(0, timeout.TotalMilliseconds)));
```
gautam-achieveai: **[Gautam's review bot] :yellow_circle: MEDIUM — Comment contradicts implementation: negative timeout handling**

The comment says "Negative timeouts mean 'wait indefinitely' per the BCL contract" but the code does `Math.Max(0, timeout.TotalMilliseconds)` — which clamps negative values to `0` (immediate return), NOT infinite wait.

The BCL's `Process.WaitForExit(int)` uses `-1` (`Timeout.Infinite`) for infinite wait, but this implementation will never pass `-1`.

**Impact today:** None — the only caller passes `TimeSpan.FromMilliseconds(5000)`. But a future caller passing `Timeout.InfiniteTimeSpan` would get an immediate `false` instead of blocking.

**Suggestion:** Either fix the comment to match the code, or implement the stated contract:
```csharp
int ms = timeout < TimeSpan.Zero
    ? Timeout.Infinite
    : (int)Math.Min(int.MaxValue, timeout.TotalMilliseconds);
return _process.WaitForExit(ms);
```

tests/ClaudeAgentSdkProvider.Tests/Agents/ClaudeAgentSdkClient.CliArguments.Tests.cs
```diff
@@ -131,4 +132,159 @@ public void ApplyStagingDirectoryEnv_DoesNotSet_WhenStagingDirectoryEmpty()
...
+        var client = new ClaudeAgentSdkClient(new ClaudeAgentSdkOptions());
+        var request = new ClaudeAgentSdkRequest { ModelId = "claude-sonnet-4-6", AllowedTools = "Read" };
+
+        var tokens = client.BuildCliArgumentTokens(request);
+
+        Assert.Contains("--include-partial-messages", tokens);
+    }
+
+    [Fact]
+    public void BuildCliArgumentTokens_EmitsNoCheckpoints_WhenDisabled()
```
gautam-achieveai, line 222: **[Gautam's review bot] :yellow_circle: MEDIUM — Missing test: `--tools` (AllowedTools) token emission**

The 11 new tests cover model, setting-sources, checkpoints, session-persistence, mcp-config, and resume — but `--tools` has no dedicated test. The two tests that set `AllowedTools = "Read"` only assert `--include-partial-messages`, not that `--tools Read` appears.

Since `BuildCliArgumentTokens` is the **sole runtime path**, this is a real regression gap. `AllowedTools` restricts agent capabilities — a security-adjacent feature.

**Suggested addition:**
```csharp
[Fact]
public void BuildCliArgumentTokens_EmitsToolsToken_WhenAllowedToolsSet()
{
    var client = new ClaudeAgentSdkClient(new ClaudeAgentSdkOptions());
    var request = new ClaudeAgentSdkRequest { ModelId = "claude-sonnet-4-6", AllowedTools = "Read,Write,Bash" };
    var tokens = client.BuildCliArgumentTokens(request);
    AssertContainsPair(tokens, "--tools", "Read,Write,Bash");
}

[Fact]
public void BuildCliArgumentTokens_OmitsToolsToken_WhenAllowedToolsIsNull()
{
    var client = new ClaudeAgentSdkClient(new ClaudeAgentSdkOptions());
    var request = new ClaudeAgentSdkRequest { ModelId = "claude-sonnet-4-6" };
    var tokens = client.BuildCliArgumentTokens(request);
    Assert.DoesNotContain("--tools", tokens);
}
```

src/ProcessLauncher/IProcessLauncher.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    Task<IProcessHandle> LaunchAsync(ProcessLaunchRequest request, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Synchronous convenience wrapper around <see cref="LaunchAsync"/>. The
+    /// default implementation blocks on the async call; sync-native launchers
+    /// may override this to avoid the extra Task allocation.
+    /// </summary>
+    IProcessHandle Launch(ProcessLaunchRequest request, CancellationToken cancellationToken = default)
+        => LaunchAsync(request, cancellationToken).GetAwaiter().GetResult();
```
gautam-achieveai, line 42: **[Gautam's review bot] :green_circle: LOW — DIM `.GetAwaiter().GetResult()` doc warning for future implementers**

This is safe today — `DefaultProcessLauncher` overrides with sync-native impl, and all production callers use `LaunchAsync`. However, a future remote launcher that only implements `LaunchAsync` could deadlock if `Launch` is called from a SynchronizationContext.

**Suggestion:** Add a `<remarks>` warning:
```xml
/// <remarks>
/// ⚠️ Implementations that perform real async I/O in LaunchAsync
/// MUST override this method to avoid thread pool starvation.
/// </remarks>
```

Not blocking — just a documentation improvement.

src/ProcessLauncher/SystemProcessHandle.cs
gautam-achieveai: **[Gautam's review bot]** Verified — the 	ry/catch(InvalidOperationException) cleanly closes the TOCTOU window. HasExited guard as early-out optimization is the right pattern. :white_check_mark:

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
gautam-achieveai: **[Gautam's review bot]** Verified — OS-managed wait via `Process.WaitForExit(int)` is a clear improvement. Thread blocks but no CPU spin. One minor note: the comment on the `WaitForExit` implementation says negative timeouts wait indefinitely but the code clamps to 0 — see my inline comment on `SystemProcessHandle.cs:78`.

src/ProcessLauncher/SystemProcessHandle.cs
gautam-achieveai: **[Gautam's review bot]** Verified — `ObjectDisposedException` + `InvalidOperationException` covers the two failure modes correctly. :white_check_mark:

src/ProcessLauncher/IProcessLauncher.cs
gautam-achieveai: **[Gautam's review bot]** Verified — `LaunchAsync` as primary with sync `Launch` as DIM is the right shape. One minor suggestion: add a `<remarks>` warning on the DIM that async-heavy implementers MUST override `Launch` to avoid deadlock. See inline comment.

src/ClaudeAgentSdkProvider/Agents/ClaudeAgentSdkClient.cs
gautam-achieveai: **[Gautam's review bot]** Good coverage addition. One gap: `--tools` (AllowedTools) token emission has no dedicated test — see my inline comment. The two tests that set `AllowedTools` only assert `--include-partial-messages`, not the tools flag itself.

src/ProcessLauncher/CliAgentKind.cs
gautam-achieveai: **[Gautam's review bot]** Fair points — the Docker/SSH launcher needing agent-kind awareness to resolve the correct base image / runtime is a compelling argument for keeping the enum centralized. I accept this trade-off. The 4th-provider scenario is real but manageable (add enum value + switch case). No further action needed.

src/ProcessLauncher/DefaultProcessLauncher.cs
gautam-achieveai: **[Gautam's review bot]** Acknowledged — keeping the resolver in both places avoids a cross-package dependency for the existing `CopilotVersionChecker` consumer. The duplication is contained and well-understood. No further action needed.

src/ProcessLauncher/ProcessLaunchRequest.cs
gautam-achieveai: **[Gautam's review bot]** Accepted — same reasoning as CliAgentKind. The launcher needs to know the runtime dependency to select the right container image. No further action needed.

src/ProcessLauncher/SystemProcessHandle.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in af9da9e. Switched to your suggested form:
```csharp
int ms = timeout < TimeSpan.Zero
    ? Timeout.Infinite
    : (int)Math.Min(int.MaxValue, timeout.TotalMilliseconds);
return _process.WaitForExit(ms);
```
Now `Timeout.InfiniteTimeSpan` (or any negative) actually waits indefinitely; comment matches behaviour.

tests/ClaudeAgentSdkProvider.Tests/Agents/ClaudeAgentSdkClient.CliArguments.Tests.cs
gautam-achieveai: [Gautam's Dev bot]

Added in af9da9e: `BuildCliArgumentTokens_EmitsToolsToken_WhenAllowedToolsSet` (asserts `--tools Read,Write,Bash` pair) and `BuildCliArgumentTokens_OmitsToolsToken_WhenAllowedToolsIsNull` (negative case). 13/13 `BuildCliArgumentTokens_*` tests pass.

src/ProcessLauncher/IProcessLauncher.cs
gautam-achieveai: [Gautam's Dev bot]

Added the `<remarks>` block in af9da9e on `IProcessLauncher.Launch` — warns implementers that async-I/O `LaunchAsync` impls (docker pull, K8s scheduling, SSH handshake) MUST override the sync method to avoid `GetAwaiter().GetResult()` deadlock under a captured `SynchronizationContext` or thread-pool starvation.

Issue #76, opened 2026-06-06
Surface sandbox-discovered subagents as spawnable types in the Agent tool


gautam-achieveai:
## Summary

The sandbox MCP gateway (SandboxedOstoolsMcpServer) discovers subagent definitions (`.claude/agents/*.md`, `.ai/agents/*.md`) in the sandbox workspace — at sandbox-create (workspace-root scan) and on every MCP `Read` (ancestor walk). Today LmDotnetTools ignores these: `SubAgentOptions.Templates` is a static dictionary fixed at `MultiTurnAgentLoop` construction, so the `Agent` tool only offers caller-hardcoded subagent types.

This feature consumes `kind=subagent` discoveries from the gateway and registers them as `SubAgentTemplate`s, so the MultiTurnAgent's `Agent` tool can spawn them by name — and spawned subagents run with full access to the session's sandbox tools (`sandbox-Read`, `sandbox-Bash`, …).

## Value

- Workspace repos can ship their own subagents (Claude Code convention) and have them work out-of-the-box in the workspace agent — parity with Claude Code's `.claude/agents` behavior (follows #70).
- Subagents execute inside the same sandbox session as the parent, preserving isolation and egress/token-injection guarantees.

## Background (gateway API surface)

- **Webhook** (rich payload — the only channel carrying `frontmatter` + `prompt`): registered per-create via `discovery.webhook`; POSTs `context_discovery` events: `{ kind: "subagent", name, description, frontmatter: {…}, prompt: "<system prompt>", path }`.
- **Query API** `GET /api/v1/sandboxes/{id}/discovered`: metadata only (kind, name, description, path) — content must be re-read from the workspace (HostPath) and the YAML frontmatter parsed client-side.
- Subagent `.md` format: YAML frontmatter (`name` — falls back to file stem, `description`, `model`, arbitrary fields incl. `tools`) + markdown body = system prompt.

### 1. LmMultiTurn core — dynamic template source

- Allow subagent templates to be added after `SubAgentOptions` creation (e.g. an `ISubAgentTemplateSource` / mutable registry consulted by `SubAgentManager` at spawn time and by `SubAgentToolProvider` when building the `Agent` tool catalog).
- MVP: registrations land **before the agent loop starts** (session-create-time discoveries). Per-turn `Agent`-tool description refresh for mid-session discoveries is a linked follow-up.

### 2. Discovery consumption — LmStreaming.Sample

[...]

Issue #77, opened 2026-06-06
Mid-session activation of sandbox-discovered subagents in the Agent tool


gautam-achieveai:
## Summary

Follow-up to #76. That issue seeds sandbox-discovered subagents at session-create time; this one activates subagents discovered **mid-conversation** (the gateway's per-`Read` ancestor walk pushes `context_discovery` webhook events as the agent explores deeper directories — e.g. after cloning a repo whose subdirs ship their own `.claude/agents/`).

## Value

- Subagents that ship inside repos cloned/explored *during* a conversation become spawnable without restarting the session — full parity with the gateway's discovery model.

## Design notes

- Webhook handler (built in #76) currently records mid-session `kind=subagent` discoveries; this issue routes them into the live template registry.
- `Agent` tool catalog (`SubAgentToolProvider.BuildTemplateCatalog`) is baked into the tool description at `MultiTurnAgentLoop` construction; the contracts/handlers built by `FunctionRegistry.BuildToolCallComponents()` are similarly snapshotted. Needs a refresh path so the next turn's tool injection reflects the updated catalog — options: recompute the Agent tool descriptor per turn, or version the registry and rebuild middleware on change.
- Thread-safety: discoveries arrive on webhook threads while the agent loop runs; spawn-time template resolution must tolerate concurrent registration.
- Running subagents are unaffected; only new spawns see the updated catalog.

## Acceptance Criteria

- [ ] A subagent `.md` discovered mid-session (via gateway per-`Read` walk → webhook) becomes spawnable on the parent's next turn without agent restart.
- [ ] The `Agent` tool description shown to the model reflects the updated catalog on the next turn.
- [ ] Concurrent webhook registration + spawn is race-free (no torn reads, no duplicate templates).
- [ ] Collision policy unchanged: first registration wins, warn on conflict.

## Testing Plan

- **Integration:** register a template while the loop is mid-run; assert next-turn catalog includes it and `SpawnAsync` resolves it.
- **E2E (Instruction-Chain):** turn 1 — parent invokes `sandbox-Read` on a file deep in a fixture tree containing `.claude/agents/<y>.md` (triggers gateway ancestor walk → webhook); turn 2 — parent calls `Agent { subagent_type: "<y>" }`; subagent runs `tools_list` echo + a `sandbox-Bash` file-write; assert host-FS effect and `RemainingTurns` drained. May need a short poll/sync barrier since the gateway walk is a detached task.

## Priority

[...]

PR #79, opened 2026-06-06
feat(subagent): surface sandbox-discovered subagents as spawnable types

gautam-achieveai:
**[Gautam's Dev bot]** Phase 2 of WI-76 — surface sandbox-discovered subagents as spawnable Agent-tool types.

## Summary
- **`SubAgentMarkdownParser`** — pure parser for `.claude/agents/*.md` (YAML frontmatter + body). Tolerates BOM, CRLF, unknown frontmatter keys; returns null on malformed input so the loader can log + skip per file.
- **`WorkspaceSubAgentLoader`** — for each `kind == "subagent"` item the gateway reports, resolves the file inside the workspace host directory (containment guard rejects `..` traversal and sibling-prefix attacks like `C:\work` vs `C:\work-evil`), parses, and maps to `SubAgentTemplate`. Per-file failures log + skip; a complete gateway failure returns an empty dictionary after warning.
- **`SandboxSessionRegistry`** — new `ListDiscoveredAsync` GET against `/api/v1/sandboxes/{id}/discovered` + an outbound `discovery` block on `CreateSandboxRequest` that hands the gateway the webhook URL + shared secret.
- **`ContextDiscoveryController`** — log-only webhook for gateway-pushed discovery events. Constant-time SHA256 auth mirroring `AuthWebhookController`; the dynamic-activation queue is deferred to #77.
- **`Program.cs`** — async-ifies `BuildProductionSubAgentOptions` and merges discovered templates under **built-in wins** collision semantics.

## Reviewer findings applied
1. Dropped `IContextDiscoveryQueue` — controller is log-only this iteration.
2. Discovery GET folded into `SandboxSessionRegistry` (reuses the same auth-aware `HttpClient`).
3. `SubAgentTemplateFactory` folded into the loader as `internal static MapToTemplate`.
4. Dropped `ParsedSubAgent.Frontmatter` dict — only the fields we actually map are surfaced.
5. Dropped the 256 KB read cap.
6. Path guard pinned: `NormalizeBasePath` always carries a trailing separator + Ordinal/OrdinalIgnoreCase selected by `OperatingSystem.IsWindows()`.
7. `description` → both `Description` AND `WhenToUse` (the Agent-tool catalog is never blank for discovered templates).
8. All gateway DTOs pinned with explicit `[JsonPropertyName]`.

## Test plan
- [x] `dotnet build samples/LmStreaming.Sample/LmStreaming.Sample.csproj` → 0 warnings, 0 errors
- [x] `dotnet test tests/LmStreaming.Sample.Tests/LmStreaming.Sample.Tests.csproj` → **107 / 107 passed** (31 new)
  - `SubAgentMarkdownParserTests` (12): well-formed/malformed/BOM/CRLF/fallback-name/unknown-keys/empty-tools vs absent
  - `WorkspaceSubAgentLoaderTests` (13): mapping table + traversal guard (`../` rejected, sibling-prefix rejected, relative/absolute under base accepted) + `NormalizeBasePath` trailing-separator pin
  - `ContextDiscoveryControllerTests` (6): 401 missing/wrong secret, 400 null/missing kind/missing name, 200 happy path
- [ ] Manual gateway smoke once the matching gateway build lands (gateway-side discovery emission is the other half of the contract).

Refs #76

### DIFF
--- a/samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
+++ b/samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
@@ -1,5 +1,3 @@
-using System.Security.Cryptography;
-using System.Text;
 using System.Text.Json.Serialization;
 using LmStreaming.Sample.Services.Auth;
 using Microsoft.AspNetCore.Mvc;
@@ -32,7 +30,7 @@ public sealed class ContextDiscoveryController(
     [HttpPost("context_discovery")]
     public IActionResult Notify([FromBody] ContextDiscoveryPayload? body)
     {
-        if (!IsAuthorized())
+        if (!sharedSecret.Matches(Request.Headers.Authorization.ToString()))
         {
             // Do not reveal whether the header was missing, malformed, or simply wrong.
             logger.LogWarning("Rejected unauthorized context-discovery webhook call.");
@@ -45,26 +43,17 @@ public sealed class ContextDiscoveryController(
             return BadRequest();
         }
 
+        // Description is logged alongside kind/name/path so the field carries diagnostic value
+        // in this iteration even though dynamic activation (the consumer that mutates the catalog
+        // from these events) is deferred to #77.
         logger.LogInformation(
-            "ContextDiscovery: kind={Kind} name={Name} path={Path}",
+            "ContextDiscovery: kind={Kind} name={Name} path={Path} description={Description}",
             body.Kind,
             body.Name,
-            body.Path);
+            body.Path,
+            body.Description);
 
         return Ok();
-    }
-
-    private bool IsAuthorized()
-    {
-        var presented = Request.Headers.Authorization.ToString();
-        if (string.IsNullOrEmpty(presented))
-        {
-            return false;
-        }
-
-        var presentedHash = SHA256.HashData(Encoding.UTF8.GetBytes(presented));
-        var expectedHash = SHA256.HashData(Encoding.UTF8.GetBytes(sharedSecret.Value));
-        return CryptographicOperations.FixedTimeEquals(presentedHash, expectedHash);
     }
 }
 
--- a/samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
+++ b/samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
@@ -1,0 +1,90 @@
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json.Serialization;
+using LmStreaming.Sample.Services.Auth;
+using Microsoft.AspNetCore.Mvc;
+
+namespace LmStreaming.Sample.Controllers;
+
+/// <summary>
+/// Webhook endpoint the sandbox gateway calls when it discovers a new context file in the
+/// workspace (sub-agent, skill, …). This iteration is <b>log-only</b>: the controller verifies
+/// the shared secret in constant time, logs a structured event, and returns 200. There is no
+/// in-memory queue and no dynamic activation — those are tracked in #77 and land with their
+/// consumer (mid-session sub-agent catalog refresh / dynamic Agent tool reload).
+/// </summary>
+/// <remarks>
+/// SECURITY: mirrors <see cref="AuthWebhookController"/>'s pattern: the gateway authenticates
+/// itself with a shared secret in the <c>Authorization</c> header, compared in constant time
+/// over fixed-width hashes. The controller NEVER logs the Authorization header value or the
+/// shared secret — only the discovery payload's kind/name/path and the allow decision.
+/// </remarks>
+[ApiController]
+[Route("api/discovery")]
+public sealed class ContextDiscoveryController(
+    AuthSharedSecret sharedSecret,
+    ILogger<ContextDiscoveryController> logger) : ControllerBase
+{
+    /// <summary>
+    /// Gateway callback for one discovered context item. Returns 200 on success, 401 if the
+    /// shared secret doesn't match, 400 if the payload is malformed.
+    /// </summary>
+    [HttpPost("context_discovery")]
+    public IActionResult Notify([FromBody] ContextDiscoveryPayload? body)
+    {
+        if (!IsAuthorized())
+        {
+            // Do not reveal whether the header was missing, malformed, or simply wrong.
+            logger.LogWarning("Rejected unauthorized context-discovery webhook call.");
+            return Unauthorized();
+        }
+
+        if (body is null || string.IsNullOrWhiteSpace(body.Kind) || string.IsNullOrWhiteSpace(body.Name))
+        {
+            logger.LogWarning("Rejected context-discovery webhook with malformed payload.");
+            return BadRequest();
+        }
+
+        logger.LogInformation(
+            "ContextDiscovery: kind={Kind} name={Name} path={Path}",
+            body.Kind,
+            body.Name,
+            body.Path);
+
+        return Ok();
+    }
+
+    private bool IsAuthorized()
+    {
+        var presented = Request.Headers.Authorization.ToString();
+        if (string.IsNullOrEmpty(presented))
+        {
+            return false;
+        }
+
+        var presentedHash = SHA256.HashData(Encoding.UTF8.GetBytes(presented));
+        var expectedHash = SHA256.HashData(Encoding.UTF8.GetBytes(sharedSecret.Value));
+        return CryptographicOperations.FixedTimeEquals(presentedHash, expectedHash);
+    }
+}
+
+/// <summary>
+/// Gateway → app payload for a single discovered context item. Field names are the gateway's
+/// wire contract (snake_case), pinned via <see cref="JsonPropertyNameAttribute"/> so they bind
+/// regardless of the app's JSON naming defaults. Unknown fields are tolerated by System.Text.Json
+/// by default so the gateway can add new fields without breaking older app builds.
+/// </summary>
+public sealed record ContextDiscoveryPayload
+{
+    [JsonPropertyName("kind")]
+    public string? Kind { get; init; }
+
+    [JsonPropertyName("name")]
+    public string? Name { get; init; }
+
+    [JsonPropertyName("description")]
+    public string? Description { get; init; }
+
+    [JsonPropertyName("path")]
+    public string? Path { get; init; }
+}
--- a/samples/LmStreaming.Sample/Controllers/AuthWebhookController.cs
+++ b/samples/LmStreaming.Sample/Controllers/AuthWebhookController.cs
@@ -1,4 +1,3 @@
-using System.Security.Cryptography;
 using System.Text;
 using System.Text.Json.Serialization;
 using LmStreaming.Sample.Services.Auth;
@@ -39,7 +38,7 @@ public sealed class AuthWebhookController(
         [FromBody] AuthWebhookRequest body,
         CancellationToken ct = default)
     {
-        if (!IsAuthorized())
+        if (!sharedSecret.Matches(Request.Headers.Authorization.ToString()))
         {
             // Do not reveal whether the header was missing, malformed, or simply wrong.
             logger.LogWarning("Rejected unauthorized auth-webhook call for provider {Provider}.", provider);
@@ -107,24 +106,6 @@ public sealed class AuthWebhookController(
                 body.DestinationHost);
             return Ok(AuthWebhookResponse.Deny($"token acquisition failed for provider '{tokenProvider.ProviderId}'"));
         }
-    }
-
-    /// <summary>
-    /// Constant-time comparison of the incoming <c>Authorization</c> header against the shared secret.
-    /// Both sides are hashed to a fixed-width SHA-256 digest first, so the comparison neither throws
-    /// on a length mismatch nor leaks the secret's length via an early-exit.
-    /// </summary>
-    private bool IsAuthorized()
-    {
-        var presented = Request.Headers.Authorization.ToString();
-        if (string.IsNullOrEmpty(presented))
-        {
-            return false;
-        }
-
-        var presentedHash = SHA256.HashData(Encoding.UTF8.GetBytes(presented));
-        var expectedHash = SHA256.HashData(Encoding.UTF8.GetBytes(sharedSecret.Value));
-        return CryptographicOperations.FixedTimeEquals(presentedHash, expectedHash);
     }
 
     /// <summary>Resolves a registered provider by the route id, matching <see cref="IOAuthTokenProvider.ProviderId"/> case-insensitively.</summary>
--- a/samples/LmStreaming.Sample/LmStreaming.Sample.csproj
+++ b/samples/LmStreaming.Sample/LmStreaming.Sample.csproj
@@ -9,6 +9,8 @@
   <ItemGroup>
     <!-- Expose internal auth-provider helpers (authorize-URL / scope builders) to the E2E test project. -->
     <InternalsVisibleTo Include="LmStreaming.Sample.E2E.Tests" />
+    <!-- Expose internal sub-agent merge helper + loader internals to the unit test project. -->
+    <InternalsVisibleTo Include="LmStreaming.Sample.Tests" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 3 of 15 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #79 (linked to #76).

## Summary
Well-designed feature with solid security, good test coverage (31 new tests), and clean separation of concerns. The parser, loader, controller, and gateway DTO changes form a cohesive unit.

### Security: PASS
- Path traversal guard is correct: trailing-separator normalization prevents sibling-prefix attacks, `Path.GetFullPath` collapses `../`, OS-aware case comparison.
- Constant-time SHA256 webhook auth mirrors existing pattern correctly.
- Symlink escape is documented as out-of-scope (correct scoping decision).

### Design
- Collision policy (built-in wins) is clear, consistent, and tested.
- Per-file error resilience (log + skip) ensures one bad `.md` file doesn't block the whole catalog.
- `ParsedSubAgent` to `SubAgentTemplate` mapping is pinned by unit tests with good edge-case coverage (null vs empty tools, absent vs present model).

### Minor Suggestions (non-blocking)

**1. MEDIUM - Sync-over-async comment missing (Program.cs:608)**
`.GetAwaiter().GetResult()` on `BuildProductionSubAgentOptionsAsync` lacks the explanatory comment that other sync-over-async sites have (e.g. line 427). Adding a brief note like:
`// Sync-over-async: acceptable in factory delegate (no SynchronizationContext, matches existing pattern)`
signals this was deliberate, not an oversight.

**2. LOW - Unbounded error body in ListDiscoveredAsync exception (SandboxSessionRegistry.cs:283)**
The exception message includes the full response body. If the gateway returns a large HTML error page (502 from reverse proxy), this could produce multi-KB log entries. Consider truncating:
`var bodySnippet = body.Length > 500 ? body[..500] + "...(truncated)" : body;`

---

**Verdict: APPROVE** - Ship it. Both suggestions are quality-of-life improvements for a future iteration.

samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
```diff
@@ -0,0 +1,90 @@
...
+/// <summary>
+/// Gateway → app payload for a single discovered context item. Field names are the gateway's
+/// wire contract (snake_case), pinned via <see cref="JsonPropertyNameAttribute"/> so they bind
+/// regardless of the app's JSON naming defaults. Unknown fields are tolerated by System.Text.Json
+/// by default so the gateway can add new fields without breaking older app builds.
+/// </summary>
+public sealed record ContextDiscoveryPayload
+{
+    [JsonPropertyName("kind")]
+    public string? Kind { get; init; }
```
gautam-achieveai, line 69: Should this really be string? No emum or something equivalent when new kinds are added?

gautam-achieveai:
[Gautam's bot]

## Review of PR #79 — focus: over-engineering, simplicity, duplication (incl. of concerns)

Deep review at commit 95a55ad against merge-base 86a6086 (9 files, +1095/−5). Verified the production code against approved plan v2 on #76 point-by-point: **all 8 plan-review findings and the `/approve-with-changes` condition are honored** (no `SubAgent.Discovery.Enabled` anywhere), zero `src/` changes, no speculative abstractions, no drive-bys. On the requested over-engineering axis, the production code is clean — explicitly cleared after scrutiny: the `FrontmatterDto` two-step (YamlDotNet needs a mutable target), the hand-rolled frontmatter split (YamlDotNet can't split fenced markdown), the payload-vs-item DTO separation (different wire contracts), and the per-class deserializer.

The problems are concentrated in two areas: **the promised test layers were not delivered** (the plan's architectural bargain was "zero src/ refactoring *because* invariants get pinned by tests" — the pins are missing), and **duplication of security/error-handling concerns** — squarely the requested focus.

**Verdict: REQUEST CHANGES** — 5 HIGH (4 blocking).

---

### Blocking

**1. [BLOCKER] "Closes #76" will auto-close the issue with its headline acceptance criterion unverified**

#76's AC says the spawned subagent must "successfully invoke sandbox tools (verified by real sandbox-Bash/file-write effect)". The only layer that proves this — the plan-v2 E2E scenario `SubAgentDiscoveryInstructionChainTests.cs` — exists nowhere on the branch, and the PR's own unchecked "Manual gateway smoke" box acknowledges it. Merging as-is closes #76 on an unverified AC, and that tracking loss is silent. **Cheapest fix in the whole review: change `Closes #76` → `Refs #76`** and track the remaining ACs — or deliver the E2E scenario.

**2. [BLOCKER] The promised integration-test layer (tests/LmMultiTurn.Tests) is entirely absent — the load-bearing invariant is unguarded**

Plan v2 promised three pins there: (a) discovered-template catalog+spawn, (b) sandbox-tool inheritance (stub `sandbox-Bash` registered pre-construction + `EnabledTools=null` ⇒ subagent handler set contains it), (c) no-recursion (no `Agent`/`SendMessage`/`CheckAgent` in the subagent's set). None exist. The mechanism the feature relies on is the null-passthrough in untouched `SubAgentManager.BuildEnabledToolSet` — existing tests there cover filter mechanics only, not the discovery scenario. A future reorder in `Program.cs` registration or an edit to that method silently breaks every discovered subagent with no failing test. The plan justified zero `src/` changes *by* these pins; without them the bargain is void.

**3. [BLOCKER] Built-in-wins merge is untested — a one-line flip silently inverts a trust boundary**

`Program.cs:1270-1290` merges discovered templates via `TryAdd` so built-ins win collisions. Change `TryAdd` to the indexer and **workspace-supplied files shadow trusted built-in templates** — untrusted sandbox content overriding trusted definitions — and nothing detects it. This is a trust-boundary guard, not a coverage nicety, and it's a one-test fix.

**4. [BLOCKER-soft] `IsAuthorized()` duplicated byte-for-byte across both webhook controllers**

`AuthWebhookController.cs:117-128` and `ContextDiscoveryController.cs:63-74` carry identical constant-time secret comparison (`SHA256.HashData` + `CryptographicOperations.FixedTimeEquals`). Security-sensitive logic in two copies is the classic divergence trap — a timing/length fix lands in one and rots in the other. Both controllers already share `AuthSharedSecret` as the secret holder; move the comparison onto it:

```csharp
// AuthSharedSecret.cs
public bool Matches(string? presented)
{
    if (string.IsNullOrEmpty(presented)) return false;
    var presentedHash = SHA256.HashData(Encoding.UTF8.GetBytes(presented));
    var expectedHash  = SHA256.HashData(Encoding.UTF8.GetBytes(Value));
    return CryptographicOperations.FixedTimeEquals(presentedHash, expectedHash);
}
```

**5. [HIGH] Registry gateway-call error handling: three divergent shapes, plus a doc that lies in new code**

`SandboxSessionRegistry.cs` now has three error-handling policies for the same concern: `CreateSessionAsync` (:196-209) logs then throws; the new `ListDiscoveredAsync` (:282-289) throws without logging; `DestroySessionAsync` swallows. And `ListDiscoveredAsync`'s XML doc says the exception carries the body "(truncated)" — the code performs no truncation. Extract a small `SendAndReadAsync<T>` private helper, or at minimum align `ListDiscoveredAsync`'s logging with `CreateSessionAsync` and fix the doc. The doc/code mismatch should not merge regardless.

### Non-blocking (fix in this PR or fast-follow)

**6. [MEDIUM] `WorkspaceSubAgentLoader.LoadAsync` — the only public method — has zero direct tests.** Kind-filter, per-file skip+warn, gateway-failure→empty-dict, and loader-level name collision are all unverified; the 13 delivered tests cover only the internal static helpers. Plan v2 promised a `MockHttpMessageHandler` loader test.

**7. [MEDIUM] `SandboxSessionRegistry.ListDiscoveredAsync` untested** (success + non-2xx) — also explicitly promised in plan v2.

**8. [MEDIUM] `MaxTurnsPerRun = 25` hardcoded ×3** — `Program.cs:1252`, `:1266`, and `MapToTemplate` — while `SubAgentTemplate`'s own default is 50, so the literal overrides the type's default in three places. One `const`.

**9. [MEDIUM] Webhook base-URL derivation duplicated** — `BuildAuthProviders` (:359) and `BuildDiscovery` (:426) both derive the callback base from `AuthOptions.PublicBaseUrl` with the same trim/fallback logic.

### Minor

- **[LOW]** `Program.cs:600-606` `BuildProductionSubAgentOptionsAsync(...).GetAwaiter().GetResult()` now blocks on an HTTP gateway call. Acceptable on the startup path, but add a comment saying why it's safe *here* so the pattern doesn't migrate into a request path. (Flagged independently by 3 review agents — visibility, not severity.)
- **[LOW]** `NormalizeTools` could be a LINQ one-liner; `bodySpan` local used once — inline; ctor guards use if/throw where `ArgumentNullException.ThrowIfNull` is used elsewhere in the same file.
- **[INFO]** `ContextDiscoveryPayload.Description` is deserialized but never read — plan-authorized for #77; consider including it in the discovery log line so the field isn't dead in the meantime.

### What's genuinely good (keep as-is)

- Scope discipline is exemplary: 14-point plan-v2 conformance, zero `src/` changes as promised, no invented abstractions.
- Delivered unit tests are high quality: real traversal attacks including the sibling-prefix case (`C:\work` vs `C:\work-evil`), empty-vs-absent `tools` pinned at both parser and loader layers, no over-mocking.
- `NormalizeBasePath` trailing-separator + OS-aware comparison pin is exactly what the plan review asked for.

### Summary

| # | Severity | Finding | Fix cost |
|---|---|---|---|
| 1 | BLOCKER | "Closes #76" with unverified AC | one word |
| 2 | BLOCKER | Integration pins absent | 3 tests |
| 3 | BLOCKER | Built-in-wins merge untested | 1 test |
| 4 | BLOCKER-soft | `IsAuthorized()` ×2 | 1 method extraction |
| 5 | HIGH | Registry error-handling ×3 shapes + doc mismatch | small helper |

CI is green (build-and-test + browser-e2e). Production code needs no structural changes — the path to approval is: tests promised by the approved plan + the two duplication extractions + the `Closes`→`Refs` word swap (or the E2E scenario).

gautam-achieveai:

Preview stopped.

samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs
gautam-achieveai: **[Gautam's Dev bot]** Deliberate. The gateway owns the discovery wire contract and the set of `kind` values it emits is expected to grow (skill, prompt, future kinds). Locking the app side to an enum would force a lock-step release every time the gateway adds a kind. STJ on `string` tolerates unknown values out of the box; the loaders kind filter (`item.Kind == "subagent"`) means anything unknown is silently skipped, which is exactly the forward-compat behaviour we want. If we promote `kind` to an enum later, the right place to put it is the gateway client library so both sides share one definition — not here.

gautam-achieveai:
**[Gautam's Dev bot]** Iteration on the [Gautam's bot] deep review pushed in 4c8baac. Sample suite now 126 / 126 (was 107 / 107; +19 new tests).

## Addressed

**BLOCKER 1 (Closes → Refs):** PR body updated above; `Refs #76` so the work item stays open after merge.

**BLOCKER 2 (untested trust-boundary at `WorkspaceSubAgentLoader.LoadAsync`):** new `WorkspaceSubAgentLoaderLoadAsyncTests` (9 tests) drives the registry's HTTP path through a stub handler against a real on-disk workspace. Pins: well-formed → mapped template; `kind != "subagent"` skipped; 502 gateway → empty dict (not throw); per-file resilience (missing file / malformed markdown skipped, batch continues); `../outside.md` rejected at the loader boundary; two discoveries with the same parsed name collapse to first-wins; empty `HostPath` short-circuits; null args throw.

**BLOCKER 3 (untested merge collision):** extracted `WorkspaceSubAgentLoader.MergeBuiltInWins` (was an inline loop in `Program.cs`) and added `SubAgentMergeTests` (4 tests). Pins built-in-wins on collision, warning log, no-op on empty discovered. Comment in the test calls out that flipping merge direction is exactly the regression this exists to catch. `InternalsVisibleTo` added so the test project can reach the internal helper.

**BLOCKER 4 (duplicated `IsAuthorized`):** hoisted into `AuthSharedSecret.Matches(string? presented)`. Both controllers (`AuthWebhookController`, `ContextDiscoveryController`) shrink to a single-line call. One canonical SHA256 / `FixedTimeEquals` implementation.

**HIGH 5 (doc/code mismatch on `(truncated)`):** `ListDiscoveredAsync` now truncates error bodies at 500 chars with the `...(truncated)` marker (matches `CreateSessionAsync` policy) and `LogError`s the failure. New `SandboxSessionRegistryListDiscoveredTests` (6 tests) cover success / empty / 502 + truncation marker / 404 / null payload / null sessionId.

**MEDIUM 8 (magic `25`):** promoted to `WorkspaceSubAgentLoader.DefaultMaxTurnsPerRun`. Replaced two `Program.cs` call sites + updated existing test assertion.

**MEDIUM 9 (webhook base-URL helper):** added `WebhookOptions.CallbackBaseUrl` computed property. Replaced two inline `PublicBaseUrl.TrimEnd('/')` usages in `SandboxSessionRegistry`.

**LOW (sync-over-async at Program.cs:600):** safety comment added explaining why `GetAwaiter().GetResult()` is safe here (DI build phase, never on a request thread).

**INFO (Description not logged):** inlined into the `ContextDiscovery` log line so we can correlate webhook-side events with what the gateway actually emitted.

**Inline question on `kind` as string (3368138870):** replied — deliberate forward-compat. The gateway owns the wire contract and adds kinds over time; locking to an enum forces lock-step releases. STJ tolerates unknown strings; the loader filter (`item.Kind == "subagent"`) silently skips unknown kinds.

## Deferred (with reason)

**Reviewer asked for E2E tests inside `tests/LmMultiTurn.Tests`** (tool inheritance, sandbox-rerun gating, no-recursion). Those touch the multi-turn sub-agent runner, which is structurally outside this PR's diff. The sample-layer loader is fully gated end-to-end here (HTTP → kind filter → containment → read → parse → map → first-wins merge). I'd like to do the multi-turn E2E in the WI-77 follow-up where the queue / dynamic-activation work lives, since that's where the inheritance + no-recursion contracts actually run. If that's a stop-ship for this PR rather than a defer, say so and I'll move them here.

## Conversation surfaces reviewed

- 13 inline threads — read every one; the only unanswered question was 3368138870 (now replied).
- All three top-level review summaries — read, addressed inline above.
- Top-level conversation comments — none open from a human; bot comments (PR-Agent / [Gautam's bot]) accounted for.

Build: 0 warnings / 0 errors. Tests: 126 / 126 green.

gautam-achieveai:
**[Gautam's review bot]** Incremental review of iteration 2 (commit 4c8baac vs 95a55ad).

## Verdict: ✅ APPROVE

All 4 blockers, 1 HIGH, and 4 MEDIUMs from the initial review have been cleanly addressed. Tests pass 126/126 (+19 new). The code quality of the fixes is high — particularly:

- **AuthSharedSecret.Matches** — clean single-responsibility extraction with proper null guard and XML doc.
- **MergeBuiltInWins** — trust-boundary comment is excellent; the test explicitly calls out that flipping merge direction is the regression it exists to catch.
- **TruncateBody** — clean helper, tested with the `...(truncated)` marker assertion pinning both the cap and the suffix.
- **DefaultMaxTurnsPerRun** const + **CallbackBaseUrl** property — good DRY refactors.
- **WorkspaceSubAgentLoaderLoadAsyncTests** (9 tests) — real on-disk workspace with stub HTTP exercising the full pipeline. Covers gateway failure, kind filter, path traversal, malformed markdown, name collision, and empty host path. Exactly what was missing.

### Minor note (non-blocking)

`StubHandler` is near-duplicated between `SandboxSessionRegistryListDiscoveredTests` and `WorkspaceSubAgentLoaderLoadAsyncTests` (the former adds `LastRequest` capture). Consider extracting to a shared test fixture in a future cleanup — not a merge blocker.

### Deferred work accepted

E2E tests in `tests/LmMultiTurn.Tests` for tool inheritance + no-recursion deferred to #77. The sample-layer pipeline is fully tested end-to-end here; runner-level contracts belong where the runner work lives. 👍

gautam-achieveai (commented):
**[Gautam's review bot]** ✅ Iteration 2 reviewed — all blockers resolved. See comment above for full details.

PR #83, opened 2026-06-07
[Gautam's Dev bot] Per-provider auth controllers + M365 (Graph) provider

### LINKED ISSUE #82: Per-provider auth token services + controllers: ADO, GitHub, new M365
## Summary

Split the monolithic OAuth surface in `LmStreaming.Sample` into **three per-provider vertical slices** — each with its own controller + token service — and add a **new M365 (Microsoft Graph) provider** alongside the existing ADO and GitHub ones:

1. **ADO** — `AdoAuthController` + `AdoTokenService` (extracted from today's shared `AuthController` + `AdoOAuthProvider`)
2. **GitHub** — `GitHubAuthController` + `GitHubTokenService` (extracted from today's shared `AuthController` + `GitHubOAuthProvider`)
3. **M365** — `M365AuthController` + `M365TokenService` / `M365OAuthProvider` (**new**, backed by the existing `EmailMcpServer` Entra ID app registration)

Today a single `AuthController` (`api/auth/{provider}/signin|status|signout`) fans out over `IEnumerable<IOAuthTokenProvider>`. This issue gives each provider a dedicated controller and service so provider-specific behavior (M365 confidential-client flow, refresh rotation, Graph scopes) has a clean home, while keeping the shared `IOAuthTokenProvider` / `OAuthProviderBase` contracts intact for the sandbox egress webhook.

## Value

- Provider-specific flows stop accreting conditionals in one shared controller.
- M365/Graph tokens become available to the sandbox egress token-injection path (`AuthWebhookController`) → sandboxed agents can call Graph (mail, calendar, meetings) without ever seeing the token.
- Clear per-provider routes: `api/auth/ado/*`, `api/auth/github/*`, `api/auth/m365/*`.

## Files touched

**Modified**
- `samples/LmStreaming.Sample/Controllers/AuthController.cs` — split into per-provider controllers (shared plumbing moves to a base or shared service; this file is either removed or reduced to shared helpers)
- `samples/LmStreaming.Sample/Controllers/AuthWebhookController.cs` — resolve M365 provider for egress token injection
- `samples/LmStreaming.Sample/Services/Auth/OAuthProviderHosts.cs` — add `m365` → `graph.microsoft.com` (and `login.microsoftonline.com` for the token endpoint) host mapping
- `samples/LmStreaming.Sample/Services/Auth/AuthOptions.cs` — add `M365` options block (ClientId, TenantId, Scopes, RedirectUri; **ClientSecret only via user-secrets/env, never committed**)
- `samples/LmStreaming.Sample/Services/Auth/GitHubOAuthProvider.cs` — rename/move per the service split (`GitHubTokenService`)
- `samples/LmStreaming.Sample/Services/Auth/AdoOAuthProvider.cs` — rename/move per the service split (`AdoTokenService`)
- `samples/LmStreaming.Sample/Program.cs` (~:168-178) — DI registration for the three services + new controllers
- `samples/LmStreaming.Sample/appsettings.Development.json` — `Auth:M365` section (client id + tenant id only; secret comes from user-secrets)
- `samples/LmStreaming.Sample/AuthProviderGuide.md` — document the per-provider endpoints + M365 setup

**New**
- `samples/LmStreaming.Sample/Controllers/AdoAuthController.cs`
- `samples/LmStreaming.Sample/Controllers/GitHubAuthController.cs`
- `samples/LmStreaming.Sample/Controllers/M365AuthController.cs`
- `samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs` (aka `M365TokenService`)
- Browser entry route `GET /auth/{provider}` — either on the per-provider controllers or a small shared `AuthPagesController` (implementation's choice)

**Tests**
- `tests/LmStreaming.Sample.E2E.Tests/Scenarios/GitHubOAuthProviderTests.cs` — update for split
- `tests/LmStreaming.Sample.E2E.Tests/Scenarios/AdoOAuthProviderTests.cs` — update for split
- `tests/LmStreaming.Sample.E2E.Tests/Scenarios/AuthWebhookControllerTests.cs` — extend for `m365` provider resolution + Graph host mapping
- `tests/LmStreaming.Sample.E2E.Tests/Scenarios/M365OAuthProviderTests.cs` *(new)* — auth-code exchange, refresh rotation persistence, status mapping
- `tests/LmStreaming.Sample.E2E.Tests/Scenarios/FileOAuthTokenStoreTests.cs` — extend if record shape changes (refresh-token rotation)

### Shared contracts (unchanged)
- Keep `IOAuthTokenProvider` + `OAuthProviderBase` + `IOAuthTokenStore`/`FileOAuthTokenStore` — the sandbox egress webhook (`AuthWebhookController`) and `OAuthTokenHydrator` already work against these. The split is at the controller/service layer, not the contract layer.
- Per-provider controllers route as `[Route("api/auth/ado")]`, `[Route("api/auth/github")]`, `[Route("api/auth/m365")]` with the same `signin`/`status`/`signout` verbs. UI-facing endpoints **never surface token material** (only `OAuthStatus` / `SignInChallenge`), same as today.

### Browser entry point — `GET /auth/{provider}`
- A top-level, browser-navigable route (NOT under `/api`): visiting `http://localhost:5000/auth/github` (or `/auth/ado`, `/auth/m365`) in a browser:
  - **Not signed in** → starts the sign-in flow immediately (redirect to the provider's authorization URL); after the callback completes, shows a simple success page (provider, account, scopes, expiry).
  - **Already signed in** → skips the flow and renders the success page directly.
- Rationale: today `POST api/auth/{provider}/signin` is unusable from the address bar (browser GET → 404). The JSON API endpoints stay for the UI; this route is the human-friendly path.

### Token lifecycle — refresh tokens persisted, access tokens refreshed on demand
- **Refresh tokens are persisted** (via `IOAuthTokenStore`/`FileOAuthTokenStore`) for every provider that issues them (ADO, M365) so sign-in survives app restarts — interactive re-auth only when the refresh token itself is invalid/expired/revoked.
- **Access tokens refresh lazily on demand**: `GetAccessTokenAsync` transparently redeems the stored refresh token when the access token is expired or near expiry (with a small skew window, e.g. 5 min) — callers (egress webhook, UI status) never trigger an interactive prompt for a refreshable session.
- GitHub: long-lived OAuth-App token, no refresh dance — persistence of the token itself suffices (already the case).
- M365: refresh redemption **rotates** the refresh token — persist the newest one on every redemption (see M365 section).

### GitHub provider — app registration (existing, already wired)
- **OAuth app:** gh CLI first-party app — client id `178c6fc778ccc68e1d6a` + client secret (GitHub documents this pair as safe to embed; already present in `appsettings.Development.json` `Auth:Github`).
- **Flow:** standard OAuth web flow with dynamic loopback callback; scopes `repo`, `read:org`. Tokens are long-lived OAuth-App tokens (no refresh-token dance).
- This slice is a refactor only (`GitHubOAuthProvider` → `GitHubTokenService` + `GitHubAuthController`) — no registration changes.

### ADO provider — app registration (existing, source: `azure-devops-mcp`)
- **App registration:** Microsoft's `azure-devops-mcp` public client — client id `0d50963b-7bb9-4fe7-94c7-a99af00b5136` (see `azure-devops-mcp/src/auth.ts`); no client secret (public client, PKCE).
- **Authority:** `https://login.microsoftonline.com/common` by default; tenant-specific authority when a tenant id is configured (current `Auth:Ado:TenantId` = `common`).
- **Scopes:** `499b84ac-1321-427f-aa17-267ca6975798/.default` (Azure DevOps resource) + `offline_access` — MSAL interactive browser flow with loopback redirect, silent refresh via cached account (~1h access tokens).
- This slice is a refactor only (`AdoOAuthProvider` → `AdoTokenService` + `AdoAuthController`) — no registration changes.

### M365 provider — backed by the existing `EmailMcpServer` app registration
- **App (client) ID:** `1d999ae2-a1f6-44ca-b733-c3df6ce8dc0c` · **Tenant (mcqdb):** `d6f8c4a4-cd9a-4551-b1cd-0f79d1f7fa12` · multi-tenant (work accounts)
- **Client model:** confidential web client — auth-code flow with **client secret** (Secret_1, expires 2027-04-23). Secret value supplied via .NET user-secrets / environment (`Auth:M365:ClientSecret`); **never committed, never logged**. PKCE on top of the secret (recommended hardening).
- **Redirect URI / ports:** OAuth callbacks should **share the application's primary port** wherever the registration permits — the M365 app registration's redirect URI will be updated from `http://localhost:3333/auth/callback` to an app-hosted callback (e.g. `http://localhost:5000/auth/m365/callback`), served by a normal controller action. No separate `HttpListener`, no fixed port 3333. GitHub/ADO keep their existing first-party loopback flows (gh CLI app dynamic loopback; MSAL owns ADO's listener) — separate ephemeral listener ports are acceptable there when not shareable with the app port.
- **Authority:** `https://login.microsoftonline.com/d6f8c4a4-cd9a-4551-b1cd-0f79d1f7fa12` (tenant-pinned; app is multi-tenant so `organizations` also works — start tenant-pinned).
- **Token lifetimes / rotation:** access tokens ~60-90 min; refresh tokens rotate on every redemption — **persist the newest refresh_token each time** (token store record update on every refresh, not just sign-in). 90-day inactivity expiry.
- **Starter delegated scopes** (of the 39 granted): `offline_access`, `User.Read`, `Mail.Read`, `Calendars.Read`, `OnlineMeetings.Read`, `OnlineMeetingTranscript.Read.All`, `OnlineMeetingRecording.Read.All` — configurable via `Auth:M365:Scopes`.
- **Caveats to document in `AuthProviderGuide.md`:** delegated Teams transcript/recording scopes are organizer-only; some scopes are admin-restricted (admin consent: `az ad app permission admin-consent --id 1d999ae2-...`).

### Egress webhook
- `OAuthProviderHosts` learns `m365` → `graph.microsoft.com`; the gateway can then route sandboxed Graph calls through `POST /api/auth/webhook/m365` for bearer injection, identical to GitHub/ADO.

## Out of scope

- App-only (client-credentials) Graph flows — delegated only.
- New UI for M365 sign-in (reuse the existing provider sign-in UI pattern).
- EmailMcpServer itself — this issue only consumes its app registration.

## Acceptance Criteria

- [ ] `POST api/auth/ado/signin`, `GET api/auth/ado/status`, `POST api/auth/ado/signout` served by a dedicated `AdoAuthController` + `AdoTokenService`; existing ADO sign-in flow still works end-to-end.
- [ ] Same for GitHub via `GitHubAuthController` + `GitHubTokenService`.
- [ ] New `M365AuthController` + `M365OAuthProvider`: full auth-code sign-in against the EmailMcpServer registration (confidential client + PKCE, callback hosted by the app on its primary port, e.g. `GET /auth/m365/callback`) → `SignedIn` status with account UPN.
- [ ] M365 access tokens auto-refresh; **rotated refresh tokens are persisted on every redemption** (verified by test).
- [ ] Client secret is sourced from user-secrets/env only — repo contains no secret value; secret never appears in logs or HTTP responses.
- [ ] `AuthWebhookController` injects M365 bearer tokens for `graph.microsoft.com` egress; deny (403) when not signed in.
- [ ] No UI-facing endpoint returns token material (status/challenge only) — preserved across the split.
- [ ] Browser-navigable `GET /auth/{provider}`: starts the sign-in flow when signed out; shows a success page (provider, account, scopes, expiry) when already signed in — works from the address bar for all three providers.
- [ ] Refresh tokens are persisted across app restarts (ADO, M365); interactive re-auth happens **only** when the refresh token is invalid/expired/revoked.
- [ ] `GetAccessTokenAsync` transparently refreshes expired/near-expiry access tokens from the stored refresh token — no interactive prompt for a refreshable session (verified by test).
- [ ] `AuthProviderGuide.md` documents all three providers incl. M365 setup, scopes, organizer-only + admin-consent caveats.

## Open questions

1. Should any service expose a raw-token endpoint for local tooling (e.g. `GET api/auth/{provider}/token`)? **Default: no** — webhook-injection only, consistent with the current security posture.

## Priority

**Medium**

## Related

- #72, #68, #71 (closed) — sandbox egress token injection + OAuth providers groundwork

gautam-achieveai:
Closes #82

## Summary

Splits the monolithic `AuthController` into per-provider controllers, extracts shared PKCE helpers, and adds a Microsoft 365 (Graph) OAuth provider so the sample can mint Graph tokens for sandbox tools that need M365 (Outlook, Teams, SharePoint, etc.). Also adds a small browser landing page at `GET /auth/{provider}` so an operator can kick off sign-in from a tab and watch the status update inline.

## Commits

1. **refactor(sample/auth): split monolithic AuthController into per-provider controllers** — `AdoAuthController` + `GitHubAuthController` each handle their own provider, with concrete-type DI and `IOAuthTokenProvider` alias-to-concrete dual registration so existing consumers keep working.
2. **refactor(sample/auth): extract shared PkceHelper for code-verifier + S256** — single source of truth for RFC 7636 PKCE so ADO/GitHub/M365 don't drift.
3. **feat(sample/auth): add Microsoft 365 (Graph) OAuth provider** — `M365OAuthProvider` over MSAL.NET `IConfidentialClientApplication` with file-backed user token cache, app-hosted callback on the primary Kestrel port (`/auth/m365/callback`), single-use state map (10-minute TTL via `TimeProvider`), `offline_access` at the authorize URL only. New `M365AuthController` mirrors the ADO/GitHub shape. Configured via `Auth:M365:*`. 11 unit tests cover authorize-URL shape, state replay, expiry, no-code, and unconfigured paths.
4. **feat(sample/auth): add browser entry route GET /auth/{provider}** — `AuthPagesController` returns a 404 on unknown providers (no enumeration leak), a "signed in" page if already authenticated, an inline "unavailable" page if the provider is not configured, or a small self-contained poll page that hits `/api/auth/{provider}/status` every 1.5s until sign-in resolves. All output is HTML-encoded; the provider id is JSON-serialized into the inline script and `encodeURIComponent`-ed into the URL.
5. **feat(sample/auth): wire M365 into sandbox auth_providers + network rules** — `OAuthProviderHosts` gains `m365 → graph.microsoft.com` as the single source of truth used by both the sandbox spec and the webhook's defense-in-depth host check. `SandboxSessionRegistry.BuildAuthProviders` emits the m365-auth webhook + 443 allow rule gated on both ClientId AND ClientSecret being present.
6. **docs(sample/auth): document per-provider controllers + M365 provider** — `AuthProviderGuide.md` refreshed with the new flow diagram, Entra app-reg prerequisites, M365 config table, env-var overrides, browser-entry section, and troubleshooting for 409 / AADSTS50011 / organizer-only Teams meetings.

## Security notes

- Browser entry returns 404 (not 400) on unknown providers so a typo can't enumerate which provider ids are valid.
- Inline poll page never surfaces token material — only provider id, sign-in state, and (for failures) an HTML-encoded OAuth error code.
- M365 webhook keeps the existing always-200 contract and applies the `OAuthProviderHosts.IsAllowed` check as defense-in-depth before minting a token.
- Stale/replay callbacks no longer demote an already-SignedIn user to Failed — the MSAL cache still holds valid tokens, so the visible status is preserved.
- M365 client secret lives in dotnet user-secrets / env vars — never in source.

## Test plan

- [x] `dotnet build LmDotnetTools.sln` — 0 warnings, 0 errors
- [x] `dotnet test tests/LmStreaming.Sample.E2E.Tests` — 58 pass / 2 skipped / 0 failed
- [x] New `M365OAuthProviderTests` (11 cases) covers authorize URL shape, single-use state, replay, expiry, unconfigured paths
- [x] New `AuthPagesControllerTests` (3 cases) covers 404, unavailable rendering, HTML-encoding
- [x] Extended `AuthControllersHostBuildTests` proves M365 concrete + `IOAuthTokenProvider` alias resolve to the same instance and route under their own base
- [x] Extended `AuthWebhookControllerTests` covers m365 not-signed-in deny + disallowed-host defense-in-depth deny
- [ ] Manual: real Entra app reg → `dotnet user-secrets set Auth:M365:ClientId/ClientSecret/TenantId` → `dotnet run` → visit `http://localhost:5000/auth/m365` → confirm browser flow + status page updates


gautam-achieveai:

Preview stopped.

gautam-achieveai (commented):
## [Gautam's review bot] PR Review — Per-provider auth controllers + M365 (Graph) provider

**Work Item:** #82 — Per-provider auth token services + controllers: ADO, GitHub, new M365

---

### Overall Assessment: :white_check_mark: APPROVE (with suggestions)

Well-structured PR with clean commit separation. The architecture is sound — the split from monolithic dispatch to typed per-provider controllers improves compile-time safety and Swagger documentation. Security implementation is solid: PKCE is RFC 7636 compliant, OAuth state is single-use + time-bounded + 256-bit random, client secrets are never logged, and all HTML output is properly encoded. The dual-registration DI pattern is a well-known .NET technique and is correctly applied.

---

### Findings

#### 1. :yellow_circle: MEDIUM — Cross-provider coupling via `AdoOAuthProvider.StripReservedScopes`

**File:** `M365OAuthProvider.cs:59`

M365OAuthProvider reaches into `AdoOAuthProvider` to call a static utility that strips OIDC reserved scopes. This creates lateral coupling between peer providers. If `AdoOAuthProvider` evolves its scope logic, M365 silently inherits it.

**Suggestion:** Move `StripReservedScopes` + the `ReservedScopes` HashSet to `OAuthProviderBase` or a standalone `MsalScopeHelper` class.

---

#### 2. :yellow_circle: MEDIUM — Route attribute inconsistency on `M365AuthController`

Ado/GitHub controllers use class-level `[Route("api/auth/{provider}")]` + relative action routes. M365 uses per-action full-path routes AND mixes two path prefixes (`api/auth/m365/*` vs `auth/m365/callback`) on the same controller.

**Suggestion:** Add `[Route("api/auth/m365")]` at class level with relative routes; use a leading-slash override (`[HttpGet("/auth/m365/callback")]`) for the callback endpoint.

---

#### 3. :speech_balloon: LOW — Controller action duplication

`AdoAuthController` and `GitHubAuthController` are character-for-character identical in their SignIn/Status/SignOut bodies (~33 lines). M365 repeats the same plus a unique Callback. For 3 providers this is defensible, but if a 4th is added or error handling changes, all must be updated in lockstep. Consider a thin `OAuthControllerBase<TProvider>` if provider count grows.

---

### Security :white_check_mark:

| Check | Status |
|-------|--------|
| OAuth state (CSRF) — single-use, 10-min TTL, 256-bit random | :white_check_mark: |
| PKCE — RFC 7636 compliant | :white_check_mark: |
| Client secret never logged | :white_check_mark: |
| XSS — all user input HTML-encoded or JSON-serialized | :white_check_mark: |
| Open redirect — none, only inline HTML | :white_check_mark: |
| Token material never surfaces to browser | :white_check_mark: |
| Defense-in-depth host check on webhook | :white_check_mark: |

---

### Tests :white_check_mark:

Comprehensive coverage: M365OAuthProviderTests (11), PkceHelperTests (6), AuthControllersHostBuildTests (3), AuthPagesControllerTests (3), AuthWebhookControllerTests (+2 m365 cases).

---

*Review agents: architecture-review, security-review, duplicate-code-detector*

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
```diff
@@ -0,0 +1,363 @@
...
+        string tokenCacheFilePath,
+        ILogger<M365OAuthProvider> logger,
+        TimeProvider? time = null)
+        : base(logger)
+    {
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _cacheFilePath = tokenCacheFilePath ?? throw new ArgumentNullException(nameof(tokenCacheFilePath));
+        _time = time ?? TimeProvider.System;
+
+        _scopes = AdoOAuthProvider.StripReservedScopes(_options.Scopes);
```
gautam-achieveai: **[Gautam's review bot]** :yellow_circle: Cross-provider coupling — `AdoOAuthProvider.StripReservedScopes` is a general MSAL utility, not ADO-specific. This creates a lateral dependency between peer providers.

Consider moving this to `OAuthProviderBase` (both MSAL providers inherit from it already) or a standalone `MsalScopeHelper`:

```csharp
// In OAuthProviderBase:
protected static string[] StripReservedScopes(IEnumerable<string> scopes) =>
    [.. scopes.Where(s => !ReservedMsalScopes.Contains(s))];

private static readonly HashSet<string> ReservedMsalScopes = new(StringComparer.OrdinalIgnoreCase)
{
    "openid", "profile", "offline_access",
};
```

samples/LmStreaming.Sample/Controllers/M365AuthController.cs
```diff
@@ -0,0 +1,95 @@
...
+/// authorization callback lands on this same app's primary port at
+/// <see cref="M365AuthOptions.RedirectPath"/> (no standalone HttpListener).
+/// </summary>
+/// <remarks>
+/// SECURITY: these endpoints never surface token material — only the UI-safe
+/// <see cref="OAuthStatus"/> record and the <see cref="SignInChallenge"/> (URL that was opened).
+/// The callback action returns an inline HTML page with no token information.
+/// </remarks>
+[ApiController]
+public sealed class M365AuthController(
```
gautam-achieveai, line 19: **[Gautam's review bot]** :yellow_circle: Routing convention inconsistency — This controller uses per-action full-path routes (`[HttpPost("api/auth/m365/signin")]`) while `AdoAuthController` and `GitHubAuthController` use class-level `[Route("api/auth/{provider}")]` + relative action routes.

Additionally, it mixes two path prefixes on one controller (`api/auth/m365/*` for API endpoints, `auth/m365/callback` for the browser callback).

Suggestion: align with siblings via class-level `[Route("api/auth/m365")]` + relative routes, and use a leading `/` on the callback to override the class route:

```csharp
[ApiController]
[Route("api/auth/m365")]
public sealed class M365AuthController(...)
{
    [HttpPost("signin")] ...
    [HttpGet("status")] ...
    [HttpPost("signout")] ...
    [HttpGet("/auth/m365/callback")]  // leading '/' overrides class route
    public async Task<IActionResult> Callback(...) { ... }
}
```

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
gautam-achieveai: [Gautam's Dev bot]

Addressed in ad81cc2 — moved `ReservedMsalScopes` + `StripReservedScopes` from `AdoOAuthProvider` to `OAuthProviderBase`. Both ADO and M365 now call the helper via inheritance, no peer-to-peer reach. Updated the `AdoOAuthProviderTests` call site to `OAuthProviderBase.StripReservedScopes`.

samples/LmStreaming.Sample/Controllers/M365AuthController.cs
gautam-achieveai: [Gautam's Dev bot]

Addressed in ad81cc2 — moved `M365AuthController` to class-level `[Route("api/auth/m365")]` with relative `[signin]`/`[status]`/`[signout]` (matches ADO/GitHub). The callback uses a leading-slash override `[/auth/m365/callback]` so the Entra-registered callback path stays the default and isn't affected by the class prefix.

samples/LmStreaming.Sample/Controllers/AuthPagesController.cs
```diff
@@ -0,0 +1,112 @@
...
+        // Inline poll script — checks /api/auth/{providerId}/status every 1.5s and updates the page
+        // when the state changes. Self-contained (no external script tag, no token material).
+        // providerId is embedded via System.Text.Json.JsonSerializer.Serialize so any unusual
+        // character is escaped as a JSON string literal (not interpolated as raw JS).
+        var script = "<script>"
+            + "const PROVIDER=" + System.Text.Json.JsonSerializer.Serialize(providerId) + ";"
+            + "async function poll(){try{const r=await fetch('/api/auth/'+encodeURIComponent(PROVIDER)+'/status');if(!r.ok)return;"
+            + "const s=await r.json();const el=document.getElementById('state');if(!el)return;"
+            + "el.textContent=s.state;"
+            + "if(s.state==='SignedIn'){document.getElementById('msg').textContent='Signed in. You can close this tab.';return;}"
```
gautam-achieveai, line 107: [Gautam's bot] **B1 — HIGH (blocker):** This comparison never matches. `OAuthSignInState` has no `[JsonConverter]` and MVC is bare `AddControllers()` (Program.cs:105), so `/api/auth/{p}/status` serializes the enum as an **integer** — `{"state":2}`, never `"SignedIn"`. Both branches are dead: the page renders "Status: 2" and polls forever for every provider, breaking the headline AC of #82.

**Fix (attribute-scoped, not global):** `[JsonConverter(typeof(JsonStringEnumConverter))]` on `OAuthSignInState`. Verified safe — the Vue client has zero `api/auth/*` references and the only tests hitting `/status` assert status codes, so this page is the JSON's only consumer. A global `AddJsonOptions` converter would change the wire format for every MVC endpoint; don't. Add a round-trip test asserting `state` is the **string** name to pin the poll contract.

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
```diff
@@ -0,0 +1,363 @@
...
+
+        // Leave the provider disabled (no MSAL app) when unconfigured — sign-in/token calls then
+        // surface a clear "not configured" error on demand instead of failing host startup.
+        if (string.IsNullOrWhiteSpace(_options.ClientId) || string.IsNullOrWhiteSpace(_options.ClientSecret))
+        {
+            return;
+        }
+
+        var authority = $"https://login.microsoftonline.com/{_options.TenantId}";
+        _app = ConfidentialClientApplicationBuilder
```
gautam-achieveai: [Gautam's bot] **B2 — HIGH (blocker):** The confidential client is built without MSAL's `IMsalHttpClientFactory` seam, so the token endpoint cannot be faked — and the issue's "verified by test" ACs (rotated refresh tokens persisted on every redemption; transparent silent refresh) plus plan commit-3's promised tests (redemption success/error, cache write-on-acquire/read-on-restart, `AcquireTokenSilent`) are all missing. `M365OAuthProviderTests` covers only pre-redemption guards. Contrast `GitHubOAuthProviderTests`, which fully exercises redemption + rotation + restart via `FakeHttpMessageHandler`.

**Fix:** optional ctor param → `.WithHttpClientFactory(...)`, then add: (a) fake redemption success → `SignedIn` + cache file written; (b) second provider over same cache dir → hydrate restores `SignedIn`; (c) silent refresh with rotated token rewrites the cache; (d) `invalid_grant` → mapped error, no token material in the error string.

**Named must-fix within this work:** `OnBeforeCacheAccess` (:332-341) does an unguarded `DeserializeMsalV3(File.ReadAllBytes(...))` — a corrupt/truncated `msal-m365.bin` throws on the token-acquire path and wedges a previously signed-in user. Guard (treat as empty cache + warn) and cover it.

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        // See SignOutAsync — single-user file cache, GetAccountsAsync is the right call.
+#pragma warning disable CS0618
+        var accounts = await _app.GetAccountsAsync().ConfigureAwait(false);
+#pragma warning restore CS0618
+        return accounts.FirstOrDefault();
+    }
+
+    private void SweepExpired()
```
gautam-achieveai, line 361: [Gautam's bot] **B3 — HIGH (blocker):** `SweepExpired` only prunes map entries — nothing ever flips the visible status when a flow is abandoned (user closes the Entra tab), so it stays `Pending` forever and the landing page has no terminal state on the timeout path. GitHub has `SignInTimeout` → `sign_in_timeout`; M365 has no equivalent. Co-requisite of B1 — even with the enum fixed, the page polls forever here.

**Fix:** when sweeping an expired entry while status is still `Pending`, `SetFailed("sign_in_timeout")`, mirroring GitHub.

samples/LmStreaming.Sample/Controllers/AuthPagesController.cs
```diff
@@ -0,0 +1,112 @@
...
+            // Provider is registered but not configured (e.g. missing ClientId/secret). Surface the
+            // unconfigured error inline rather than a 409 so the user sees what to fix.
+            logger.LogInformation("OAuth sign-in unavailable for provider {ProviderId}.", provider.ProviderId);
+            return Content(BuildUnavailableHtml(provider.ProviderId, ex.Message), "text/html");
+        }
+
+        return Content(BuildPendingHtml(provider.ProviderId), "text/html");
+    }
+
+    private static string BuildSignedInHtml(string providerId, string? account)
```
gautam-achieveai: [Gautam's bot] **MEDIUM (non-blocking):** The AC enumerates "(provider, account, scopes, expiry)" for the success page; `BuildSignedInHtml` renders 2 of 4. `OAuthStatus` already carries `Scopes` and `ExpiresAtUtc` — trivial to add both lines.

samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
```diff
@@ -670,6 +670,37 @@ internal static string StripLongPathPrefix(string path)
             );
         }
 
+        // M365 (Microsoft Graph). Gate on both ClientId and ClientSecret: the provider stays
+        // disabled when the secret is missing, so emitting the gateway rule + webhook entry in that
+        // case would point to a webhook that always denies — confusing rather than helpful.
+        if (!string.IsNullOrWhiteSpace(_authOptions.M365.ClientId)
```
gautam-achieveai, line 685: [Gautam's bot] **B5 — MEDIUM (blocker):** Plan commit 5 promised to "narrowly assert `BuildAuthProviders` output via a small unit" — nothing covers the `m365-auth` DTO emission, endpoint string, hosts list, or this ClientId+ClientSecret dual gate. This is the production-pattern path deciding what credentials cross the sandbox boundary, and the secret-gating predicate is exactly the kind that silently degrades under future edits.

Two small cases suffice: both-present → emitted with expected shape; secret-missing → not emitted.

gautam-achieveai (commented):
## [Gautam's bot] PR Review — Per-provider auth controllers + M365 (Graph) provider

**Work item:** #82 · **Head:** ad81cc2 · **Baseline:** approved plan v2 · Prior round's 2 findings verified fixed in ad81cc2 ✅

---

### Verdict: 🔴 REQUEST CHANGES — 5 blockers (4 HIGH, 1 MEDIUM)

The structure is right and stays faithfully inside the approved plan (11/13 items fully compliant, zero over-engineering found, all 8 security claims in the PR body verified TRUE). But the headline AC of #82 — the browser-navigable `/auth/{provider}` flow — is functionally broken as shipped, and the plan's commit-3/4/5 test promises didn't land. The misses are connected: the bug shipped exactly through the biggest untested hole.

---

### Blockers

#### B1 — HIGH — Landing page never resolves: poll script compares strings, API serializes the enum as an integer

`Controllers/AuthPagesController.cs:92` · `Program.cs:105` · `Services/Auth/IOAuthTokenProvider.cs:4`

The poll script checks `s.state==='SignedIn'` / `'Failed'`, but `OAuthSignInState` has no `[JsonConverter]` and MVC is bare `AddControllers()` — System.Text.Json's MVC default serializes enums as **integers**. `/api/auth/{p}/status` returns `{"state":2}`; both comparisons are dead. The page renders "Status: 2" and polls forever — it never flips to success or failure for any provider. That breaks the AC: "Browser-navigable `GET /auth/{provider}` … works from the address bar for all three providers."

**Fix (attribute-scoped, not a global converter):** `[JsonConverter(typeof(JsonStringEnumConverter))]` on `OAuthSignInState`. Verified safe: the Vue client has zero references to `api/auth/*`, and the only tests hitting `/status` assert status codes, not JSON shape — this page is the JSON's only consumer. A global `AddJsonOptions` converter would change the wire format for every MVC endpoint; don't. Add a test that round-trips `GET /api/auth/{p}/status` and asserts `state` is the **string** name — that pins the poll contract.

#### B2 — HIGH — "verified by test" ACs are unmet: redemption, cache round-trip, rotation, and silent refresh are all untested (and currently untestable)

`Services/Auth/M365OAuthProvider.cs:70-78`

Issue ACs say "rotated refresh tokens are persisted on every redemption (**verified by test**)" and "GetAccessTokenAsync transparently refreshes … (**verified by test**)"; plan commit 3 promised tests for redemption success + error, MSAL cache write-on-acquire/read-on-restart, and `AcquireTokenSilent` lazy refresh. None exist — `M365OAuthProviderTests` covers only pre-redemption guards (state/expiry/no-code/not-configured). Root cause: the confidential client is built without MSAL's `IMsalHttpClientFactory` seam, so the token endpoint cannot be faked. Contrast `GitHubOAuthProviderTests`, which fully exercises redemption + rotation + restart via `FakeHttpMessageHandler`.

**Fix:** optional ctor param → `.WithHttpClientFactory(...)`, then add: (a) fake auth-code redemption success → `SignedIn` + cache file written; (b) second provider over the same cache dir → hydrate restores `SignedIn` (read-on-restart); (c) silent refresh returning a rotated token rewrites the cache file; (d) token-endpoint `invalid_grant` → mapped error code, no token material in the error string.

**Named must-fix within this work:** `OnBeforeCacheAccess` (`M365OAuthProvider.cs:332-341`) does an unguarded `DeserializeMsalV3(File.ReadAllBytes(...))` — a corrupt/truncated `msal-m365.bin` (crash mid-write, disk full) throws on the token-acquire path and wedges a previously signed-in user. Guard it (treat as empty cache + warn) and cover it.

#### B3 — HIGH — Abandoned M365 sign-in is stuck `Pending` forever; the landing page has no terminal state

`Services/Auth/M365OAuthProvider.cs:320-330`

`SweepExpired` only prunes map entries — nothing ever flips the visible status to `Failed` when a flow is abandoned (user closes the Entra tab). GitHub has `SignInTimeout` → `sign_in_timeout`; M365 has no equivalent. Co-requisite of B1: even with the enum fixed, the page polls forever on the timeout path.

**Fix:** when sweeping an expired entry while status is still `Pending`, `SetFailed("sign_in_timeout")`, mirroring GitHub.

#### B4 — HIGH — Plan commit-4 tests missing: the signed-in and landing branches of `AuthPagesController` have zero coverage

`tests/LmStreaming.Sample.E2E.Tests/Scenarios/AuthPagesControllerTests.cs`

Only unknown→404 and unconfigured→unavailable are tested. The promised "signed-in → 200 + success HTML" and "signed-out → `BeginSignInAsync` called once + landing HTML" tests are absent — and the untested landing branch is exactly where B1 shipped. These must land with B1's fix or the same blind spot persists. Suggested: a stub `IOAuthTokenProvider` seeded to `SignedIn` → assert success HTML (incl. the B1 contract); a configured stub recording `BeginSignInAsync` calls → assert called exactly once + poll script present.

#### B5 — MEDIUM (blocking) — Sandbox egress m365 block unpinned: `BuildAuthProviders` has no test

`Services/SandboxSessionRegistry.cs:676-677`

Plan commit 5: "narrowly assert `BuildAuthProviders` output via a small unit." Nothing covers the `m365-auth` DTO emission, endpoint string, hosts list, or the ClientId+ClientSecret dual gate. This is the production-pattern path deciding what credentials cross the sandbox boundary; the secret-gating predicate is exactly the kind that silently degrades under future edits. Two small cases suffice: both-present → emitted with expected shape; secret-missing → not emitted.

---

### Non-blocking (fix here or note for follow-up)

- **[MEDIUM] Success page omits scopes + expiry** — the AC enumerates "(provider, account, scopes, expiry)"; `BuildSignedInHtml` (`AuthPagesController.cs:61-70`) renders 2 of 4. `OAuthStatus` already carries both — trivial.
- **[LOW] Guide misses the state-map limitation** — the in-memory state→{verifier, expiry} map is single-process (plan risk item + commit-6 doc requirement); it's in the XML-doc but not `AuthProviderGuide.md`. While there: add that for any non-localhost lift, the plain-HTTP unauthenticated callback becomes a hard HTTPS/loopback-only requirement.
- **[LOW] Stale callback can demote a concurrent `Pending` flow** — `CompleteSignInAsync` guards only `SignedIn` (`M365OAuthProvider.cs:209-216`); a foreign/stale state flips a live `Pending` to `Failed` mid-flow. Availability-only, no token impact.
- **[LOW] Two divergent HTML emitters in one slice** — `M365AuthController.BuildHtml/WebEncode` (`:87-94`) vs `AuthPagesController.Page/Encode` (`:104-111`); share one tiny internal helper so callback and landing markup/encoding can't drift (~8 lines of dedup; Razor stays correctly out).
- **[LOW] Status surfaces unstripped scopes** — `OAuthStatus.Scopes` gets `_options.Scopes` (incl. `offline_access`) while every MSAL call uses stripped `_scopes` (`:59` vs `:101/:135/:246`). Surface `_scopes`. (ADO has the same pre-existing pattern — aligning it is optional here.)
- **[LOW] Dead/over-exposed members** — `CallbackBaseUrl` is `internal` + "Test-visible" with zero external consumers (`:84-85`); `PendingSignInTtl` `internal` but externally unused (`:31`); `ManualTimeProvider.Advance` dead in tests (`M365OAuthProviderTests.cs:244`).

---

### Verified good

- **Plan compliance:** 11/13 fully compliant; both prior-round findings genuinely fixed in ad81cc2 (`StripReservedScopes` on the base with no peer reach from M365; M365 routes aligned with the callback kept at the Entra-registered path).
- **All 8 security claims in the PR body verified TRUE** with line-level evidence: 404 anti-enumeration · token-free pages · always-200 webhook with defense-in-depth host gate · SignedIn demotion guard · secret never committed/logged · single-use state + 10-min TTL via `TimeProvider` · `offline_access` at authorize-URL only · dual ClientId+ClientSecret sandbox gate.
- **Security posture:** PKCE 256-bit + S256 with an RFC 7636 Appendix B vector test (exemplary), XSS encoding correct everywhere it matters, no open redirect, authorize URL via `QueryHelpers`, constant-time shared-secret compare, msal/token files gitignored, `login.microsoftonline.com` correctly absent from sandbox egress.
- **No over-engineering:** the implementation stayed inside the approved, already-trimmed plan — no speculative abstractions, all new options consumed, DI dual-registration correct (same-instance pinned by `AuthControllersHostBuildTests`, which genuinely exercises controller activation, not just `BuildServiceProvider`).
- **Build/test:** 0 warnings with `-warnaserror`; 58/60 pass — the 2 failures are pre-existing Docker-gated `SandboxWorkspaceGatewayE2ETests`, unrelated to this PR.

---

### Why REQUEST CHANGES rather than approve-with-comments

B1 is a verified functional break of the headline AC; B2 means the PR claims test coverage that doesn't exist. B3/B4 are not separable from B1 — fixing the enum without the timeout path and the branch tests leaves the same AC half-broken and the same blind spot open, burning a second round. One bundled fix round closes all five.

*Lenses: over-engineering + class design (requested focus), plan-v2 compliance, OAuth security, test coverage; findings graded before verdict.*

gautam-achieveai (commented):
## [Gautam's review bot] Incremental Review — Iteration 7 (ad81cc2)

### Addressed Findings ✅

Both round-1 suggestions are correctly fixed in `ad81cc2`:

1. **StripReservedScopes → OAuthProviderBase** — `ReservedMsalScopes` + `StripReservedScopes` now live on the shared base. M365 calls via inheritance (`StripReservedScopes(_options.Scopes)`), no lateral reach. Test updated to `OAuthProviderBase.StripReservedScopes`. Clean. ✅

2. **M365AuthController routing aligned** — Class-level `[Route("api/auth/m365")]` + relative action routes, with a leading-slash `[HttpGet("/auth/m365/callback")]` override for the Entra callback. Matches ADO/GitHub exactly. XML doc explains the override clearly. ✅

### No New Issues

The 5-file, 26-line diff is surgical — no regressions, no new concerns.

### Outstanding Blockers (B1–B5)

The 5 findings from the deep review (posted earlier today) remain unaddressed — they were filed after this commit landed:

| ID | Severity | Summary |
|----|----------|---------|
| B1 | HIGH | `OAuthSignInState` serializes as integer → poll script dead, landing page broken for all providers |
| B2 | HIGH | Redemption/cache/rotation tests missing and untestable (no `IMsalHttpClientFactory` seam) |
| B3 | HIGH | Abandoned M365 sign-in stuck `Pending` forever — no timeout → `Failed` transition |
| B4 | HIGH | `AuthPagesController` signed-in + landing branches have zero test coverage |
| B5 | MEDIUM | `BuildAuthProviders` m365 block (sandbox egress gate) has no unit test |

### Verdict: 🔴 REQUEST CHANGES

The two addressed fixes are clean and correct. The PR remains blocked on B1–B5 from the prior review round. One bundled fix iteration should close all five.

samples/LmStreaming.Sample/Controllers/AuthPagesController.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in ba42714. OAuthSignInState now carries an attribute-scoped `[JsonStringEnumConverter]` (samples/LmStreaming.Sample/Services/Auth/IOAuthTokenProvider.cs:12) so the wire shape is pinned to the string name regardless of host MVC defaults. AuthControllersHostBuildTests gains a `[Theory]` over `/api/auth/{github,ado,m365}/status` that asserts `state` is a JSON string and one of the four documented names — that's the contract the landing-page poll script depends on.

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in ba42714 across two parts.

**Part 1 — corrupt-cache guard** (samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs:362-396): `OnBeforeCacheAccess` now reads the file under a try/catch for IO errors and a separate try/catch around `DeserializeMsalV3` for malformed bytes. Either path logs a warning and continues with an empty cache — the user re-signs once and the cache is rewritten. New test `Hydrate_with_corrupt_cache_file_is_tolerated_and_leaves_not_started` writes garbage bytes and verifies `HydrateFromStoreAsync` returns without throwing.

**Part 2 — `IMsalHttpClientFactory` seam** (M365OAuthProvider.cs:52-89): optional ctor parameter threads through to `.WithHttpClientFactory(...).WithInstanceDiscovery(false)` on the confidential-client builder. Three new tests in M365OAuthProviderTests cover the seam:
- `CompleteSignInAsync_routes_token_exchange_through_injected_msal_http_factory` — pins that MSAL's POST to `/oauth2/v2.0/token` is actually routed through the injected handler.
- `CompleteSignInAsync_maps_invalid_grant_response_to_status_failed_with_error_code` — covers AC (d): the OAuth error code (not the description) is what surfaces.
- `CompleteSignInAsync_maps_token_endpoint_5xx_to_exchange_failed` — covers the catch-all branch and asserts the response body is not leaked into `Status.Error`.

Deferred from the original AC: a full happy-path test that fabricates a valid id_token + writes the cache file end-to-end (AC b/c). MSAL's id_token validation needs JWT construction + JWKS staging which is a larger harness; the current three tests cover the seam contract and the error-mapping path that previously had zero coverage. Happy to add the JWT fixture in a follow-up if you want it pinned here.

samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in ba42714. `BeginSignInAsync` (samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs:111-145) now schedules a `TimeProvider`-driven background deadline via `StartBackgroundSignInAsync`. `RunSignInDeadlineAsync` awaits `Task.Delay(PendingSignInTtl, _time, token)` then, if the state entry is still pending and Status is still `Pending`, calls `SetFailed("sign_in_timeout")`.

Race-safety: `_pending.TryRemove` is the gate — `CompleteSignInAsync` always removes the entry on success/failure, so an in-flight redemption clears it before the deadline fires; the additional `Status.State == Pending` check stops the deadline from demoting a `SignedIn` status. The pattern mirrors GitHub's `CancelAfter` deadline, just routed through the controller-driven callback path instead of a loopback listener.

samples/LmStreaming.Sample/Controllers/AuthPagesController.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in ba42714. `BuildSignedInHtml` (samples/LmStreaming.Sample/Controllers/AuthPagesController.cs:71-89) now renders the OAuthStatus's `Scopes` (space-joined) and `ExpiresAtUtc` (ISO-8601 invariant) in addition to the provider id + account. Both fields are HTML-encoded; the lines collapse to empty strings when the provider exposed no scopes or no expiry (e.g. M365 hydrate). New `AuthPagesControllerTests.Signed_in_branch_renders_provider_account_scopes_and_expiry` pins all four fields show up in the rendered body.

samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in ba42714. Added an internal test seam `SandboxSessionRegistry.GetAuthProviderIdsForTest()` (samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs:614-621) that returns just the provider ids `BuildAuthProviders` would emit — no need to expose the private DTO records.

New `SandboxSessionRegistryBuildAuthProvidersTests` (5 tests, all green) pins the m365 gating contract:
- Both ClientId + ClientSecret set → `m365-auth` emitted.
- ClientSecret null → not emitted.
- ClientId null → not emitted.
- ClientSecret whitespace → not emitted (matches the `IsNullOrWhiteSpace` gate so appsettings placeholders don't slip through).
- Empty AuthOptions → no providers emitted at all.

Why the seam over making `BuildAuthProviders` internal: the DTOs are private, so exposing the method would leak record types into the test surface. The id-only seam is the minimum useful surface for testing the gating logic.

gautam-achieveai:
[Gautam's Dev bot]

**iter3 (ba42714) — addresses all 5 review-bot blockers + 1 MED**

| Finding | Status | Where |
|---|---|---|
| B1 (HIGH) — `OAuthSignInState` serializes as int | ✅ Fixed | `IOAuthTokenProvider.cs` `[JsonStringEnumConverter]`; theory pinning test |
| B2 (HIGH) — corrupt MSAL cache wedges user / no `IMsalHttpClientFactory` test seam | ✅ Fixed | `M365OAuthProvider.OnBeforeCacheAccess` guarded; ctor accepts `IMsalHttpClientFactory`; 3 new redemption/error tests |
| B3 (HIGH) — abandoned M365 sign-in stays `Pending` forever | ✅ Fixed | `BeginSignInAsync` schedules a `TimeProvider`-driven background deadline → `SetFailed("sign_in_timeout")` |
| B4 (HIGH) — `AuthPagesController` has no SignedIn / Landing tests | ✅ Fixed | 3 new unit tests over stub provider (signed-in, landing, unavailable) |
| B5 (MED-blocker) — `BuildAuthProviders` m365 dual-gate untested | ✅ Fixed | New `GetAuthProviderIdsForTest` seam + 5 gating tests |
| MED — success page omits scopes/expiry | ✅ Fixed | `BuildSignedInHtml` now renders both, HTML-encoded, ISO-8601 expiry |

**Build + tests:** 0 warn / 0 err. Unit tests **179/0**. E2E **68/0/2 skipped** (Docker-gated). Up from 58/0/2 — 10 net new tests across the auth area.

**Deferred (call-outs, not gating):**
- B2 AC parts (b) cache-roundtrip across two provider instances and (c) silent-refresh-rotates-cache need a JWT-fixture harness (id_token construction + JWKS) — happy to land that in a follow-up; the new seam tests already cover the error-mapping path that had zero coverage before.
- LOW `OAuthControllerBase<TProvider>` extraction still YAGNI for 3 providers.

Individual thread replies are in place for B1/B2/B3/B5/MED.

samples/LmStreaming.Sample/Controllers/AdoAuthController.cs
```diff
@@ -0,0 +1,54 @@
...
+    public IActionResult Status() => Ok(provider.Status);
+
+    /// <summary>Clears stored tokens and resets to the not-started state.</summary>
+    [HttpPost("signout")]
+    public async Task<IActionResult> SignOut(CancellationToken ct = default)
+    {
+        await provider.SignOutAsync(ct);
+        logger.LogInformation("Signed out OAuth provider {ProviderId}.", provider.ProviderId);
+        return NoContent();
+    }
```
gautam-achieveai, line 53: **[LOW — class design, non-blocking]** This `signin` / `status` / `signout` trio is byte-for-byte identical to `GitHubAuthController`, and the same three actions are repeated again in `M365AuthController` — only the route prefix and injected provider type differ. The copy-paste has **already drifted**: this file's catch comment reads `missing ClientId`, while `GitHubAuthController.cs:36` reads `missing ClientId / ClientSecret` for the identical code path. That divergence is exactly the maintenance cost duplication invites.

Consider collapsing the shared trio into one provider-agnostic controller routed by `{providerId}`, resolving the provider from keyed DI:

```csharp
[ApiController]
[Route("api/auth/{providerId}")]
public sealed class OAuthController(IOAuthProviderRegistry registry, ILogger<OAuthController> logger)
    : ControllerBase
{
    [HttpPost("signin")]
    public async Task<IActionResult> SignIn(string providerId, CancellationToken ct) { /* shared */ }
    // status / signout ...
}
```

`M365AuthController` would then shrink to just its `/auth/m365/callback` action.

Caveat: this is a **sample** app, where a little explicit per-provider duplication is a defensible teaching choice — hence non-blocking. Flagging it because the review brief was class-design / over-engineering, and the divergent comment shows the drift is real, not hypothetical.

samples/LmStreaming.Sample/Services/Auth/AuthOptions.cs
```diff
@@ -61,6 +64,47 @@ public sealed class AdoAuthOptions
...
+    /// Delegated Microsoft Graph scopes requested during the authorization-code flow. MSAL injects
+    /// reserved OIDC scopes (<c>openid</c>, <c>profile</c>, <c>offline_access</c>) itself.
+    /// </summary>
+    public string[] Scopes { get; set; } =
+    [
+        "User.Read",
+        "Mail.Read",
+        "Calendars.Read",
+        "OnlineMeetings.Read",
+    ];
```
gautam-achieveai: **[LOW — correctness nit, non-blocking]** These in-code default scopes combine with the *same* scopes in `appsettings.Development.json` to produce **duplicates** at runtime: the .NET configuration binder **appends** bound array elements onto a pre-populated `string[]` rather than replacing it. Live repro: `GET /api/auth/m365/status` returned all four scopes **twice** (8 entries). The same latent gotcha applies to `Github.Scopes` (line 48) and `Ado.Scopes` (line 64).

Sign-in itself is unaffected because `BuildAuthorizeUrl` does `.Distinct()`, so this only corrupts the UI-facing `status.scopes[]`.

Two fixes, ideally both:
1. **Report granted, not requested, scopes** — on `SignedIn`, surface MSAL's `result.Scopes` instead of `_options.Scopes` (`M365OAuthProvider.cs:283`). Reflects what was actually consented and sidesteps the dup entirely.
2. **Pick one source of truth** — drop the in-code initializer (`= ["User.Read", ...]`) so config replaces cleanly, or remove the scopes from `appsettings.Development.json` and keep the code default.

gautam-achieveai (commented):
## Review — live-tested; approving with two non-blocking nits

I built and ran the sample locally (`samples/LmStreaming.Sample` on `:5000`, Development) and drove all three providers through real interactive sign-in. Functionally this checks out end-to-end:

- **Wire shape:** `OAuthSignInState` serializes as a JSON **string** on both `NotStarted` and `SignedIn` — the enum-converter fix holds.
- **Status DTO:** account, scopes, and ISO-8601 `expiresAtUtc` surface correctly on `SignedIn`.
- **Security:** no token material crosses the UI boundary — `status`/`signin` return only `OAuthStatus` + the authorize URL, and the M365 callback page echoes only HTML-encoded OAuth error codes. The M365 client secret binds from env (`Auth__M365__ClientSecret`) and is never logged or surfaced.
- **Tokens:** MSAL code-redemption succeeds for M365 and ADO; both persist a **refresh token** (`offline_access` honored). GitHub correctly has **no** refresh token — it issues a classic non-expiring `gho_` user-to-server token, so there is nothing to refresh.

Two **LOW / non-blocking** items inline. Neither blocks merge.

_(I'm the PR author on this account, so GitHub won't let me click **Approve** — treat this as a ✅-with-nits.)_

samples/LmStreaming.Sample/Services/Auth/AuthOptions.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 7a10dbf — both halves of the suggestion.

**1. Report granted, not requested, scopes.** `M365OAuthProvider.CompleteSignInAsync` (samples/LmStreaming.Sample/Services/Auth/M365OAuthProvider.cs:283-289) now surfaces `result.Scopes.ToArray()` — MSAL's granted set from the token response — instead of `_options.Scopes`. Mirrored the same change on the ADO interactive path (samples/LmStreaming.Sample/Services/Auth/AdoOAuthProvider.cs:119-123). GitHub already wrote granted scopes from the token-record path.

**2. Pick one source of truth.** Dropped the in-code default `Scopes` arrays for `GitHubAuthOptions` / `AdoAuthOptions` / `M365AuthOptions` (samples/LmStreaming.Sample/Services/Auth/AuthOptions.cs:48,64,93). They now default to `[]`, so the config binder replaces cleanly instead of appending. Moved the Github + Ado defaults into `appsettings.Development.json` (the M365 section already had them).

Sign-in itself was always fine — `BuildAuthorizeUrl` already `.Distinct()`-deduped — so only the UI-facing `status.scopes[]` was affected. With both halves applied, the post-sign-in path is dup-proof at the source (granted scopes, not options) AND the hydrate-from-store / pending-state paths are dup-proof structurally (empty in-code default).

Build 0/0; unit 179/0; E2E 68/0/2 skipped.

samples/LmStreaming.Sample/Controllers/AdoAuthController.cs
gautam-achieveai: [Gautam's Dev bot]

Acknowledged. Two-part reply: addressed the drift you flagged; deferring the extraction.

**Drift fixed in 7a10dbf.** Standardised the catch comment across all three controllers to a single uniform line — `// Provider is registered but not configured (e.g. missing client id / secret).` — at samples/LmStreaming.Sample/Controllers/AdoAuthController.cs:36, samples/LmStreaming.Sample/Controllers/GitHubAuthController.cs:36, samples/LmStreaming.Sample/Controllers/M365AuthController.cs (newly added). The `InvalidOperationException.Message` from each provider already carries the provider-specific specifics, so the catch site stays neutral.

**Deferring the `OAuthController(string providerId)` extraction.** Concur with your caveat on the sample-app teaching-value angle, and adding the registry + keyed-DI plumbing changes the public DI surface for a sample where 3-of-a-kind reads more clearly than a `{providerId}`-templated dispatch. If/when a 4th interactive provider lands, extract `OAuthControllerBase<TProvider>` (or your registry shape) then — the trio is small enough that the YAGNI cost stays below the maintenance cost for now, and the catch-comment fix removes the immediate drift signal.

gautam-achieveai:
[Gautam's Dev bot]

**iter4 (7a10dbf) — addresses both LOW nits from the live-tested review**

| Finding | Status | Where |
|---|---|---|
| LOW — duplicate scopes in `status.scopes[]` (live repro: 8 entries) | ✅ Fixed | M365 + ADO surface MSAL granted scopes; AuthOptions defaults emptied (single source of truth = config) |
| LOW — controller body duplication + drift | ✅ Drift fixed / ⏸ Extraction deferred | Uniform catch comment across all three; `OAuthControllerBase<TProvider>` deferred pending a 4th provider (YAGNI for 3) |

Both fixes verified locally — build 0/0, unit 179/0, E2E 68/0/2 skipped (Docker-gated, unchanged). PR ready for re-review.

Issue #13, opened 2026-04-26
[Gautam's Dev bot] Codex E2E using MockProviderHost + scripted scenarios


gautam-achieveai:
[Gautam's Dev bot]

Follow-up to #11 / PR #12.

## Goal
Run the OpenAI Codex CLI (`@openai/codex` or successor) against `samples/MockProviderHost` and assert it routes through the host the same way the Claude Agent SDK CLI does.

## Why
PR #12 proved the in-process provider clients (`OpenClient`, `AnthropicClient`) work end-to-end against the host. The natural next step is exercising the *real CLI* path for OpenAI-flavored agents — the same shape we have scaffolding for with the Claude Agent SDK CLI in `tests/MockProviderHost.E2E.Tests/`, but we are blocked on the protocol-level handshake the Claude CLI performs against `/v1/messages`.

A Codex E2E will:
- Validate the host's OpenAI compatibility against a *real* CLI, not just a programmatic LmCore client.
- Give us a second data point on how production CLIs behave so we can refine the gating story for the Claude Agent SDK E2E.
- Establish a reusable pattern for "spawn CLI, point it at MockProviderHost, assert scripted turn was consumed".

## Suggested scope
- `tests/MockProviderHost.E2E.Tests/Scenarios/CodexAgainstMockTests.cs` modeled on `ClaudeAgentSdkAgainstMockTests`.
- New `CodexCliPrerequisites` (mirroring `ClaudeCliPrerequisites`) that detects Node + the codex binary.
- Gate behind a dedicated env var (e.g. `LMDOTNET_RUN_CODEX_E2E=1`) so default CI continues to skip.
- Use `OPENAI_BASE_URL` / `OPENAI_API_KEY` (or the Codex equivalent) overrides — likely a new additive options bag on whatever LmDotnetTools surface owns Codex spawning, paralleling the `ClaudeAgentSdkOptions` extension landed in PR #12.

## Out of scope
- Production Codex provider implementation (separate workstream).
- Mocking out OpenAI tool-calling round-trips — start with a single scripted assistant turn.

gautam-achieveai:
[Gautam's Dev bot]

## Phase 1 — Design plan: Codex E2E against MockProviderHost

### Context recap

[...]

PR #10, opened 2026-04-24
[Gautam's Dev bot] test: Playwright browser E2E tests for LmStreaming.Sample (#8)

gautam-achieveai:
## Summary

Closes #8 — Phase 2 of the Playwright browser E2E initiative. Adds a new xUnit test project that drives the Vue SPA against a real Kestrel `WebApplicationFactory<Program>` and the scripted-SSE responder stack from #7. Mock-only (`test` + `test-anthropic`); Codex/Copilot modes stay out of scope since they require external subprocess binaries.

### Infrastructure
- `PlaywrightFixture` boots Chromium once per xUnit collection.
- `BrowserWebAppFactory` preserves the `LM_PROVIDER_MODE` handshake and applies the Martin Costello Kestrel workaround for real HTTP binding.
- `ScriptedScenario.OpenAsync` collapses ~8 lines of per-test boilerplate into a single call, and disposes factory + context if partial construction throws.
- `DomAssertions` uses pure Playwright auto-wait (no `Task.Delay` loops).

### Scenarios (AC1-AC6)
- `ModeSwitchingTests` — provider-mode parametrized theory + mode-dropdown reconnect.
- `MultiTurnConversationTests` — multi-assistant-turn render order.
- `SubAgentLifecycleTests` — `Agent` tool-call pill + parent final text.
- `RecursiveInstructionChainTests` — parent + sub-agent instruction chains render end-to-end.
- `ErrorHandlingTests` — upstream 5xx surfaces on the error banner.
- `CancellationTests` — mid-stream stop button returns UI to idle.

### Sample changes
- Added `data-testid` hooks across `ChatInput`/`ChatLayout`/`ChatView`/`MessageList`/`MetadataPill`/`ModeSelector` so selectors stay stable across refactors.
- `useChat.ts` `handleRunCompleted` reads `IsError`/`ErrorMessage` tolerantly (PascalCase wire format vs camelCase TS types) and clears stale banner on clean completions.
- `BuildClientApp` MSBuild target in `LmStreaming.Sample.csproj` keeps the Vue bundle fresh before the test host boots.

### CI
- `.github/workflows/playwright-e2e.yml` caches `PLAYWRIGHT_BROWSERS_PATH` and runs the suite fully offline.

## Test plan

- [x] `dotnet test tests/LmStreaming.Sample.Browser.E2E.Tests` → **Passed 13, Failed 0, Duration 15s**
- [x] `dotnet build LmDotnetTools.sln` → 0 errors, 0 warnings
[...]

PR #87, opened 2026-06-11
Defer sandbox auth: hold webhook and prompt sign-in over WebSocket

### LINKED ISSUE #86: Defer sandbox auth: hold webhook and prompt sign-in over WebSocket
## Problem

The sandbox gateway's auth webhook (`POST /api/auth/webhook/{provider}`) denies immediately when the user hasn't pre-authenticated, so the sandboxed agent's network request fails and the user must sign in *before* starting work.

## Desired behavior

Remove the pre-auth requirement. When the webhook finds no token, hold the call open, push an `auth_required` event to connected chat clients, let the user sign in on demand, and resolve the held call with the token once sign-in completes. Time out to a deny if the user never signs in.

## Scope

`samples/LmStreaming.Sample/` and its test projects only.

## Acceptance

- Held webhook broadcasts `auth_required`; resolves `allow` once a token lands; `auth_completed` dismisses the prompt.
- Configurable hold timeout (`Auth:Webhook:HoldTimeoutSeconds`, default 120, `0` disables) → deny on expiry.
- Functional tests cover the flow without a real gateway (unit, WS-level E2E, Playwright browser E2E).

gautam-achieveai:
## Summary

When the sandbox gateway's auth webhook finds no token, it now **holds** the call while connected chat clients are prompted to sign in over the WebSocket, instead of denying immediately. The held call resolves `allow` the moment a token lands (the user completed sign-in), or `deny` after a configurable timeout. This removes the requirement to pre-authenticate before the sandboxed agent touches GitHub/ADO/Graph.

## Changes

- **`PendingAuthCoordinator`** — ref-counted per-provider holds (one prompt per provider), polls `GetAccessTokenAsync`, denies on timeout or a fresh sign-in failure, releases on gateway abort; replays pending prompts to clients that connect mid-hold.
- **`WebSocketConnectionRegistry`** — per-connection send gate so out-of-band pushes never race the agent stream; all `ChatWebSocketManager` writes route through it.
- **`WebSocketAuthEventNotifier` / `IAuthEventNotifier`** — broadcast `auth_required` / `auth_completed` frames.
- **`AuthWebhookController`** defers on the not-signed-in path; new config `Auth:Webhook:HoldTimeoutSeconds` (default 120, `0` disables) and `PollIntervalSeconds`.
- **Vue client** — `AuthRequiredBanner.vue` (testids `auth-required-banner` / `auth-signin-button` / `auth-dismiss-button`), `auth_required`/`auth_completed` dispatch in `wsClient.ts`, `pendingAuth` state in `useChat.ts`, status-poll fallback dismissal.
- **Docs** — `AuthProviderGuide.md` documents the deferred-auth flow and config keys.

## Testing

All without a real sandbox gateway (the tests play the gateway by POSTing the webhook, and seed the token store to stand in for an interactive sign-in):

- 8 `PendingAuthCoordinator` unit tests (token mid-wait, timeout, concurrent dedupe, fresh-vs-stale failure, disabled mode, cancellation cleanup).
- 6 WS-level E2E (`DeferredAuthWebhookTests`): hold → prompt → seed → allow → completed; timeout deny; single prompt for concurrent holds; replay-on-connect; disallowed-host immediate deny; gateway-abort cleanup.
- 1 Playwright browser E2E (`DeferredAuthBannerTests`): banner appears, popup opens to `/auth/github`, banner survives popup close, webhook resolves allow after token seed, banner dismisses via `auth_completed`.

Full solution builds clean (0 warnings); sample test projects + 200 vitest pass. Two existing not-signed-in deny tests pinned to `HoldTimeoutSeconds=0`; fixed a stale `ChatLayout.test.ts` mock.

## Related Issues

Fixes #86


samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
```diff
@@ -0,0 +1,185 @@
...
+                return;
+            }
+
+            _ = _pending.Remove(providerId);
+            notifyCompleted = entry.TokenObtained;
+        }
+
+        if (notifyCompleted)
+        {
+            await notifier.NotifyAuthCompletedAsync(providerId, CancellationToken.None);
```
gautam-achieveai, line 200: **[HIGH · BLOCKER] Deny path has no terminal client signal → banner stuck forever.**

`ExitAsync` broadcasts `auth_completed` only when `TokenObtained` is true. When a hold resolves **deny** (timeout `:100`, fresh failure `:117`, non-gateway cancel `:128`) the client gets nothing. The banner's only dismissal paths are `auth_completed` (token-only) and the status-poll on `state === 'SignedIn'` — neither fires on a timeout-without-signin or a failed sign-in, so the yellow "Sign in required" banner stays up indefinitely with a now-dead button (manual ✕ is the only escape). This fails #86's explicit "time out to a deny… `auth_completed` dismisses the prompt" on the deny side, and is untested.

**Fix:** add a terminal `auth_denied` to `IAuthEventNotifier` and emit it here when the last waiter leaves without a token; wire it into the client `AuthEvent` union, `isAuthEventPayload`, and `handleAuthEvent` to delete the prompt:
```csharp
if (notifyCompleted)
    await notifier.NotifyAuthCompletedAsync(providerId, CancellationToken.None);
else
    await notifier.NotifyAuthDeniedAsync(providerId, "sign-in not completed", CancellationToken.None);
```
This one chokepoint covers timeout, fresh-failure, and disabled-deferral exits.

gautam-achieveai (commented):
## 🤖 Automated multi-agent code review — PR #87

**Assessment: REQUEST CHANGES** (posted as a comment review since GitHub doesn't allow an author to formally request changes on their own PR).

Reviewers run: over-engineering · class-design · exception-handling · performance/concurrency · test-coverage · silent-failure · temp-code, then a severity grader.

This is well-built work — disciplined scope against #86, strong server-side test coverage (the fresh-vs-stale failure branch and replay-on-connect both have dedicated tests), correct concurrency primitives, and **zero temp-code/debug artifacts**. The two abstractions worth scrutinizing (`WebSocketConnectionRegistry` send-gate, `IAuthEventNotifier`) both earn their keep. The blockers are concentrated in the **deny / unhappy path** of the client dismissal flow, plus one exception-handling trap.

### 🔴 Blockers (2 HIGH)
1. **Deny path leaves the sign-in banner stuck forever** — a held call that resolves *deny* (timeout / failed sign-in / non-gateway cancel) sends no terminal frame to the client. `auth_completed` fires only on a token; the status-poll dismisses only on `SignedIn`. So on timeout/failed-signin the banner stays up with a dead "Sign in" button (only escape is manual ✕). Fails #86's explicit "deny on timeout → dismiss prompt". Fix: add an `auth_denied` terminal frame from the `ExitAsync` chokepoint, wired into the client union/sniff/handler.
2. **`WaitForTokenAsync` awaited inside a `catch` block** — the deferred call runs inside `catch (InvalidOperationException)`, so the sibling broad `catch (Exception)` can't catch it and the always-200 deny contract doesn't hold for the deferred branch (any non-OCE failure → opaque 500 to the gateway). Fix: hoist the defer decision out of the catch.

### 🟡 Should-fix (5 MEDIUM)
3. `BroadcastAsync` is sequential + `CancellationToken.None` → unbounded head-of-line blocking; fan out with `Task.WhenAll` + a `CancelAfter`.
4. The send-gate's whole reason to exist (serialize concurrent socket writes) is untested.
5. Client `pendingAuth` map logic covered only by one slow Playwright path — the `useChat.test.ts` harness already exposes the `onAuthEvent` seam.
6. Status-poll fallback is out-of-scope (beyond #86) **and** incomplete (never covers the timeout case) — resolve as part of #1, then delete it.
7. `SendProviderUnavailableErrorAsync` takes both the raw socket and its wrapper; add a close method to `RegisteredWebSocketConnection` and drop the raw param.

### 🟢 Minor (6 LOW)
8. `Enter()` outside the `try/finally` — latent ref-count leak footgun.
9. Per-connection `SemaphoreSlim` never disposed.
10. `auth_completed` only on last-waiter-exit → bounded dismissal latency with concurrent holds.
11. `statusAtEntry` reference-equality early-deny couples to `SetFailed` semantics (verified to hold) — keep consciously or cut to match #86.
12. A few name-restating doc one-liners amid genuinely valuable "why" comments.
13. `isAuthEventPayload` substring sniff has no unit test.

**Confirmed correct-by-design:** temp-code clean (no debug/secrets/disabled tests); controller broad-catch deny; best-effort `TrySendTextAsync` swallow; OCE-vs-timeout disambiguation; register-before-snapshot replay (no frame can be lost). FYI: your uncommitted working-tree files (`.gitignore`, `.mcp.json`, `.vscode/settings.json`, `scratchPad/.gitkeep`, `worktrees/`) are not part of this PR's diff.

**Bottom line:** Two small edits (add `auth_denied`; hoist the defer out of the catch) clear both blockers and let you delete the out-of-scope poll (#6) and unblock the client unit test (#5).

samples/LmStreaming.Sample/Controllers/AuthWebhookController.cs
```diff
@@ -84,6 +85,24 @@ public async Task<IActionResult> Evaluate(
...
             logger.LogInformation(
                 ex,
+                "Auth-webhook found no valid token for provider {ProviderId} (host {DestinationHost}); deferring.",
+                tokenProvider.ProviderId,
+                body.DestinationHost);
+
+            // Deferred auth: hold this call open while connected chat clients are prompted to
+            // sign in (auth_required WebSocket frame). Resolves allow the moment a token lands;
+            // resolves deny when the hold times out, sign-in fails, or deferral is disabled.
+            var deferred = await pendingAuth.WaitForTokenAsync(tokenProvider, body.RequiredScopes, ct);
```
gautam-achieveai, line 131: **[HIGH · BLOCKER] Deferred call runs inside a `catch` block, so the always-200 contract doesn't cover it.**

This `await` is inside `catch (InvalidOperationException)` (`:82`). By C# semantics, an exception thrown here is **not** caught by the sibling `catch (OperationCanceledException when …)` (`:111`) or `catch (Exception)` (`:116`) of the same `try` — it escapes `Evaluate`. Today the gateway-abort OCE propagates (intended, and tested), but **any non-OCE failure** from the deferred path (token-store IO, refresh HTTP, MSAL, JSON) becomes an opaque HTTP 500 to the gateway instead of the clean always-200 deny the broad catch at `:116` exists to guarantee. It looks protected; it provably isn't.

**Fix:** hoist the defer decision out of the catch and run the hold after the try/catch with its own OCE-rethrow + generic-deny guard:
```csharp
bool defer = false;
try { /* GetAccessTokenAsync + allow */ }
catch (InvalidOperationException) { defer = true; }
catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }
catch (Exception ex) { /* log + return deny */ }

if (defer)
{
    try {
        var deferred = await pendingAuth.WaitForTokenAsync(tokenProvider, body.RequiredScopes, ct);
        return Ok(deferred is not null ? AuthWebhookResponse.Allow(...) : AuthWebhookResponse.Deny(...));
    }
    catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }
    catch (Exception ex) { logger.LogWarning(ex, "..."); return Ok(AuthWebhookResponse.Deny(...)); }
}
```

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// <summary>Removes a connection from the registry (idempotent).</summary>
+    public void Unregister(string connectionId) => _connections.TryRemove(connectionId, out _);
+
+    /// <summary>Point-in-time view of the live connections.</summary>
+    public IReadOnlyList<RegisteredWebSocketConnection> Snapshot() => [.. _connections.Values];
+
+    /// <summary>Best-effort broadcast of a text frame to every live connection.</summary>
+    public async Task BroadcastAsync(string json, CancellationToken ct)
```
gautam-achieveai, line 158: **[MEDIUM] Sequential broadcast + `CancellationToken.None` → unbounded head-of-line blocking.**

`BroadcastAsync` awaits each send sequentially, and the coordinator passes `CancellationToken.None`, so a single half-open/wedged socket can stall the `auth_required`/`auth_completed` broadcast to **all** other clients with no time bound. Each connection has its own send-gate, so concurrent fan-out is safe:
```csharp
await Task.WhenAll(Snapshot().Select(c => c.TrySendTextAsync(json, ct)));
```
Ideally also wrap the broadcast in a `CreateLinkedTokenSource(ct)` + `CancelAfter(...)` so the auth path can never wedge (`TrySendTextAsync` already turns a cancel into a quiet `false`).

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// <summary>The conversation thread this connection streams.</summary>
+    public string ThreadId { get; }
+
+    /// <summary>
+    /// Sends a single text frame, serialized behind the connection's send gate. Returns false (and
+    /// never throws) when the socket is closed or the send fails — a dying connection must not
+    /// fault a broadcast or the agent stream loop. Cancellation of <paramref name="ct"/> is also
+    /// reported as false rather than thrown so best-effort broadcasts stay quiet on teardown.
+    /// </summary>
+    public async Task<bool> TrySendTextAsync(string json, CancellationToken ct)
```
gautam-achieveai, line 38: **[MEDIUM] The send-gate's entire reason to exist is untested.**

`RegisteredWebSocketConnection` was introduced specifically to serialize concurrent writes to one socket (`WebSocket.SendAsync` isn't concurrency-safe — agent stream loop + out-of-band auth broadcasts now write the same socket). No test verifies the `SemaphoreSlim` actually serializes them, so accidentally dropping the gate would pass CI and produce intermittent interleaved-frame corruption that's miserable to diagnose.

**Add:** a unit test firing ~50 concurrent `TrySendTextAsync` over a fake `WebSocket` whose `SendAsync` `Interlocked.Increment`s an in-flight counter, delays, decrements — assert the recorded max never exceeds 1 and all calls return true. ~20 lines, high value.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
```diff
@@ -158,6 +160,33 @@ export function useChat(options: UseChatOptions = {}) {
   // Persistent WebSocket connection for full-duplex communication
   let wsConnection: import('@/api/wsClient').WebSocketConnection | null = null;
 
+  // Deferred-auth prompts pushed by the backend while a sandbox webhook call is held
+  // (providerId -> auth_required event). Replaced wholesale on change for Vue reactivity.
+  const pendingAuth = ref<Map<string, AuthRequiredEvent>>(new Map());
+
+  /** Handle an out-of-band deferred-auth frame from the WebSocket. */
+  function handleAuthEvent(event: AuthEvent): void {
```
gautam-achieveai, line 168: **[MEDIUM] Client `pendingAuth` map logic covered only by one slow Playwright happy-path.**

`handleAuthEvent` / `dismissAuthRequest` / `pendingAuthRequests` are the client-side feature state machine, but the existing `useChat.test.ts` mock harness already captures the options object and drives `options.onMessage(...)` — the identical `options.onAuthEvent` seam is right there and unused. Drive it to assert: `auth_required` adds an entry; a second provider coexists; `auth_completed` removes the **right** key; `dismissAuthRequest` is a no-op for an unknown provider; and the `new Map(...)` replacement re-triggers the computed. A regression here (wrong key deleted, in-place mutation breaking reactivity) silently leaves a stuck banner — and this same test naturally covers the deny-dismissal once the `auth_denied` frame from blocker #1 exists.

samples/LmStreaming.Sample/ClientApp/src/components/AuthRequiredBanner.vue
```diff
@@ -0,0 +1,149 @@
...
+}>();
+
+function openSignIn(request: AuthRequiredEvent): void {
+  window.open(request.signinUrl, `auth-${request.providerId}`, 'popup,width=640,height=760');
+}
+
+const POLL_INTERVAL_MS = 2000;
+let pollTimer: ReturnType<typeof setInterval> | null = null;
+
+async function pollStatuses(): Promise<void> {
```
gautam-achieveai: **[MEDIUM] Status-poll fallback is both out-of-scope and incomplete.**

This polls `/api/auth/{provider}/status` every 2s as a second dismissal path beyond #86's `auth_completed` mechanism — untested, and it will drift from the frame contract. Crucially it does **not** even cover the case it appears aimed at: on a timeout/failed sign-in the provider status never reaches `SignedIn`, so the poll never dismisses (that's blocker #1). 

**Recommendation:** add the proper `auth_denied` terminal frame (blocker #1) and **delete this poll**, rather than shipping a second incomplete dismissal path. If dropped-WS-frame recovery is genuinely wanted, split it into its own ticket with a test.

samples/LmStreaming.Sample/WebSocket/ChatWebSocketManager.cs
```diff
@@ -320,15 +339,11 @@ private async Task ProcessClientMessageAsync(
 
     private async Task SendProviderUnavailableErrorAsync(
         System.Net.WebSockets.WebSocket webSocket,
+        RegisteredWebSocketConnection connection,
```
gautam-achieveai, line 341: **[MEDIUM] `SendProviderUnavailableErrorAsync` takes both the raw socket and the wrapper that owns it.**

The method receives `webSocket` (used only for the final `CloseAsync` at `:370`) **and** the `connection` that wraps that same socket. Passing both invites the two to diverge and contradicts the wrapper's own contract ("all outbound text frames MUST go through `TrySendTextAsync`") — closing is an outbound op that currently bypasses the wrapper.

**Fix:** add a `CloseAsync(status, description, ct)` (or `TryCloseAsync`) to `RegisteredWebSocketConnection`, route the close through it, and drop the raw `webSocket` parameter. While there, broaden the close's `catch (WebSocketException)` to also cover `ObjectDisposedException`/`InvalidOperationException`, which `CloseAsync` can throw on a raced socket.

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        var pollInterval = TimeSpan.FromSeconds(Math.Max(0.05, options.Webhook.PollIntervalSeconds));
+
+        // Snapshot the status REFERENCE at entry: a pre-existing Failed (from an older attempt)
+        // must not block deferral, but a NEW failure during the hold (fresh status instance)
+        // means the user tried and failed — deny early instead of waiting out the clock.
+        var statusAtEntry = provider.Status;
+
+        var entry = Enter(provider.ProviderId, out var isFirstWaiter);
```
gautam-achieveai: **[LOW] `Enter()` sits outside the `try` whose `finally` calls `ExitAsync` — latent ref-count leak.**

Leak-free today (nothing between `Enter` and the `try` can throw), but the placement is exactly the pattern that becomes a permanent orphan under future edits: any throwing statement inserted here (a log, a metric, another await) would leave the entry in `_pending` with no decrement, pinning `auth_required` broadcast/replay forever. Make the enter/exit symmetry structural by moving `Enter` to the first line inside the `try`.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
```diff
@@ -0,0 +1,119 @@
...
+/// A live chat WebSocket connection registered with <see cref="WebSocketConnectionRegistry"/>.
+/// All outbound text frames MUST go through <see cref="TrySendTextAsync"/> — it serializes writers
+/// (the per-connection agent stream loop and out-of-band broadcasts such as auth events) behind a
+/// single async gate, because <see cref="System.Net.WebSockets.WebSocket.SendAsync(ArraySegment{byte}, WebSocketMessageType, bool, CancellationToken)"/>
+/// is not safe for concurrent calls.
+/// </summary>
+public sealed class RegisteredWebSocketConnection
+{
+    private readonly System.Net.WebSockets.WebSocket _socket;
+    private readonly SemaphoreSlim _sendGate = new(1, 1);
```
gautam-achieveai, line 17: **[LOW] Per-connection `SemaphoreSlim` is never disposed.**

`RegisteredWebSocketConnection` allocates `new SemaphoreSlim(1,1)` per connection but isn't `IDisposable`, and `Unregister` only removes the dictionary entry. Negligible for a single-user demo (no unmanaged handle unless `AvailableWaitHandle` is read, which it isn't), but for correctness make the type `IDisposable`, dispose `_sendGate`, and dispose it in `Unregister` — tie this into the `CloseAsync` work from the `SendProviderUnavailableErrorAsync` finding. (Disposal is safe after in-flight sends since `TrySendTextAsync` already swallows `ObjectDisposedException`.)

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
```diff
@@ -0,0 +1,185 @@
...
+            isFirstWaiter = true;
+            return entry;
+        }
+    }
+
+    /// <summary>Leaves the per-provider entry; the last waiter out clears it and, when a token was obtained, tells clients to dismiss the prompt.</summary>
+    private async Task ExitAsync(string providerId, PendingEntry entry)
+    {
+        bool notifyCompleted;
+        lock (_gate)
```
gautam-achieveai, line 181: **[LOW] `auth_completed` only on last-waiter-exit → bounded dismissal latency.**

With concurrent holds for the same provider, an early-obtaining waiter sets `TokenObtained` and exits while a slower waiter keeps `WaiterCount > 0`, so `auth_completed` isn't broadcast until that slower waiter's next poll — the prompt lingers up to one `PollIntervalSeconds` after sign-in already succeeded. End state is correct; only the latency is the artifact (and the status-poll currently masks it). If you want immediate dismissal, broadcast on the `TokenObtained` 0→1 transition instead of on last-waiter-exit. Optional.

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
```diff
@@ -0,0 +1,185 @@
...
+                        entry.TokenObtained = true;
+                    }
+
+                    return token;
+                }
+                catch (InvalidOperationException)
+                {
+                    // Still not signed in — keep holding unless a NEW failure occurred.
+                    var status = provider.Status;
+                    if (status.State == OAuthSignInState.Failed && !ReferenceEquals(status, statusAtEntry))
```
gautam-achieveai, line 121: **[LOW] Fresh-vs-stale failure early-deny is beyond #86's "time out to deny" — keep consciously.**

The `!ReferenceEquals(status, statusAtEntry)` early-deny is a real, tested UX win (fail fast when the user cancels the popup instead of holding the full timeout), and the invariant it relies on genuinely holds (`OAuthProviderBase.SetFailed` uses record `with`, so a fresh failure is always a new reference). But it couples the coordinator to that invariant and to `GetAccessTokenAsync` throwing `InvalidOperationException` when not signed in — if a future provider sets `Failed` in place or throws a different type, this silently stops working with no test pointing at the cause. Keep it (it's the first thing to cut if you want to hew exactly to #86); the explanatory comment earns its place.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// <summary>Registers a newly-accepted connection; the caller must <see cref="Unregister"/> it on teardown.</summary>
+    public RegisteredWebSocketConnection Register(string threadId, System.Net.WebSockets.WebSocket socket)
+    {
+        var connection = new RegisteredWebSocketConnection(Guid.NewGuid().ToString("N"), threadId, socket);
+        _connections[connection.ConnectionId] = connection;
+        return connection;
+    }
+
+    /// <summary>Removes a connection from the registry (idempotent).</summary>
+    public void Unregister(string connectionId) => _connections.TryRemove(connectionId, out _);
```
gautam-achieveai: **[LOW] Comment density — trim name-restating one-liners.**

The *why* comments across these new files (polling rationale, the ref-equality trick, the send-gate "MUST go through" contract) are genuinely valuable — keep them. A handful just paraphrase the member name, e.g. this "Removes a connection from the registry (idempotent)" over `Unregister`, and the `Register`/`Snapshot` one-liners. Optional trim to keep the signal of the valuable comments high. Not blocking.

samples/LmStreaming.Sample/ClientApp/src/types/auth.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export interface AuthCompletedEvent {
+  $type: 'auth_completed';
+  providerId: string;
+}
+
+export type AuthEvent = AuthRequiredEvent | AuthCompletedEvent;
+
+export function isAuthEventPayload(data: string): boolean {
```
gautam-achieveai, line 31: **[LOW] `isAuthEventPayload` substring sniff has no unit test.**

This stringly-typed `data.includes('"$type":"auth_required"')` check is the single routing gate for the whole client feature and runs *before* the done/error sniff — if it returns false for a real frame the banner never appears; if it returns true for a chat message a real message is dropped. It silently breaks if the `$type` value ever changes. Add cheap pure-function tests (true for both auth frames, false for done/error/normal text). (The JSON.parse-after-match path is already correctly surfaced via `onError` — verified, no change needed there.)

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — ExitAsync now emits a terminal auth_denied frame on the deny side (timeout/failed/disabled) at the last-waiter chokepoint, alongside auth_completed. Wired through IAuthEventNotifier, the client AuthEvent union, isAuthEventPayload, and handleAuthEvent, so the banner dismisses on timeout.

samples/LmStreaming.Sample/Controllers/AuthWebhookController.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — the deferred WaitForTokenAsync is hoisted out of the catch into its own try/catch (OCE-rethrow + generic-deny), so a non-OCE failure on the deferred path returns a clean 200 deny instead of a 500.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — BroadcastAsync now fans out with Task.WhenAll and wraps the sends in a linked CancelAfter(5s), so one wedged socket cannot head-of-line-block delivery to the others.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
gautam-achieveai: [Gautam's bot] Added in e216b18 — WebSocketConnectionRegistryTests fires 50 concurrent TrySendTextAsync over a fake socket that records peak overlap and asserts max in-flight == 1.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
gautam-achieveai: [Gautam's bot] Added in e216b18 — useChat.test.ts drives the onAuthEvent seam: auth_required adds an entry, a second provider coexists, auth_completed and auth_denied remove the right key, and dismissAuthRequest no-ops an unknown provider.

samples/LmStreaming.Sample/ClientApp/src/components/AuthRequiredBanner.vue
gautam-achieveai: [Gautam's bot] Done in e216b18 — deleted the status-poll. The new auth_denied terminal frame is the proper deny-side dismissal, so the second incomplete path is gone.

samples/LmStreaming.Sample/WebSocket/ChatWebSocketManager.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — added TryCloseAsync to RegisteredWebSocketConnection (catching WebSocketException, ObjectDisposedException, InvalidOperationException), routed the SendProviderUnavailableErrorAsync close through it, and dropped the raw socket parameter.

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — Enter() moved to the first line inside the try; the finally guards on a null entry, so the enter/exit ref-count pairing is now structural.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
gautam-achieveai: [Gautam's bot] Fixed in e216b18 — RegisteredWebSocketConnection is now IDisposable and Unregister disposes the send gate.

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
gautam-achieveai: [Gautam's bot] Kept the last-waiter chokepoint (it now emits completed-or-denied). The bounded sub-poll-interval latency with concurrent same-provider holds is the simple, correct trade-off; not switching to the 0 to 1 transition. Acknowledged as optional.

samples/LmStreaming.Sample/Services/Auth/PendingAuthCoordinator.cs
gautam-achieveai: [Gautam's bot] Kept as-is per your recommendation — the fresh-vs-stale early-deny and its explanatory comment stay.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
gautam-achieveai: [Gautam's bot] Left the XML docs intact — after the blocker edits the remaining one-liners carry non-obvious info (Unregister documents disposal, Register the teardown contract, Snapshot the point-in-time semantics), so I did not strip public-member docs.

samples/LmStreaming.Sample/ClientApp/src/types/auth.ts
gautam-achieveai: [Gautam's bot] Added in e216b18 — auth.test.ts covers isAuthEventPayload: true for auth_required/completed/denied, false for done/error/normal-text/empty.

gautam-achieveai:
## Review fixes pushed — e216b18

Both HIGH blockers and all actionable MEDIUM/LOW items addressed (per-thread replies above).

**Blockers**
- Deny path now sends a terminal `auth_denied` frame (`ExitAsync` chokepoint → `auth_completed` on token, `auth_denied` on timeout/failed/disabled), wired through the client union/`isAuthEventPayload`/`handleAuthEvent`. Banner dismisses on timeout instead of lingering.
- Deferred `WaitForTokenAsync` hoisted out of the `catch` into its own try/catch, restoring the always-200 deny contract for the deferred path.

**Also:** parallel broadcast with bounded `CancelAfter`; `TryCloseAsync` on the connection wrapper (raw-socket param dropped); `IDisposable` send gate; `Enter()` inside the try; **deleted the out-of-scope status-poll** (the `auth_denied` frame replaces it).

**Consciously kept:** the fresh-vs-stale early-deny (#11, per your note) and the value-carrying XML docs (#12). #10's sub-poll-interval dismissal latency left as the simple correct trade-off (optional).

**Tests:** send-gate concurrency test (50 concurrent sends → max in-flight 1), client `isAuthEventPayload` + `pendingAuth` state-machine tests, coordinator `auth_denied` assertions. Full sweep green: 188 unit · 74 WS-E2E · 14 browser-E2E · 210 vitest; solution builds clean.

samples/LmStreaming.Sample/WebSocket/WebSocketConnectionRegistry.cs
```diff
@@ -0,0 +1,173 @@
...
+    public async Task<bool> TrySendTextAsync(string json, CancellationToken ct)
+    {
+        if (_socket.State != WebSocketState.Open)
+        {
+            return false;
+        }
+
+        try
+        {
+            await _sendGate.WaitAsync(ct);
```
gautam-achieveai, line 47: **[LOW · NEW — introduced by the #9 dispose fix] Use-after-dispose race on the send gate.**

Making `Unregister` dispose `_sendGate` (good intent) created a race with the now-parallel `BroadcastAsync`: `Unregister` runs from a connection's own teardown, but a broadcast that already called `Snapshot()` can be mid-flight on that same connection. `TrySendTextAsync` isn't tolerant of a disposed gate:

- **Line 47** `await _sendGate.WaitAsync(ct)` is wrapped only in `catch (OperationCanceledException)`. A disposed gate throws **`ObjectDisposedException`** → uncaught → faults the `Task.WhenAll` in `BroadcastAsync`.
- **Line 83** `finally { _ = _sendGate.Release(); }` — if the gate is disposed while a send is in-flight, `Release()` throws **`ObjectDisposedException` from a `finally`**.

The `_socket.State != Open` check at line 40 masks the common teardown ordering, but it's a TOCTOU and doesn't protect the in-flight `Release` at all. Worst case the faulted broadcast turns an `auth_completed` into a spurious deny via the controller's deny-catch. Narrow race in a single-user demo, hence LOW — but it's a real correctness regression from a fix to a cosmetic item.

**Simplest remedy — don't dispose (revert #9):** this `SemaphoreSlim` never has `AvailableWaitHandle` read, so it holds no unmanaged handle; disposing it reclaims nothing and not disposing is strictly safer than this race. If you'd rather keep `Dispose`, make the send path tolerant instead:
```csharp
try { await _sendGate.WaitAsync(ct); }
catch (OperationCanceledException) { return false; }
catch (ObjectDisposedException) { return false; }   // gate disposed mid-teardown
...
finally { try { _ = _sendGate.Release(); } catch (ObjectDisposedException) { } }
```

gautam-achieveai (commented):
## 🔁 Re-review — delta `f8c9f50` → `e216b18`

Verified each fix against the code (not just the thread replies). **Both HIGH blockers are genuinely resolved, with real regression tests.**

**✅ Blockers**
- **#1 stuck banner** — `ExitAsync` now emits a terminal frame at the last-waiter chokepoint: `auth_completed` on token, else `auth_denied` (timeout / failed / disabled), wired through `IAuthEventNotifier` → the client union / `isAuthEventPayload` / `handleAuthEvent`. Status-poll deleted. Coordinator tests assert `auth_denied` on timeout / fresh-failure / gateway-abort and empty on token/disabled.
- **#2 catch-in-catch** — deferred `WaitForTokenAsync` hoisted into its own try/catch (OCE-rethrow + generic-deny); the always-200 deny contract now holds on the deferred path.

**✅ Also resolved:** #3 parallel broadcast + `CancelAfter(5s)` · #4 send-gate concurrency test (50 sends → max in-flight 1) · #5 `useChat` `onAuthEvent` state-machine tests · #6 status-poll removed · #7 `TryCloseAsync` + raw-socket param dropped · #8 `Enter()` inside the try · #13 `isAuthEventPayload` tests. #10/#11/#12 consciously kept with sound rationale.

**⚠️ One NEW finding (LOW)** — the #9 fix (dispose the send gate) introduced a use-after-dispose race with the now-parallel broadcast. See the inline comment on `WebSocketConnectionRegistry.cs`. Narrow race, single-user demo, non-blocking — but worth tightening (or just reverting the dispose) before merge.

**Verdict: 🟢 APPROVE WITH COMMENTS.** Suite green (188 unit · 74 WS-E2E · 14 browser-E2E · 210 vitest). Nice, thorough turnaround.

PR #91, opened 2026-06-19
feat(sample): workspace picker, marketplace catalog & Copilot keychain auth + history-recovery and per-conversation sandbox isolation

gautam-achieveai:
## Summary

Bundles the sample-app work layered on top of the deferred-sandbox-auth foundation (the deferred-auth itself already landed in `main` via #87). Net-new since #87:

- **Workspace picker + CRUD** (merged here via #90) — create/edit workspaces, each mapping to its own working directory and a **per-workspace sandbox session**.
- **Gateway-sourced marketplace catalog** — the workspace marketplace selection is now driven by the live gateway catalog (`GET /api/marketplaces`) and threaded through to the sandbox-create request, so picking marketplaces for a workspace enables exactly those for that conversation (verified by a deterministic browser E2E).
- **Copilot keychain auth** — `CliCredentialCopilotTokenProvider` resolves the Copilot token from the macOS keychain.
- **History recovery on (re)create** — `MultiTurnAgentBase` auto-recovers persisted messages when an agent is recreated.
- **Per-conversation sandbox isolation** — `SandboxSessionRegistry` bindings are scoped per conversation (with a `tools_echo` probe + isolation tests).
- Chores: ignore `worktrees/` and `scratchPad/`, trim local `.mcp.json`.

## Tests

- Backend `LmStreaming.Sample.Tests`: 250 passing.
- Frontend (vitest): 235 passing.
- Browser E2E (Playwright, scripted/mock provider): green in CI, including the new per-workspace marketplace-selection proof.

## Notes for reviewers

- #87 was **squash-merged**, so the merge-base predates it — the diff may surface already-merged deferred-auth files. Resolving against current `main` (which also carries #85) removes that overlap.


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #91 — workspace picker, marketplace catalog, Copilot keychain auth, history-recovery, and per-conversation sandbox isolation.

---

## 📋 Scope Concern

This PR bundles **5+ distinct features** across **83 files / ~7800 lines** with **no commit messages and no PR description**. Individual components show solid engineering (deferred-auth polling, connection send-gate, single-flight session creation), but aggregate scope makes thorough review difficult and future `git bisect` painful.

**Recommendation:** Consider splitting into separate PRs for future work of this magnitude.

---

## 🔴 HIGH — Actionable Issues

### 1. Case-sensitive path traversal check on Windows
**`Services/SandboxGatewayOptions.cs:101`**

```csharp
if (!fullPath.StartsWith(baseWithSeparator, StringComparison.Ordinal))
```

On Windows (case-insensitive filesystem), different casing after canonicalization could bypass this containment check. Should use `StringComparison.OrdinalIgnoreCase` on Windows:

```csharp
var comparison = OperatingSystem.IsWindows()
    ? StringComparison.OrdinalIgnoreCase
    : StringComparison.Ordinal;
if (!fullPath.StartsWith(baseWithSeparator, comparison))
```

### 2. Sync-over-async in `MultiTurnAgentPool`
**`Agents/MultiTurnAgentPool.cs:416, 497`**

Both `LoadPersistedProviderId` and `LoadPersistedWorkspaceId` use `.GetAwaiter().GetResult()`. If the `IConversationStore` implementation ever awaits on a context-bound continuation, this deadlocks. The code comment acknowledges the risk but doesn't mitigate it.

**Consider:** Making `GetOrCreateAgent` async, or ensuring the store implementation uses `ConfigureAwait(false)` throughout, or running on a dedicated thread pool thread.

---

## 🟡 MEDIUM — Worth Addressing

### 3. Silent data loss in `FileWorkspaceStore`
**`Persistence/FileWorkspaceStore.cs:218-220`**

```csharp
catch (JsonException)
{
    // If the file is corrupted, start fresh.
    return [];
}
```

A corrupted file permanently loses all user workspaces with **no log or diagnostic**. The user just sees their workspaces vanish. At minimum, emit a warning log so the loss is discoverable.

### 4. Multiple bare `new HttpClient()` without `IHttpClientFactory`
**`Program.cs:152, 163`**

Long-lived singleton `HttpClient` instances don't honor DNS TTL. For a sample app this is acceptable, but since this code serves as a reference implementation, using `IHttpClientFactory` would be the better example.

### 5. `SandboxSessionRegistry` accumulates too many concerns (40KB)
Session CRUD, thread↔session routing, sub-agent bindings, context-discovery dedup all live in one class. **Consider extracting** `SubAgentBindingRegistry` and `SessionThreadRouter` as follow-up.

### 6. `SandboxGatewayOptions.ResolveWorkspace` returns raw `relPathOverride` as the leaf
**`Services/SandboxGatewayOptions.cs:108`**

After the traversal check passes, the returned leaf is the *raw user input* (which could contain `./` or redundant separators). Downstream code that uses this leaf directly in path construction (rather than `fullPath`) could diverge from the canonicalized result.

---

## 🟢 Well-Designed Aspects

- **`PendingAuthCoordinator`** — Clean ref-counted hold with try/finally pairing, deliberate polling design
- **`RegisteredWebSocketConnection`** — Send-gate preventing frame interleaving
- **`BroadcastAsync`** — Fan-out with per-connection timeout cap
- **`SandboxSessionRegistry.GetOrCreateSessionAsync`** — Single-flight via `ConcurrentDictionary<string, Lazy<Task<>>>` with evict-on-failure
- **`FileWorkspaceStore`** — Atomic write-then-rename pattern
- **`MultiTurnAgentBase` history recovery** — Idempotent via `_historyRecovered` flag, no double-restore
- **`CliCredentialCopilotTokenProvider.TryGetTokenFromMacKeychain`** — Proper process timeout handling + token format validation

---

## 🏗️ Architecture Notes

| Area | Assessment |
|------|-----------|
| Concurrency | ✅ Well-designed primitives throughout |
| Security | ✅ Constant-time secret comparison, path-traversal prevention (with Windows caveat above) |
| DI Registration | ⚠️ Program.cs is 1949 lines — extract `AddAuthServices()`, `AddSandboxServices()`, etc. |
| Frontend | ✅ No XSS (all Vue `{{ }}` interpolation), clean composable patterns |
| Tests | ✅ Extensive coverage with 250+ test scenarios added |

---

**Verdict: COMMENT** — Non-blocking. The individual components are well-implemented. Primary concerns are the path-traversal case sensitivity (actionable bug on Windows) and the sync-over-async pattern (latent risk). Structural issues (scope, god files) are tech debt, not blockers.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review (since `1124563`) - commit `f440d5ec`.

---

## Merge of Main

This increment is a clean **merge of main** into the feature branch, incorporating:
- **#85** - tool_schema Test (Mock) directive + provider-override telemetry
- **#87** - Defer sandbox auth (squash-merged overlap resolution)

## Changes Verified

| Area | Assessment |
|------|-----------|
| Provider override warning log | Structured logging, OrdinalIgnoreCase comparison |
| GetEffectiveProviderId public API | Clean delegation to ResolveProviderId with exception fallback |
| __TOOL_SCHEMA__ directive (OpenAI + Anthropic) | Consistent implementation, shared ToolSchemaFormatter |
| SSE chunk boundary space fix | Correct i > 0 guard, tested with 6-word and JSON payloads |
| ExtractRequestParams to Dictionary<string, JsonElement> | Preserves JSON structure fidelity |
| Test coverage | 9 new tests all passing |

## Previous Comments Still Apply

The HIGH/MEDIUM findings from iteration 1 (path traversal case-sensitivity, sync-over-async, silent data loss in FileWorkspaceStore) were not addressed by this merge - they remain valid for the full PR.

---

**Verdict: APPROVE** - No new issues in this increment. Clean merge of already-reviewed code.

gautam-achieveai (commented):
## 🛡️ Code Review — multi-agent (architecture · performance · exception-handling · test-coverage · schema-compat · secret/EUII)

**Verdict: 🔴 REQUEST CHANGES** — 3 small, well-contained blockers; everything else is optional hardening. Build + ~860 tests + both CI checks are green, and two whole dimensions came back clean (no secret/EUII leaks, no backward-incompatible schema changes). The blockers are all in the **history-recovery** change plus one timeout one-liner.

---

### 🚧 Blockers (fix these, then this is a clean merge)

**B1 · HIGH — Best-effort history recovery can now abort agent startup**
`src/LmMultiTurn/MultiTurnAgentBase.cs:876` (`RunAsync`) — `_ = await RecoverAsync(ct);` has no guard. `RecoverAsync` awaits `Store.LoadMetadataAsync`/`LoadMessagesAsync`; a transient IO/deserialization fault now propagates out of `RunAsync`, so **the agent fails to start at all**. Before this PR, startup never recovered, so a flaky store couldn't block it — this is a behavior regression. Worse, `_historyRecovered = true` is set at `:494` *before* the awaited IO, so it's also a poisoned-retry trap (a future retry would skip recovery).
**Fix:** make startup recovery non-fatal, and set/reset the flag around success:
```csharp
if (Store != null && !_historyRecovered)
{
    try { _ = await RecoverAsync(ct); }
    catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }
    catch (Exception ex)
    {
        _historyRecovered = false; // allow a later explicit retry
        Logger.LogWarning(ex, "History recovery failed for thread {ThreadId}; starting with empty history", ThreadId);
    }
}
```
(and/or move the `_historyRecovered = true` in `RecoverAsync` to *after* a successful load.)

**B2 · HIGH — The double-restore guard is untested**
`MultiTurnAgentBase.cs:494,876` + `tests/LmMultiTurn.Tests/MultiTurnAgentBaseTests.cs`. The `_historyRecovered` flag exists specifically to stop a double-recover (`RestoreHistory` appends → a second recover would duplicate the *entire* conversation and corrupt LLM context). Only the auto-recover happy path is tested; the idempotency guard — the actual bug-class — has zero coverage, so a flag regression would silently duplicate history with the suite green.
**Fix:** add one test — call `RecoverAsync()` explicitly, then `RunAsync()`, and assert the history count equals the persisted message count (no doubling).

**B3 · MEDIUM (blocking — same-PR convention violation) — Sandbox-session `HttpClient` has no timeout**
`Program.cs:220` registers the registry's client as bare `new HttpClient()` (100s default), while this *same PR* gave the sibling marketplace client an explicit `Timeout = 10s` (`Program.cs:163`). `CreateSessionAsync`'s POST runs sync-over-async on the WebSocket request thread (`Program.cs:384`), so a stalled (not cleanly-offline) gateway blocks that thread ~100s.
**Fix:** one line — `new HttpClient { Timeout = TimeSpan.FromSeconds(30) }` (or a bounded linked token on the POST).

---

### 🟡 Optional hardening (non-blocking; good follow-ups)

- **Architecture** — `SandboxSessionRegistry` is a ~914-line god class with ~5 responsibilities; this PR added 3 more state maps to it. Extract the session→thread routing/dedup and the sub-agent-binding catalog before the *next* sandbox feature lands here. Also: the agent-factory closure in `Program.cs` resolves singletons via `sp.GetRequiredService` at invocation time (some already captured as locals) — hoist them.
- **Performance** — `DestroySessionAsync` takes no token and disposals run serially (slow shutdown on a stalled gateway); `WorkspaceSelector.loadAvailableMarketplaces` refetches on every form-open with no `AbortController` (write-after-unmount) — mirror the `useMarketplaces` abort fix and consider caching the catalog.
- **Test coverage** — add cheap tests for: `CliCredentialCopilotTokenProvider` source *priority ordering* (the shell-out itself is fine to leave uncovered); `useMarketplaces` abort-race guards; `useWorkspaces` selection-fallback + `workspacesApi`.
- **Robustness (LOW)** — normalize a persisted `"marketplaces": null` → `[]` on read (and add a legacy-blob round-trip test); `FileWorkspaceStore` swallows `JsonException` with no log breadcrumb; consider a `MarketplaceAliases.Format` companion; a wire-casing test pinning the default camelCase contract the SPA depends on.

---

### ✅ Verified clean
- **Secrets/EUII:** credential handling is disciplined — Copilot/keychain tokens and the shared secret are never logged or placed in exceptions; request-logging captures no bodies/auth headers. (OAuth account usernames at Info are EUII but intended for this sample.)
- **Schema/compat:** no field/enum renames; persisted workspaces written before `marketplaces` existed round-trip safely (missing → `[]`, verified empirically); the gateway `marketplaces` field is correctly **omitted-when-unset** (not `[]`), so the gateway keeps its default set; C#↔TS casing matches.

---
*Reviewed at merge-base `8ffcf9e` → PR head `f440d5e`. Findings consolidated from 6 specialized review agents + a grading pass.*

gautam-achieveai:
Addressed all 3 blocking findings in c34a6f6.

**B1 (HIGH) — history recovery can abort agent startup.** `MultiTurnAgentBase.RecoverAsync` set `_historyRecovered = true` before awaiting `Store.LoadMetadataAsync`/`LoadMessagesAsync`, so a transient store/IO/deserialization fault crashed `RunAsync` startup AND poisoned the flag (a retry then skipped recovery). Fixed both: moved the flag set to after a successful or definitively-empty load, and wrapped the `RunAsync` startup call so failures degrade to empty history (resetting the flag for a later explicit retry) while genuine caller-cancellation still propagates. The double-restore guard is preserved.

**B2 (HIGH) — idempotency guard untested.** Added `RunAsync_DoesNotReRecover_AfterExplicitRecoverAsync` to `MultiTurnAgentBaseTests`: persists N messages, calls `RecoverAsync()` explicitly, then `RunAsync()`, and asserts history stays at N (not 2N) since `RestoreHistory` appends. Mirrors the existing recovery test's setup and uses a poll-until-deadline (no fixed `Task.Delay` sleeps).

**B3 (MEDIUM) — sandbox HttpClient had no timeout.** Gave the `SandboxSessionRegistry` `HttpClient` in `Program.cs` `Timeout = TimeSpan.FromSeconds(30)`, matching the sibling marketplace catalog client convention from this PR. Its create-POST runs sync-over-async on the WebSocket request thread, so the 100s default could stall it. Only the registry's client was touched.

Verification: `dotnet build LmDotnetTools.sln` clean (0 warnings/0 errors); `LmMultiTurn.Tests` 269 passing (incl. the new test); `LmStreaming.Sample.Tests` 252 passing.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review (since `f440d5ec`) - commit `c34a6f68`.

---

## All Blockers Resolved

This commit cleanly addresses the 3 blockers raised in the multi-agent review:

| Blocker | Fix | Verification |
|---------|-----|-------------|
| **B1** - Recovery can abort agent startup | RunAsync wraps RecoverAsync in try/catch; only genuine cancellation propagates; _historyRecovered deferred to after successful load | Code review |
| **B2** - Double-restore guard untested | New test `RunAsync_DoesNotReRecover_AfterExplicitRecoverAsync` asserts history count = N (not 2N) | Test passes |
| **B3** - Sandbox HttpClient unbounded timeout | `new HttpClient { Timeout = FromSeconds(30) }` with rationale comment | Code review |

## Implementation Quality

- **Flag placement is correct:** `_historyRecovered = true` is now set only at the 3 definitive exit points of `RecoverAsync` (no-metadata, no-messages, successful-load), never before IO that could throw transiently.
- **Retry semantics are sound:** On catch in `RunAsync`, `_historyRecovered = false` allows a later explicit `RecoverAsync()` call to retry.
- **Test is well-structured:** Uses `InMemoryConversationStore` with pre-seeded history, exercises the explicit-then-implicit double-recover path, and asserts both count and content identity.

## Build and Test
- `dotnet build` - 0 warnings, 0 errors
- New test - 1/1 passed (454ms)

---

**Verdict: APPROVE** - All blockers resolved, implementation is correct, no new issues. Previous MEDIUM/optional items (god class, IHttpClientFactory, etc.) remain valid tech debt for follow-up but are non-blocking.

PR #92, opened 2026-06-21
fix(sandbox): recreate gateway session on 404 so marketplace tools persist

gautam-achieveai:
## Problem

The Workspace Agent silently loses its marketplace-provided tools (e.g. `sandbox-Skill`) after the sandbox gateway has been idle for a while. A `tool_schema`/`tools_list` probe shows only the 10 base sandbox tools; `tool_schema{name:"sandbox-Skill"}` returns *"Tool 'sandbox-Skill' not found"*.

## Root cause

`SandboxSessionRegistry` caches sandbox sessions per-workspace in memory (a `Lazy<Task<SandboxSession>>`) and reuses them without re-validating. The gateway evicts idle sessions on its own schedule — after which every call for that id returns `404 "Session not found"`. The app kept handing back the dead handle, so the gateway no longer mounted the workspace's marketplaces and only its always-on base tools remained. The session was only re-created by a full app restart (which clears the cache).

Confirmed via logs + live gateway: the same session that successfully called `sandbox-Skill` in the morning later 404'd (`GET /api/v1/sandboxes/{id}` → 404; `GET /api/v1/sandboxes` → `{"sandboxes": []}`), and a fresh Demo2 session (marketplaces `ClaudePlugins, knowledge-work-plugins, superpowers`) restored `sandbox-Skill`.

## Fix

- **`GetOrCreateLiveSessionAsync`** probes the gateway for the cached session and, on a definitive `404`, evicts the stale cache entry and re-creates the session with the same workspace + marketplace selection. Non-404 / transient probe failures are treated as "still alive" so a flaky gateway never churns a healthy session.
- **`Program.cs`** uses it at workspace-agent session acquisition (before the MCP client connects).
- **`InvalidateSession`** evicts only the exact dead `Lazy` (matching `AwaitAndEvictOnFailureAsync`'s discipline) so a concurrent caller's fresh session is never clobbered / orphaned.
- The liveness probe is capped at 5s (linked CTS) so a hung gateway can't stall a turn for the full client timeout.

## Tests

`SandboxSessionRegistryRecreateOnGateway404Tests` (6): recreate-on-404, reuse-when-alive, no-churn-on-503, **marketplace preservation on recreate**, reverse-map cleanup, concurrent-recreate convergence.

All registry tests (38) and the full `LmStreaming.Sample.Tests` project (261) pass; concurrency test 5/5; sample builds 0 warnings / 0 errors.

### DIFF
--- a/samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
+++ b/samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
@@ -1,4 +1,5 @@
 using System.Collections.Concurrent;
+using System.Net;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using AchieveAi.LmDotnetTools.LmCore.Agents;
@@ -69,6 +70,12 @@ public sealed class SandboxSessionRegistry : IAsyncDisposable
     /// <see cref="SandboxGatewayOptions.Workspace"/>. Also the seeded id used by the workspace store.
     /// </summary>
     public const string DefaultWorkspaceId = "default";
+
+    /// <summary>
+    /// Upper bound on the gateway session-liveness probe so a wedged gateway degrades to "assume
+    /// alive" quickly rather than blocking the turn for the full shared-client timeout.
+    /// </summary>
+    private static readonly TimeSpan SessionLivenessProbeTimeout = TimeSpan.FromSeconds(5);
 
     private static readonly JsonSerializerOptions JsonOptions = new()
     {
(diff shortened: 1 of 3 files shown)


gautam-achieveai:
**[Gautam's review bot]** Deep review complete.

## Summary
Clean, well-documented fix with thorough test coverage and correct concurrency handling. The liveness probe + invalidate + recreate pattern is sound - `InvalidateSession`'s `ReferenceEquals` guard prevents clobbering, and `GetOrAdd` single-flights recreation.

## Findings

### [MEDIUM] Unconditional HTTP probe on every agent creation adds latency
Every call to `GetOrCreateLiveSessionAsync` now issues an HTTP GET before returning. Under the existing sync-over-async pattern, this blocks a thread pool thread for the full round-trip (up to 5s timeout on a wedged gateway).

**Suggestions for follow-up:**
1. **Freshness TTL** - skip the probe if session was created/verified within the last 30-60s. A newly-created session won't be evicted immediately.
2. **`HttpCompletionOption.ResponseHeadersRead`** - since only the status code matters, don't buffer the response body.
3. Consider reducing the timeout from 5s to 2s on the creation path.

### [LOW] `LogDebug` for probe failures may be invisible in production
Other similar "degraded but operational" scenarios in this file use `LogWarning`. If the gateway becomes persistently unreachable, probe failures will be silent at typical production log levels. Consider `LogInformation` as a middle ground.

## What looked good
- Exception handling in `IsSessionAliveAsync` is textbook: proper cancellation token filter, justified broad catch, fail-safe "assume alive" on error
- `InvalidateSession` correctly uses exact-match removal so concurrent recreations are never clobbered
- Test suite covers 6 key scenarios (recreate-on-404, reuse-when-alive, no-churn-on-503, marketplace preservation, reverse-map cleanup, concurrent convergence)
- Thread-safe `CallLog` stub models gateway behavior cleanly

## Verdict: **APPROVE** 
The latency concern is real but acceptable for a correctness fix, and can be optimized in a follow-up.

samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
```diff
@@ -229,6 +236,129 @@ private async Task<SandboxSession> AwaitAndEvictOnFailureAsync(string workspaceI
...
+    )
+    {
+        ObjectDisposedException.ThrowIf(_disposed, this);
+        ArgumentNullException.ThrowIfNull(workspaceRef);
+
+        var workspaceId = string.IsNullOrWhiteSpace(workspaceRef.Id) ? DefaultWorkspaceId : workspaceRef.Id;
+        var effectiveRef = workspaceRef with { Id = workspaceId };
+
+        var session = await GetOrCreateSessionAsync(effectiveRef, ct).ConfigureAwait(false);
+        if (await IsSessionAliveAsync(session.SessionId, ct).ConfigureAwait(false))
```
gautam-achieveai, line 275: **[Gautam's review bot]** [MEDIUM] Every turn now pays an HTTP round-trip here before returning the cached session. Consider a freshness TTL - e.g. track when the session was last verified and skip the probe if it was within the last 30-60s. A session that was just created/verified won't be evicted immediately by the gateway.

Also: since only the status code matters, `HttpCompletionOption.ResponseHeadersRead` in `IsSessionAliveAsync` would let the probe return as soon as the status line arrives.

samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
```diff
@@ -229,6 +236,129 @@ private async Task<SandboxSession> AwaitAndEvictOnFailureAsync(string workspaceI
...
+        }
+        catch (OperationCanceledException) when (ct.IsCancellationRequested)
+        {
+            throw; // genuine caller cancellation — propagate
+        }
+        catch (Exception ex)
+        {
+            // Probe timeout or transient error → assume alive so a flaky/slow gateway never churns a
+            // healthy session (recreating wouldn't help if the gateway is unreachable anyway).
+            _logger.LogDebug(
```
gautam-achieveai, line 323: **[Gautam's review bot]** [LOW] `LogDebug` is typically disabled in production. If the gateway becomes persistently unreachable, these probe failures will be invisible - the system will silently assume all sessions are alive with no observable signal. Other similar degradation paths in this file use `LogWarning` (e.g. the recreation warning above at line 280). Consider `LogInformation` here, or at minimum a counter/metric.

Issue #93, opened 2026-06-21
sandbox: optimize per-turn gateway liveness probe (freshness TTL)


gautam-achieveai:
Follow-up from the #92 review (deferred, non-blocking).

`SandboxSessionRegistry.GetOrCreateLiveSessionAsync` issues an HTTP `GET /api/v1/sandboxes/{id}` liveness probe on every workspace-agent session acquisition (per new turn/agent-creation). Under the sync-over-async call site this blocks a thread-pool thread for the round-trip (capped at 5s).

Already addressed in the follow-up commit:
- `HttpCompletionOption.ResponseHeadersRead` (status-only, no body buffering).
- Probe-failure log raised Debug → Information.

Still open (this issue):
- **Freshness TTL** — track when a session was last created/verified and skip the probe if within ~30–60s; a just-created session won't be evicted immediately by the gateway. This removes the probe from the hot path for back-to-back turns.
- Consider reducing the probe timeout from 5s → 2s on the creation path.

Source: review of #92 ("Gautam's review bot", MEDIUM finding).

PR #94, opened 2026-06-21
refactor(sandbox): status-only liveness probe + observable probe-failure log

gautam-achieveai:
Follow-up to #92 addressing the reviewer's two non-blocking inline findings (the PR merged before these landed).

## Changes
- **[MEDIUM, partial]** `IsSessionAliveAsync` only checks the status code, so it now uses `HttpCompletionOption.ResponseHeadersRead` instead of buffering the (404) response body.
- **[LOW]** Probe-failure log raised `Debug` → `Information` so a persistently unreachable gateway (which makes every session "assume alive") is observable at typical production log levels.

## Deferred
The reviewer's freshness-TTL suggestion (skip the probe when a session was verified <30–60s ago) and the 5s→2s timeout tuning are tracked in #93 — they add per-session state and are out of scope for this small follow-up.

## Verification
Build 0 warnings / 0 errors; `SandboxSessionRegistry` tests 38/38 pass.

### DIFF
--- a/samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
+++ b/samples/LmStreaming.Sample/Services/SandboxSessionRegistry.cs
@@ -309,7 +309,9 @@ public sealed class SandboxSessionRegistry : IAsyncDisposable
         probeCts.CancelAfter(SessionLivenessProbeTimeout);
         try
         {
-            using var response = await _httpClient.GetAsync(requestUri, probeCts.Token).ConfigureAwait(false);
+            using var response = await _httpClient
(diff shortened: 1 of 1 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** ✅ LGTM

Clean, well-scoped follow-up to #92. Both changes are correct:

1. **`ResponseHeadersRead`** — proper optimization since only `StatusCode` is inspected; avoids buffering potentially large error bodies on 404s.
2. **`Debug` → `Information`** — makes persistently failing probes observable at production log levels, which is important since the assume-alive fallback silently masks gateway outages.

No issues found. Behavior is unchanged, just more efficient and more observable.

PR #105, opened 2026-06-24
fix: message-identity (pillbox grouping) + incremental render on conversation switch + Anthropic empty-id result guard

### LINKED ISSUE #104: Message-identity defects break tool-call pillbox grouping & incremental render (LmStreaming.Sample)
## Summary

A cluster of message-identity defects in the streaming pipeline (plus two related frontend state bugs) cause two visible problems in **LmStreaming.Sample**, and one replay failure on Anthropic-compatible providers.

### 1. Tool calls render in the wrong "pillbox" (GPT-5.5 / Copilot)
For the GPT-5.5 (Copilot / OpenAI-reasoning) provider, tool-call messages are streamed with the **wrong `GenerationId`**. Evidence from a recorded WebSocket stream:

- `run_assignment` advertises `GenerationId` = a clean 32-hex run id.
- Every `tool_call` / `tool_call_result` carries `generation_id` = the provider's **opaque encrypted response token** (many distinct values), *not* the run id.
- `text_update` / `usage` carry no `generation_id`.

The client groups pills by the merge key `kind-runId-generationId-messageOrderIdx`, so a mismatched `generationId` means tool-call pills can't be grouped to their run and land in the wrong pillbox.

**Root cause:** the run mints a `generationId` but never threads it into `GenerateReplyOptions`; providers stamp their own per-message id instead. (`toolCallIdx` always being `0` in the recording is a *downstream symptom* of this, not a separate bug.)

### 2. Incremental rendering breaks after switching conversations and back
After switching to another conversation and back, streaming no longer renders incrementally.

**Root cause (two contributing frontend defects):**
- `useChat.loadMessagesFromBackend` indexes rehydrated messages by `pm.id`, while live streaming indexes by `getMergeKey(...)` — so a later streaming update can't merge into a rehydrated message and duplicates instead (violates the documented merge-key contract).
- The WebSocket reuse path reused a socket bound to a *different* conversation's thread (no `threadId` check), and the switch-back path never disconnected it.

### 3. Anthropic empty-id server-tool *result* orphan (replay 400)
A prior fix dropped empty-id `server_tool_use` blocks (Kimi omits the id), but the **singular `ToolCallResultMessage` branch** had no matching guard, so an orphaned empty-id `tool_result` could survive and 400 on replay.

## Acceptance
- Every message in a run carries the run's `GenerationId`, regardless of provider.
- Locally-executed tool results carry a `MessageOrderIdx`.
- Rehydrated conversations merge subsequent streaming updates in place (no duplicate bubbles).
- A socket bound to a different thread is not reused after a conversation switch.
- Empty-id server-tool results are dropped on request rebuild.
- All fixes covered by regression tests (TDD).

gautam-achieveai:
Closes #104.

Fixes a cluster of message-identity defects in the streaming pipeline plus two related frontend state bugs. All changes are TDD (failing test first, then minimal fix). Branches were developed in parallel by file-disjoint area and merged.

### Backend — message identity
- **H1 — tool-call `GenerationId` mis-stamped.** The run mints a `generationId` but never threads it into `GenerateReplyOptions`, so providers stamp their own per-message id (for GPT-5.5/Copilot, an opaque encrypted response token). The client merge key `kind-runId-generationId-messageOrderIdx` then can't group tool-call pills to their run → tools render in the wrong pillbox.
  - Add `GenerateReplyOptions.GenerationId`; `MessageExtensions.WithIds(options)` now stamps the run's `GenerationId` onto every message arm (overriding the provider's, preserving the provider value when the run advertises none). This is the single seam every provider already calls to apply run identity, so it is provider-agnostic — Anthropic benefits too with no provider edits. `MultiTurnAgentLoop` threads the run's generation into options.
- **H3b — local tool-result `MessageOrderIdx` was `null`.** Locally-executed tool results are published out-of-band, bypassing the ordering middleware. Stamp `MessageOrderIdx = callOrderIdx + 1` in `MultiTurnAgentLoop`.
- **H3a — `toolCallIdx` always `0`:** investigated and intentionally **not** changed — proven (via a test against the real reconstruction pipeline) to be a downstream symptom of H1's `generationId` fragmentation, already resolved by H1.

### Frontend — `useChat` (incremental render after switching conversations)
- **Rehydration merge-key mismatch.** `loadMessagesFromBackend` indexed rehydrated messages by `pm.id` while live streaming indexes by `getMergeKey(...)`; a later streaming update couldn't merge and duplicated. Now rehydrate by `getMergeKey(parsedMessage)` (restores the documented contract).
- **Cross-thread socket reuse.** The reuse path reused a socket bound to a different conversation's thread. Now the reuse guard requires `wsConnection.threadId === threadId.value`; otherwise it closes and recreates with the current callbacks.

### Anthropic — replay 400
- Guard the **singular `ToolCallResultMessage`** branch in `AnthropicRequest` with the existing `IsEmptyIdServerToolCall` check so an orphaned empty-id server-tool `tool_result` is dropped on request rebuild (the aggregate path was already guarded).

## Tests (TDD, red→green verified)
- `tests/LmCore.Tests/Messages/MessageExtensionsWithIdsTests.cs` — run `GenerationId` overrides the provider's opaque id; preserved when the run advertises none.
- `tests/LmMultiTurn.Tests/MultiTurnAgentLoopTests.cs` — local tool result carries `MessageOrderIdx`.
- `tests/AnthropicProvider.Tests/Models/ServerToolRequestTests.cs` — empty-id server-tool result dropped.
- `ClientApp/.../useChat.test.ts` — rehydrated message merges a later streaming update in place (no duplicate); a socket bound to a different thread is not reused. (Each verified to fail against the pre-fix code.)

## Verification
- `dotnet build LmDotnetTools.sln`: clean, 0 warnings / 0 errors.
- Changed projects (isolated): LmCore 746, LmMultiTurn 276, OpenAIProvider 89, AnthropicProvider 169, LmStreaming.Sample 279 — all green. Frontend: 237/237.
- Pre-existing/unrelated and **not** caused by this PR: `McpIntegrationTests` ×4 (hardcoded `McpSampleServer.exe`, can't run on macOS); `LmMultiTurn` ×2 only under full-parallel `dotnet test` load (passes 276/276 in isolation).

## Notes / follow-ups (out of scope here)
- The two frontend fixes restore the documented merge-key contract and are correct, but reproducing the *exact* on-screen symptom would benefit from a `Browser.E2E.Tests` switch-away-and-back scenario (proposed follow-up).
- Unrelated to code: a live `GITHUB_COPILOT_TOKEN` in `samples/LmStreaming.Sample/.env` should be rotated; the GitHub-auth Prod→Dev workaround is a config/symptom fix (missing prod `ClientId`).

### DIFF
--- a/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
@@ -1080,8 +1080,16 @@ export function useChat(options: UseChatOptions = {}) {
           isStreaming: false,
         };
 
+        // Stream persistence can hold several records that collapse to one logical merge key
+        // (e.g. an intermediate update record beside the finalizing message, same
+        // run/generation/messageOrderIdx). Append to messageOrder only on FIRST insert; otherwise
+        // overwrite the existing messageIndex entry in place so the final record wins WITHOUT
+        // accumulating a duplicate key that would render/sort the same message multiple times.
+        const isFirstInsert = !messageIndex.value.has(mergeKey);
         messageIndex.value.set(mergeKey, chatMessage);
-        messageOrder.value.push(mergeKey);
+        if (isFirstInsert) {
+          messageOrder.value.push(mergeKey);
+        }
       } catch (e) {
         log.warn('Failed to parse persisted message', { messageId: pm.id, error: e });
       }
--- a/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
@@ -97,17 +97,21 @@ function getMergeKey(msg: Message): string {
   const messageOrderIdx = msg.messageOrderIdx ?? 0;
   const mergeKind = getMergeKind(msg);
   
-  // For tool call updates, include toolCallIdx
-  if (isToolCallUpdateMessage(msg) || isToolsCallUpdateMessage(msg)) {
-    if (isToolCallUpdateMessage(msg)) {
-      // Individual tool call - use tool_call_id as unique identifier
-      return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}-${msg.tool_call_id || 'tc'}`;
-    } else {
-      // ToolsCallUpdate - use messageOrderIdx only (tools are accumulated in array)
-      return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`;
-    }
+  // Individual tool calls — streaming OR finalized — are keyed by tool_call_id. Several concurrent
+  // tool calls in one turn share runId/generationId/messageOrderIdx and differ only by tool_call_id
+  // (e.g. GPT-5.5 via the OpenAI Responses API); without it they collapse into a single pill.
+  if (isToolCallUpdateMessage(msg) || isToolCallMessage(msg)) {
+    return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}-${msg.tool_call_id || 'tc'}`;
   }
-  
+
+  // A finalized single-call ToolsCallMessage is the same case wrapped in the aggregate type —
+  // disambiguate by its tool_call_id too. Multi-call aggregates already carry every call in one
+  // message (rendered as N pills), so they key on messageOrderIdx only.
+  if (isToolsCallMessage(msg) && msg.tool_calls?.length === 1 && msg.tool_calls[0]?.tool_call_id) {
+    return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}-${msg.tool_calls[0].tool_call_id}`;
+  }
+
+  // ToolsCallUpdate accumulates tools into one array → key on messageOrderIdx only.
   return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`;
 }
 
--- a/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
@@ -854,13 +854,21 @@ export function useChat(options: UseChatOptions = {}) {
   ): Promise<void> {
     const effectiveThreadId = getOrCreateThreadId();
     
-    // Check if we have an open connection
-    if (wsConnection && wsConnection.isConnected && wsConnection.socket.readyState === WebSocket.OPEN) {
-      log.info('Reusing existing WebSocket connection', { 
+    // Check if we have an open connection that belongs to the current thread.
+    // A socket bound to a previously-viewed conversation must not be reused for a
+    // different thread (e.g. after switching conversations); close it and fall
+    // through to create a fresh connection with the current callbacks instead.
+    if (
+      wsConnection &&
+      wsConnection.isConnected &&
+      wsConnection.socket.readyState === WebSocket.OPEN &&
+      wsConnection.threadId === threadId.value
+    ) {
+      log.info('Reusing existing WebSocket connection', {
         connectionId: wsConnection.connectionId,
-        threadId: wsConnection.threadId 
+        threadId: wsConnection.threadId
       });
-      
+
       // Send message on existing connection
       const { sendWebSocketMessage } = await import('@/api/wsClient');
       sendWebSocketMessage(wsConnection, text);
@@ -1042,9 +1050,21 @@ export function useChat(options: UseChatOptions = {}) {
           }
         }
 
+        // Ensure the parsed message carries the persisted identity fields so the
+        // merge key matches what live streaming computes for the same logical message.
+        parsedMessage.runId = parsedMessage.runId ?? pm.runId;
+        parsedMessage.parentRunId = parsedMessage.parentRunId ?? pm.parentRunId ?? undefined;
+        parsedMessage.generationId = parsedMessage.generationId ?? pm.generationId ?? undefined;
+        parsedMessage.messageOrderIdx = parsedMessage.messageOrderIdx ?? pm.messageOrderIdx ?? undefined;
+
+        // Index rehydrated messages by the same merge key used by live streaming
+        // (kind-runId-generationId-messageOrderIdx) so a subsequent streaming update
+        // sharing that identity merges in place instead of creating a duplicate bubble.
+        const mergeKey = getMergeKey(parsedMessage);
+
         // Create chat message
         const chatMessage: InternalChatMessage = {
-          id: pm.id,
+          id: mergeKey,
           role,
           status: 'completed',
           content: parsedMessage,
@@ -1056,8 +1076,8 @@ export function useChat(options: UseChatOptions = {}) {
           isStreaming: false,
         };
 
-        messageIndex.value.set(pm.id, chatMessage);
-        messageOrder.value.push(pm.id);
+        messageIndex.value.set(mergeKey, chatMessage);
+        messageOrder.value.push(mergeKey);
       } catch (e) {
         log.warn('Failed to parse persisted message', { messageId: pm.id, error: e });
       }
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -710,37 +710,10 @@ try
                         ? allBuiltInTools
                         : ModeToolFilter.FilterBuiltInTools(allBuiltInTools, mode.EnabledTools);
 
-                    // Surface model reasoning. Anthropic-format providers (incl. the Copilot-backed
-                    // sonnet/haiku — the Copilot proxy accepts the thinking parameter, verified by
-                    // CopilotAnthropicLiveTests.Thinking_param_is_accepted_by_copilot_backend) get
-                    // extended thinking; the OpenAI Responses providers (gpt-5.5/gpt-5.5-mini) get a
-                    // reasoning summary. Without this, Copilot-backed models returned no thinking blocks.
-                    var extraProperties = ImmutableDictionary<string, object?>.Empty;
-                    if (
-                        string.Equals(normalizedProviderId, "anthropic", StringComparison.Ordinal)
-                        || string.Equals(normalizedProviderId, "test-anthropic", StringComparison.Ordinal)
-                        || string.Equals(normalizedProviderId, "sonnet", StringComparison.Ordinal)
-                        || string.Equals(normalizedProviderId, "haiku", StringComparison.Ordinal)
-                    )
-                    {
-                        var budgetTokens = int.TryParse(
-                            Environment.GetEnvironmentVariable("ANTHROPIC_THINKING_BUDGET"),
-                            out var parsed
-                        )
-                            ? parsed
-                            : 2048;
-                        extraProperties = extraProperties.Add("Thinking", new AnthropicThinking(budgetTokens));
-                    }
-                    else if (
-                        string.Equals(normalizedProviderId, "gpt-5.5", StringComparison.Ordinal)
-                        || string.Equals(normalizedProviderId, "gpt-5.5-mini", StringComparison.Ordinal)
-                    )
-                    {
-                        extraProperties = extraProperties.Add(
-                            "Reasoning",
-                            new ResponseReasoningOptions { Summary = "auto" }
-                        );
-                    }
+                    // Surface model reasoning (provider→Thinking/Reasoning mapping). Extracted to a
+                    // testable helper so the per-provider wiring is regression-guarded; see
+                    // ProgramReasoningExtraPropertiesTests.
+                    var extraProperties = BuildReasoningExtraProperties(normalizedProviderId);
 
                     // Sub-agent orchestration options. Only the middleware providers reach this
                     // path — the CLI providers (codex/claude/copilot and their *-mock variants)
@@ -1056,6 +1029,44 @@ finally
 
 public partial class Program
 {
+    /// <summary>
+    ///     Maps a normalized provider id to the reasoning/thinking request options that surface a
+    ///     model's reasoning. Anthropic-format providers (incl. the Copilot-backed sonnet/haiku — the
+    ///     Copilot proxy accepts the thinking parameter, verified by
+    ///     CopilotAnthropicLiveTests.Thinking_param_is_accepted_by_copilot_backend) get an extended-
+    ///     thinking budget; the OpenAI Responses providers (gpt-5.5/gpt-5.5-mini) get a reasoning-
+    ///     summary request (CopilotResponsesLiveTests confirms gpt-5.5 returns reasoning). Other
+    ///     providers get none. Without this wiring, Copilot-backed models return no thinking blocks.
+    /// </summary>
+    internal static ImmutableDictionary<string, object?> BuildReasoningExtraProperties(string normalizedProviderId)
+    {
+        var extraProperties = ImmutableDictionary<string, object?>.Empty;
+        if (
+            string.Equals(normalizedProviderId, "anthropic", StringComparison.Ordinal)
+            || string.Equals(normalizedProviderId, "test-anthropic", StringComparison.Ordinal)
+            || string.Equals(normalizedProviderId, "sonnet", StringComparison.Ordinal)
+            || string.Equals(normalizedProviderId, "haiku", StringComparison.Ordinal)
+        )
+        {
+            var budgetTokens = int.TryParse(
+                Environment.GetEnvironmentVariable("ANTHROPIC_THINKING_BUDGET"),
+                out var parsed
+            )
+                ? parsed
+                : 2048;
+            extraProperties = extraProperties.Add("Thinking", new AnthropicThinking(budgetTokens));
+        }
+        else if (
+            string.Equals(normalizedProviderId, "gpt-5.5", StringComparison.Ordinal)
+            || string.Equals(normalizedProviderId, "gpt-5.5-mini", StringComparison.Ordinal)
+        )
+        {
+            extraProperties = extraProperties.Add("Reasoning", new ResponseReasoningOptions { Summary = "auto" });
+        }
+
+        return extraProperties;
+    }
+
     /// <summary>
     ///     Creates a test-mode agent using an <see cref="ITestAgentBuilder"/>-supplied handler.
     /// </summary>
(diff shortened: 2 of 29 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review (iterations 7-8, commits `856c2bc...a11f1bd`)

## Verdict: APPROVE :white_check_mark:

### What changed since last review (`9714f80`)

| Commit | Summary |
|--------|---------|
| `856c2bc` | **H1 (Responses path)**: Threads `GenerationId` from `GenerateReplyOptions` into `EventStreamToMessages`. Run's GenerationId now wins over the provider's opaque `resp_...` id, fixing tool-call pillbox grouping for GPT-5.5 (Copilot). Fallback to provider id preserved when run advertises none. |
| `a11f1bd` | **Reasoning support**: Maps `ExtraProperties["Reasoning"]` to `ResponseCreateRequest.Reasoning` in `MessageMapper`, enabling thinking/reasoning summaries for GPT-5.5. Extends thinking-budget gate to Copilot-backed `sonnet`/`haiku` providers. Canary live test confirms Copilot proxy accepts the thinking param. |

### Test verification
- OpenAiResponsesProvider.Tests: **59/59 passed** (includes 2 new GenerationId tests + 2 new reasoning-mapping tests)

### Observations (non-blocking)

1. **[GOOD]** The `ScriptedClient` test helper in `OpenAiResponsesAgentGenerationIdTests` is clean and reusable - tests the real `EventStreamToMessages` path rather than testing `WithIds` in isolation, which wouldn't have caught this bug.

2. **[GOOD]** The canary test `Thinking_param_is_accepted_by_copilot_backend` is a smart regression signal: if Copilot ever starts rejecting the thinking param (like `web_search`), this test fails - signal to gate it off. Doc comment explains this clearly.

3. **[INFO]** Provider-ID string comparisons in `Program.cs` are growing (6 cases now). Sample code, so acceptable - but may benefit from a config-driven approach if more providers arrive.

No blocking issues. Changes are well-structured, well-tested, and correctly extend the H1 fix to the Responses path that `WithIds` couldn't reach.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review of iteration 11 (`65f6cf0...82abf42`).

## Summary

Clean testability refactoring: extracts the inline provider->reasoning extra-properties mapping into a `BuildReasoningExtraProperties` internal static helper and adds 9 regression tests covering all branches (anthropic-format, responses-format, other). No behavioral change.

## Findings

**[LOW]** `ProgramReasoningExtraPropertiesTests.Build()` uses reflection (`GetType(""Program"")` + `GetMethod(...)`), but `InternalsVisibleTo Include=""LmStreaming.Sample.Tests""`) is already configured in the sample's `.csproj`. You can call `Program.BuildReasoningExtraProperties(providerId)` directly — simpler, refactor-safe, and eliminates the null-check ceremony.

**Verdict: APPROVE** :white_check_mark: — no blocking issues.

gautam-achieveai (commented):
[Gautam's review bot] Incremental review of commits `82abf42...3f9e912` (iterations 12-13).

## Summary

1. **ddf7410 - test(responses): mock emits real reasoning_summary streaming grammar** - Makes mock SSE writer faithful to real GPT-5.5 behavior (empty summary array + streaming delta events). Agent test now exercises the WHOLE reasoning flow and asserts delta concatenation equals final text.

2. **3f9e912 - fix(useChat): key finalized tool calls by tool_call_id** - Fixes concurrent tool calls collapsing into a single pill (they share runId/generationId/messageOrderIdx after H1 unification). Clean three-tier merge-key approach: individual calls keyed by tool_call_id, single-call wrapper keyed by inner tool_call_id, multi-call aggregates keyed by messageOrderIdx only. Regression test drives 3 concurrent calls and asserts 3 distinct pills in 1 pillbox.

## Verdict: APPROVE

No blocking issues. Both changes are well-tested, minimal, and correct. The merge-key fix is a natural consequence of the H1 GenerationId unification.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
```diff
@@ -1056,8 +1080,8 @@ export function useChat(options: UseChatOptions = {}) {
           isStreaming: false,
         };
 
-        messageIndex.value.set(pm.id, chatMessage);
-        messageOrder.value.push(pm.id);
+        messageIndex.value.set(mergeKey, chatMessage);
+        messageOrder.value.push(mergeKey);
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] Rehydrating by `mergeKey` fixes live-update lookup, but this still pushes every persisted record into `messageOrder`. Stream persistence can contain many updates with the same logical merge key, so `messageIndex` overwrites while `messageOrder` accumulates duplicates and later render/sort can process the same final message many times. Please only append on first insert, merge/overwrite existing entries, or hydrate through the live update-merging path.

src/OpenAiResponsesProvider/Agents/OpenAiResponsesAgent.cs
```diff
@@ -190,6 +204,34 @@ [EnumeratorCancellation] CancellationToken cancellationToken
 
                     break;
 
+                case ResponseReasoningSummaryTextDeltaEvent reasoningDelta:
+                    // Reasoning summary streams as its own event channel (the reasoning output_item's
+                    // summary array stays empty until done) — surface each delta as a reasoning update.
+                    yield return new ReasoningUpdateMessage
+                    {
+                        Reasoning = reasoningDelta.Delta,
+                        Visibility = ReasoningVisibility.Plain,
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] Reasoning summary events should not be emitted as `ReasoningVisibility.Plain`. These are provider summaries; persisting them as plain reasoning means Anthropic replay can serialize them as unsigned thinking instead of summary text. Please use `ReasoningVisibility.Summary` here and for the done event / summary-array extraction, with regression tests.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
```diff
@@ -97,17 +97,21 @@ function getMergeKey(msg: Message): string {
...
-      return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}-${msg.tool_call_id || 'tc'}`;
-    } else {
-      // ToolsCallUpdate - use messageOrderIdx only (tools are accumulated in array)
-      return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`;
-    }
+  // Individual tool calls — streaming OR finalized — are keyed by tool_call_id. Several concurrent
+  // tool calls in one turn share runId/generationId/messageOrderIdx and differ only by tool_call_id
+  // (e.g. GPT-5.5 via the OpenAI Responses API); without it they collapse into a single pill.
+  if (isToolCallUpdateMessage(msg) || isToolCallMessage(msg)) {
+    return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}-${msg.tool_call_id || 'tc'}`;
```
gautam-achieveai, line 104: [gautam-achieveai's bot] [BLOCKER] The merge key depends on `msg.generationId`, but backend/tool JSON can use `generation_id` and the WebSocket normalizer only aliases leading-uppercase keys, not snake_case. Those messages fall back to `default` and fail to merge/group correctly. Please alias `generation_id -> generationId` at normalization or dual-read here, and test both wire shapes.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -709,21 +710,10 @@
...
-                            out var parsed
-                        )
-                            ? parsed
-                            : 2048;
-                        extraProperties = extraProperties.Add("Thinking", new AnthropicThinking(budgetTokens));
-                    }
+                    // Surface model reasoning (provider→Thinking/Reasoning mapping). Extracted to a
+                    // testable helper so the per-provider wiring is regression-guarded; see
+                    // ProgramReasoningExtraPropertiesTests.
+                    var extraProperties = BuildReasoningExtraProperties(normalizedProviderId);
```
gautam-achieveai, line 716: [gautam-achieveai's bot] This PR appears to roll out broader reasoning/thinking enablement for additional providers, beyond the stated message-identity/incremental-render bugfix. Unless required for the fix, please split this into a separate PR/work item to reduce review and rollout risk.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
```diff
@@ -97,17 +97,21 @@ function getMergeKey(msg: Message): string {
...
   }
-  
+
+  // A finalized single-call ToolsCallMessage is the same case wrapped in the aggregate type —
+  // disambiguate by its tool_call_id too. Multi-call aggregates already carry every call in one
+  // message (rendered as N pills), so they key on messageOrderIdx only.
+  if (isToolsCallMessage(msg) && msg.tool_calls?.length === 1 && msg.tool_calls[0]?.tool_call_id) {
```
gautam-achieveai, line 110: [gautam-achieveai's bot] Please add a regression test for this finalized single-call `ToolsCallMessage` branch: two `MessageType.ToolsCall` messages sharing run/generation/order but with different `tool_call_id` should not collide. The current added coverage exercises individual `ToolCall` messages, not this aggregate single-call path.

src/LmCore/Core/GenerateReplyOptions.cs
```diff
@@ -147,6 +157,10 @@ public GenerateReplyOptions Merge(GenerateReplyOptions? other)
             BuiltInTools = other.BuiltInTools ?? BuiltInTools,
             ToolChoice = other.ToolChoice ?? ToolChoice,
             ContainerId = other.ContainerId ?? ContainerId,
+            RunId = other.RunId ?? RunId,
```
gautam-achieveai, line 160: [gautam-achieveai's bot] Please add a direct unit test for `GenerateReplyOptions.Merge` preserving/overriding the identity fields (`RunId`, `ParentRunId`, `ThreadId`, `GenerationId`). These are now part of the merge contract but appear untested.

src/LmCore/Messages/MessageExtensions.cs
```diff
@@ -214,32 +214,129 @@ public static bool IsTransformableToolCall(this IMessage message)
...
-            ToolCallMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ToolCallUpdateMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ToolCallResultMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ToolsCallMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ToolsCallUpdateMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ToolsCallResultMessage m => m with { RunId = runId, ThreadId = threadId },
-            ReasoningMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            ReasoningUpdateMessage m => m with { RunId = runId, ParentRunId = parentRunId, ThreadId = threadId },
-            UsageMessage m => m with { RunId = runId, ThreadId = threadId },
+            TextMessage m => m with
```
gautam-achieveai, line 268: [gautam-achieveai's bot] The `WithIds` generation-id stamping touches many message arms, but the tests only sample a few. Please add parameterized coverage across the message variants, especially update/result/usage arms, so future additions do not silently miss `GenerationId`.

gautam-achieveai (commented):
[gautam-achieveai's bot] ## PR review: would request changes

GitHub does not allow this account to request changes on its own PR, so I am posting this as a comment review. Substantively, this is a **REQUEST CHANGES** review.

I found three blocking correctness/schema issues that should be addressed before merge, plus a few scope/testing follow-ups:

- [BLOCKER] Rehydration now indexes persisted streamed updates by merge key but still appends every persisted record to `messageOrder`, allowing duplicate render/process entries for the same logical message.
- [BLOCKER] OpenAI Responses reasoning summaries are emitted/extracted as `Plain` visibility, which can corrupt cross-provider replay semantics.
- [BLOCKER] Frontend merge keys depend on camelCase `generationId`, but incoming tool JSON may provide `generation_id` and the normalizer only aliases Pascal/PascalCase-like keys.
- Medium: reasoning/thinking rollout appears broader than the stated bugfix.
- Medium/Low: several identity/merge-key paths need direct regression coverage.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review of commit `6d48950` (iteration 14).

## Verdict: APPROVE

### What changed since last review (`3f9e912`)

| Commit | Summary |
|--------|---------|
| `6d48950` | **fix(anthropic)**: Demote unsigned thinking blocks to `text` on multi-turn replay. Kimi (and similar Anthropic-compatible backends) emit thinking content without a signature; replaying that as a `thinking` block triggers `400 invalid_request_error`. The fix adds a second reverse-iteration pass in `MergeAdjacentThinkingBlocks` that: (1) demotes unsigned blocks with content to `text`, and (2) removes empty unsigned blocks. |

### Test verification
- ThinkingReplayTests: **2/2 passed**
- ThinkingWithToolCallSerializationTests: **5/5 passed** (updated assertions)

### Observations (non-blocking)

1. **[GOOD]** Two-pass design is correct - merge pass runs first, so valid text+signature pairs still merge. Only *remaining* unsigned blocks are demoted. Reverse iteration avoids index-shift bugs during removal.

2. **[GOOD]** `JsonIgnoreCondition.WhenWritingNull` on `Thinking`/`ThinkingSignature` properties ensures demoted `text` blocks don't leak stale `thinking`/`signature` fields in serialized JSON.

3. **[INFO]** This is a defensive guard at the Anthropic request-building layer. It complements (but does not replace) fixing the emission layer if OpenAI Responses reasoning-summary events are incorrectly tagged as `ReasoningVisibility.Plain` - that concern was raised in the previous review and remains a separate item.

No blocking issues in this iteration. The fix is focused, well-tested, and correctly handles the Kimi interop failure mode.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
gautam-achieveai: [Gautam Bhakar's bot] Fixed in 4999998. `loadMessagesFromBackend` now appends to `messageOrder` only on the first insert of a merge key and overwrites the existing `messageIndex` entry in place afterwards, so persisted records that collapse to one logical merge key (e.g. an intermediate update record beside the finalizing message) no longer accumulate duplicate keys or render/sort the same final message multiple times. Regression test added: loading two persisted records sharing run/generation/messageOrderIdx yields exactly one rendered bubble holding the last value.

src/OpenAiResponsesProvider/Agents/OpenAiResponsesAgent.cs
gautam-achieveai: [Gautam Bhakar's bot] Fixed at the source in cca02f6. The `reasoning_summary_text.delta` (ReasoningUpdateMessage) and `.done` (ReasoningMessage) events now emit `ReasoningVisibility.Summary` instead of Plain, and `TryExtractReasoning`'s summary[].text path also returns Summary (encrypted_content stays Encrypted). `AnthropicRequest` already maps Summary -> a text block, so replays no longer serialize these as unsigned thinking (400). `OpenAiResponsesReasoningSummaryTests` and `OpenAiResponsesAgentTests` now assert Visibility==Summary on the streamed updates, the final reasoning message, and the summary-array extraction path.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
gautam-achieveai: [Gautam Bhakar's bot] Fixed in the normalizer in 4999998 so all consumers benefit. `normalizeKeys` (in src/api/wsClient.ts) now aliases the snake_case identity fields used by the merge key — `generation_id`/`run_id`/`parent_run_id`/`message_order_idx` -> camelCase — write-once alongside the existing PascalCase aliasing (`tool_call_id` is consumed directly in snake_case by getMergeKey, so it needs no alias). Added a wsClient unit test driving snake_case `generation_id` and asserting it resolves to `generationId` (and that an explicit camelCase value is not clobbered, plus nested recursion).

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
gautam-achieveai: [Gautam Bhakar's bot] Regression test added in 4999998. The new case drives two `MessageType.ToolsCall` messages that share runId/generationId/messageOrderIdx but carry different `tool_calls[0].tool_call_id`, and asserts they render as two distinct pills under one pillbox (not collapsed). This exercises the aggregate single-call `ToolsCallMessage` branch of `getMergeKey` specifically, distinct from the existing individual-`ToolCall` coverage.

src/LmCore/Core/GenerateReplyOptions.cs
gautam-achieveai: [Gautam Bhakar's bot] Unit test added in 9f1d797 (tests/LmCore.Tests/Core/GenerateReplyOptionsMergeIdentityTests.cs). Three cases cover `GenerateReplyOptions.Merge` for the identity fields: other overrides RunId/ParentRunId/ThreadId/GenerationId when set; original is preserved when other leaves them null; and they are applied when the original had none.

src/LmCore/Messages/MessageExtensions.cs
gautam-achieveai: [Gautam Bhakar's bot] Parameterized coverage added in 9f1d797 (MessageExtensionsWithIdsTests). A `[Theory]` now drives `WithIds` across every message variant — including the update/result/usage arms (TextUpdateMessage, ToolCallUpdateMessage, ToolsCallUpdateMessage, ToolCallResultMessage, ToolsCallResultMessage, ReasoningUpdateMessage, UsageMessage) — asserting the run GenerationId is stamped and the concrete type is preserved. A future arm that forgets GenerationId (and falls through to the default `_ => message` case) now fails this test.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [Gautam Bhakar's bot] Keeping this in the PR intentionally. The reasoning/thinking enablement (the sonnet/haiku Copilot gate + gpt-5.5 reasoning request) was explicitly requested by the maintainer in the same working session as the direct fix for the user-visible "no thinking blocks" bug — it is the root-cause half of that same defect, not separate scope. It is covered by unit tests (OpenAiResponsesReasoningSummaryTests, the provider->reasoning extra-properties wiring guard) and live canaries (the gpt-5.5 reasoning-summary canary). Happy to split it into a follow-up PR/work item if you'd prefer a smaller rollout surface, but it is deliberately bundled here.

PR #119, opened 2026-06-29
feat: WebFetch and WebSearch tools (Jina-backed, Markdown output) (#113)

### LINKED ISSUE #113: Add WebFetch and WebSearch tools (Jina-backed, Markdown output)
## Summary

Add two LLM-callable `[Function]` tools so the streaming agent can read web pages and run web searches, returning **Markdown**. The tools are **provider-agnostic** behind a small abstraction; **Jina** (Reader `r.jina.ai` + Search `s.jina.ai`) is the first concrete backend, with room to add more later.

- **`WebFetch(url, …)`** — fetch a single URL and return its content as Markdown.
- **`WebSearch(query, …)`** — run a web search and return the top results as a Markdown list.

**Why these exist (purpose):** they are a **fallback** that gives web access to LLMs/providers that lack a native web capability — e.g. Anthropic or OpenAI accessed **via GitHub Copilot**, or the plain OpenAI APIs. When the active provider already ships native web tools (Anthropic built-in `web_search`/`web_fetch`), those are used instead and these tools stay off for that mode (see decision #1).

**Architecture — extensibility first.** The tool surface depends on backend **interfaces**, not Jina directly, so a new backend is a drop-in implementation + config switch:
- `src/Misc/Web/IWebFetchProvider.cs` — `Task<WebFetchResult> FetchAsync(string url, WebFetchOptions opts, CancellationToken ct)`.
- `src/Misc/Web/IWebSearchProvider.cs` — `Task<WebSearchResult> SearchAsync(string query, WebSearchOptions opts, CancellationToken ct)`.
- `src/Misc/Web/Jina/JinaWebProvider.cs` (+ `JinaHttpClient`) — the first implementation of both interfaces.
- `src/Misc/Utils/WebFetchTool.cs` + `WebSearchTool.cs` — `[Function]` tools that hold the interface(s) and do the Markdown formatting/truncation; **backend-agnostic**.
- `src/Misc/Configuration/WebToolsOptions.cs` — selects the active backend (default `jina`), `JINA_API_KEY` constants/helpers (via `EnvironmentVariableHelper`), and the output cap.

**Placement** mirrors the existing `TaskManager` utility-tool pattern. `JinaHttpClient` extends `BaseHttpService` (retry/backoff via `HttpRetryHelper`) and takes a **constructor-injectable `HttpClient`** so tests can swap the transport.

**Output:** Markdown-focused. The Jina backend requests `Accept: application/json` and reads `data.content` (robust parsing); the tool returns Markdown to the model. WebSearch formats results as a Markdown list (title, URL, snippet).

## Value

Gives the LmStreaming sample agent first-class web read/search **defined in this repo**, available to **any** provider — including those routed through GitHub Copilot or OpenAI that have no native web tools. The provider abstraction means future backends (other search/reader services) plug in without touching the tool surface. Markdown output is LLM-ready; a single `JINA_API_KEY` keeps the default config simple.

## API reference (verified against live endpoints)

**Reader** — `GET https://r.jina.ai/{url}` (also `POST https://r.jina.ai/` with `{"url":…}`)
- Auth: optional `Authorization: Bearer ${JINA_API_KEY}`. **Works without a key (~20 RPM)**; key raises the limit.
- Headers: `X-Return-Format: markdown` (default), `X-Target-Selector`, `X-No-Cache`, `X-Timeout`, `X-With-Generated-Alt`.
- `Accept: application/json` → `{ "code":200, "data": { "title", "url", "content" (markdown), "publishedTime", "warning", "usage": { "tokens" } } }`.

**Search** — `GET https://s.jina.ai/?q={query}` (also `POST` with `{"q":…}`)
- Auth: **API key REQUIRED** — returns HTTP **401 `AuthenticationRequiredError`** without it.
- Params: `q`, `num`/`count`, `gl` (country), `hl` (language); `Accept: application/json` for structured results.

## Design decisions (resolved)

1. **Tool names: keep `WebFetch` / `WebSearch`.** The Anthropic provider ships built-in `web_search`/`web_fetch` (`src/AnthropicProvider/Models/AnthropicBuiltInTools.cs`) and the Claude CLI path advertises `WebSearch`/`WebFetch` (`samples/LmStreaming.Sample/Program.cs`). The Jina-backed tools and the provider built-in web tools must be **mutually exclusive per chat mode** — enable these fallback tools only when the active provider has **no native** web capability (e.g. Copilot-routed Anthropic/OpenAI, plain OpenAI); when the provider has native web tools, use those instead. The model never sees two tools with the same name.
2. **Extensibility via a backend abstraction.** The tools depend on `IWebFetchProvider` / `IWebSearchProvider`; Jina is the first implementation. `WebToolsOptions` selects the active backend (default `jina`). Adding a backend = new implementation + config value, no change to the tool surface or tests for the tool layer.
3. **Single `JINA_API_KEY`** for the Jina backend. WebFetch uses it when present and still works without it; WebSearch requires it. (Other backends define their own option keys.)
4. **Scope: library + unit tests + sample wiring.** Tools register through the per-conversation **function-registry** path in `samples/LmStreaming.Sample/Program.cs`, and their names are added to the relevant mode `EnabledTools` so mode filtering actually applies (note: `ModeToolFilter.FilterBuiltInTools` only filters `AnthropicBuiltInTool`s — the function-tool path is the per-conversation `filteredRegistry`, not `ModeToolFilter`).
5. **Cancellation: custom handler within this feature.** `TypeFunctionProvider` drops the `CancellationToken` (`TypeFunctionProvider.cs:213`), so register the web tools via explicit `AddFunction(contract, handler, …)` whose handler threads the token into the provider's `…Async(…, ct)` → `HttpClient.SendAsync(…, ct)`. No shared-framework change.
6. **Output truncation: configurable cap with a sane default** (`WebToolsOptions`, e.g. ~50k chars) with a `…[truncated]` marker; surface the backend's token usage when available (Jina's `usage.tokens`).

### WebFetch
- [ ] `WebFetch(url)` returns the page as Markdown parsed from Jina's `data.content` (JSON envelope), for a successful fetch.
- [ ] Request targets `https://r.jina.ai/{url}` with `Accept: application/json` and `X-Return-Format: markdown`; `Authorization: Bearer` is included **iff** `JINA_API_KEY` is set and omitted otherwise.
- [ ] Optional params are honored: `targetSelector` → `X-Target-Selector`, `noCache` → `X-No-Cache`.
- [ ] The target URL is validated: only `http`/`https` accepted; malformed/empty/non-http input returns a clear error **without** issuing an HTTP call; URLs with query strings/fragments are encoded correctly and fetch successfully.
- [ ] Non-2xx upstream (e.g. 404), network error, and timeout each return a graceful, bounded error string (no exception propagates to the model).

### WebSearch
- [ ] `WebSearch(query)` returns the top results as a Markdown list (title, URL, snippet) parsed from Jina's JSON.
- [ ] Request sends URL-encoded `q` plus optional `count`/`country`(`gl`)/`language`(`hl`) and the `Authorization: Bearer` header.
- [ ] **Missing `JINA_API_KEY`** → a clear, actionable message ("WebSearch unavailable: set JINA_API_KEY") returned up front **without** an HTTP call (and the tool is not advertised when no key is present); an **invalid** key (401) maps to a clear error rather than a raw status dump.
- [ ] Empty results → a friendly "no results" message.

### Extensibility / architecture
- [ ] `WebFetchTool`/`WebSearchTool` depend only on `IWebFetchProvider`/`IWebSearchProvider` (and the options/formatting helpers) — **no direct Jina references** in the tool classes. Verified by a test that drives the tools with a fake/in-memory provider implementation.
- [ ] `WebToolsOptions` selects the active backend (default `jina`); an unknown/unconfigured backend yields a clear error.
- [ ] Markdown formatting and output truncation live in the **backend-agnostic** tool layer, not in the Jina provider, so they apply to any future backend.

### Cross-cutting
- [ ] **Provider gating:** the fallback tools are enabled only for modes/providers without native web tools; when the active provider has native web tools, the Jina-backed tools are not advertised (mutually exclusive — see decision #1).
- [ ] **Output cap:** results longer than the configured `WebToolsOptions` max are truncated with a `…[truncated]` marker; default applies when unset.
- [ ] **Cancellation:** cancelling the run/clicking Stop aborts an in-flight Jina request (token threaded to `HttpClient.SendAsync`).
- [ ] **Retry/limits:** 429 and 5xx are retried with backoff (existing `HttpRetryHelper`); a defined per-call timeout budget is applied. (Honoring `Retry-After` is optional — call it out if implemented.)
- [ ] **Secret/error hygiene:** the `JINA_API_KEY` value never appears in tool output or logs; upstream error bodies are bounded before being surfaced. Asserted by a test.
- [ ] **Mode filtering:** a mode with the web tools disabled does not expose them; Jina tools and provider built-in web tools are never both active in the same mode.

### Tests (MessageHandler mocks — no real network)
- [ ] Tests inject `HttpClient` backed by `src/LmTestUtils/FakeHttpMessageHandler` (request-capture to assert URL/query/headers; status-code-sequence/retry helpers for 429→200; error handler for timeout/network/cancellation).
- [ ] Coverage includes: WebFetch happy path + headers, key present/absent, optional params, 404/network/timeout, URL validation; WebSearch happy path + params, missing key, 401, empty results; output truncation; secret-not-leaked assertion.
- [ ] A `FunctionRegistry` contract test verifies both tools register with the expected name/description/parameter schema (mirrors `tests/LmCore.Tests/Middleware/FunctionRegistryTests.cs`).
- [ ] Tests live under `tests/Misc.Tests/Http/` and `tests/Misc.Tests/Utils/`.

### Build/quality
- [ ] Zero-warning build (`TreatWarningsAsErrors`); CSharpier-formatted.

## Out of scope / follow-ups
- **Additional concrete backends** beyond Jina — the abstraction is built now (in scope), but only the Jina implementation ships; other backends are follow-up issues.
- Forwarding the `CancellationToken` through `TypeFunctionProvider` for **all** reflective tools (broader shared change) — separate issue.
- Screenshot/HTML return formats, image alt captioning (`X-With-Generated-Alt`), and `Retry-After`-aware backoff unless trivially added.

gautam-achieveai:
## Summary

Adds two LLM-callable tools — **`WebFetch(url, …)`** and **`WebSearch(query, …)`** — that return **Markdown**, as a fallback giving web access to providers without a native web capability (Copilot-routed Sonnet/Haiku/GPT-5.5, plain OpenAI). They sit behind provider-agnostic interfaces (`IWebFetchProvider` / `IWebSearchProvider`); **Jina** (Reader `r.jina.ai` + Search `s.jina.ai`) is the first concrete backend. Wired into the LmStreaming.Sample agent via a provider/mode-gated registration policy.

Fixes #113

## What's included

**Library (`src/Misc`):**
- `Web/IWebFetchProvider`, `IWebSearchProvider` + result/option records — the backend abstraction (tools depend only on these; verified by a fake-provider test).
- `Web/Jina/JinaWebProvider` — single `BaseHttpService` subclass implementing both interfaces. **POST + JSON body** to Jina (sidesteps URL query/fragment encoding), per-request headers (`Accept: application/json`, `X-Return-Format: markdown`, optional `X-Target-Selector`/`X-No-Cache`, `Authorization: Bearer` iff `JINA_API_KEY` set), tolerant parsing (JSON envelope or plain markdown; missing/partial fields tolerated), retry via the existing `HttpRetryHelper` (429/5xx), and a per-call timeout via a linked `CancellationTokenSource`.
- `Web/WebInputValidator` — URL safety (http/https only; rejects userinfo, localhost/loopback/private/link-local/multicast/unspecified IPs and `.local`/`.internal`/`.localhost` hosts; strips fragments) and query validation; **no HTTP on invalid input**.
- `Utils/WebFetchTool`, `WebSearchTool` — backend-agnostic adapters: URL/query validation, Markdown formatting, untrusted-content framing (prompt-injection mitigation), secret sanitization, output truncation, and **bounded/sanitized error mapping** (never surfaces the raw upstream body or the API key). Registered via explicit `AddFunction(contract, handler)` so the handler threads the `CancellationToken`.
- `Utils/WebToolOutput`, `Configuration/WebToolsOptions` — formatting/truncation helpers and options (env-driven; validated backend selector, output cap, per-call timeout).

**Sample (`samples/LmStreaming.Sample`):**
- `Services/WebToolRegistrationPolicy` — central, testable gating: registers the Jina tools only for no-native-web providers (`openai`, `sonnet`, `haiku`, `gpt-5.5`, `gpt-5.5-mini`), gated by the mode's `EnabledTools`, with case-insensitive collision detection and a `JINA_API_KEY` registration gate for `WebSearch`.
- `Program.cs` wiring (one process-lifetime provider; policy invoked per-conversation after MCP tools), `Prompts.yaml` (research-assistant mode), `ManualTesting.md` (+ privacy note).

## Key decisions

1. **Tool names `WebFetch`/`WebSearch` (PascalCase)** and **fallback only for no-native-web providers** — confirmed with the maintainer. Native web (Anthropic built-ins, Claude CLI, Codex `WebSearchMode`) and the Jina fallback are **never both active** for one conversation, enforced by provider gating. Plain `copilot` is intentionally excluded (its early-return `CopilotAgentLoop` path bypasses this registration seam — deferred follow-up).
2. **Two-list mode model preserved:** Jina function tools are gated by `EnabledTools`; native provider built-ins remain gated by `EnabledBuiltInTools ?? EnabledTools` (untouched).
3. **Single `JinaWebProvider : BaseHttpService`** (collapsed from a proposed transport/mapping split) to match the existing `AnthropicClient`/`OpenClient` convention; backend selector kept minimal (string + switch, no factory).
4. **POST-JSON to Jina** instead of path-appended GET, for correct handling of URLs with query strings/fragments.
5. **Secret/content hygiene end-to-end:** the tool layer maps exceptions to bounded, sanitized strings (the shared `HttpRetryHelper` appends the raw upstream body to its exception message); the provider suppresses inherited HTTP/retry body-logging (passes `NullLogger` to its base). Asserted by tests that drive a real provider and assert the key/body never reach tool output **or** logs.

## Testing

- `dotnet build LmDotnetTools.sln` — **0 errors, 0 new warnings** (the 6 warnings are pre-existing `NU1902` Scriban advisories in LmCore).
- `dotnet test tests/Misc.Tests` — **177/177 passed**.
- `dotnet test tests/LmStreaming.Sample.Tests` — **352/352 passed**.
- CSharpier/format gate clean. All tests use `FakeHttpMessageHandler` (no real network). Each acceptance criterion maps to a RED→GREEN test (request shape/headers, auth-iff-key, optional params, URL validation incl. SSRF/private-host, 404/network/timeout bounded errors, retry, cancellation threading, WebSearch no-key/401/empty, output cap, secret hygiene incl. logs, contract registration, provider/mode gating incl. `copilot`-excluded and two-conversation isolation).

## Follow-ups (out of scope here)

- Live-verify the Jina **POST search** parameter field names (`num`/`gl`/`hl`), the per-result snippet key, and `usage.tokens` nesting against the real endpoint (parsing is tolerant; an env-gated smoke test can pin it).
- Wire the fallback tools for plain **`copilot`** (separate `CopilotAgentLoop` dynamic-tool path).
- Additional backends behind the same interfaces.

### DIFF
--- a/src/Misc/Utils/WebFetchTool.cs
+++ b/src/Misc/Utils/WebFetchTool.cs
@@ -1,0 +1,176 @@
+using System.Net;
+using System.Text.Json;
+using AchieveAi.LmDotnetTools.LmCore.Core;
+using AchieveAi.LmDotnetTools.LmCore.Messages;
+using AchieveAi.LmDotnetTools.LmCore.Middleware;
+using AchieveAi.LmDotnetTools.LmCore.Models;
+using AchieveAi.LmDotnetTools.Misc.Configuration;
+using AchieveAi.LmDotnetTools.Misc.Web;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace AchieveAi.LmDotnetTools.Misc.Utils;
+
+/// <summary>
+///     Backend-agnostic <c>WebFetch</c> tool. Validates the URL, delegates to an
+///     <see cref="IWebFetchProvider" />, then frames, sanitizes, and bounds the Markdown it returns
+///     to the model. Provider exceptions are mapped to bounded, sanitized error strings so that raw
+///     upstream bodies, exception messages, or the API key are never surfaced.
+/// </summary>
+public sealed class WebFetchTool
+{
+    /// <summary>
+    ///     The tool name advertised to the model and used for registration.
+    /// </summary>
+    public const string ToolName = "WebFetch";
+
+    private readonly IWebFetchProvider _provider;
+    private readonly WebToolsOptions _options;
+    private readonly ILogger _logger;
+
+    /// <summary>
+    ///     Initializes a new <see cref="WebFetchTool" />.
+    /// </summary>
+    /// <param name="provider">The backend used to fetch pages.</param>
+    /// <param name="options">Web tools configuration (API key, output cap).</param>
+    /// <param name="logger">Optional logger; defaults to <see cref="NullLogger{T}.Instance" />.</param>
+    public WebFetchTool(IWebFetchProvider provider, WebToolsOptions options, ILogger? logger = null)
+    {
+        _provider = provider ?? throw new ArgumentNullException(nameof(provider));
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _logger = logger ?? NullLogger<WebFetchTool>.Instance;
+        Contract = BuildContract();
+    }
+
+    /// <summary>
+    ///     The function contract describing this tool to the model.
+    /// </summary>
+    public FunctionContract Contract { get; }
+
+    /// <summary>
+    ///     The tool handler delegate that threads the cancellation token through to the provider.
+    /// </summary>
+    public ToolHandler Handler => HandleAsync;
+
+    private static FunctionContract BuildContract()
+    {
+        return new FunctionContract
+        {
+            Name = ToolName,
+            Description = "Fetch a single web page and return its content as Markdown.",
+            Parameters =
+            [
+                new FunctionParameterContract
+                {
+                    Name = "url",
+                    Description = "The absolute http/https URL of the page to fetch.",
+                    ParameterType = JsonSchemaObject.String("The absolute http/https URL of the page to fetch."),
+                    IsRequired = true,
+                },
+                new FunctionParameterContract
+                {
+                    Name = "targetSelector",
+                    Description = "Optional CSS selector identifying the region of the page to extract.",
+                    ParameterType = JsonSchemaObject.String(
+                        "Optional CSS selector identifying the region of the page to extract."
+                    ),
+                    IsRequired = false,
+                },
+                new FunctionParameterContract
+                {
+                    Name = "noCache",
+                    Description = "When true, bypasses any cached copy and fetches fresh content.",
+                    ParameterType = JsonSchemaObject.Boolean(
+                        "When true, bypasses any cached copy and fetches fresh content."
+                    ),
+                    IsRequired = false,
+                },
+            ],
+            ReturnType = typeof(string),
+        };
+    }
+
+    private async Task<ToolHandlerResult> HandleAsync(
+        string argsJson,
+        ToolCallContext context,
+        CancellationToken cancellationToken
+    )
+    {
+        string? url;
+        string? targetSelector;
+        bool? noCache;
+        try
+        {
+            using var document = JsonDocument.Parse(string.IsNullOrWhiteSpace(argsJson) ? "{}" : argsJson);
+            var root = document.RootElement;
+            url = WebToolArgs.ReadString(root, "url");
+            targetSelector = WebToolArgs.ReadString(root, "targetSelector");
+            noCache = WebToolArgs.ReadBool(root, "noCache");
+        }
+        catch (JsonException)
+        {
+            return Error("WebFetch error: arguments were not valid JSON.");
+        }
+
+        var validation = WebInputValidator.ValidateUrl(url);
+        if (!validation.IsValid)
+        {
+            return Error("WebFetch error: " + validation.Error);
+        }
+
+        try
+        {
+            var result = await _provider.FetchAsync(
+                validation.Value!,
+                new WebFetchOptions { TargetSelector = targetSelector, NoCache = noCache },
+                cancellationToken
+            );
+
+            var markdown = WebToolOutput.FormatFetch(result);
+            var framed = WebToolOutput.WrapUntrusted(markdown, url!);
+            var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
+            var bounded = WebToolOutput.Truncate(sanitized, _options.OutputCap);
+            return ToolHandlerResult.FromText(bounded);
+        }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // Caller-initiated cancellation propagates so the agent loop can unwind.
+            throw;
+        }
+        catch (OperationCanceledException)
+        {
+            // The linked per-call timeout fired without the caller cancelling.
+            return Error("WebFetch error: request timed out.");
+        }
+        catch (HttpRequestException ex)
+        {
+            return Error(MapHttpError(ex));
+        }
+        catch (Exception)
+        {
+            return Error("WebFetch error: could not fetch the page.");
+        }
+    }
+
+    /// <summary>
+    ///     Maps an HTTP failure to a bounded message. The status code is the only upstream detail
+    ///     surfaced; the exception message (which may carry the raw response body) is never used.
+    /// </summary>
+    private static string MapHttpError(HttpRequestException ex)
+    {
+        return ex.StatusCode == HttpStatusCode.NotFound
+            ? "WebFetch error: page not found (404)."
+            : $"WebFetch error: upstream request failed ({(int?)ex.StatusCode}).";
+    }
+
+    /// <summary>
+    ///     Builds an error result, redacting the API key as a defense-in-depth measure and logging a
+    ///     sanitized reason (never the raw exception, body, or key).
+    /// </summary>
+    private ToolHandlerResult Error(string message)
+    {
+        var sanitized = WebToolOutput.Sanitize(message, _options.JinaApiKey);
+        _logger.LogWarning("WebFetch returned an error to the model: {Reason}", sanitized);
+        return ToolHandlerResult.FromError(sanitized);
+    }
+}
--- a/src/Misc/Utils/WebFetchTool.cs
+++ b/src/Misc/Utils/WebFetchTool.cs
@@ -120,9 +120,12 @@ public sealed class WebFetchTool
 
         try
         {
+            // targetSelector is LLM-controlled and is later added to the X-Target-Selector header via
+            // TryAddWithoutValidation, so validate it at this trust boundary: a control character (CR/LF)
+            // or an overlong value could enable header injection. Omit it rather than failing the call.
             var result = await _provider.FetchAsync(
                 validation.Value!,
-                new WebFetchOptions { TargetSelector = targetSelector, NoCache = noCache },
+                new WebFetchOptions { TargetSelector = NormalizeTargetSelector(targetSelector), NoCache = noCache },
                 cancellationToken
             );
 
@@ -153,6 +156,21 @@ public sealed class WebFetchTool
             return Error("WebFetch error: could not fetch the page.");
         }
     }
+
+    /// <summary>
+    ///     The maximum length accepted for <c>targetSelector</c>; longer values are omitted.
+    /// </summary>
+    private const int MaxTargetSelectorLength = 256;
+
+    /// <summary>
+    ///     Accepts a <c>targetSelector</c> only when it carries no control characters (including CR/LF,
+    ///     which could enable header injection) and stays within <see cref="MaxTargetSelectorLength" />;
+    ///     otherwise omits it (<c>null</c>) so no <c>X-Target-Selector</c> header is sent.
+    /// </summary>
+    private static string? NormalizeTargetSelector(string? selector) =>
+        selector is not null && selector.Length <= MaxTargetSelectorLength && !selector.Any(char.IsControl)
+            ? selector
+            : null;
 
     /// <summary>
     ///     Maps an HTTP failure to a bounded message. The status code is the only upstream detail
--- a/src/Misc/Utils/WebFetchTool.cs
+++ b/src/Misc/Utils/WebFetchTool.cs
@@ -127,8 +127,9 @@ public sealed class WebFetchTool
             );
 
             var markdown = WebToolOutput.FormatFetch(result);
-            // Use the validated, fragment-stripped URL actually fetched as the source label (not the raw arg).
-            var framed = WebToolOutput.WrapUntrusted(markdown, validation.Value!);
+            // Minimize the URL for display: the full URL is still fetched, but the source label drops the
+            // query, fragment, and userinfo so secrets/PII in those parts are never echoed to the model.
+            var framed = WebToolOutput.WrapUntrusted(markdown, WebToolOutput.MinimizeUrl(validation.Value));
             var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
             var bounded = WebToolOutput.Truncate(sanitized, _options.OutputCap);
             return ToolHandlerResult.FromText(bounded);
--- a/src/Misc/Utils/WebFetchTool.cs
+++ b/src/Misc/Utils/WebFetchTool.cs
@@ -127,7 +127,8 @@ public sealed class WebFetchTool
             );
 
             var markdown = WebToolOutput.FormatFetch(result);
-            var framed = WebToolOutput.WrapUntrusted(markdown, url!);
+            // Use the validated, fragment-stripped URL actually fetched as the source label (not the raw arg).
+            var framed = WebToolOutput.WrapUntrusted(markdown, validation.Value!);
             var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
             var bounded = WebToolOutput.Truncate(sanitized, _options.OutputCap);
             return ToolHandlerResult.FromText(bounded);
--- a/src/Misc/Utils/WebSearchTool.cs
+++ b/src/Misc/Utils/WebSearchTool.cs
@@ -136,13 +136,16 @@ public sealed class WebSearchTool
 
         try
         {
+            // The optional parameters are LLM-controlled, so validate them at this trust boundary before
+            // they flow into the provider request: clamp the count and drop malformed locale codes
+            // (rather than failing the whole call).
             var result = await _provider.SearchAsync(
                 validation.Value!,
                 new WebSearchOptions
                 {
-                    Count = count,
-                    Country = country,
-                    Language = language,
+                    Count = ClampCount(count),
+                    Country = NormalizeCountry(country),
+                    Language = NormalizeLanguage(language),
                 },
                 cancellationToken
             );
@@ -181,6 +184,49 @@ public sealed class WebSearchTool
         {
             return Error("WebSearch error: could not run the search.");
         }
+    }
+
+    /// <summary>
+    ///     Clamps a requested result count into the supported <c>[1, 20]</c> range, leaving an absent
+    ///     count (<c>null</c>) untouched so the backend default applies.
+    /// </summary>
+    private static int? ClampCount(int? count) => count is int value ? Math.Clamp(value, 1, 20) : null;
+
+    /// <summary>
+    ///     Accepts a country (Jina <c>gl</c>) only when it is a two-letter alpha code; otherwise omits it.
+    /// </summary>
+    private static string? NormalizeCountry(string? country) =>
+        country is { Length: 2 } && char.IsAsciiLetter(country[0]) && char.IsAsciiLetter(country[1])
+            ? country
+            : null;
+
+    /// <summary>
+    ///     Accepts a language (Jina <c>hl</c>) only when it matches <c>^[A-Za-z]{2,5}(-[A-Za-z0-9]{1,8})?$</c>
+    ///     (a short primary subtag with an optional single secondary subtag); otherwise omits it.
+    /// </summary>
+    private static string? NormalizeLanguage(string? language) => IsValidLanguage(language) ? language : null;
+
+    private static bool IsValidLanguage(string? language)
+    {
+        if (string.IsNullOrEmpty(language))
+        {
+            return false;
+        }
+
+        var dash = language.IndexOf('-');
+        var primary = dash < 0 ? language : language[..dash];
+        if (primary.Length is < 2 or > 5 || !primary.All(char.IsAsciiLetter))
+        {
+            return false;
+        }
+
+        if (dash < 0)
+        {
+            return true;
+        }
+
+        var secondary = language[(dash + 1)..];
+        return secondary.Length is >= 1 and <= 8 && secondary.All(char.IsAsciiLetterOrDigit);
     }
 
     /// <summary>
--- a/src/Misc/Utils/WebSearchTool.cs
+++ b/src/Misc/Utils/WebSearchTool.cs
@@ -1,0 +1,206 @@
+using System.Net;
+using System.Text.Json;
+using AchieveAi.LmDotnetTools.LmCore.Core;
+using AchieveAi.LmDotnetTools.LmCore.Messages;
+using AchieveAi.LmDotnetTools.LmCore.Middleware;
+using AchieveAi.LmDotnetTools.LmCore.Models;
+using AchieveAi.LmDotnetTools.Misc.Configuration;
+using AchieveAi.LmDotnetTools.Misc.Web;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace AchieveAi.LmDotnetTools.Misc.Utils;
+
+/// <summary>
+///     Backend-agnostic <c>WebSearch</c> tool. Gates on the presence of an API key, validates the
+///     query, delegates to an <see cref="IWebSearchProvider" />, then frames, sanitizes, and bounds
+///     the Markdown it returns to the model. Provider exceptions are mapped to bounded, sanitized
+///     error strings so that raw upstream bodies, exception messages, or the API key are never
+///     surfaced.
+/// </summary>
+public sealed class WebSearchTool
+{
+    /// <summary>
+    ///     The tool name advertised to the model and used for registration.
+    /// </summary>
+    public const string ToolName = "WebSearch";
+
+    private readonly IWebSearchProvider _provider;
+    private readonly WebToolsOptions _options;
+    private readonly ILogger _logger;
+
+    /// <summary>
+    ///     Initializes a new <see cref="WebSearchTool" />.
+    /// </summary>
+    /// <param name="provider">The backend used to run searches.</param>
+    /// <param name="options">Web tools configuration (API key, output cap, query length).</param>
+    /// <param name="logger">Optional logger; defaults to <see cref="NullLogger{T}.Instance" />.</param>
+    public WebSearchTool(IWebSearchProvider provider, WebToolsOptions options, ILogger? logger = null)
+    {
+        _provider = provider ?? throw new ArgumentNullException(nameof(provider));
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _logger = logger ?? NullLogger<WebSearchTool>.Instance;
+        Contract = BuildContract();
+    }
+
+    /// <summary>
+    ///     The function contract describing this tool to the model.
+    /// </summary>
+    public FunctionContract Contract { get; }
+
+    /// <summary>
+    ///     The tool handler delegate that threads the cancellation token through to the provider.
+    /// </summary>
+    public ToolHandler Handler => HandleAsync;
+
+    private static FunctionContract BuildContract()
+    {
+        return new FunctionContract
+        {
+            Name = ToolName,
+            Description = "Run a web search and return the ranked results as Markdown.",
+            Parameters =
+            [
+                new FunctionParameterContract
+                {
+                    Name = "query",
+                    Description = "The search query.",
+                    ParameterType = JsonSchemaObject.String("The search query."),
+                    IsRequired = true,
+                },
+                new FunctionParameterContract
+                {
+                    Name = "count",
+                    Description = "Optional maximum number of results to return.",
+                    ParameterType = JsonSchemaObject.Integer("Optional maximum number of results to return."),
+                    IsRequired = false,
+                },
+                new FunctionParameterContract
+                {
+                    Name = "country",
+                    Description = "Optional ISO country code used to localize results (for example, \"US\").",
+                    ParameterType = JsonSchemaObject.String(
+                        "Optional ISO country code used to localize results (for example, \"US\")."
+                    ),
+                    IsRequired = false,
+                },
+                new FunctionParameterContract
+                {
+                    Name = "language",
+                    Description = "Optional language code used to localize results (for example, \"en\").",
+                    ParameterType = JsonSchemaObject.String(
+                        "Optional language code used to localize results (for example, \"en\")."
+                    ),
+                    IsRequired = false,
+                },
+            ],
+            ReturnType = typeof(string),
+        };
+    }
+
+    private async Task<ToolHandlerResult> HandleAsync(
+        string argsJson,
+        ToolCallContext context,
+        CancellationToken cancellationToken
+    )
+    {
+        // Up-front key gate: WebSearch requires an API key, so fail fast without touching the provider.
+        if (string.IsNullOrWhiteSpace(_options.JinaApiKey))
+        {
+            return ToolHandlerResult.FromError("WebSearch unavailable: set JINA_API_KEY.");
+        }
+
+        string? query;
+        int? count;
+        string? country;
+        string? language;
+        try
+        {
+            using var document = JsonDocument.Parse(string.IsNullOrWhiteSpace(argsJson) ? "{}" : argsJson);
+            var root = document.RootElement;
+            query = WebToolArgs.ReadString(root, "query");
+            count = WebToolArgs.ReadInt(root, "count");
+            country = WebToolArgs.ReadString(root, "country");
+            language = WebToolArgs.ReadString(root, "language");
+        }
+        catch (JsonException)
+        {
+            return Error("WebSearch error: arguments were not valid JSON.");
+        }
+
+        var validation = WebInputValidator.ValidateQuery(query, _options.MaxQueryLength);
+        if (!validation.IsValid)
+        {
+            return Error("WebSearch error: " + validation.Error);
+        }
+
+        try
+        {
+            var result = await _provider.SearchAsync(
+                validation.Value!,
+                new WebSearchOptions
+                {
+                    Count = count,
+                    Country = country,
+                    Language = language,
+                },
+                cancellationToken
+            );
+
+            if (result.Items.Count == 0)
+            {
+                return ToolHandlerResult.FromText("No results found.");
+            }
+
+            var markdown = WebToolOutput.FormatSearch(result);
+            var framed = WebToolOutput.WrapUntrusted(markdown, "web search: " + validation.Value);
+            var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
+            var bounded = WebToolOutput.Truncate(sanitized, _options.OutputCap);
+            return ToolHandlerResult.FromText(bounded);
+        }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // Caller-initiated cancellation propagates so the agent loop can unwind.
+            throw;
+        }
+        catch (OperationCanceledException)
+        {
+            // The linked per-call timeout fired without the caller cancelling.
+            return Error("WebSearch error: request timed out.");
+        }
+        catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.Unauthorized)
+        {
+            return Error("WebSearch error: the JINA_API_KEY is invalid (401).");
+        }
+        catch (HttpRequestException ex)
+        {
+            return Error(MapHttpError(ex));
+        }
+        catch (Exception)
+        {
+            return Error("WebSearch error: could not run the search.");
+        }
+    }
+
+    /// <summary>
+    ///     Maps an HTTP failure to a bounded message. The status code is the only upstream detail
+    ///     surfaced; the exception message (which may carry the raw response body) is never used.
+    /// </summary>
+    private static string MapHttpError(HttpRequestException ex)
+    {
+        return ex.StatusCode == HttpStatusCode.NotFound
+            ? "WebSearch error: page not found (404)."
+            : $"WebSearch error: upstream request failed ({(int?)ex.StatusCode}).";
+    }
+
+    /// <summary>
+    ///     Builds an error result, redacting the API key as a defense-in-depth measure and logging a
+    ///     sanitized reason (never the raw exception, body, or key).
+    /// </summary>
+    private ToolHandlerResult Error(string message)
+    {
+        var sanitized = WebToolOutput.Sanitize(message, _options.JinaApiKey);
+        _logger.LogWarning("WebSearch returned an error to the model: {Reason}", sanitized);
+        return ToolHandlerResult.FromError(sanitized);
+    }
+}
--- a/src/Misc/Utils/WebSearchTool.cs
+++ b/src/Misc/Utils/WebSearchTool.cs
@@ -153,7 +153,8 @@ public sealed class WebSearchTool
             }
 
             var markdown = WebToolOutput.FormatSearch(result);
-            var framed = WebToolOutput.WrapUntrusted(markdown, "web search: " + validation.Value);
+            // Use a generic source label: the query may carry PII/secrets and must not be echoed back.
+            var framed = WebToolOutput.WrapUntrusted(markdown, "web search");
             var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
             var bounded = WebToolOutput.Truncate(sanitized, _options.OutputCap);
             return ToolHandlerResult.FromText(bounded);
--- a/src/Misc/Utils/WebToolOutput.cs
+++ b/src/Misc/Utils/WebToolOutput.cs
@@ -111,7 +111,9 @@ public static class WebToolOutput
                 .Append(". [")
                 .Append(item.Title)
                 .Append("](")
-                .Append(item.Url)
+                // Minimize the result URL so query/fragment/userinfo (which may carry PII/secrets) are
+                // never echoed back to the model; only scheme://host[:port]/path survives.
+                .Append(MinimizeUrl(item.Url))
                 .Append(')');
 
             if (!string.IsNullOrWhiteSpace(item.Snippet))
--- a/src/Misc/Utils/WebToolOutput.cs
+++ b/src/Misc/Utils/WebToolOutput.cs
@@ -1,0 +1,158 @@
+using System.Text;
+using AchieveAi.LmDotnetTools.Misc.Web;
+
+namespace AchieveAi.LmDotnetTools.Misc.Utils;
+
+/// <summary>
+///     Formatting helpers that turn web tool results into Markdown, frame untrusted content,
+///     redact secrets, and bound output length. All methods are pure and perform no I/O.
+/// </summary>
+public static class WebToolOutput
+{
+    /// <summary>
+    ///     The deterministic marker appended when <see cref="Truncate" /> shortens text.
+    /// </summary>
+    public const string TruncationMarker = "\n\n…[truncated]";
+
+    private const string Redaction = "***";
+
+    /// <summary>
+    ///     Renders a <see cref="WebFetchResult" /> as Markdown: an optional title heading, a source line,
+    ///     the content, and an optional warning.
+    /// </summary>
+    /// <param name="result">The fetch result to render.</param>
+    /// <returns>The Markdown representation.</returns>
+    public static string FormatFetch(WebFetchResult result)
+    {
+        var builder = new StringBuilder();
+
+        if (!string.IsNullOrWhiteSpace(result.Title))
+        {
+            _ = builder.Append("# ").Append(result.Title.Trim()).Append("\n\n");
+        }
+
+        if (!string.IsNullOrWhiteSpace(result.Url))
+        {
+            _ = builder.Append("Source: ").Append(result.Url.Trim()).Append("\n\n");
+        }
+
+        _ = builder.Append(result.Content);
+
+        if (!string.IsNullOrWhiteSpace(result.Warning))
+        {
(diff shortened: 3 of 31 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR !119 (linked to #113)

## Summary

Well-architected implementation of WebFetch/WebSearch tools backed by Jina Reader/Search APIs. Clean layer separation, strong security boundaries (SSRF protection, secret sanitization, untrusted content framing), comprehensive test coverage (92 tests passing), and thoughtful error handling that never leaks upstream details to the model.

## Findings

### :yellow_circle: MEDIUM — Consider adding RFC 6761 special-use domains to SSRF blocklist
**File:** `src/Misc/Web/WebInputValidator.cs` (line 36)

Current blocklist: `.local`, `.internal`, `.localhost`

Missing RFC 6761 reserved domains: `.test`, `.example`, `.invalid`. Low-risk since requests go to Jina (not your infra), but cheap defense-in-depth:
```csharp
private static readonly string[] BlockedHostSuffixes = [
    ".local", ".internal", ".localhost",
    ".test", ".example", ".invalid"  // RFC 6761
];
```

### :yellow_circle: MEDIUM — ReadAsStringAsync buffers full response before JSON parsing
**File:** `src/Misc/Web/Jina/JinaWebProvider.cs` (lines 103, 131)

For large pages (>5MB), `ReadAsStringAsync` + `JsonDocument.Parse(string)` allocates content twice. Consider `ReadAsStreamAsync` + `JsonDocument.ParseAsync(stream)` for better memory. Not blocking since `OutputCap` truncates downstream, but worth noting for memory-sensitive deployments.

### :green_circle: LOW — WebToolsOptions mixes tool and provider concerns
**File:** `src/Misc/Configuration/WebToolsOptions.cs`

`OutputCap`/`MaxQueryLength` (tool) and `JinaApiKey`/`TimeoutMs` (provider) share one record. Acceptable with a single backend, but consider splitting if a second provider is added.

### :green_circle: LOW — Shared timeout budget across retries is intentional but undocumented
**File:** `src/Misc/Web/Jina/JinaWebProvider.cs` (line 95)

`CancelAfter(_options.TimeoutMs)` bounds ALL retries collectively — a slow first attempt leaves less time for retries. Design is correct (caps total user-facing latency), but a brief code comment would help future maintainers.

## Positive Observations

:white_check_mark: **Secret hygiene is exemplary** — NullLogger suppresses inherited retry logging that could embed raw upstream bodies. Separate logger used only for sanitized host-only diagnostics. `Sanitize()` applied on both success and error paths.

:white_check_mark: **Untrusted content framing** — `WrapUntrusted()` adds clear delimiters and defensive instructions, consistently applied in both tools.

:white_check_mark: **Tolerant JSON arg parsing** — `WebToolArgs` degrades gracefully (returns null) on malformed model arguments.

:white_check_mark: **Test quality** — Secret hygiene tests drive real JinaWebProvider -> WebFetchTool chains against mock HTTP handlers that echo secrets, proving end-to-end redaction. SSRF tests cover IPv4/IPv6, boundary conditions.

:white_check_mark: **Registration policy** — Well-bounded with collision detection, provider allow-list, and API-key gate. Thorough tests cover the full provider matrix.

**Verdict: APPROVE** — Solid implementation with only minor, non-blocking suggestions above.

src/Misc/Web/WebInputValidator.cs
```diff
@@ -0,0 +1,196 @@
...
+/// <summary>
+///     Validates URLs and search queries for the web tools without performing any network I/O.
+///     URL validation blocks SSRF-prone targets (localhost, loopback, private, link-local, multicast,
+///     and internal host suffixes) and strips the fragment from the returned value.
+/// </summary>
+public static class WebInputValidator
+{
+    private const int MaxUrlLength = 2048;
+
+    private static readonly string[] BlockedHostSuffixes = [".local", ".internal", ".localhost"];
```
gautam-achieveai: **[Gautam's review bot]** :yellow_circle: Consider extending with RFC 6761 special-use domains (\.test\, \.example\, \.invalid\) as defense-in-depth. Low-risk since Jina does the actual fetch, but these domains should never resolve to real infrastructure.

src/Misc/Web/Jina/JinaWebProvider.cs
```diff
@@ -0,0 +1,364 @@
...
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        cts.CancelAfter(_options.TimeoutMs);
+
+        // The request is built inside the operation lambda so each retry attempt gets a fresh
+        // HttpRequestMessage (a sent request cannot be reused).
+        return await ExecuteHttpWithRetryAsync(
+            () => HttpClient.SendAsync(BuildFetchRequest(url, options), cts.Token),
+            async response =>
+            {
+                var body = await response.Content.ReadAsStringAsync(cts.Token);
```
gautam-achieveai, line 104: **[Gautam's review bot]** :yellow_circle: \ReadAsStringAsync\ buffers the entire response before JSON parsing. For large pages this allocates content twice. Consider \ReadAsStreamAsync\ + \JsonDocument.ParseAsync(stream)\ to avoid the intermediate string buffer. Not blocking since \OutputCap\ truncates downstream.

src/Misc/Web/Jina/JinaWebProvider.cs
```diff
@@ -0,0 +1,364 @@
...
+    )
+    {
+        ArgumentNullException.ThrowIfNull(url);
+        ArgumentNullException.ThrowIfNull(options);
+
+        LogTargetHost("WebFetch", url);
+
+        // Bound the whole call (including retries) by the configured per-call timeout.
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        cts.CancelAfter(_options.TimeoutMs);
```
gautam-achieveai, line 96: **[Gautam's review bot]** :green_circle: Nit: Consider a brief comment like \// Bounds ALL retries collectively (not per-attempt) to cap total user-facing latency.\ for future readers.

gautam-achieveai:
[gautam-achieveai's bot] **Implementation review for #119 / #113: REQUEST CHANGES**

I reviewed the implementation against issue #113 and the approved v3 + copilot-amendment plan, using the PR-review workflow with parallel lenses for over-engineering, blind spots, architecture/design, code simplicity, test coverage, performance, exception handling, EUII/privacy, and rollout/feature-flag risk.

Good news: I did **not** find over-engineering, architecture, code-simplicity, performance, exception-handling, or feature-flag blockers. The provider allow-list/c `copilot` exclusion and sample-scoped rollout controls look aligned with the approved plan.

## Blocking finding

### HIGH — Tool output echoes raw user URLs/search queries into the transcript

**Files:**
- `src/Misc/Utils/WebFetchTool.cs:129-132`
- `src/Misc/Utils/WebSearchTool.cs:155-157`
- `src/Misc/Utils/WebToolOutput.cs:34-36`
- `src/Misc/Utils/WebToolOutput.cs:98-102`

`WebFetch` wraps the returned content with the validated URL as the source label, and `WebSearch` wraps results with `"web search: " + validation.Value`. The sanitizer only receives `_options.JinaApiKey`, so user-provided URL query strings and search queries are returned verbatim to the model-visible transcript. `FormatFetch` can also emit provider-returned `result.Url` verbatim as `Source: ...`.

That means values like these can be persisted/displayed in the conversation:

```text
https://example.com/reset?email=user@example.com&token=...
web search: <name/email/account id/medical or financial query>
```

This conflicts with the approved plan’s URL/query/echoed-query sanitization requirement. It is also a privacy/EUII risk because URLs and search terms often contain identifiers, secrets, or sensitive intent.

**Recommendation:**
- Do not include raw search query text in returned tool output; use a generic source label such as `web search`.
- For fetch source labels, display a minimized/sanitized URL, e.g. scheme + host + path, or redact/omit query values.
- Sanitize provider-returned `result.Url` before `FormatFetch` emits `Source: ...`.
- Add tests proving URL query tokens and sensitive search queries are not echoed in tool output/logs.

## Other findings

### MEDIUM — `OutputCap` is not enforced as a final output cap

**Files:**
- `src/Misc/Utils/WebToolOutput.cs:144-156`
- `tests/Misc.Tests/Utils/WebToolOutputTests.cs:20-32`
- `tests/Misc.Tests/Utils/WebFetchToolTests.cs:163-175`

`Truncate` currently returns `text[..cap] + TruncationMarker`, so the final output can exceed `WebToolsOptions.OutputCap` by the marker length. The issue/plan described `OutputCap` as the final Markdown character cap applied after formatting/sanitization, and the current tests encode the looser behavior.

**Recommendation:** reserve marker space when truncating and update tests to assert final length `<= OutputCap`, e.g. keep `Math.Max(0, cap - TruncationMarker.Length)` characters before appending the marker.

### MEDIUM — Provider-level cancellation/timeout and WebSearch retry behavior need direct coverage

**Files:**
- `src/Misc/Web/Jina/JinaWebProvider.cs:93-136`
- `tests/Misc.Tests/Http/JinaWebProviderTests.cs:138-237`

Existing tests cover Fetch retries and tool-level token forwarding through fake providers, but they do not directly prove that `JinaWebProvider` passes cancellation into `HttpClient.SendAsync`, that `TimeoutMs` cancels the HTTP operation, or that `SearchAsync` retries 429/5xx.

**Recommendation:** add provider-level tests with a blocking `HttpMessageHandler` for caller cancellation and timeout, plus `SearchAsync_RetriesOn429ThenSucceeds` / `SearchAsync_RetriesOn500ThenSucceeds`.

### LOW — WebSearch contract handler routing is not exercised

**File:**
- `tests/Misc.Tests/Utils/WebToolContractTests.cs:40-106`

The contract tests verify both schemas, but only invoke the `WebFetch` handler. Add a WebSearch handler routing test that invokes `handlers[WebSearchTool.ToolName]` and asserts the fake search provider receives the query/options.

## Verification note

I could not independently run `.NET` verification in this sandbox because `dotnet` is not installed here (`dotnet: not found`). I reviewed the diff and tests statically; the issue comment says the build/tests passed, but this review did not re-run them locally.

src/Misc/Utils/WebFetchTool.cs
```diff
@@ -0,0 +1,177 @@
...
+            var result = await _provider.FetchAsync(
+                validation.Value!,
+                new WebFetchOptions { TargetSelector = targetSelector, NoCache = noCache },
+                cancellationToken
+            );
+
+            var markdown = WebToolOutput.FormatFetch(result);
+            // Use the validated, fragment-stripped URL actually fetched as the source label (not the raw arg).
+            var framed = WebToolOutput.WrapUntrusted(markdown, validation.Value!);
+            var sanitized = WebToolOutput.Sanitize(framed, _options.JinaApiKey);
```
gautam-achieveai, line 136: [gautam-achieveai's bot] [HIGH] This only redacts `JINA_API_KEY`, while the returned output above still includes the full validated URL as the untrusted-content source label. Because `ValidateUrl` preserves query strings, URLs such as password-reset links (`?email=...&token=...`) will be written into the model-visible transcript. The approved plan called out URL/query sanitization; please use a display-safe source label (for example scheme + host + path, with query values omitted/redacted) and add a regression test that a sensitive query parameter is not echoed.

src/Misc/Utils/WebSearchTool.cs
```diff
@@ -0,0 +1,206 @@
...
+                cancellationToken
+            );
+
+            if (result.Items.Count == 0)
+            {
+                return ToolHandlerResult.FromText("No results found.");
+            }
+
+            var markdown = WebToolOutput.FormatSearch(result);
+            var framed = WebToolOutput.WrapUntrusted(markdown, "web search: " + validation.Value);
```
gautam-achieveai: [gautam-achieveai's bot] [HIGH] This source label echoes the raw user search query into the tool result/transcript. Search queries can contain emails, account IDs, names, medical/legal/financial terms, etc., and the subsequent sanitize call only redacts the Jina API key. Please avoid including the raw query in returned output (for example use a generic `web search` label, or a redacted/minimized display value) and add coverage proving sensitive query text is not returned/logged.

src/Misc/Utils/WebToolOutput.cs
```diff
@@ -0,0 +1,158 @@
...
+        var builder = new StringBuilder();
+
+        if (!string.IsNullOrWhiteSpace(result.Title))
+        {
+            _ = builder.Append("# ").Append(result.Title.Trim()).Append("\n\n");
+        }
+
+        if (!string.IsNullOrWhiteSpace(result.Url))
+        {
+            _ = builder.Append("Source: ").Append(result.Url.Trim()).Append("\n\n");
```
gautam-achieveai: [gautam-achieveai's bot] [HIGH] `result.Url` comes from the backend and is emitted verbatim. If Jina returns a URL containing sensitive query parameters, this bypasses the caller-side source-label sanitization even after that is fixed. Please sanitize/minimize provider-returned URLs before formatting (or have `FormatFetch` accept a display-safe URL) so query values/tokens are not persisted in tool output.

src/Misc/Utils/WebToolOutput.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            return string.Empty;
+        }
+
+        if (cap < 0)
+        {
+            cap = 0;
+        }
+
+        return text.Length <= cap ? text : text[..cap] + TruncationMarker;
```
gautam-achieveai: [gautam-achieveai's bot] [MEDIUM] The approved plan describes `OutputCap` as the final Markdown character cap, but this returns `cap + TruncationMarker.Length` characters. Reserve room for the marker when truncating (for example keep `Math.Max(0, cap - TruncationMarker.Length)` characters before appending it) and update the tests to assert final returned length is `<= OutputCap`.

tests/Misc.Tests/Http/JinaWebProviderTests.cs
```diff
@@ -0,0 +1,238 @@
...
+    [Fact]
+    public async Task SearchAsync_EmptyResults_ReturnsEmpty()
+    {
+        var handler = FakeHttpMessageHandler.CreateSimpleJsonHandler("{\"data\":[]}");
+        var provider = CreateProvider(handler);
+
+        var result = await provider.SearchAsync("q", new WebSearchOptions(), CancellationToken.None);
+
+        result.Items.Should().BeEmpty();
+    }
```
gautam-achieveai, line 237: [gautam-achieveai's bot] [MEDIUM] Coverage is missing for provider-level cancellation/timeout and for WebSearch retry behavior. The tools' fake-provider tests prove token forwarding, but they do not prove `JinaWebProvider` passes cancellation into `HttpClient.SendAsync` or that `TimeoutMs` cancels an in-flight HTTP request. Also Fetch has 429/500 retry tests, but Search does not. Please add a blocking `HttpMessageHandler` test for caller cancellation and timeout, plus `SearchAsync_RetriesOn429ThenSucceeds` / `SearchAsync_RetriesOn500ThenSucceeds`.

tests/Misc.Tests/Utils/WebToolContractTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            "{\"url\":\"https://e.com\"}",
+            new ToolCallContext(),
+            CancellationToken.None
+        );
+
+        fetch.Called.Should().BeTrue();
+        // ValidateUrl normalizes the authority-only URL by appending the root path.
+        fetch.ReceivedUrl.Should().Be("https://e.com/");
+        result.ResultText.Should().Contain("routed-content");
+    }
```
gautam-achieveai, line 106: [gautam-achieveai's bot] [LOW] This contract suite verifies both schemas but only invokes the WebFetch handler. Please add a WebSearch handler-routing check that invokes `handlers[WebSearchTool.ToolName]` and asserts the fake search provider receives the query/options, so a future wrong-handler registration is caught.

gautam-achieveai (commented):
[gautam-achieveai's bot] Inline review findings from the implementation review. I could not submit this as a formal Request Changes review because GitHub does not allow requesting changes on your own PR; please treat the HIGH privacy/EUII comments as blocking.

gautam-achieveai:
[gautam-achieveai's bot] **Review feedback addressed — pushed `a38a410`**

Thanks for the thorough review. All blocking + actionable findings are fixed (TDD, RED→GREEN). **Misc.Tests 177 → 200**, LmStreaming.Sample.Tests 352 (no regression), solution build 0 errors / 0 new warnings.

### HIGH — raw URLs/queries echoed into the transcript ✅ fixed
This was a real gap against the plan's "sanitize URLs/queries" requirement. The user's URL/query is now never echoed into model-visible output:
- New `WebToolOutput.MinimizeUrl(...)` returns `scheme://host[:port]/path` only (drops query, fragment, userinfo; placeholder for unparsable input).
- `WebFetchTool` labels the wrapped content with `MinimizeUrl(validatedUrl)` (the **full** URL is still sent to Jina so fetching works); `FormatFetch` minimizes the provider-returned `result.Url` on the `Source:` line.
- `WebSearchTool` now uses a generic `"web search"` label — the query text is no longer echoed.
- Tests prove `https://example.com/reset?email=…&token=…#frag` and a sensitive search query do not appear in tool output (while the provider still receives the full URL).

### MEDIUM — `OutputCap` not a final cap ✅ fixed
`Truncate` now reserves marker room (`cap - marker.Length`), so the final string length is `<= OutputCap`; tests updated to assert `<= cap` (plus a small-cap edge case).

### MEDIUM — provider-level cancel/timeout + WebSearch retry coverage ✅ added
Added `SearchAsync_RetriesOn429ThenSucceeds`/`…On500…`, `FetchAsync_CallerCancellation_PropagatesToSendAsync` (blocking handler awaits the received token → proves the caller token threads into `HttpClient.SendAsync`), and `FetchAsync_PerCallTimeout_CancelsRequest` (`TimeoutMs=30` → linked-CTS `CancelAfter` fires). Deterministic.

### LOW — WebSearch contract handler routing ✅ added
`Build_WebSearchHandler_RoutesToProvider` invokes `handlers[WebSearchTool.ToolName]` and asserts the fake provider receives the query.

### MEDIUM (approving review) — RFC 6761 reserved domains ✅ added
`.test`, `.example`, `.invalid` added to the SSRF blocked-suffix list (+ rejecting tests). Note `example.com` is unaffected — `.example` only matches a `*.example` TLD label.

### LOW (approving review) — shared timeout budget ✅ documented
Added a comment at the `CancelAfter(TimeoutMs)` sites explaining it bounds all retries collectively (caps total user-facing latency).

### Intentionally not changed (non-blocking)
- **`ReadAsStringAsync` → stream parsing:** kept. The body string is required for the plain-markdown (non-JSON) fallback, and `OutputCap` bounds downstream size; this is a perf-only note for very large pages.
- **Splitting `WebToolsOptions`:** kept as one record — acceptable for the single Jina backend; worth revisiting if/when a second backend lands (already an out-of-scope follow-up).

The two pre-existing follow-ups from the PR description still stand: live-verify the Jina POST search field names against the real endpoint, and wiring plain `copilot` (separate `CopilotAgentLoop` path).

gautam-achieveai (commented):
[Gautam's review bot] **Incremental re-review of a38a410 - APPROVE**

All blocking and actionable findings from the previous review are thoroughly addressed. Build passes, **200 tests green** (up from 177).

## Fixes verified

| Prior finding | Status |
|---|---|
| **HIGH** - Raw URLs/queries echoed to model | Fixed via MinimizeUrl (scheme+host+path only) and generic 'web search' label |
| **MEDIUM** - OutputCap not a final cap | Fixed - marker room reserved, final length <= cap guaranteed |
| **MEDIUM** - Provider cancel/timeout + search retry coverage | 4 new provider-level tests with blocking handlers |
| **MEDIUM** - RFC 6761 reserved TLDs | .test, .example, .invalid added to blocklist |
| **LOW** - WebSearch contract routing | Build_WebSearchHandler_RoutesToProvider added |
| **LOW** - Timeout budget undocumented | Comments added at both CancelAfter sites |

## Quality observations

- MinimizeUrl handles all edge cases cleanly (null, whitespace, non-http schemes, relative URIs, userinfo stripping)
- Tests are well-structured: data-driven Theory cases for URL minimization, deterministic blocking handlers for timeout/cancellation
- The Truncate edge case (cap smaller than marker length) is thoughtfully handled rather than throwing

No new issues found. Ship it.

gautam-achieveai:
[gautam-achieveai's bot] **Implementation re-review for #119 / #113: REQUEST CHANGES**

Reviewed the latest PR head `a38a410` in a fresh worktree at `/workspace/LmDotNetTools/pr119-impl-review-latest` against merge-base `927328cc`. I used the PR review workflow and parallel lenses for over-engineering, blind spots, architecture/design, code simplicity, class design, test coverage, performance, exception handling, EUII/privacy, and rollout risk.

Good news: I did **not** find over-engineering, scope creep, architecture, class-design, code-simplicity, performance, or rollout/feature-flag blockers. The previous raw user URL/query echo and final output-cap issues appear addressed. However, one security issue remains blocking, plus several non-blocking robustness/privacy issues.

## Blocking finding

### HIGH — SSRF validation misses IPv4-mapped IPv6 literals

**Files:**
- `src/Misc/Web/WebInputValidator.cs:144-170`
- `src/Misc/Utils/WebFetchTool.cs:123-126`
- `src/Misc/Web/Jina/JinaWebProvider.cs:146-151`

`ValidateUrl` parses bracketed IPv6 literals and delegates them to `IsBlockedIp`. For IPv6 addresses, `IsBlockedIp` rejects link-local, multicast, and unique-local addresses, but it does not normalize IPv4-mapped IPv6 literals back to IPv4 before applying the private/loopback/link-local checks.

That leaves inputs such as these able to bypass the intended private/loopback checks:

```text
http://[::ffff:127.0.0.1]/
http://[::ffff:10.0.0.1]/
http://[::ffff:192.168.1.1]/
http://[::ffff:169.254.169.254]/
```

Once accepted, the full validated URL is passed to the fetch provider and serialized into the Jina request payload. This is a material gap against the PR’s SSRF hygiene requirement.

**Recommendation:** in `IsBlockedIp`, detect `ip.IsIPv4MappedToIPv6`, map with `ip.MapToIPv4()`, and run the existing IPv4 block checks on the mapped address. Add regression tests for IPv4-mapped loopback, private, and link-local ranges.

## Other findings

### MEDIUM — WebSearch result URLs are still emitted verbatim into tool output

**File:** `src/Misc/Utils/WebToolOutput.cs:108-115`

The previous fix correctly minimizes the WebFetch display URL and uses a generic `web search` source label, but `FormatSearch` still writes each upstream result URL directly into the Markdown link target:

```csharp
.Append("](")
.Append(item.Url)
.Append(')');
```

Those URLs come from Jina search results and can contain query-string identifiers, fragments, redirect targets, tracking tokens, or email addresses. They are then returned as model-visible tool output and can be persisted/displayed in the transcript.

**Recommendation:** render search-result links with `WebToolOutput.MinimizeUrl(item.Url)` or another sanitizer that strips query, fragment, and userinfo before emitting them. Add a regression test using a result URL like `https://example.com/reset?email=user@example.com&token=secret#frag`.

### MEDIUM — WebSearch optional parameters are forwarded without range/format validation

**Files:**
- `src/Misc/Utils/WebSearchTool.cs:120-147`
- `src/Misc/Utils/WebToolArgs.cs:47-60`
- `src/Misc/Web/Jina/JinaWebProvider.cs:175-196`

The query itself is validated, but the LLM-controlled optional fields are not. `count` accepts any `int`, including zero, negative, or very large values. `country` and `language` accept arbitrary strings, then flow directly into the Jina JSON payload as `num`, `gl`, and `hl`.

**Recommendation:** enforce a sane documented range for `count` (for example `1..20`, or Jina’s supported limit) and validate `country`/`language` with short documented patterns. Add tests for negative/zero/huge counts and malformed/overlong locale strings.

### MEDIUM — Jina timeout/caller cancellation is bypassed while reading final error bodies

**Files:**
- `src/Misc/Web/Jina/JinaWebProvider.cs:95-108`
- `src/Misc/Web/Jina/JinaWebProvider.cs:127-138`
- `src/LmCore/Http/HttpRetryHelper.cs:142-143`

`JinaWebProvider` creates a linked CTS and passes it through `ExecuteHttpWithRetryAsync`, but the retry helper’s final/non-retryable error path reads the response body with `response.Content.ReadAsStringAsync()` without the provided cancellation token. If Jina returns error headers and then stalls or streams a slow body, the configured `TimeoutMs` and caller cancellation may not interrupt that read.

**Recommendation:** use `ReadAsStringAsync(cancellationToken)` in the retry helper’s error-body path, or add a Jina-specific path that does not read upstream error bodies. Add a regression test with a handler that returns an error response whose content blocks until the token cancels.

### LOW — Non-retryable Jina statuses can still be retried based on raw body text

**Files:**
- `src/LmCore/Http/HttpRetryHelper.cs:143-174`
- `src/LmCore/Http/HttpRetryHelper.cs:212-271`
- used by `src/Misc/Web/Jina/JinaWebProvider.cs:100-108` and `src/Misc/Web/Jina/JinaWebProvider.cs:130-138`

For a non-retryable status, the helper builds an `HttpRequestException` whose message includes the raw response body. The catch filter then calls `IsRetryableError(ex)`, which substring-matches terms like `timeout`, `500`, and `503`. A `400`/`401` body containing those strings can trigger avoidable retries before the correct error is surfaced.

This appears pre-existing in the shared helper, but this PR newly relies on it for Jina requests.

**Recommendation:** when `HttpRequestException.StatusCode` is present, classify retryability from the status code first and do not fall back to body/message substring matching for non-retryable statuses.

## Question

### Should `targetSelector` be bounded/validated before becoming an HTTP header?

**Files:**
- `src/Misc/Utils/WebFetchTool.cs:120-125`
- `src/Misc/Web/Jina/JinaWebProvider.cs:157-160`

`targetSelector` is parsed from tool JSON and added to `X-Target-Selector` via `TryAddWithoutValidation`. Is the lack of length/control-character validation intentional? Since the value is LLM-controlled, a conservative length cap and control-character rejection would make this boundary more robust.

## Verification note

I could not run `.NET` verification in this sandbox because `dotnet` is not installed (`dotnet: not found`). This review is based on source/diff inspection plus parallel reviewer agents. The final severity pass kept the IPv4-mapped IPv6 SSRF bypass as a blocking HIGH issue and recommended **REQUEST CHANGES**.

src/Misc/Web/WebInputValidator.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            return true;
+        }
+
+        if (IPAddress.IsLoopback(ip))
+        {
+            return true;
+        }
+
+        if (ip.AddressFamily == AddressFamily.InterNetworkV6)
```
gautam-achieveai, line 175: @/workspace/Memory/tasks/review-lmdotnettools-issue-119-artifacts/inline-1.md

src/Misc/Utils/WebToolOutput.cs
```diff
@@ -0,0 +1,203 @@
...
+                _ = builder.Append("\n\n");
+            }
+
+            _ = builder
+                .Append("### ")
+                .Append(i + 1)
+                .Append(". [")
+                .Append(item.Title)
+                .Append("](")
+                .Append(item.Url)
```
gautam-achieveai: @/workspace/Memory/tasks/review-lmdotnettools-issue-119-artifacts/inline-2.md

src/Misc/Utils/WebSearchTool.cs
```diff
@@ -0,0 +1,207 @@
...
+        if (!validation.IsValid)
+        {
+            return Error("WebSearch error: " + validation.Error);
+        }
+
+        try
+        {
+            var result = await _provider.SearchAsync(
+                validation.Value!,
+                new WebSearchOptions
```
gautam-achieveai, line 144: @/workspace/Memory/tasks/review-lmdotnettools-issue-119-artifacts/inline-3.md

src/Misc/Web/Jina/JinaWebProvider.cs
```diff
@@ -0,0 +1,367 @@
...
+        LogTargetHost("WebFetch", url);
+
+        // This timeout bounds ALL retry attempts collectively (not per-attempt), capping total
+        // user-facing latency. Intentional trade-off: a slow first attempt leaves less time for retries.
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        cts.CancelAfter(_options.TimeoutMs);
+
+        // The request is built inside the operation lambda so each retry attempt gets a fresh
+        // HttpRequestMessage (a sent request cannot be reused).
+        return await ExecuteHttpWithRetryAsync(
```
gautam-achieveai, line 100: @/workspace/Memory/tasks/review-lmdotnettools-issue-119-artifacts/inline-4.md

src/Misc/Web/Jina/JinaWebProvider.cs
```diff
@@ -0,0 +1,367 @@
...
+        var payload = JsonSerializer.Serialize(new Dictionary<string, string> { ["url"] = url });
+        var request = new HttpRequestMessage(HttpMethod.Post, ReaderUrl)
+        {
+            Content = new StringContent(payload, Encoding.UTF8, "application/json"),
+        };
+
+        request.Headers.Accept.ParseAdd("application/json");
+        _ = request.Headers.TryAddWithoutValidation("X-Return-Format", "markdown");
+
+        if (!string.IsNullOrEmpty(options.TargetSelector))
```
gautam-achieveai, line 157: @/workspace/Memory/tasks/review-lmdotnettools-issue-119-artifacts/inline-5.md

gautam-achieveai:
[gautam-achieveai's bot] **Round-2 re-review addressed — pushed `e9dc2be`**

Thanks for the deep re-review. All actionable findings are fixed (TDD, RED→GREEN). **Misc.Tests 200 → 219**, **LmCore.Tests 764**, LmStreaming.Sample.Tests 352; solution build 0 errors / 0 new warnings.

### HIGH — SSRF: IPv4-mapped IPv6 literals bypass ✅ fixed
`WebInputValidator.IsBlockedIp` now normalizes mapped addresses first: `if (ip.IsIPv4MappedToIPv6) ip = ip.MapToIPv4();`, then runs the existing IPv4 private/loopback/link-local/multicast/unspecified checks. Added reject tests for `[::ffff:127.0.0.1]`, `[::ffff:10.0.0.1]`, `[::ffff:192.168.1.1]`, and `[::ffff:169.254.169.254]` (the cloud-metadata case was the dangerous miss — `::ffff:127.0.0.1` was already caught by `IPAddress.IsLoopback`, the private/link-local ones were not).

### MEDIUM — WebSearch result URLs emitted verbatim ✅ fixed (one tradeoff to confirm)
`FormatSearch` now renders each result link target via `MinimizeUrl(item.Url)` (strips query, fragment, userinfo) — consistent with the WebFetch `Source:`/label treatment. Regression test added with a `?email=…&token=…#frag` result URL.

> **Tradeoff to flag:** minimizing result links drops their query string, so a result whose canonical address depends on query params (`…/article?id=123`) renders as `…/article` and may not be directly re-fetchable via `WebFetch`. I went with privacy-over-reusability for consistency. If you'd prefer to **keep the query on result links** (so the model can re-fetch them) while still stripping userinfo/fragment, say the word and I'll switch `FormatSearch` to a query-preserving sanitizer.

### MEDIUM — LLM-controlled optional params unvalidated ✅ fixed
Validated at the tool boundary, gracefully (invalid optional values are omitted/clamped, the call still proceeds):
- `count` → `Math.Clamp(value, 1, 20)`.
- `country` (`gl`) → accepted only as 2 ASCII letters, else omitted.
- `language` (`hl`) → `^[A-Za-z]{2,5}(-[A-Za-z0-9]{1,8})?$`, else omitted (implemented with char checks, no `Regex`, to avoid SYSLIB1045 under `TreatWarningsAsErrors`).
- `targetSelector` → omitted if it contains any control char (incl. CR/LF) or exceeds 256 chars, **before** it reaches the `X-Target-Selector` header (closes the `TryAddWithoutValidation` header-injection vector — answers the open question too).

### MEDIUM — timeout/cancellation bypassed while reading the final error body ✅ fixed
`HttpRetryHelper`'s final error-body read now uses `ReadAsStringAsync(cancellationToken)`. One related correctness fix was needed in the same block: the fallback `catch (Exception ex) when (ex is not HttpRequestException)` is now `… and not OperationCanceledException`, otherwise the token-aware read's `OperationCanceledException` was reclassified into an `HttpRequestException` by `EnsureSuccessStatusCode()` and cancellation never surfaced. Both changes only affect the already-cancelled-token path. Verified with a new `tests/LmCore.Tests/Http/HttpRetryHelperCancellationTests.cs` using a blocking-body `HttpContent` (RED = 15s hang → GREEN = `OperationCanceledException` in ~70ms). (This is the one deliberate shared-`LmCore` change, made per your recommendation; full `LmCore.Tests` suite stays green.)

### LOW — non-retryable statuses retried on raw body substring match — deferred
This lives in the shared `HttpRetryHelper.IsRetryableError` and changing it alters retry classification for **every** provider in the repo. As you noted it's pre-existing; I've left it out of this feature PR and will track it as a separate follow-up (classify retryability from `StatusCode` first when present). Happy to do it in a dedicated PR.

---
Updated branch: 7 commits, head `e9dc2be`. Existing follow-ups still stand (live-verify Jina search field names; plain-`copilot` wiring; and now the retry-classification cleanup above).

gautam-achieveai:
[gautam-achieveai's bot] **Re-review for #119 / #113: APPROVE WITH COMMENTS**

Reviewed the latest delta from `a38a410` to `e9dc2bed` in `/workspace/LmDotNetTools/pr119-rereview-clean-e9dc2bed`, focusing on the previously posted findings and question.

The previous blocking/security issue is addressed:

- IPv4-mapped IPv6 addresses are now mapped back to IPv4 before the IPv4 private/loopback/link-local checks run (`src/Misc/Web/WebInputValidator.cs:155-161`), with regression cases for mapped loopback/private/link-local addresses (`tests/Misc.Tests/Web/WebInputValidatorTests.cs:50-53`).
- WebSearch result URLs are now minimized before Markdown emission (`src/Misc/Utils/WebToolOutput.cs:108-117`), with coverage for dropping query/fragment/userinfo (`tests/Misc.Tests/Utils/WebToolOutputTests.cs:173-197`).
- WebSearch optional parameters are now bounded/normalized before reaching the provider (`src/Misc/Utils/WebSearchTool.cs:139-149`, `:189-230`), with tests for count clamping and malformed locale omission (`tests/Misc.Tests/Utils/WebSearchToolTests.cs:154-218`).
- The final error-body read in `HttpRetryHelper` now honors the cancellation token and avoids masking cancellation (`src/LmCore/Http/HttpRetryHelper.cs:142-160`), with a blocking-body cancellation regression test (`tests/LmCore.Tests/Http/HttpRetryHelperCancellationTests.cs:15-40`).
- The `targetSelector` question was addressed by bounding the value and rejecting control characters before it reaches the `X-Target-Selector` header path (`src/Misc/Utils/WebFetchTool.cs:123-129`, `:160-173`), with tests for control characters and overlong selectors (`tests/Misc.Tests/Utils/WebFetchToolTests.cs:141-168`).

## Remaining non-blocking comment

### LOW — Retry classification can still be misled by non-retryable response body text

**File:** `src/LmCore/Http/HttpRetryHelper.cs:142-174`, `src/LmCore/Http/HttpRetryHelper.cs:215-274`

The cancellation part of the prior finding is fixed, but the lower-severity retry-classification issue remains: for a non-retryable status, the helper builds an `HttpRequestException` whose message includes the raw response body, then the catch filter calls `IsRetryableError(ex)`, which substring-matches terms like `timeout`, `429`, `500`, or `503`. A `400`/`401` response body containing those strings can still trigger avoidable retries before surfacing the correct error.

This is mostly pre-existing and bounded to unnecessary delay/extra calls, so it is not blocking this PR.

**Suggestion:** when `HttpRequestException.StatusCode` is present, classify retryability from the status code first and do not fall back to body/message substring matching for explicitly non-retryable statuses.

## Verification note

I could not run the .NET test suite in this sandbox because `dotnet` is unavailable (`dotnet: not found`). This re-review is based on source/diff inspection plus focused parallel reviewers for tests, privacy/EUII, exception handling, and blind spots. No new blocking findings were found.

gautam-achieveai (commented):
[gautam-achieveai's bot] Re-review completed at `e9dc2bed`: previous blocking/security findings are addressed. Remaining LOW retry-classification note is non-blocking; see summary comment: https://github.com/achieveai/LmDotnetTools/pull/119#issuecomment-4845290857

gautam-achieveai (commented):
[gautam-achieveai's bot] Re-review completed at `e9dc2bed`: previous blocking/security findings are addressed. Remaining LOW retry-classification note is non-blocking; see summary comment: https://github.com/achieveai/LmDotnetTools/pull/119#issuecomment-4845290857

gautam-achieveai:
[gautam-achieveai's bot] **LOW retry-classification finding — fixed in `f1e9f07`**

Resolved rather than deferred — the fix is narrow and the PR already touches `HttpRetryHelper`.

`HttpRetryHelper.IsRetryableError` now classifies by status code first: when `HttpRequestException.StatusCode` is present it returns `IsRetryableStatusCode(statusCode)` and does not consult the message/body at all. The existing substring detection (`timeout`/`network`/connection/5xx) is retained only as the fallback for status-less transport exceptions from `SendAsync`. Net effect: 429/5xx still retry; non-retryable 4xx never retry regardless of body text; genuine network errors are unchanged.

Coverage (`tests/LmCore.Tests/Http/HttpRetryHelperClassificationTests.cs`): a 400 whose body contains `Internal Server Error`/`500`/`timeout` is no longer retryable (unit) and is invoked exactly once then thrown as `HttpRequestException(BadRequest)` (integration) — both RED before the change (the 400 was retried 3×), GREEN after; plus 5xx/429 → retryable, status-less `timeout` → retryable, 401 → not.

Verified across the retry-dependent suites, all green and unmodified: LmCore 770, OpenAIProvider 92, AnthropicProvider 174, LmEmbeddings 370, Misc 219. Solution build 0 errors / 0 new warnings.

This clears the last open comment; the two product follow-ups (live-verify Jina search field names; plain-\`copilot\` wiring) remain tracked separately.

PR #120, opened 2026-06-30
fix(responses+copilot): retry transient 502s on streaming Responses + WebSocket connect (#114)

### LINKED ISSUE #114: OpenAiResponsesClient + Copilot WebSocket don't retry transient 502 Bad Gateway
## Summary

Streaming calls to the **OpenAI Responses API** and the **GitHub Copilot Responses WebSocket** transport do not retry transient upstream failures. A **502 Bad Gateway** from the Copilot proxy (observed intermittently in real agent runs) fails the whole turn instead of being retried, even though the repo already has a retry framework that classifies 502 as retryable and other providers already use it.

## Impact / Motivation

- Copilot-backed agents (GPT‑5.5 via `/responses`) intermittently hit **502 Bad Gateway** from the Copilot gateway. These are transient — a short retry-with-backoff would absorb them — but today the run dies on the first 502.
- The **WebSocket transport is the *default*** for Copilot Responses (`CopilotResponsesAgentFactory.Create` defaults `transport = WebSocket`; the provider README example uses it). Its failure mode is worse: a transient mid-turn socket close is silently swallowed and the turn is returned **as if it completed** — a partial answer with no error.

## Affected Paths

| Transport | Client | Current behavior on transient 502 / drop |
|---|---|---|
| Copilot SSE / OpenAI Responses (GPT‑5.5) | `OpenAiResponsesClient.StreamResponseAsync` | **Throws `HttpRequestException` on first non‑2xx, no retry** (`src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:104-127`) |
| Direct OpenAI Responses provider | same `OpenAiResponsesClient` (`OpenAiResponsesClient.cs:69`) | Same gap — shared client, so one fix covers both |
| Copilot Responses **WebSocket** (default) | `CopilotResponsesWebSocketClient` | **No retry on connect** (`src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs:146-168`); **silent truncation** on mid-turn close via `yield break` (`...WebSocketClient.cs:110-115`) |
| Copilot Anthropic (Sonnet) | `AnthropicClient` | **Already retries 502** — verify-only, not a fix target (see Root Cause) |

## Steps to Reproduce

**SSE / Responses:**
1. Create a Copilot-backed Responses agent over SSE (`CopilotResponsesAgentFactory.Create(..., transport: Sse)`), or any direct `OpenAiResponsesClient`.
2. Have the upstream return `502 Bad Gateway` for the `POST /responses` (reproducible deterministically with a mock handler that returns 502 then a valid SSE body).
3. Call `StreamResponseAsync`.
4. **Observed:** it throws `HttpRequestException` immediately on the 502; no second attempt is made.

**WebSocket:**
1. Create a Copilot Responses agent with the default transport (WebSocket).
2. Drop the socket mid-turn (after some events, before completion).
3. **Observed:** the enumerable ends cleanly (`yield break`) — the caller receives a partial turn with no error and no retry.

## Expected Behavior

- A transient **502 (and other retryable 5xx / 429 / connection faults)** on the initial `POST /responses` is retried automatically with exponential backoff before the error surfaces — consistent with `AnthropicClient` and `OpenClient`.
- A transient WebSocket **connect** failure is retried; a mid-turn socket drop **surfaces as an error** instead of being silently presented as a complete turn.

## Actual Behavior

- `OpenAiResponsesClient` throws on the first non‑2xx with zero retries.
- `CopilotResponsesWebSocketClient` has no connect retry and silently truncates mid-turn drops as if the turn completed successfully.

## Root Cause

The repo already has a complete retry framework that **already classifies 502 as retryable**:
- `HttpRetryHelper.IsRetryableStatusCode` → 429 + all 5xx (`src/LmCore/Http/HttpRetryHelper.cs:200-204`); `IsRetryableError` matches `"502"` / `"Bad Gateway"` / `"Service Unavailable"` / timeouts / `"response ended prematurely"` / connection-closed (`HttpRetryHelper.cs:218-271`).
- `RetryOptions` defaults: `MaxRetries=2` (3 total attempts), `InitialDelayMs=1000`, `MaxDelayMs=30000`, `BackoffMultiplier=2.0` (`src/LmCore/Http/RetryOptions.cs:29-48`).

It is wired into `AnthropicClient` (streaming path `StreamingChatCompletionsAsync` wraps the POST in `ExecuteHttpWithRetryAsync`, `src/AnthropicProvider/Agents/AnthropicClient.cs:192-223`; defaults `_retryOptions ?? RetryOptions.Default`, line 53) and `OpenClient` (`src/OpenAIProvider/Agents/OpenClient.cs:290-324`).

**The gap is that `OpenAiResponsesClient` was never given retry.** `StreamResponseAsync` builds one request and, on `!IsSuccessStatusCode`, throws straight away (`OpenAiResponsesClient.cs:104-127`). Neither construction site supplies retry:
- `CopilotResponsesAgentFactory.CreateSseClient` calls `CopilotHttpClientFactory.Create(...)` without the optional `innerHandler` (`src/GithubCopilotProvider/Agents/CopilotResponsesAgentFactory.cs:70-71`).
- The static `OpenAiResponsesClient.Create` (direct OpenAI) likewise wraps a plain client.

The **Copilot-Anthropic** path is *not* affected: `CopilotAnthropicAgentFactory` builds an `AnthropicClient`, which already retries 502 on its streaming POST. It only needs a regression test to lock that in.

The **WebSocket** path is a separate root cause: `CopilotResponsesWebSocketClient` doesn't use `HttpClient`/`HttpRetryHelper` at all, has no connect retry, and treats a mid-turn read returning null as normal end-of-stream (`yield break`).

## Proposed Fix (scope)

**Decided design — in-client retry, not handler-level.** Retry belongs inside `OpenAiResponsesClient.StreamResponseAsync`, because handler-level retry (via the unused `CopilotHttpClientFactory.innerHandler`) would only cover the Copilot SSE path — not the direct-OpenAI client and not WebSocket (which doesn't use `HttpClient`).

1. **`OpenAiResponsesClient` (SSE + direct OpenAI):** wrap **only the initial POST + status check (pre‑stream)** in `HttpRetryHelper.ExecuteHttpWithRetryAsync`, reusing `RetryOptions`. Iterate the SSE stream **outside** the retry scope.
   - **Build the `HttpRequestMessage` + `JsonContent` *inside* the retry lambda** — an `HttpRequestMessage`/`HttpContent` can't be re-sent; a naive re-send throws `InvalidOperationException` ("request already sent"), which is **not** an `HttpRequestException` and would slip past the retry filter and crash. Mirror `OpenClient`/`AnthropicClient` streaming. (Bonus: a fresh request per attempt also refreshes the Copilot bearer token + `x-interaction-id` via `CopilotHeadersHandler`.)
   - Add an optional `RetryOptions? retryOptions = null` ctor param (default `RetryOptions.Default`) so tests can inject `RetryOptions.FastForTests` and callers can tune interactive backoff. Mirrors `AnthropicClient`.
   - Thread a **real logger** into the Copilot SSE client — it's currently constructed with `logger: null` (`CopilotResponsesAgentFactory.cs:71`), so retry warnings would be invisible on the exact path that motivated this bug. `CopilotResponsesAgentFactory` already has a `logger` param; forward it to `CreateSseClient`.
2. **`CopilotResponsesWebSocketClient`:** add retry on connect (`EnsureConnectedAsync`/`ConnectAsync`) for transient faults, **and fix the silent truncation** — a mid-turn socket close must surface as an error rather than `yield break`-ing a partial turn as if complete.
3. **`AnthropicClient` (Copilot-Anthropic):** no code change — add a regression test confirming it retries a pre-stream 502.

## Scope Boundary (explicit)

- **In scope:** transient faults **before the first SSE byte** (the POST/handshake) — 502/503/504/500/429 and connection-refused/DNS/TLS failures; WebSocket connect; WebSocket silent-truncation fix.
- **Out of scope (follow-up):** **mid-stream** resilience for SSE — a reset *after* a 200 OK while reading events (`OpenAiResponsesClient.cs:130-136`) is not retried and surfaces as an error. True mid-stream resume would need response buffering + `previous_response_id` replay and is a much larger change. Track separately.
- **Idempotency caveat:** `/responses` POST carries no `Idempotency-Key`, and the SSE path doesn't force `Store=false` (the WS path does, `...WebSocketClient.cs:92`). A 502 returned *after* the backend processed the request would, on retry, create a duplicate stored response / double-bill. Window is tiny (retry only before first byte ≈ request likely never reached the backend). Document as accepted risk unless the backend dedupes; consider sending an `Idempotency-Key`.
- **Interactive latency:** with `RetryOptions.Default` a fully-failing sequence adds ~1s + ~2s ≈ 3s before the error surfaces to an interactive chat user. Acceptable per the "reuse defaults" decision; consider surfacing a "retrying…" status to the UI rather than a silent pause (optional follow-up).

## Acceptance Criteria

- [ ] A transient 502 on the initial `POST /responses` is retried (default 3 attempts, exp backoff) for both the Copilot SSE and direct OpenAI Responses paths; a persistent 502 still throws `HttpRequestException` carrying `StatusCode == BadGateway` after exhausting retries.
- [ ] Non‑retryable statuses (e.g. 400/404) still throw immediately — exactly one POST, no added latency.
- [ ] Cancellation during backoff throws `OperationCanceledException` and issues no further POST.
- [ ] `OpenAiResponsesClient` accepts an optional `RetryOptions` and logs each retry via a non-null logger on the Copilot SSE path.
- [ ] WebSocket connect failures are retried; a mid-turn socket drop surfaces as an error instead of a silently-truncated successful turn.
- [ ] A regression test confirms the Copilot-Anthropic path retries a pre-stream 502.

## Tests (required)

New fail-injection test infra is needed — existing `OpenAiResponsesTestSseMessageHandler` has no fail mode and `MockHttpHandlerBuilder.RetryScenario` returns a JSON (not `text/event-stream`) success body, so it can't drive *502‑then‑SSE*. Add a "fail first N with status X, then stream valid SSE" handler. Then, using `RetryOptions.FastForTests`:

1. 502 once → then valid SSE → `StreamResponseAsync` yields the full event sequence; assert exactly **2** POSTs.
2. 502 on every attempt → throws `HttpRequestException` with `StatusCode == BadGateway` after `MaxRetries+1` POSTs.
3. Non‑retryable 400 → throws immediately, exactly 1 POST.
4. Cancellation during backoff → `OperationCanceledException`, no further POST.
5. WebSocket: transient connect failure retried; mid-turn drop surfaces an error (no silent `yield break`).
6. Copilot-Anthropic: pre-stream 502 retried (locks in existing behavior).

## Repro Proof

The 502 is an upstream-transient condition, so it can't be reproduced deterministically against the live Copilot gateway. The defect is **structural and verified by direct code reading**: `OpenAiResponsesClient.StreamResponseAsync` contains no retry around the throw at `OpenAiResponsesClient.cs:123`, and neither construction site injects any. The deterministic repro is test #1 above (mock handler returns 502 then SSE): it fails today (client throws on the first 502) — **RED proves the repro** — and passes once retry is added.

## Key Files

- `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs` (fix target)
- `src/GithubCopilotProvider/Agents/CopilotResponsesAgentFactory.cs` (logger + SSE wiring)
- `src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs` (connect retry + truncation fix)
- `src/LmCore/Http/HttpRetryHelper.cs`, `src/LmCore/Http/RetryOptions.cs` (existing retry framework — reuse)
- `src/AnthropicProvider/Agents/AnthropicClient.cs`, `src/OpenAIProvider/Agents/OpenClient.cs` (streaming-retry precedents)
- `src/LmTestUtils/TestMode/OpenAiResponsesTestSseMessageHandler.cs`, `tests/OpenAiResponsesProvider.Tests/OpenAiResponsesTestSseMessageHandlerTests.cs` (test infra to extend)

gautam-achieveai:
## Summary

Transient **502 Bad Gateway** responses from the GitHub Copilot proxy were killing an agent turn instead of being retried, and the Copilot **WebSocket** transport silently truncated a mid-turn drop as if the turn had completed. This adds **pre-stream retry** to the streaming Responses client (covering both the Copilot SSE path and the direct OpenAI Responses provider, reusing the existing retry framework), gives the WebSocket transport its own **connect-retry**, and surfaces a mid-turn drop as a real error. Also fixes a latent response-disposal leak in `HttpRetryHelper`.

Fixes #114

## Changes

- **`OpenAiResponsesClient.StreamResponseAsync`** — wraps the **initial POST + status check only** in `HttpRetryHelper.ExecuteHttpWithRetryAsync` (reusing `RetryOptions`, which already classifies 502/5xx/429 as retryable). The request is rebuilt **fresh inside the retry lambda** (a sent `HttpRequestMessage` can't be re-sent — and this refreshes the bearer token + `x-interaction-id` per attempt); the `(Response, Stream)` tuple is enumerated **outside** the retry scope in a `try/finally` that disposes both, so early cancel / consumer break / parser failure never pins the connection. Single public ctor: logger → non-generic `ILogger?`, appended `RetryOptions?`.
- **`CopilotResponsesAgentFactory`** — `Create` accepts an optional `RetryOptions`, forwarded to **both** transports; the SSE path now forwards a real logger (was `logger: null`, so retries were invisible).
- **`CopilotResponsesWebSocketClient`** — a bounded **connect-retry** loop (fresh socket + token + headers per attempt; failed sockets disposed; `_socket` assigned only on success; retry **only before** `response.create`). A mid-turn peer close before a terminal event now **throws** ("closed before turn completion") instead of `yield break`; `ReceiveTextAsync` lets a `WebSocketException` propagate (with detail) and reserves `null` for a genuine peer close; `_previousResponseId` advances only on `response.completed`.
- **`HttpRetryHelper`** — the final non-retryable / max-retries branch now **disposes the failed `HttpResponseMessage`** (capture status/body → dispose in `finally` → throw), fixing a leak that affected every caller.

## Key Decisions

1. **In-client retry, not handler-level.** Retrying inside `OpenAiResponsesClient` fixes the Copilot SSE path *and* the direct OpenAI Responses provider in one place; a handler-level retry would have missed both the direct path and the (non-`HttpClient`) WebSocket transport.
2. **Pre-stream only.** Only faults before the first SSE byte / before `response.create` are retried; a mid-stream drop surfaces as an error. True mid-stream resume (buffering + `previous_response_id` replay) is intentionally **out of scope** (follow-up).
3. **Reuse the existing retry framework** with its default `RetryOptions` (3 attempts, exponential backoff), consistent with `AnthropicClient`/`OpenClient`. The Copilot-Anthropic path already retried — covered here only by a verify-only regression test.
4. **Source-compatible API change (not binary).** A single public constructor (no overload ambiguity); the logger parameter type change + appended `RetryOptions` recompile cleanly for in-tree callers — flagged explicitly rather than claimed as binary-compatible.
5. **DNS `HostNotFound` is not retried** — an authoritative name-resolution failure is permanent; the transient DNS case stays covered by `TryAgain`.

## Testing

- `dotnet build LmDotnetTools.sln` — **0 errors, 6 warnings** (all pre-existing `NU1902`; zero new).
- `dotnet test` (touched projects, all green): `OpenAiResponsesProvider.Tests` **76**, `GithubCopilotProvider.Tests` **30**, `LmCore.Tests` **765**, `AnthropicProvider.Tests` **175**.
- New coverage: 502-once-then-SSE retry (distinct fresh request per attempt, `Accept: text/event-stream`), persistent-502 → `BadGateway` after `MaxRetries+1` with the final response disposed, non-retryable 400 → immediate single POST, early-cancel/consumer-break disposes response + stream, cancel-during-backoff, SSE logger propagation; WS connect-retry (fresh socket per attempt), persistent/non-retryable connect, `HostNotFound` not retried, cancel during connect & backoff, mid-turn close → error, `WebSocketException` propagation, stale-socket disposal on reconnect, `previous_response_id` not advanced after a truncated turn; `HttpRetryHelper` final-failed-response disposal; Copilot-Anthropic pre-stream 502 (verify-only).

## Out of scope (follow-ups)

Mid-stream SSE resume; `Idempotency-Key` for `/responses`; backoff jitter / `Retry-After`; a UI "retrying…" indicator.

### DIFF
--- a/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
+++ b/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
@@ -238,13 +238,15 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
 
         return wsEx.InnerException switch
         {
-            // TryAgain (WSATRY_AGAIN) is the non-authoritative "DNS hiccup, retry" case. HostNotFound
-            // (WSAHOST_NOT_FOUND) is an authoritative "no such host" name-resolution failure and is NOT
-            // retried — see the doc comment above.
+            // Transient transport-layer failures. TryAgain (WSATRY_AGAIN) is the non-authoritative
+            // "DNS hiccup, retry" case; HostUnreachable/NetworkUnreachable are transient routing
+            // failures. HostNotFound (WSAHOST_NOT_FOUND, authoritative "no such host") is NOT here —
+            // it is a permanent name-resolution failure (see the doc comment above).
             SocketException socketEx => socketEx.SocketErrorCode
                 is SocketError.ConnectionRefused
                     or SocketError.TimedOut
                     or SocketError.HostUnreachable
+                    or SocketError.NetworkUnreachable
                     or SocketError.TryAgain,
             HttpRequestException { StatusCode: { } status } => HttpRetryHelper.IsRetryableStatusCode(status),
             _ => false,
--- a/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
+++ b/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
@@ -162,6 +162,16 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
             return _socket;
         }
 
+        // A previously-stored socket that is no longer connected (e.g. after a faulted/truncated turn)
+        // must be disposed before we replace it, otherwise its underlying connection leaks until
+        // finalization. Null it out first so a failed reconnect never leaves a half-open socket stored.
+        if (_socket is not null)
+        {
+            var stale = _socket;
+            _socket = null;
+            await DisposeSocketSafelyAsync(stale).ConfigureAwait(false);
+        }
+
         // Pre-turn connect-retry: this runs ONLY before the first response.create frame is sent, so a
         // retry is safe (idempotent). Each attempt refreshes the token + headers (fresh x-interaction-id)
         // and creates a NEW socket; a failed socket is disposed locally and _socket is assigned ONLY
@@ -228,12 +238,14 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
 
         return wsEx.InnerException switch
         {
+            // TryAgain (WSATRY_AGAIN) is the non-authoritative "DNS hiccup, retry" case. HostNotFound
+            // (WSAHOST_NOT_FOUND) is an authoritative "no such host" name-resolution failure and is NOT
+            // retried — see the doc comment above.
             SocketException socketEx => socketEx.SocketErrorCode
                 is SocketError.ConnectionRefused
                     or SocketError.TimedOut
                     or SocketError.HostUnreachable
-                    or SocketError.TryAgain
-                    or SocketError.HostNotFound,
+                    or SocketError.TryAgain,
             HttpRequestException { StatusCode: { } status } => HttpRetryHelper.IsRetryableStatusCode(status),
             _ => false,
         };
--- a/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
+++ b/src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
@@ -1,8 +1,10 @@
+using System.Net.Sockets;
 using System.Net.WebSockets;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
 using AchieveAi.LmDotnetTools.GithubCopilotProvider.Auth;
+using AchieveAi.LmDotnetTools.LmCore.Http;
 using AchieveAi.LmDotnetTools.OpenAiResponsesProvider.Agents;
 using AchieveAi.LmDotnetTools.OpenAiResponsesProvider.Models;
 using Microsoft.Extensions.Logging;
@@ -36,6 +38,7 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
     private readonly CopilotOptions _options;
     private readonly Func<ICopilotResponsesSocket> _socketFactory;
     private readonly ILogger _logger;
+    private readonly RetryOptions _retryOptions;
     private readonly SemaphoreSlim _turnGate = new(1, 1);
 
     private ICopilotResponsesSocket? _socket;
@@ -53,13 +56,18 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
     /// <param name="socketFactory">
     ///     Optional socket factory (override in tests). Defaults to a real <see cref="ClientWebSocket"/>.
     /// </param>
+    /// <param name="retryOptions">
+    ///     Optional connect-retry configuration (bounds the pre-turn connect-retry loop). Defaults to
+    ///     <see cref="RetryOptions.Default"/>.
+    /// </param>
     public CopilotResponsesWebSocketClient(
         Uri endpoint,
         ICopilotTokenProvider tokenProvider,
         CopilotSessionContext session,
         CopilotOptions? options = null,
         ILogger? logger = null,
-        Func<ICopilotResponsesSocket>? socketFactory = null
+        Func<ICopilotResponsesSocket>? socketFactory = null,
+        RetryOptions? retryOptions = null
     )
     {
         _endpoint = endpoint ?? throw new ArgumentNullException(nameof(endpoint));
@@ -68,6 +76,7 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
         _options = options ?? new CopilotOptions();
         _logger = logger ?? NullLogger.Instance;
         _socketFactory = socketFactory ?? (() => new ClientWebSocketResponsesSocket());
+        _retryOptions = retryOptions ?? RetryOptions.Default;
     }
 
     /// <inheritdoc />
@@ -107,12 +116,15 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
-                var text = await socket.ReceiveTextAsync(cancellationToken).ConfigureAwait(false);
-                if (text is null)
-                {
-                    // Socket closed mid-turn.
-                    yield break;
-                }
+                // A null is a genuine peer Close frame. Reaching here means the socket closed BEFORE a
+                // terminal response.completed/response.failed event, i.e. the turn was truncated. Surface
+                // it as an error instead of silently ending the stream as though the turn had completed
+                // (which would also wrongly leave _previousResponseId chained to a partial turn).
+                var text =
+                    await socket.ReceiveTextAsync(cancellationToken).ConfigureAwait(false)
+                    ?? throw new IOException(
+                        "WebSocket closed before turn completion (no response.completed/response.failed received)."
+                    );
 
                 if (string.IsNullOrWhiteSpace(text))
                 {
@@ -150,21 +162,93 @@ public sealed class CopilotResponsesWebSocketClient : IOpenAiResponsesClient, IA
             return _socket;
         }
 
-        var token = await _tokenProvider.GetTokenAsync(cancellationToken).ConfigureAwait(false);
+        // Pre-turn connect-retry: this runs ONLY before the first response.create frame is sent, so a
+        // retry is safe (idempotent). Each attempt refreshes the token + headers (fresh x-interaction-id)
+        // and creates a NEW socket; a failed socket is disposed locally and _socket is assigned ONLY
+        // after a successful connect, so a half-open socket is never observed by a turn.
+        var attempt = 0;
+        while (true)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
 
-        var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            var token = await _tokenProvider.GetTokenAsync(cancellationToken).ConfigureAwait(false);
+            var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                ["Authorization"] = $"Bearer {token}",
+            };
+            foreach (var header in CopilotRequestHeaders.Build(_options, _session, Guid.NewGuid().ToString()))
+            {
+                headers[header.Key] = header.Value;
+            }
+
+            var socket = _socketFactory();
+            try
+            {
+                await socket.ConnectAsync(_endpoint, headers, cancellationToken).ConfigureAwait(false);
+                _socket = socket;
+                return socket;
+            }
+            catch (Exception ex) when (attempt < _retryOptions.MaxRetries && IsRetryableConnect(ex))
+            {
+                attempt++;
+                var delay = _retryOptions.CalculateDelay(attempt);
+                _logger.LogWarning(
+                    "WebSocket connect failed (attempt {Attempt}/{MaxRetries}), retrying in {Delay}ms: {Error}",
+                    attempt,
+                    _retryOptions.MaxRetries + 1,
+                    delay.TotalMilliseconds,
+                    ex.Message
+                );
+
+                await DisposeSocketSafelyAsync(socket).ConfigureAwait(false);
+                await Task.Delay(delay, cancellationToken).ConfigureAwait(false);
+            }
+            catch
+            {
+                // Non-retryable, retries exhausted, or cancelled: dispose the failed socket and surface.
+                await DisposeSocketSafelyAsync(socket).ConfigureAwait(false);
+                throw;
+            }
+        }
+    }
+
+    /// <summary>
+    ///     Classifies a connect/upgrade failure as transient (retryable). Retries when a
+    ///     <see cref="WebSocketException"/> wraps a transient <see cref="SocketException"/>
+    ///     (connection refused / timed out / host unreachable / DNS hiccup) or an inner
+    ///     <see cref="HttpRequestException"/> whose status is a retryable 5xx/429. Auth/permanent
+    ///     failures (401/403/400/404, DNS name-resolution failure) are NOT retried.
+    /// </summary>
+    private static bool IsRetryableConnect(Exception exception)
+    {
+        if (exception is not WebSocketException wsEx)
         {
-            ["Authorization"] = $"Bearer {token}",
-        };
-        foreach (var header in CopilotRequestHeaders.Build(_options, _session, Guid.NewGuid().ToString()))
-        {
-            headers[header.Key] = header.Value;
+            return false;
         }
 
-        var socket = _socketFactory();
-        await socket.ConnectAsync(_endpoint, headers, cancellationToken).ConfigureAwait(false);
-        _socket = socket;
-        return socket;
+        return wsEx.InnerException switch
+        {
+            SocketException socketEx => socketEx.SocketErrorCode
+                is SocketError.ConnectionRefused
+                    or SocketError.TimedOut
+                    or SocketError.HostUnreachable
+                    or SocketError.TryAgain
+                    or SocketError.HostNotFound,
+            HttpRequestException { StatusCode: { } status } => HttpRetryHelper.IsRetryableStatusCode(status),
+            _ => false,
+        };
+    }
+
+    private static async Task DisposeSocketSafelyAsync(ICopilotResponsesSocket socket)
+    {
+        try
+        {
+            await socket.DisposeAsync().ConfigureAwait(false);
+        }
+        catch (Exception ex) when (ex is WebSocketException or OperationCanceledException)
+        {
+            // Best-effort cleanup of a failed socket — never mask the original connect failure.
+        }
     }
 
     private static string? TryGetResponseId(JsonElement response)
@@ -245,8 +329,10 @@ public interface ICopilotResponsesSocket : IAsyncDisposable
     Task SendTextAsync(string text, CancellationToken cancellationToken);
 
     /// <summary>
-    ///     Receives the next complete text message (reassembling fragments). Returns null when the
-    ///     peer closes the connection.
+    ///     Receives the next complete text message (reassembling fragments). Returns null ONLY on a
+    ///     graceful peer Close frame; an abnormal/abrupt close surfaces as a thrown
+    ///     <see cref="WebSocketException"/> rather than null, so the caller can distinguish a clean
+    ///     close from a fault.
     /// </summary>
     Task<string?> ReceiveTextAsync(CancellationToken cancellationToken);
 }
@@ -292,15 +378,10 @@ public sealed class ClientWebSocketResponsesSocket : ICopilotResponsesSocket
 
         while (true)
         {
-            WebSocketReceiveResult result;
-            try
-            {
-                result = await _socket.ReceiveAsync(buffer, cancellationToken).ConfigureAwait(false);
-            }
-            catch (WebSocketException)
-            {
-                return null;
-            }
+            // A WebSocketException (abnormal/abrupt close, protocol error) is a FAULT, not a graceful
+            // close — let it propagate so the caller can surface it instead of silently truncating the
+            // turn. A null is reserved for a genuine peer Close frame below.
+            var result = await _socket.ReceiveAsync(buffer, cancellationToken).ConfigureAwait(false);
 
             if (result.MessageType == WebSocketMessageType.Close)
             {
--- a/src/LmCore/Http/HttpRetryHelper.cs
+++ b/src/LmCore/Http/HttpRetryHelper.cs
@@ -134,30 +134,34 @@ public static class HttpRetryHelper
                     continue;
                 }
 
-                // Not retryable or max retries exceeded, throw
+                // Not retryable or max retries exceeded. Capture the diagnostics BEFORE disposing, read
+                // the body in a try, dispose the failed response in the finally (so the connection is
+                // not leaked — the success path at :116 keeps the response alive for the caller, and the
+                // retryable branch at :132 already disposes), then throw with the captured values.
+                var statusCode = response.StatusCode;
+                var reasonPhrase = response.ReasonPhrase;
+                string responseBody;
                 try
                 {
-                    if (!response.IsSuccessStatusCode)
-                    {
-                        // Try to read the response body for better error information
-                        var responseBody = await response.Content.ReadAsStringAsync();
-                        var errorMessage =
-                            $"HTTP request failed with status {response.StatusCode} ({response.ReasonPhrase})";
-                        if (!string.IsNullOrWhiteSpace(responseBody))
-                        {
-                            errorMessage += $". Response body: {responseBody}";
-                        }
-
-                        throw new HttpRequestException(errorMessage, null, response.StatusCode);
-                    }
+                    responseBody = await response.Content.ReadAsStringAsync();
                 }
                 catch (Exception ex) when (ex is not HttpRequestException)
                 {
-                    // If reading response body fails, fall back to standard behavior
-                    _ = response.EnsureSuccessStatusCode();
+                    // If reading the body fails, fall back to status-only diagnostics.
+                    responseBody = string.Empty;
+                }
+                finally
+                {
+                    response.Dispose();
                 }
 
-                return default!; // This line should never be reached
+                var errorMessage = $"HTTP request failed with status {statusCode} ({reasonPhrase})";
+                if (!string.IsNullOrWhiteSpace(responseBody))
+                {
+                    errorMessage += $". Response body: {responseBody}";
+                }
+
+                throw new HttpRequestException(errorMessage, null, statusCode);
             }
             catch (HttpRequestException ex) when (attempt < options.MaxRetries && IsRetryableError(ex))
             {
--- a/src/LmCore/Http/HttpRetryHelper.cs
+++ b/src/LmCore/Http/HttpRetryHelper.cs
@@ -145,11 +145,13 @@ public static class HttpRetryHelper
                 {
                     responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
                 }
-                catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
+                catch (Exception ex) when (ex is not OperationCanceledException)
                 {
-                    // If reading the body fails, fall back to status-only diagnostics. Deliberate
-                    // cancellation is NOT swallowed here — it must surface as cancellation, not as a
-                    // synthesized HttpRequestException transport failure.
+                    // A failed body read is diagnostic-only: fall back to status-only diagnostics and throw
+                    // with the ALREADY-CAPTURED original status below. This deliberately also swallows an
+                    // HttpRequestException raised by the body read itself — otherwise it would escape to the
+                    // outer retry catch and could retry a non-retryable status (or mask the real status).
+                    // Deliberate cancellation is NOT swallowed — it must surface as cancellation.
                     responseBody = string.Empty;
                 }
                 finally
--- a/src/LmCore/Http/HttpRetryHelper.cs
+++ b/src/LmCore/Http/HttpRetryHelper.cs
@@ -143,11 +143,13 @@ public static class HttpRetryHelper
                 string responseBody;
                 try
                 {
-                    responseBody = await response.Content.ReadAsStringAsync();
+                    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
                 }
-                catch (Exception ex) when (ex is not HttpRequestException)
+                catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
                 {
-                    // If reading the body fails, fall back to status-only diagnostics.
+                    // If reading the body fails, fall back to status-only diagnostics. Deliberate
+                    // cancellation is NOT swallowed here — it must surface as cancellation, not as a
+                    // synthesized HttpRequestException transport failure.
                     responseBody = string.Empty;
                 }
                 finally
--- a/src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
+++ b/src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
@@ -94,9 +94,12 @@ public sealed class OpenAiResponsesClient : IOpenAiResponsesClient
         // the processor must NOT call EnsureSuccessStatusCode — it hands back the still-open response
         // and its stream so enumeration happens OUTSIDE the retry scope.
         var setup = await HttpRetryHelper.ExecuteHttpWithRetryAsync(
-            () =>
+            async () =>
             {
-                var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
+                // `using` so the per-attempt request + JsonContent are disposed once the response
+                // headers are read; with ResponseHeadersRead the response stream is owned by the
+                // returned HttpResponseMessage, so disposing the request here does not affect it.
+                using var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
                 {
                     Content = JsonContent.Create(streamingRequest, options: s_serializerOptions),
                 };
@@ -109,12 +112,24 @@ public sealed class OpenAiResponsesClient : IOpenAiResponsesClient
                     streamingRequest.Input.Count
                 );
 
-                return _http.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+                return await _http
+                    .SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                    .ConfigureAwait(false);
             },
             async response =>
             {
-                var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
-                return (Response: response, Stream: stream);
+                // If acquiring the stream fails/cancels, the tuple is never returned and the iterator's
+                // finally never runs — dispose the response here so the connection is not pinned.
+                try
+                {
+                    var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+                    return (Response: response, Stream: stream);
+                }
+                catch
+                {
+                    response.Dispose();
+                    throw;
+                }
             },
             _logger,
             _retryOptions,
--- a/src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
+++ b/src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
@@ -3,6 +3,7 @@ using System.Net.Http.Json;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
+using AchieveAi.LmDotnetTools.LmCore.Http;
 using AchieveAi.LmDotnetTools.OpenAiResponsesProvider.Models;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
@@ -32,19 +33,22 @@ public sealed class OpenAiResponsesClient : IOpenAiResponsesClient
     private readonly HttpClient _http;
     private readonly bool _disposeClient;
     private readonly string _responsesPath;
-    private readonly ILogger<OpenAiResponsesClient> _logger;
+    private readonly ILogger _logger;
+    private readonly RetryOptions _retryOptions;
 
     public OpenAiResponsesClient(
         HttpClient httpClient,
         bool disposeClient = false,
-        ILogger<OpenAiResponsesClient>? logger = null,
-        string? responsesPath = null
+        ILogger? logger = null,
+        string? responsesPath = null,
+        RetryOptions? retryOptions = null
     )
     {
         _http = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
         _disposeClient = disposeClient;
         _responsesPath = string.IsNullOrWhiteSpace(responsesPath) ? DefaultResponsesPath : responsesPath!;
-        _logger = logger ?? NullLogger<OpenAiResponsesClient>.Instance;
+        _logger = logger ?? NullLogger.Instance;
+        _retryOptions = retryOptions ?? RetryOptions.Default;
     }
 
     /// <summary>
@@ -55,8 +59,9 @@ public sealed class OpenAiResponsesClient : IOpenAiResponsesClient
     public static OpenAiResponsesClient Create(
         Uri baseAddress,
         string? apiKey = null,
-        ILogger<OpenAiResponsesClient>? logger = null,
-        string? responsesPath = null
+        ILogger? logger = null,
+        string? responsesPath = null,
+        RetryOptions? retryOptions = null
     )
     {
         ArgumentNullException.ThrowIfNull(baseAddress);
@@ -66,7 +71,7 @@ public sealed class OpenAiResponsesClient : IOpenAiResponsesClient
             client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", apiKey);
         }
 
-        return new OpenAiResponsesClient(client, disposeClient: true, logger, responsesPath);
+        return new OpenAiResponsesClient(client, disposeClient: true, logger, responsesPath, retryOptions);
     }
 
     /// <inheritdoc />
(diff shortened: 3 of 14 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR !120 (fixes #114)

## Summary
Solid implementation of transient-fault retry for both the SSE and WebSocket Copilot transports. The retry is correctly scoped to the pre-stream/pre-turn phase (idempotent), disposal is tight, and the test coverage is comprehensive.

## Findings

### :yellow_circle: MEDIUM - Duplicate TrackingHttpResponseMessage class
The test helper is defined identically in both:
- `tests/OpenAiResponsesProvider.Tests/RetryTestInfra.cs`
- `tests/LmCore.Tests/Http/HttpRetryHelperDisposalTests.cs`
Consider extracting to `LmTestUtils` so both test projects reference a single definition — avoids drift and aligns with the project's existing shared infra pattern.

### :blue_circle: LOW - NetworkUnreachable not in IsRetryableConnect
`SocketError.NetworkUnreachable` is arguably as transient as `HostUnreachable` (which IS retried). If intentionally excluded (e.g. because network-layer recovery requires user/infra action), a brief comment explaining the rationale would help future readers. If unintentional, it's a one-liner to add.

## Positives
- Fresh `HttpRequestMessage` / `ICopilotResponsesSocket` per retry attempt (correct — requests can't be re-sent)
- Stale socket disposed on reconnect (commit 6 fix)
- Mid-turn peer close surfaced as `IOException` instead of silent truncation
- `HostNotFound` correctly excluded from retry (authoritative DNS failure vs transient `TryAgain`)
- Response disposal leak in `HttpRetryHelper` fixed with captured-status-then-dispose pattern
- `RetryOptions.FastForTests` keeps test execution fast without flaky timers

:white_check_mark: Build passes, all 19 new retry tests pass.

**Verdict: APPROVE**

tests/LmCore.Tests/Http/HttpRetryHelperDisposalTests.cs
```diff
@@ -0,0 +1,82 @@
...
+    )
+    {
+        var response = new TrackingHttpResponseMessage(status) { Content = new StringContent($"error {(int)status}") };
+        sink.Add(response);
+        return Task.FromResult<HttpResponseMessage>(response);
+    }
+}
+
+/// <summary>An <see cref="HttpResponseMessage"/> that records whether it has been disposed.</summary>
+internal sealed class TrackingHttpResponseMessage : HttpResponseMessage
```
gautam-achieveai: **[Gautam's review bot]** :yellow_circle: This `TrackingHttpResponseMessage` class is duplicated verbatim in `tests/OpenAiResponsesProvider.Tests/RetryTestInfra.cs`. Consider extracting to `LmTestUtils` so both test projects reference a single definition — avoids drift and aligns with the project's existing shared infra pattern.

src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs
```diff
@@ -150,21 +162,105 @@ private async Task<ICopilotResponsesSocket> EnsureConnectedAsync(CancellationTok
...
+        return wsEx.InnerException switch
+        {
+            // TryAgain (WSATRY_AGAIN) is the non-authoritative "DNS hiccup, retry" case. HostNotFound
+            // (WSAHOST_NOT_FOUND) is an authoritative "no such host" name-resolution failure and is NOT
+            // retried — see the doc comment above.
+            SocketException socketEx => socketEx.SocketErrorCode
+                is SocketError.ConnectionRefused
+                    or SocketError.TimedOut
+                    or SocketError.HostUnreachable
+                    or SocketError.TryAgain,
```
gautam-achieveai, line 250: **[Gautam's review bot]** :blue_circle: Minor: `SocketError.NetworkUnreachable` is not included here but is arguably as transient as `HostUnreachable`. If intentionally excluded (e.g. because the network layer won't recover without user/infra action), a brief comment explaining the rationale would help future readers.

gautam-achieveai:
[gautam-achieveai's bot] Implementation review for #114 — **REQUEST CHANGES**

Thanks for the implementation. The scope matches the approved plan v3 — I did **not** find over-engineering or unrelated scope creep. The remaining issues are localized to resource/cancellation behavior and a couple of coverage gaps in the new retry paths.

## Blocking revisions

### 1. Dispose each per-attempt `HttpRequestMessage` / `JsonContent`

- File: `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:96-113`

The retry operation now creates a fresh `HttpRequestMessage` and `JsonContent` per attempt, but the request is never disposed. The previous implementation used `using var httpRequest`; `HttpClient.SendAsync` does not dispose the request/content for the caller.

Suggested shape:

```csharp
async () =>
{
    using var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
    {
        Content = JsonContent.Create(streamingRequest, options: s_serializerOptions),
    };

    httpRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));

    return await _http
        .SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
        .ConfigureAwait(false);
}
```

This keeps the returned `HttpResponseMessage` alive while cleaning up per-attempt request resources.

### 2. Dispose the successful response if stream acquisition fails

- File: `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:114-117`

`responseProcessor` reads the successful response stream before returning `(Response, Stream)`. If `ReadAsStreamAsync(cancellationToken)` throws or is canceled, the tuple is never assigned and the iterator `finally` at lines 131-135 never runs. With `ResponseHeadersRead`, that can leave the response undisposed and pin the connection.

Suggested shape:

```csharp
async response =>
{
    try
    {
        var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
        return (Response: response, Stream: stream);
    }
    catch
    {
        response.Dispose();
        throw;
    }
}
```

### 3. Do not swallow cancellation while reading final error body

- File: `src/LmCore/Http/HttpRetryHelper.cs:143-151`

The final failed-response diagnostic path catches almost everything while reading the response body:

```csharp
catch (Exception ex) when (ex is not HttpRequestException)
{
    responseBody = string.Empty;
}
```

That also catches `OperationCanceledException` / `TaskCanceledException`, then the helper later throws `HttpRequestException`. Cancellation should remain cancellation.

Suggested change:

```csharp
catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
{
    responseBody = string.Empty;
}
```

Also consider passing the provided cancellation token into `ReadAsStringAsync(cancellationToken)`.

### 4. Add direct WebSocket 502 upgrade retry tests

- Files:
  - `src/GithubCopilotProvider/Agents/CopilotResponsesWebSocketClient.cs:239-250`
  - `tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesWebSocketClientRetryTests.cs:105-223`

The implementation classifies this shape as retryable:

```csharp
HttpRequestException { StatusCode: { } status } => HttpRetryHelper.IsRetryableStatusCode(status),
```

But tests currently cover retryable socket errors and non-retryable HTTP 401, not the issue's core WebSocket gateway case: `WebSocketException` wrapping `HttpRequestException` with `StatusCode == BadGateway`.

Please add tests for:

1. `WebSocketException(inner: HttpRequestException(..., HttpStatusCode.BadGateway))` fails once, then succeeds; assert two socket attempts and no `response.create` sent before successful connect.
2. Persistent `BadGateway` upgrade failure exhausts `MaxRetries + 1` attempts and throws.

### 5. Strengthen Copilot SSE factory wiring coverage

- Files:
  - `src/GithubCopilotProvider/Agents/CopilotResponsesAgentFactory.cs:61-87`
  - `tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesSseRetryTests.cs:54-108`

The substantive Copilot SSE retry/logging test manually constructs `CopilotHttpClientFactory` + `OpenAiResponsesClient`. The public factory test only checks that `Create(... retryOptions)` returns non-null. If `CreateSseClient` stopped forwarding `retryOptions` or `logger`, the meaningful retry/logging test would still pass.

Please add a factory-level seam/integration test that exercises `CopilotResponsesAgentFactory.Create(... transport: Sse, logger, retryOptions)` through the returned path, or otherwise explicitly verify the private wiring via a supported injectable transport/handler.

## Notes

- I am intentionally **not** raising the public API binary/backward compatibility concern, per maintainer clarification that backward compatibility is not a requirement here.
- Over-engineering review found the implementation well-scoped to #114 and approved plan v3.
- I could not run local `dotnet test` in the sandbox because the .NET SDK is not installed (`dotnet: not found`).

src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
```diff
@@ -82,57 +87,51 @@ public async IAsyncEnumerable<ResponseEvent> StreamResponseAsync(
...
+        // Send (and pre-stream retry transient 502/5xx/429 + connection faults) via the shared
+        // HttpRetryHelper. The HttpRequestMessage + JsonContent are rebuilt fresh on every attempt
+        // because a single request message cannot be re-sent (and a fresh attempt also re-stamps the
+        // Copilot auth token + x-interaction-id). The helper performs the status check internally, so
+        // the processor must NOT call EnsureSuccessStatusCode — it hands back the still-open response
+        // and its stream so enumeration happens OUTSIDE the retry scope.
+        var setup = await HttpRetryHelper.ExecuteHttpWithRetryAsync(
+            () =>
+            {
+                var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Resource lifetime)

This retry operation creates a fresh `HttpRequestMessage` + `JsonContent` per attempt, but the request is never disposed. The previous implementation used `using var httpRequest`; `HttpClient.SendAsync` does not dispose request/content for the caller, so retry scenarios can retain request content buffers/resources longer than necessary.

**Suggestion:** make the operation lambda async and wrap the request in `using var` while returning the response:

```csharp
async () =>
{
    using var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
    {
        Content = JsonContent.Create(streamingRequest, options: s_serializerOptions),
    };
    httpRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
    return await _http.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
}
```

src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
```diff
@@ -82,57 +87,51 @@ public async IAsyncEnumerable<ResponseEvent> StreamResponseAsync(
...
+                    _responsesPath,
+                    streamingRequest.Model,
+                    streamingRequest.Input.Count
+                );
+
+                return _http.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            },
+            async response =>
+            {
+                var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Resource lifetime / exception handling)

If `ReadAsStreamAsync(cancellationToken)` throws or is canceled here, the `(Response, Stream)` tuple is never assigned and the iterator `finally` below never runs. With `ResponseHeadersRead`, that can leave the successful `HttpResponseMessage` undisposed and pin the connection.

**Suggestion:** dispose the response on stream-acquisition failure:

```csharp
async response =>
{
    try
    {
        var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
        return (Response: response, Stream: stream);
    }
    catch
    {
        response.Dispose();
        throw;
    }
}
```

src/LmCore/Http/HttpRetryHelper.cs
```diff
@@ -134,30 +134,34 @@ public static async Task<T> ExecuteHttpWithRetryAsync<T>(
...
-                        if (!string.IsNullOrWhiteSpace(responseBody))
-                        {
-                            errorMessage += $". Response body: {responseBody}";
-                        }
-
-                        throw new HttpRequestException(errorMessage, null, response.StatusCode);
-                    }
+                    responseBody = await response.Content.ReadAsStringAsync();
                 }
                 catch (Exception ex) when (ex is not HttpRequestException)
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Cancellation semantics)

This filter catches `OperationCanceledException` / `TaskCanceledException` while reading the final error body, then the helper later throws `HttpRequestException`. Cancellation should remain cancellation so callers do not treat deliberate cancellation as a transport failure.

**Suggestion:** exclude cancellation and consider passing the provided token into the body read:

```csharp
catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
{
    responseBody = string.Empty;
}
```

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesWebSocketClientRetryTests.cs
```diff
@@ -0,0 +1,375 @@
...
+
+        public ValueTask DisposeAsync()
+        {
+            DisposeCount++;
+            IsConnected = false;
+            return ValueTask.CompletedTask;
+        }
+    }
+
+    private static WebSocketException RetryableConnect() =>
```
gautam-achieveai, line 105: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test coverage)

The implementation retries `WebSocketException -> HttpRequestException(StatusCode=retryable)`, but this test helper only models retryable socket errors. The issue's core WebSocket gateway case is a 502 during upgrade/connect, so the exact `HttpStatusCode.BadGateway` classification path should be pinned.

**Suggestion:** add tests where `ConnectError` is `new WebSocketException(…, new HttpRequestException(…, null, HttpStatusCode.BadGateway))`: one transient-then-success case and one persistent-exhaustion case. Assert no `response.create` is sent before successful connect.

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesSseRetryTests.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        // The 502 was retried and the SSE stream was read end-to-end.
+        events.Should().NotBeEmpty();
+        events[^1].Type.Should().Be(ResponseEventTypes.ResponseCompleted);
+
+        // The forwarded logger captured the retry warning (it only fires on a retry).
+        logger.Entries.Should().Contain(e => e.Level == LogLevel.Warning);
+    }
+
+    [Fact]
```
gautam-achieveai, line 80: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test coverage / factory wiring)

The substantive Copilot SSE retry/logging test manually constructs `CopilotHttpClientFactory` + `OpenAiResponsesClient`; this public factory test only proves `Create(... retryOptions)` returns non-null. If `CreateSseClient` stopped forwarding `retryOptions` or `logger`, the meaningful retry/logging test would still pass.

**Suggestion:** add a factory-level seam/integration test that exercises `CopilotResponsesAgentFactory.Create(... transport: Sse, logger, retryOptions)` through the returned path, or otherwise directly verifies the private wiring via a supported injectable transport/handler.

gautam-achieveai (commented):
[gautam-achieveai's bot] Inline implementation review for #114. Backward/binary compatibility concern intentionally omitted per maintainer guidance. GitHub does not allow this bot to request changes on its own PR, so posting as review comments; please treat the blocker-tagged items below as requested changes.

src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Resource lifetime)

This retry operation creates a fresh `HttpRequestMessage` + `JsonContent` per attempt, but the request is never disposed. The previous implementation used `using var httpRequest`; `HttpClient.SendAsync` does not dispose request/content for the caller, so retry scenarios can retain request content buffers/resources longer than necessary.

**Suggestion:** make the operation lambda async and wrap the request in `using var` while returning the response:

```csharp
async () =>
{
    using var httpRequest = new HttpRequestMessage(HttpMethod.Post, _responsesPath)
    {
        Content = JsonContent.Create(streamingRequest, options: s_serializerOptions),
    };
    httpRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
    return await _http.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
}
```

src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Resource lifetime / exception handling)

If `ReadAsStreamAsync(cancellationToken)` throws or is canceled here, the `(Response, Stream)` tuple is never assigned and the iterator `finally` below never runs. With `ResponseHeadersRead`, that can leave the successful `HttpResponseMessage` undisposed and pin the connection.

**Suggestion:** dispose the response on stream-acquisition failure:

```csharp
async response =>
{
    try
    {
        var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
        return (Response: response, Stream: stream);
    }
    catch
    {
        response.Dispose();
        throw;
    }
}
```

src/LmCore/Http/HttpRetryHelper.cs
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Cancellation semantics)

This filter catches `OperationCanceledException` / `TaskCanceledException` while reading the final error body, then the helper later throws `HttpRequestException`. Cancellation should remain cancellation so callers do not treat deliberate cancellation as a transport failure.

**Suggestion:** exclude cancellation and consider passing the provided token into the body read:

```csharp
catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
{
    responseBody = string.Empty;
}
```

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesWebSocketClientRetryTests.cs
gautam-achieveai, line 105: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test coverage)

The implementation retries `WebSocketException -> HttpRequestException(StatusCode=retryable)`, but this test helper only models retryable socket errors. The issue's core WebSocket gateway case is a 502 during upgrade/connect, so the exact `HttpStatusCode.BadGateway` classification path should be pinned.

**Suggestion:** add tests where `ConnectError` is `new WebSocketException(…, new HttpRequestException(…, null, HttpStatusCode.BadGateway))`: one transient-then-success case and one persistent-exhaustion case. Assert no `response.create` is sent before successful connect.

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesSseRetryTests.cs
gautam-achieveai, line 80: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test coverage / factory wiring)

The substantive Copilot SSE retry/logging test manually constructs `CopilotHttpClientFactory` + `OpenAiResponsesClient`; this public factory test only proves `Create(... retryOptions)` returns non-null. If `CreateSseClient` stopped forwarding `retryOptions` or `logger`, the meaningful retry/logging test would still pass.

**Suggestion:** add a factory-level seam/integration test that exercises `CopilotResponsesAgentFactory.Create(... transport: Sse, logger, retryOptions)` through the returned path, or otherwise directly verifies the private wiring via a supported injectable transport/handler.

gautam-achieveai (commented):
[gautam-achieveai's bot] Inline implementation review for #114. Backward/binary compatibility concern intentionally omitted per maintainer guidance. GitHub does not allow this bot to request changes on its own PR, so posting as review comments; please treat the blocker-tagged items below as requested changes.

gautam-achieveai:
[gautam-achieveai's bot] Thanks for the review — all blocking items and both advisory notes addressed in `ccfa8de`. Build green (0 errors, 6 pre-existing `NU1902` warnings, none new); affected suites pass: OpenAiResponsesProvider **77**, GithubCopilotProvider **33**, LmCore **766**, AnthropicProvider **175**, LmTestUtils **91**.

| # | Finding | Resolution (`ccfa8de`) |
|---|---|---|
| 1 | Per-attempt `HttpRequestMessage`/`JsonContent` not disposed | `OpenAiResponsesClient` retry op-lambda is now `async` with `using var httpRequest` (returns the response after headers are read). |
| 2 | Successful response leaked if stream acquisition fails/cancels | `responseProcessor` now wraps `ReadAsStreamAsync` in `try/catch { response.Dispose(); throw; }`. + regression test (`...disposes_response_when_stream_acquisition_fails`). |
| 3 | Cancellation swallowed while reading final error body | Filter is now `ex is not HttpRequestException and not OperationCanceledException`; token passed to `ReadAsStringAsync`. + regression test (`...cancellation_reading_error_body_surfaces_as_cancellation`). |
| 4 | WS 502-during-upgrade path not pinned | Added `Connect_retries_after_transient_bad_gateway_upgrade` and `Connect_persistent_bad_gateway_upgrade_throws_after_bounded_attempts` (assert no `response.create` before a successful connect). |
| 5 | Factory SSE wiring not truly exercised | Made `CreateSseClient` `internal` with an injectable `innerHandler` (`InternalsVisibleTo` the test project) and added two tests that drive the **real** path — one asserts retry + forwarded-logger warning, one uses `MaxRetries=0` to pin `retryOptions` forwarding (would pass on a Default fallback otherwise). |
| 6 | Duplicate `TrackingHttpResponseMessage` | Extracted `TrackingHttpResponseMessage`/`TrackingStream` to `LmTestUtils` (`src/LmTestUtils/Http/`); both test projects reference the single definition. |
| 7 | `NetworkUnreachable` missing from `IsRetryableConnect` | Added `SocketError.NetworkUnreachable` to the transient set (alongside `HostUnreachable`); `HostNotFound` remains excluded (authoritative DNS). |

Re-review welcome.

gautam-achieveai:
[gautam-achieveai's bot] Re-review of #120 at `ccfa8def` — **APPROVE WITH COMMENTS**

Thanks for addressing the prior blockers. I re-reviewed the latest head with the requested over-engineering, blind-spot, architecture, code-simplicity, performance/resource, exception-handling, and test-coverage lenses. Architecture/scope look sound: retry remains scoped to pre-stream/pre-turn work, WebSocket duplicate-send risk is avoided, resource ownership is much tighter, and the new coverage pins the important 502/upgrade/disposal/cancellation paths.

I found no new blockers. A few follow-ups are still worth tightening:

## Medium — Preserve original HTTP status when reading the error body fails

- File: `src/LmCore/Http/HttpRetryHelper.cs:144-148`

The final failed-response branch now captures status/reason and disposes the response, which is good. However, the body-read fallback still excludes `HttpRequestException`:

```csharp
try
{
    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
}
catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
{
    responseBody = string.Empty;
}
```

If `ReadAsStringAsync` itself throws `HttpRequestException`, that exception bypasses the fallback. In this helper it can then be caught by the outer retry catch as a transport failure, which may retry a non-retryable status based on the body-read exception rather than the response status already captured. It can also lose the original status diagnostics on the final failure.

Suggested fix: keep cancellation propagating, but treat all other body-read failures as diagnostic-only failures:

```csharp
try
{
    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
}
{
    responseBody = string.Empty;
}
finally
{
    response.Dispose();
}
```

## Medium — Avoid `StreamReader.EndOfStream` on live SSE streams

- File: `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:167`

`ReadSseAsync` still drives the loop with:

```csharp
while (!reader.EndOfStream)
```

For a network/SSE stream that is idle but still open, `EndOfStream` can synchronously read to determine EOF before the code reaches the async/cancellable `ReadLineAsync(cancellationToken)`. Since this PR is improving streaming retry/cancellation behavior in this client, it is worth making the read loop fully async/cancellation-driven:

```csharp
while (true)
{
    cancellationToken.ThrowIfCancellationRequested();

    var line = await reader.ReadLineAsync(cancellationToken).ConfigureAwait(false);
    if (line is null)
    {
        break;
    }

    // existing line handling...
}
```

## Low — Consider sharing the small test logger helper

- Files:
  - `tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesSseRetryTests.cs:29`
  - `tests/OpenAiResponsesProvider.Tests/RetryTestInfra.cs:92`

The retry tests now have duplicate minimal `ListLogger` implementations. This is small and non-blocking, but moving the logger to shared test utilities (similar to the new shared tracking response/stream helper) would avoid drift.

## Low — Optional simplification: parameterize repeated WebSocket retry cases

- File: `tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesWebSocketClientRetryTests.cs:145-254`

The retry tests are valuable, but the first four tests repeat the same setup/assertion shape for `RetryableConnect()` and `BadGatewayUpgrade()` plus persistent variants. A `[Theory]`/`MemberData` or a couple of helper methods could keep the same coverage with less duplication.

## Verification notes

- Local `dotnet build/test` could not be run in this sandbox (`dotnet: not found`).
- Review artifacts were generated from latest PR head `ccfa8def09f3aaff8cc6978256ee8ff035f44789` in `/workspace/LmDotNetTools/issue120-impl-review-latest`.
- Agent lenses: architecture found no blockers; over-engineering/blind-spot/test agents did not return actionable latest-head findings; code-simplifier/performance/exception lenses informed the comments above.

src/LmCore/Http/HttpRetryHelper.cs
```diff
@@ -134,30 +134,36 @@ public static async Task<T> ExecuteHttpWithRetryAsync<T>(
...
-                        if (!string.IsNullOrWhiteSpace(responseBody))
-                        {
-                            errorMessage += $". Response body: {responseBody}";
-                        }
-
-                        throw new HttpRequestException(errorMessage, null, response.StatusCode);
-                    }
+                    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception ex) when (ex is not HttpRequestException and not OperationCanceledException)
```
gautam-achieveai: [gautam-achieveai's bot] **MEDIUM** (Exception handling / diagnostics)

The body-read fallback still excludes `HttpRequestException`. If `ReadAsStringAsync` itself throws `HttpRequestException`, this bypasses the fallback and can be caught by the outer retry catch as a transport failure, which may retry a non-retryable response based on the body-read exception rather than the HTTP status already captured. It can also lose the original status diagnostics on the final failure.

**Suggestion:** keep cancellation propagating, but treat all other body-read failures as diagnostic-only failures:

```csharp
try
{
    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
}
{
    responseBody = string.Empty;
}
finally
{
    response.Dispose();
}
```

gautam-achieveai:
[gautam-achieveai's bot] **MEDIUM** (Performance / async cancellation)

`StreamReader.EndOfStream` can synchronously read to determine EOF. On an idle-but-open SSE/network stream, this can block before the code reaches the async/cancellable `ReadLineAsync(cancellationToken)`, which weakens cancellation behavior in the streaming path this PR is hardening.

**Suggestion:** drive the loop from the async read instead:

```csharp
while (true)
{
    cancellationToken.ThrowIfCancellationRequested();

    var line = await reader.ReadLineAsync(cancellationToken).ConfigureAwait(false);
    if (line is null)
    {
        break;
    }

    // existing line handling...
}
```

_Fallback: intended inline location `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:167`._

src/LmCore/Http/HttpRetryHelper.cs
gautam-achieveai: [gautam-achieveai's bot] **MEDIUM** (Exception handling / diagnostics)

The body-read fallback still excludes `HttpRequestException`. If `ReadAsStringAsync` itself throws `HttpRequestException`, this bypasses the fallback and can be caught by the outer retry catch as a transport failure, which may retry a non-retryable response based on the body-read exception rather than the HTTP status already captured. It can also lose the original status diagnostics on the final failure.

**Suggestion:** keep cancellation propagating, but treat all other body-read failures as diagnostic-only failures:

```csharp
try
{
    responseBody = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
}
{
    responseBody = string.Empty;
}
finally
{
    response.Dispose();
}
```

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesSseRetryTests.cs
```diff
@@ -0,0 +1,165 @@
...
+///     handler to prove it forwards both <c>retryOptions</c> and <c>logger</c> into the client.
+/// </summary>
+public sealed class CopilotResponsesSseRetryTests
+{
+    private sealed class StubTokenProvider : ICopilotTokenProvider
+    {
+        public Task<string> GetTokenAsync(CancellationToken cancellationToken = default) => Task.FromResult("gho_test");
+    }
+
+    private sealed class ListLogger : ILogger
```
gautam-achieveai: [gautam-achieveai's bot] **LOW** (Test maintainability)

This PR now has another minimal `ListLogger` while `tests/OpenAiResponsesProvider.Tests/RetryTestInfra.cs` also defines one for the retry tests. The helper is small, but duplicating it in touched retry-test areas invites drift.

**Suggestion:** move the logger to a shared test utility (similar to the new shared tracking response/stream helper) and reuse it from both retry test projects.

gautam-achieveai:
[gautam-achieveai's bot] **MEDIUM** (Performance / async cancellation)

`StreamReader.EndOfStream` can synchronously read to determine EOF. On an idle-but-open SSE/network stream, this can block before the code reaches the async/cancellable `ReadLineAsync(cancellationToken)`, which weakens cancellation behavior in the streaming path this PR is hardening.

**Suggestion:** drive the loop from the async read instead:

```csharp
while (true)
{
    cancellationToken.ThrowIfCancellationRequested();

    var line = await reader.ReadLineAsync(cancellationToken).ConfigureAwait(false);
    if (line is null)
    {
        break;
    }

    // existing line handling...
}
```

_Fallback: intended inline location `src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs:167`._

tests/GithubCopilotProvider.Tests/Agents/CopilotResponsesWebSocketClientRetryTests.cs
```diff
@@ -0,0 +1,431 @@
...
+        );
+
+    private static ResponseCreateRequest Request() =>
+        new()
+        {
+            Model = "gpt-5.5",
+            Input = [new ResponseInputItem { Role = "user", Content = [new ResponseInputContent { Text = "hi" }] }],
+        };
+
+    [Fact]
```
gautam-achieveai: [gautam-achieveai's bot] **LOW** (Test maintainability)

The first four connect-retry tests repeat the same setup/assertion shape for `RetryableConnect()` and `BadGatewayUpgrade()` plus persistent variants. The coverage is valuable, but the duplication makes this 431-line test file harder to scan.

**Suggestion:** parameterize the retryable exception cases with `[Theory]`/`MemberData`, or extract small helpers for `first failure then success`, `persistent failure`, and `failed sockets disposed without send` assertions.

gautam-achieveai:
[gautam-achieveai's bot] Thanks for the re-review — all four follow-ups addressed in `7131242`. Build green (0 errors, 6 pre-existing `NU1902` warnings); suites pass: OpenAiResponsesProvider **77**, GithubCopilotProvider **33**, LmCore **767**, LmTestUtils **91**.

| Severity | Finding | Resolution (`7131242`) |
|---|---|---|
| Medium | `HttpRetryHelper`: a body-read `HttpRequestException` bypassed the fallback and could retry a non-retryable status / mask the original status | Catch filter is now `ex is not OperationCanceledException` — a failed final-error-body read is diagnostic-only for **all** non-cancellation exceptions, so it can't escape to the outer retry catch; the already-captured original status is thrown. Cancellation still propagates. + regression test (`..._body_read_failure_keeps_status_and_does_not_retry`) asserting a 400 with a retryable-looking body-read exception stays 1 attempt / `BadRequest`. |
| Medium | `ReadSseAsync` gated on `StreamReader.EndOfStream` (synchronous, can block on an idle-but-open SSE stream) | Loop is now `while (true)` driven by `await ReadLineAsync(ct)` (null = EOF) — fully async/cancellation-driven. |
| Low | Duplicate `ListLogger` across two test projects | Extracted to `src/LmTestUtils/Logging/ListLogger.cs`; both projects reference it (mirrors the shared tracking-response helper). |
| Low | Repetitive WebSocket connect-retry tests | Collapsed the four near-duplicate transient/persistent tests into two `[Theory]` cases over `SocketConnectionRefused` + `BadGatewayUpgrade` (same coverage, less duplication). |

Re-review welcome.

src/OpenAiResponsesProvider/Agents/OpenAiResponsesClient.cs
```diff
@@ -150,7 +164,10 @@ [EnumeratorCancellation] CancellationToken cancellationToken
         using var reader = new StreamReader(stream, Encoding.UTF8, detectEncodingFromByteOrderMarks: false, 1024, leaveOpen: true);
         var dataBuffer = new StringBuilder();
 
-        while (!reader.EndOfStream)
+        // Fully async/cancellation-driven: do NOT gate on StreamReader.EndOfStream, which performs a
```
gautam-achieveai, line 167: [gautam-achieveai's bot] **MEDIUM** (Test coverage)

This fixes the `EndOfStream` issue by making the SSE loop async/cancellation-driven, but I don't see a regression test for the failure mode. The existing `OpenAiResponsesClientSseEdgeTests` use finite `MemoryStream` inputs, so they would also pass with the old `while (!reader.EndOfStream)` implementation and don't prove that an idle-but-open SSE stream can be cancelled promptly.

**Suggestion:** add a targeted test using a custom stream or pipe that remains open without EOF and either fails on synchronous `Read` or verifies cancellation completes within a short timeout while enumerating `ReadSseAsync`. The old implementation should hang/fail; the new `ReadLineAsync(cancellationToken)` path should throw `OperationCanceledException` promptly.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #120 at `7131242`

Thanks for the update. The prior exception-handling issue in `HttpRetryHelper` is fixed, the SSE loop no longer uses `EndOfStream`, the shared `ListLogger` removes the duplicate helper, and the WebSocket retry tests are simpler without losing the key 502/transport coverage.

I found one remaining non-blocking follow-up: the SSE `EndOfStream` cancellation fix should have a regression test using an open non-EOF stream/pipe, because the current finite `MemoryStream` edge tests would not fail under the old implementation. Local test execution was attempted but the sandbox still has no `dotnet` installed.

**Verdict:** APPROVE WITH COMMENTS — no blockers.

gautam-achieveai:
[gautam-achieveai's bot] Thanks — addressed the remaining follow-up in `217c6cd`.

**Medium (test coverage) — SSE cancellation regression test.** Added `ReadSseAsync_cancels_promptly_on_idle_open_stream`, which enumerates `ReadSseAsync` over a `NeverEndingStream` that never yields data or EOF (and whose **synchronous** `Read` throws). With `CancelAfter(100ms)` the new `ReadLineAsync(cancellationToken)` loop throws `OperationCanceledException` promptly (test completes in ~116ms). Under the old `while (!reader.EndOfStream)` implementation the synchronous EOF probe would hit the forbidden synchronous `Read` (and otherwise block), so the test fails on the old code — a genuine RED→GREEN.

Build green (0 errors, 6 pre-existing `NU1902` warnings); `OpenAiResponsesProvider.Tests` now **78** passing. Re-review welcome.

PR #121, opened 2026-06-30
Code-Review Daemon sample: unattended GitHub + ADO PR reviewer (#118)

### LINKED ISSUE #118: Code-Review Daemon sample: unattended GitHub + ADO PR reviewer
## Summary

Add a new sample app under `samples/` — a **Code-Review Daemon**: a long-running,
unattended service that reviews pull requests on **both GitHub and Azure DevOps**
using LLM agents. It reuses the agent, sandbox, and OAuth/token-injection
machinery currently living inside `LmStreaming.Sample`, which is first **refactored
into a shared `src/` library** so both apps consume it.

The daemon **polls** PR APIs for new/updated PRs, runs an in-process
`MultiTurnAgentLoop` per PR to decide and execute work (Review, Re-Review, Extract
Knowledge, Judge), and **uses Sandboxes for effectively all filesystem/git work** —
including the per-PR repo setup and the agent's bash/file commands. Git/`gh`/`az`
calls from the sandbox are authorized via the existing **auth webhook + token
injection**; the embedded ASP.NET server exists **only** to serve that webhook (no
inbound PR-event webhooks). A **SQLite database** on the daemon host is the single
source of truth for PR state and drives orchestration.

> **Delivery constraint:** this is to be built in a **single PR / single shot** —
> not split across multiple PRs. The shared-library extraction and the new daemon
> land together. The tracking checklist below scopes that one PR.

## Value

- **Who:** the maintainers, who already drive code reviews manually through
  `LmStreaming.Sample`. The daemon turns that into an unattended reviewer that
  works across the team's GitHub **and** Azure DevOps PRs.
- **Why:** automated, consistent PR review + a feedback loop (Judge grades) that
  lets us measure and tune review quality, plus a knowledge base that accumulates
  what the reviewer learns about each project/developer/scenario.
- **Bonus:** extracting the reusable agent/sandbox/auth code into a shared library
  pays down duplication and makes future agent-driven apps cheap to build.

## Workspace structure (ReviewBot repo + submodule)

All work happens inside a **ReviewBot git repository** (the "Reviewer Workspace /
Reviewer Repo"), checked out in a sandbox. Its layout:

```
ReviewBot repo (root)            # versioned; pushed to a ReviewBot remote so it persists
├── KnowledgeBase/               # knowledge collected across reviews
├── PRs/                         # all PRs + PR-based learning
│   └── <provider>-<repo>-<prId>/  # "Current PR directory" — workspace for this PR's work
└── <target repo>                # the Repository-to-Review, as a git SUBMODULE
```

**Per-PR setup (driven by the daemon via Sandbox APIs/tools):**
1. Create a **new branch on the root ReviewBot repo** for this PR.
2. Set the **submodule** (target repo) to the PR's branch at the **right commit**.
3. Create the **Current PR directory** under `PRs/` as the workspace for the PR's work.

**Persistence:** the ReviewBot repo is committed and **pushed to a remote**, so the
Knowledge Base and PR learnings survive sandbox teardown; the daemon re-clones it on
start. Git is the durable store for knowledge/learnings; **SQLite** tracks PR
orchestration state.

**Isolation:** each PR runs in its **own sandbox session with its own ReviewBot
checkout** (root on the PR branch, submodule at the PR commit), so concurrent PRs do
not contend on the shared submodule checkout or root branch.

## Architecture (per the design diagram)

```
Review-Daemon (polls for new PRs / updates to existing PRs; SQLite = state of record)
  -> PR Handler        : in-process MultiTurnAgentLoop reads PR comments + SQLite stage state,
                         decides the action: Review | Re-Review | Extract Knowledge | Judge
  -> Setup Workspace   : per-PR sandbox; branch the ReviewBot root repo, point the target-repo
                         submodule at the PR branch@commit, create the Current PR directory under PRs/
  -> Review Agent      : prompt built from PR stage + ReviewBot checkout + Current PR directory;
                         2 modes: (a) Post Comments, (b) Collect Comments but don't post (experimental)
  -> Knowledge Agent   : extract knowledge not previously known re Project / Developer / Scenario
  -> PR Review Judge    : grade the Review Agent's work; persist the grade + rationale + variant id
  If PR is closed       : skip Review; branch directly to Knowledge + Judge
```

### Locked design decisions
1. **Single PR delivery.** Extraction + daemon ship together.
2. **Reuse = refactor into a shared library.** Extract the reusable pieces out of
   `LmStreaming.Sample` into a new `src/` library consumed by both apps:
   - agent pool/factory (`MultiTurnAgentPool` + factory closure → injectable `IAgentFactory`)
   - sandbox gateway + session registry (`SandboxGatewayLifetime`, `SandboxSessionRegistry`, `SandboxGatewayOptions`)
   - OAuth providers + token store (`GitHubOAuthProvider`, `AdoOAuthProvider`, `FileOAuthTokenStore`, `IOAuthTokenStore`, `OAuthProviderHosts`)
   - auth webhook + deferred-auth (`AuthWebhookController`, `PendingAuthCoordinator`)
   - context-discovery injection (`ContextDiscovery*`)
   `LmStreaming.Sample` is rebuilt on top of the library and **must stay green**.
3. **PR watching = polling.** Per-repo persisted cursor; the embedded ASP.NET
   server serves **only** the sandbox auth webhook.
4. **Agent harness runs in-process** (`MultiTurnAgentLoop`); the **sandbox runs all
   git/filesystem work** — per-PR repo/submodule setup *and* the agent's bash/file
   commands — with tokens injected via the auth webhook so git/`gh`/`az`
   authenticate. **SQLite** on the daemon host is the orchestration source of truth.
5. **Auth = one-time human OAuth, then daemon refreshes.** Reuse
   `GitHubOAuthProvider` / `AdoOAuthProvider` refresh. **Single bot identity per
   provider.** If refresh fails, the daemon surfaces an **auth-required** signal
   (log/alert) and stops acting on that provider rather than silently blocking.
6. **Judge feedback v1 = persist only.** Judge writes a graded review record
   (score, rationale, skill/prompt **variant id**). A/B testing is a variant-id
   stamp + a report; **no** auto-routing and **no** auto-rewriting of skills in v1.

### Shared library extraction
- [ ] A new `src/` library contains the extracted agent/sandbox/auth/context
      machinery; `LmStreaming.Sample` references it and contains no copy of the
      moved types.
- [ ] `dotnet build LmDotnetTools.sln` succeeds with **zero new warnings**.
- [ ] The existing `LmStreaming.Sample` regression + browser E2E suites pass
      unchanged (sample behavior is preserved).

### Daemon scaffolding & hosting
- [ ] New `samples/CodeReviewDaemon.Sample` (or agreed name) builds and runs as a
      long-running host (`BackgroundService` poller).
- [ ] The embedded ASP.NET server exposes **only** the sandbox auth webhook
      endpoint; no inbound PR-event webhook routes exist.

### Polling, dedup & state (mockable, no live API)
- [ ] Against a **mocked** GitHub and a **mocked** ADO PR API, the poller
      discovers a new PR and processes it.
- [ ] A persisted **per-repo poll cursor** advances; a poll that finds nothing new
      does no work.
- [ ] A durable **SQLite review-state DB** (on the daemon host) records (provider,
      repo, prId, branch, last-reviewed commit SHA, last-processed comment
      id/timestamp, stage, grade, status) and is the orchestration source of truth.
- [ ] **Restart safety:** restarting the daemon with an already-reviewed PR open
      produces **no** duplicate review and **no** duplicate comment (state recovered
      from SQLite).

### PR Handler stage machine
- [ ] A deterministic **per-PR stage state machine** drives transitions; the LLM
      loop only fills gaps it can't decide deterministically.
- [ ] A **closed/merged/abandoned** PR skips Review and runs Knowledge + Judge.
- [ ] A new push to a PR mid-review serializes (per-PR single-flight); the latest
      commit wins.

### Review Agent
- [ ] Review Agent runs **in-process** and executes its bash/file commands in the
      **sandbox** against the PR's ReviewBot checkout (submodule at the PR commit)
      and Current PR directory.
- [ ] **Collect-Comments (non-posting)** mode is supported and is the **default in
      all automated tests** (no test posts to a real PR).
- [ ] **Post-Comments** mode posts via an **idempotent** path (hidden
      marker/fingerprint; re-runs at the same commit do not double-post).

### ReviewBot workspace & per-PR setup (sandbox-driven)
- [ ] On start, the daemon clones the **ReviewBot repo** (with its target-repo
      **submodule**) into a sandbox; `KnowledgeBase/` and `PRs/` exist.
- [ ] Per PR: the daemon **branches the root repo**, points the **submodule at the
      PR branch@commit**, and creates the **Current PR directory** under `PRs/` —
      all via Sandbox APIs/tools. Re-processing an existing PR reuses its branch/dir.
- [ ] Each PR is isolated in its **own sandbox session / ReviewBot checkout** so
      concurrent PRs don't contend on the submodule checkout or root branch.
- [ ] ReviewBot state (KnowledgeBase + PR learnings) is **committed and pushed to a
      remote** so it survives sandbox teardown and is re-cloned on next start.
- [ ] A closed PR's branch/Current-PR-directory is **cleaned up** after Knowledge +
      Judge complete so the repo doesn't grow unbounded. (TTL sweep optional.)

### Auth
- [ ] git/`gh`/`az` calls from the sandbox are authorized via the existing **auth
      webhook + token injection**, using the refreshed OAuth token for the
      single bot identity per provider.
- [ ] When token refresh fails, the daemon emits an **auth-required** log/alert and
      stops acting on that provider (no silent indefinite block).

### Knowledge & Judge
- [ ] Knowledge Agent writes to the **`KnowledgeBase/`** store in the ReviewBot
      repo, keyed by Project / Developer / Scenario, with a "not previously known"
      dedup check (committed + pushed so it persists).
- [ ] Judge writes a **graded review record** (score, rationale, skill/prompt
      **variant id**) — persisted in SQLite and/or under the PR's directory. (No
      auto-routing / no skill rewriting in v1; a grade-by-variant aggregation
      report is optional.)

### Resilience & concurrency
- [ ] On stage failure a PR is marked **failed** and surfaced (logged/alerted),
      never silently swallowed; transient errors are retried.
- [ ] **Sandbox-gateway-down** during a review is retried later without dropping
      the PR or double-posting.
- [ ] Per-PR single-flight is enforced (above). A **global max-concurrent cap** is
      configurable (optional for v1; serial processing is acceptable).

## Out of scope (v1)
- Inbound GitHub/ADO PR-event webhooks (polling only).
- Auto-routing of A/B prompt variants by score; auto-rewriting skill files.
- Per-repo/per-org credentials (single bot identity per provider).
- A headless (non-OAuth) credential provider.

## References
- Reuses the sandbox + OAuth token-injection line: #72, #68, #82, #86, #78, #77, #76.
- Conceptually adjacent: #106 (LmWorkflow), #107 (Wait/trigger primitive), #21 (RevoBot per-PR app).

---

# Design (approved 2026-06-28)

> Elaborates the requirements above into an implementation-ready design. The
> **Locked decisions** and **Acceptance Criteria** sections still hold; this adds the
> concrete architecture, components, data model, and flows. Brainstorming decisions
> baked in: daemon drives the sandbox as a **direct MCP client** for setup (LLM
> reserved for review/knowledge/judge); shared lib is **one cohesive
> `src/LmAgentInfra`**; Knowledge Base is **markdown + generated ToC**; A/B = a
> **primary** that posts plus an optional **variant** that only collects for
> comparison.

## Architecture overview

```
┌──────────────────────────────────────────────────────────────────────┐
│  CodeReviewDaemon.Sample  (single long-running .NET host)              │
│                                                                        │
│  ┌────────────┐   polls    ┌──────────────────────────────────────┐   │
│  │  Pollers   │──────────▶ │  PrOrchestrator (per-PR state machine)│   │
│  │ GH + ADO   │            │  SQLite = source of truth            │   │
│  └────────────┘            └───────────────┬──────────────────────┘   │
│        ▲                                    │ drives stages           │
│        │ IPrProvider                        ▼                          │
│  ┌─────┴───────┐   ┌───────────────────────────────────────────────┐  │
│  │ OAuth (lib) │   │ SandboxOrchestrator = DIRECT MCP CLIENT        │  │
│  │ GH + ADO    │   │ JSON-RPC bash tool: clone/branch/submodule/dir │  │
│  └─────────────┘   └───────────────────────────────────────────────┘  │
│        │                                    │  (agent's bash runs here)│
│  ┌─────▼─────────────────┐  in-process      ▼                          │
│  │ ASP.NET (webhook ONLY)│  ┌─────────────────────────────────────┐   │
│  │ auth token injection  │◀─┤ Agents: MultiTurnAgentLoop (in-proc) │   │
│  └───────────────────────┘  │  Handler · Review · Knowledge · Judge│   │
│                             └─────────────────────────────────────┘    │
└──────────────────────────────────────────────────────────────────────┘
        │ references
        ▼
   src/LmAgentInfra  (shared library — also consumed by LmStreaming.Sample)
```

- **One process.** A `BackgroundService` poller per provider; SQLite is the
  orchestration source of truth.
- **Two distinct sandbox callers:** (1) the **daemon as a direct MCP client** for
  mechanical setup (branch / submodule / dir), and (2) the **in-process review agent**
  whose tool calls also execute in the sandbox. Both authenticate via the egress
  auth-webhook + token injection.
- **ASP.NET exists only** to serve the sandbox auth webhook.

## Shared library (`src/LmAgentInfra`)

Single new project, folder-organized:

| Folder | Moved from `LmStreaming.Sample` |
|---|---|
| `Sandbox/` | `SandboxGatewayLifetime`, `SandboxSessionRegistry`, `SandboxGatewayOptions`, egress-proxy spawn |
| `Auth/` | `GitHubOAuthProvider`, `AdoOAuthProvider`, `FileOAuthTokenStore`, `IOAuthTokenStore`, `OAuthProviderHosts`, `AuthWebhookController`, `PendingAuthCoordinator` |
| `Agents/` | `MultiTurnAgentPool` + provider factory closure, refactored behind an injectable `IAgentFactory` |
| `Context/` | `ContextDiscovery*` |

`LmStreaming.Sample` deletes its copies and references the library (no daemon→sample
dependency). The **direct MCP client** wrapper starts in the daemon project (YAGNI);
it can move into `Sandbox/` when a second consumer appears.

## Daemon components

1. **Pollers (`GitHubPoller`, `AdoPoller` : `BackgroundService`).** Wake on an
   interval, call `IPrProvider.ListChangedPrsAsync(sinceCursor)`, advance a persisted
   per-repo cursor, enqueue work. A poll that finds nothing does no work.
2. **`IPrProvider` abstraction** (GH/ADO seam):
   ```
   ListChangedPrsAsync(cursor)          // new/updated PRs since cursor
   GetPrAsync(id) -> {branch, headSha, state, isClosed}
   GetCommentsAsync(id, sinceMarker)
   PostCommentAsync(id, body, fingerprint)   // idempotent
   GetDiffAsync(id)
   GetCloneUrlAsync(id)                  // auth via injected token
   ```
   Two implementations; divergences hidden behind it. Tests use **mocked** providers.
3. **`PrOrchestrator` + deterministic stage machine.** SQLite is source of truth.
   Stages: `Discovered → WorkspaceReady → Reviewing → KnowledgeExtracting → Judging →
   Done` (+ `Failed`). Transitions computed deterministically from (PR state,
   last-reviewed SHA, last-processed comment, stage). The **LLM PR Handler is invoked
   only to fill a genuine gap** (e.g. reading new comments to decide whether to
   re-review). Closed/merged/abandoned PRs **skip Reviewing** → Knowledge + Judge.
4. **`SandboxOrchestrator` (direct MCP client).** Owns the JSON-RPC MCP client. Does,
   idempotently: clone ReviewBot repo (if session new) →
   `git checkout -b pr/<provider>-<repo>-<id>` on root → set submodule to PR branch @
   `headSha` → create `PRs/<provider>-<repo>-<id>/`. Re-processing reuses the
   branch/dir. Each PR gets its **own sandbox session / ReviewBot checkout**.
5. **Agents (in-process `MultiTurnAgentLoop` via `IAgentFactory`).**
   - **Review Agent** — prompt from PR stage + ReviewBot checkout + Current PR dir;
     bash/file tools execute **in the sandbox**. Two modes wired to the A/B model.
   - **Knowledge Agent** — extracts project/developer/scenario knowledge, writes
     markdown to `KnowledgeBase/`, dedups ("not previously known"), regenerates
     `_toc.md`.
   - **Judge** — grades the review work; persists score + rationale + variant id.
6. **A/B variant model.** A **primary** skill always produces the real review in
   **Post-Comments** mode. An optional **variant** (single active, per-repo mappable
   in config) runs in **Collect-but-don't-post** mode purely for comparison — it never
   posts. The Judge grades both and stamps the variant id. No auto-routing / no skill
   rewriting in v1.

## Data flow (one PR)

```
poll → IPrProvider reports PR @ SHA (newer than SQLite)
  → SQLite upsert (stage=Discovered)
  → SandboxOrchestrator: own session, branch root, submodule@SHA, make PR dir  (stage=WorkspaceReady)
  → PrOrchestrator computes action; if ambiguous, LLM Handler reads comments to decide
  → if PR open & SHA unreviewed:
        Review Agent (primary, Post mode) → idempotent PostComment(fingerprint=SHA)
        [if variant configured for repo] Review Agent (variant, Collect-only) → not posted
                                                                        (stage=Reviewing→done)
  → Knowledge Agent → write/dedup KnowledgeBase md + ToC; commit+push ReviewBot repo
  → Judge → grade primary (+variant if present); persist score+rationale+variantId
  → SQLite: last-reviewed SHA, last comment marker, stage=Done
  → cleanup PR branch/dir after Knowledge+Judge (closed PRs); push
If PR closed at poll time → skip Review, go straight to Knowledge + Judge.
```

## Data model

**SQLite (daemon host — orchestration source of truth)**

`pr_state` — one row per PR:

| column | purpose |
|---|---|
| `provider, repo, pr_id` | composite key |
| `branch, head_sha` | PR head being reviewed |
| `last_reviewed_sha` | dedup — skip if unchanged |
| `last_comment_marker` | last processed comment id/timestamp |
| `stage` | `Discovered…Done` / `Failed` |
| `status` | open / closed / merged / abandoned |
| `sandbox_session_id` | the PR's own session |
| `updated_at, attempt_count` | retry / visibility |

`poll_cursor` — per `(provider, repo)`: last poll position.
`review_grade` — per run: `pr_id, run_sha, variant_id, mode (primary/variant), score,
rationale, created_at`.

**ReviewBot git repo (durable knowledge store — committed + pushed)**

```
ReviewBot/ (root, pushed to a ReviewBot remote)
├── KnowledgeBase/
│   ├── project/<repo>.md
│   ├── developer/<author>.md
│   ├── scenario/<topic>.md
│   └── _toc.md                # generated from md headers
├── PRs/<provider>-<repo>-<prId>/   # Current PR dir: context, collected comments, variant output
└── <target repo>             # git submodule, set to PR branch @ head_sha
```

Knowledge persists in git (re-cloned on start); SQLite holds only
orchestration/grades. **Two stores, no overlap.**

## Error handling, concurrency & auth

- **Per-PR single-flight.** New push mid-review serializes; latest SHA wins. A global
  max-concurrent cap is configurable; serial is acceptable for v1.
- **Stage failure** → row marked `Failed`, surfaced (log/alert), `attempt_count++`;
  transient errors retried with backoff. Never silently swallowed.
- **Sandbox-gateway-down** during a review → retried later from SQLite state without
  dropping the PR or double-posting.
- **Idempotent posting** → hidden marker/fingerprint keyed on `head_sha`; re-runs at
  the same SHA never double-post.
- **Auth.** Refreshed OAuth token for the single bot identity per provider, injected
  via the egress auth-webhook. **Refresh failure → loud `auth-required` log/alert and
  the daemon stops acting on that provider.**
- **Restart safety.** All progress in SQLite; restart produces no duplicate review /
  comment.
- **Cleanup.** Closed PRs' branch + Current-PR-dir removed after Knowledge + Judge;
  optional TTL sweep.

## Testing strategy

- **Mocked providers** for GH and ADO — no live API, no real PR posting, in any test.
- **Collect-but-don't-post is the default mode in all automated tests.**
- Tests: new-PR discovery + cursor advance; **restart safety** (no dup review /
  comment); idempotent posting at same SHA; closed-PR → Knowledge + Judge skip;
  stage-machine transitions; `SandboxOrchestrator` setup against a mock MCP endpoint;
  auth-required signal on refresh failure.
- **Hard gate:** existing `LmStreaming.Sample` regression + browser E2E suites pass
  unchanged after the extraction.
- Build: `dotnet build LmDotnetTools.sln` with **zero new warnings**; CSharpier
  formatting.

gautam-achieveai:
Closes #118.

## Code-Review Daemon sample: unattended GitHub + ADO PR reviewer

A new `samples/CodeReviewDaemon.Sample` — a console-launchable daemon that polls GitHub (and optionally Azure DevOps) for open PRs and runs an unattended, in-process review pipeline. Delivered as a single PR per the approved design.

### Architecture (as locked in the issue)

- **Reusable machinery extracted to `src/LmAgentInfra`** (agent pool, sandbox gateway lifetime/registry, OAuth providers + token store, the auth webhook controller, context discovery), consumed by **both** `LmStreaming.Sample` and the new daemon. `LmStreaming.Sample` stays green.
- **PR watching = polling.** ASP.NET Core hosts exactly **one** route: `POST /api/auth/webhook/{provider}` (the sandbox gateway's post-auth callback). No other route is exposed (enforced by a route-exposure test).
- **In-process agent harness** (`MultiTurnAgentLoop`); the **sandbox runs all git/fs work** (driven as a direct MCP client); **SQLite is the orchestration source of truth** (`review_run` identity, opaque poll cursors, crash-safe outbox, append-compatible artifacts).
- **One-time human OAuth, then unattended refresh**; on refresh failure the daemon raises an operator "auth-required" signal and fails fast (it never holds a webhook call open for an absent human).
- **Judge persists only** `{score, rationale, variant_id}` — never auto-routes or rewrites anything.

### Pipeline

`PrPollingService` → `PrOrchestrator` (deterministic, resumable `StageMachine`) → `DaemonReviewStageExecutor`:

1. **ContextReady** — fetch the PR diff in the sandbox → persist a `review-context` artifact.
2. **Reviewed** — primary `ReviewAgent` → `review` artifact; optional collect-only A/B **B** arm (`VariantReviewer`, bounded 2-way); optional `KnowledgeAgent` (writes a Knowledge Base entry + regenerates its ToC).
3. **Judged** — optional `JudgeAgent` grades the review.
4. **Posted** — idempotent `ReviewPoster` (versioned key + outbox + provider-side backstop scan) via the provider-matched publisher.

### Safe by default

Every feature flag defaults to its conservative setting: **collect-only** (no comments posted), **GitHub-only** (ADO opt-in), and an **empty repo allow-list** (reviews nothing). With no allow-listed repos the poller has no targets and the whole pipeline — including the live OpenAI agent loop — is **inert and dead-by-default**. Live posting against a real repository remains an explicit operator action.

### Tests & verification

- **Daemon: 221 tests** (SQLite store/migrations, stage machine, orchestrator, poller, providers/publishers over fake HTTP, agents, A/B capability enforcement, idempotent posting, the stage executor, DI/route-exposure boot).
- **Build:** `dotnet build LmDotnetTools.sln` → 0 errors / 0 warnings (`TreatWarningsAsErrors`).
- **Regression gate:** `LmStreaming.Sample.Tests` unit **325/325**; `LmStreaming.Sample.E2E.Tests` **76/76** (the only 2 failures are pre-existing `SandboxWorkspaceGatewayE2ETests` environment-dependent 400s that require a live gateway/workspace path).


samples/CodeReviewDaemon.Sample/Workspace/OperationPolicy.cs
```diff
@@ -0,0 +1,319 @@
...
+            _ => PolicyDecision.Deny($"unknown operation '{request.Operation}'"),
+        };
+    }
+
+    /// <summary>
+    /// Whether a credential may be injected for this request. Deliberately identical to
+    /// <see cref="Decide"/> so a denied operation can never be credential-injected (fail closed
+    /// both ways, plan §4).
+    /// </summary>
+    public bool ShouldInjectCredential(OperationRequest request) => Decide(request).IsAllowed;
```
gautam-achieveai, line 160: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Security / plan contract)

Plan v5 says `OperationPolicy` is the single source of truth for both sandbox network rules and webhook token resolution, and that a denied operation must both block egress and withhold credentials. The policy is implemented here, but in the reviewed runtime path it is only consumed by `SubmoduleInitializer`; `Program.cs` creates `SandboxOrchestrator` with just gateway/session, `AuthWebhookController` still gates token injection by provider host allowlist, and `SandboxSessionRegistry` emits broad host rules with empty method/path filters. That leaves the approved operation/method/path policy unenforced for target/fork fetches, provider metadata, comment posting, and ReviewBot pushes.

**Suggestion:** Wire this policy into the daemon's actual sandbox/auth path: classify each webhook/network request into `FetchTarget`, `FetchForkHead`, `FetchSubmodule`, `PostReviewComment`, `PushReviewBot`, or `ReadProviderMetadata` using provider/host/method/path; block denied requests and return no credential. Add integration tests proving denied post/push/fetch requests are both egress-blocked and credential-denied.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+            Provider = provider,
+            Payload = JsonSerializer.Serialize(
+                new ContextArtifactPayload(run.PrId, run.BaseSha, run.HeadSha, result.Stdout)),
+        });
+
+        _logger.LogInformation("Run {RunId}: persisted {Kind} ({Length} char diff).",
+            run.Id, ContextArtifactKind, result.Stdout.Length);
+    }
+
+    private async Task ReviewAsync(ReviewRun run, CancellationToken cancellationToken)
```
gautam-achieveai, line 221: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Completeness / plan contract)

Plan v5 requires every primary review to retain `PRs/...` artifacts and `KnowledgeBase/...` updates in the ReviewBot repo default branch via the durable one-commit sequence, then persist `reviewbot_push` or keep the review branch on `GitSyncFailed`. The PR includes `ReviewBotRepoManager` and tests for that sequence, but this runtime stage never calls it: it writes context/review artifacts to SQLite, optionally writes a KB entry, and posts/collects the review. Scoped usage search found `ReviewBotRepoManager` referenced only by its own class/tests, so the core durable retention contract appears unreachable in production.

**Suggestion:** Integrate `ReviewBotRepoManager.PublishAsync` into the primary review/knowledge path, writing both `PRs/...` and `KnowledgeBase/...` artifacts through it, and persist `reviewbot_push` / `GitSyncFailed` in SQLite. Add executor/orchestrator tests showing a real review run invokes the durable commit sequence and handles sync failure.

samples/CodeReviewDaemon.Sample/Workspace/ReviewBot/ReviewBotInitCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+        var probe = await git
+            .RunAsync(["rev-parse", "--is-inside-work-tree"], workdir, cancellationToken)
+            .ConfigureAwait(false);
+        if (probe.Succeeded)
+        {
+            logger.LogInformation("Reusing existing ReviewBot checkout at {Workdir}.", workdir);
+            return true;
+        }
+
+        var clone = await git
```
gautam-achieveai: [gautam-achieveai's bot] **MEDIUM** (Completeness / setup UX)

The v5 plan and approval notes put ReviewBot setup ownership on `reviewbot init` and specifically call out clear behavior when `--url` points at a GitHub repo that does not exist. This path currently just runs `git clone` and reports a generic clone/connect failure. That gives operators no clear distinction between not-found, missing repo-create permission, bad credentials, or transient gateway failure.

**Suggestion:** Either implement the provider-specific repo creation path for setup identities with repo-create permission, or explicitly narrow the contract/docs/tests to require a pre-created remote and surface precise not-found/permission diagnostics from clone failures.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/ReviewBotInitializerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace CodeReviewDaemon.Sample.Tests.Scenarios;
+
+/// <summary>
+/// P3.2 — the idempotent ReviewBot setup decision. An empty checkout is seeded + committed + pushed; a
+/// fully-seeded checkout is a no-op; a partially-seeded checkout is reported malformed (with the exact
+/// gaps) and is never silently mutated.
+/// </summary>
+public sealed class ReviewBotInitializerTests
```
gautam-achieveai: [gautam-achieveai's bot] **MEDIUM** (Testing / observability)

The approved implementation notes asked daemon tests to follow the repo's structured logging convention (`LoggingTestBase`, named properties, logs under `.logs/tests/tests.jsonl`). In the reviewed `CodeReviewDaemon.Sample.Tests` scope I found no `LoggingTestBase`/`ITestOutputHelper` usage; auth/git/orchestration tests commonly pass `NullLogger`. For this feature area, failures in token resolution, sandbox git commands, and state transitions will be harder to diagnose than the rest of the repo's tests.

**Suggestion:** Convert the scenario tests that exercise auth, git, orchestration, and persistence transitions to inherit `LoggingTestBase` and pass test loggers/logger factories into components. Keep `NullLogger` only for trivial pure-value tests.

gautam-achieveai:
[gautam-achieveai's bot] [BLOCKER] **HIGH** (Security / webhook contract)

The approved v5 plan requires the daemon's only runtime route to validate HMAC over the raw body, enforce a ±5 minute timestamp, reject replayed nonces/delivery ids, cap bodies at 1 MB, accept only `application/json`, and reject unknown provider paths. This action currently authenticates with the shared `Authorization` secret and then binds `[FromBody]` JSON; I don't see the raw-body signature, timestamp/replay, size, or content-type enforcement in this path, and scoped searches did not find the required endpoint test matrix.

**Suggestion:** Add the full webhook verification layer before token resolution: raw-body HMAC validation, timestamp tolerance, delivery-id replay cache, provider allowlist, JSON-only and max-body checks. Cover invalid signature, stale timestamp, duplicate delivery, malformed JSON, oversized body, wrong provider path, and valid signature over altered body.

_File/line fallback: intended for `src/LmAgentInfra/Controllers/AuthWebhookController.cs` around the `Evaluate` webhook action._

gautam-achieveai:
[gautam-achieveai's bot] # PR Review: Code-Review Daemon sample: unattended GitHub + ADO PR reviewer (#118)

## Verdict: REQUEST CHANGES

I reviewed PR #121 against the approved v5 implementation plan for #118. The shape is directionally aligned, and the over-engineering / architecture / simplification passes did not surface extra scope concerns. However, three approved plan contracts appear implemented as isolated helpers/tests but not integrated into the runtime path, so this needs changes before merge.

### Blocking findings

1. **HIGH — `OperationPolicy` is not wired into actual egress/token enforcement**  
   The policy exists, but the runtime sandbox/auth path does not appear to use it as the shared source of truth for method/path-scoped egress and credential injection.

2. **HIGH — Webhook security contract from v5 is missing**  
   The webhook uses the shared Authorization secret, but the plan-required HMAC/timestamp/replay/body/content-type validation is not visible in the endpoint path or tests.

3. **HIGH — ReviewBot durable retention is not invoked by runtime**  
   `ReviewBotRepoManager` exists and is tested, but the stage executor does not appear to call it, so the `PRs/...` + `KnowledgeBase/...` durable one-commit retention contract is unreachable.

### Non-blocking findings

4. **MEDIUM — `reviewbot init` lacks explicit remote creation / permission diagnostics**
5. **MEDIUM — New daemon tests do not follow structured logging convention**

See inline comments for details and suggested fixes.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in dd74c19. The primary review path now invokes `ReviewBotRepoManager.PublishAsync`, landing `PRs/{provider}/{repo}/{pr}-{sha}/review.md` plus the regenerated `KnowledgeBase/...` onto the ReviewBot default branch via the one-commit durable sequence, and records the `reviewbot_push` outcome in the outbox (terminal on success; left non-terminal on `GitSyncFailed` so the reconcile path retries). Gated on the new `CodeReviewDaemonOptions.ReviewBotRepoUrl` — unset (default) skips retention so a freshly-deployed daemon stays inert (review output still lands in SQLite). Added executor tests for the durable publish and the sync-failure path. (AC#6)

samples/CodeReviewDaemon.Sample/Workspace/OperationPolicy.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in dd74c19. Every provider-API request the daemon issues is now tagged with its `SandboxOperation` (`ReadProviderMetadata`/`PostReviewComment`) and routed through a new `OperationPolicyHandler` (`DelegatingHandler`) that classifies it against the canonical `OperationPolicy` and, on denial (or an untagged request), BOTH blocks egress AND strips the `Authorization` header — the same fail-closed-both-ways guarantee the policy already makes for git transport. Wired at the daemon's own HTTP seam (providers/publishers + Program.cs), so no shared `LmAgentInfra` type changed and the LmStreaming suites stay green. Added `OperationPolicyHandlerTests` proving a denied post/fetch is both egress-blocked and credential-denied. Git-transport classification (FetchTarget/FetchSubmodule/PushReviewBot) remains enforced at the sandbox/git seam as before.

samples/CodeReviewDaemon.Sample/Workspace/ReviewBot/ReviewBotInitCommand.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in dd74c19. `reviewbot init` now requires a pre-created remote and runs clone failures through a new `CloneFailureClassifier` that distinguishes not-found, permission-denied, bad-credential, and transient/network failures from git's stderr, surfacing a precise diagnostic + distinct exit code instead of a generic clone/connect error. Narrowed the documented contract accordingly (provider repo-creation is explicitly out of scope). Added `CloneFailureClassifierTests`.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/ReviewBotInitializerTests.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in dd74c19. The daemon scenario tests exercising auth, git, orchestration, and persistence transitions now inherit `LoggingTestBase` and pass real test loggers / `ILoggerFactory` into the components under test, so logs route to `.logs/tests/tests.jsonl` with `testClassName`/`testCaseName` correlation. `NullLogger` is kept only for trivial pure-value tests (pure parsers / static builders with no component logging).

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #121

I re-reviewed current head `dd74c19` against the prior review findings.

### Resolved from the prior review

- `OperationPolicy` is now wired into the daemon provider/publisher HTTP path via `OperationPolicyHandler`, and provider requests are tagged with `ReadProviderMetadata` / `PostReviewComment` before egress.
- ReviewBot durable retention is now invoked from `DaemonReviewStageExecutor` when `ReviewBotRepoUrl` is configured, with `reviewbot_push` outbox recording and `GitSyncFailed` left non-terminal for retry.
- `reviewbot init` now narrows the contract to a pre-created remote and classifies clone failures into actionable diagnostics.
- The daemon scenario tests now use `LoggingTestBase` / test loggers for auth, git, orchestration, and persistence paths.

### Still blocking

The webhook security contract from the approved v5 plan is still not implemented. `AuthWebhookController` still accepts the gateway callback with the shared `Authorization` secret and `[FromBody]` JSON binding, but the plan-required raw-body HMAC signature validation, timestamp tolerance, nonce/delivery replay protection, max-body/content-type enforcement, and matching negative test matrix are still absent in the reviewed scope.

**Verdict remains: REQUEST CHANGES** until that webhook verification layer and tests are added.

gautam-achieveai:
[gautam-achieveai's bot] **Addressed** in eaba9e6 — the plan §9 webhook security layer is now implemented.

A daemon-only `WebhookVerificationMiddleware` runs in front of the shared `AuthWebhookController` for `POST /api/auth/webhook/{provider}`, before MVC binds the body or any token is resolved:
- **Raw-body HMAC-SHA256** over `{timestamp}.{rawBody}` (timestamp bound into the signed payload, Stripe-style), constant-time compare;
- **±5 min timestamp** tolerance;
- **single-use delivery id** via a TTL-bounded replay cache;
- **`application/json` only** + **1 MiB body cap**;
- **provider allow-list** (github; ado when enabled), unknown → reject.

All checks fail closed. The signing secret is shared with the gateway via `Auth:Webhook:SigningSecret` (random fallback so an unconfigured daemon rejects every callback); the gateway must sign its callbacks with these `X-Sandbox-Signature`/`X-Sandbox-Timestamp`/`X-Sandbox-Delivery-Id` headers.

Scope note (consistent with the accepted OperationPolicy fix): enforcement is at the **daemon's own seam** — the shared `AuthWebhookController` (also used by LmStreaming.Sample) is untouched, so the LmStreaming hard gate stays green. Because it's added via `UseWhen` on the webhook path it registers **no route**, so AC#4's single-endpoint surface and `RouteExposureTests` are preserved; the controller's shared-secret check remains as defence-in-depth.

**Negative-test matrix** (`WebhookRequestVerifierTests` + end-to-end `WebhookVerificationEndpointTests`): invalid signature, valid-signature-over-altered-body, stale/future/unparseable timestamp, missing headers, non-JSON content-type, oversized body, unknown provider path, duplicate delivery (replay), and malformed JSON (passes the edge → 400 at model binding). Build 0 warnings; daemon tests 261/261; LmStreaming unit 325/325, E2E 76/76 (2 pre-existing env-400s).

samples/CodeReviewDaemon.Sample/Auth/WebhookSigningSecret.cs
```diff
@@ -0,0 +1,79 @@
...
+            ? RandomNumberGenerator.GetHexString(64)
+            : configuredSecret;
+        _key = Encoding.UTF8.GetBytes(Value);
+    }
+
+    /// <summary>The resolved signing secret. SECRET — never log this value.</summary>
+    public string Value { get; }
+
+    /// <summary>
+    /// Computes the lowercase-hex HMAC-SHA256 over <c>{timestamp}.{body}</c> (Stripe-style). Binding the
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Security / replay protection)

The replay cache key is not authenticated. The signature currently covers only `{timestamp}.{rawBody}`, while `X-Sandbox-Delivery-Id` is required separately and then trusted as the replay-cache key. A captured valid callback can be resent within the timestamp window with the same body/signature/timestamp but a different delivery id, bypassing `DeliveryReplayCache` and reaching the controller/token-resolution path again. That leaves the plan §9 replay/nonce contract only partially satisfied.

**Suggestion:** Include the delivery id in the canonical signed payload (for example `{timestamp}.{deliveryId}.{rawBody}`, or a length-delimited equivalent), verify with the presented delivery id, and add a regression test where the same signed body/timestamp/signature is replayed with a changed delivery id and is rejected.

samples/CodeReviewDaemon.Sample/Auth/WebhookVerificationMiddleware.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        // Cheap pre-read guard: a declared length over the cap is rejected before buffering the body.
+        if (request.ContentLength is { } declared && declared > _maxBodyBytes)
+        {
+            await RejectAsync(context, StatusCodes.Status413PayloadTooLarge, "body too large").ConfigureAwait(false);
+            return;
+        }
+
+        var (body, overflowed) = await ReadCappedAsync(request, context.RequestAborted).ConfigureAwait(false);
```
gautam-achieveai, line 83: [gautam-achieveai's bot] **MEDIUM** (Testing / webhook contract)

The middleware now has real HTTP-level size enforcement: it rejects an over-limit declared `Content-Length` before buffering and also rejects overflow while reading the request stream. The final endpoint matrix covers invalid signature, altered body, stale timestamp, missing signature, non-JSON content, unknown provider, replay, and malformed JSON, but I could not find an endpoint-level oversized-body test. The only oversized-body coverage is the direct `WebhookRequestVerifier` unit test, which bypasses the request-body buffering/cap/rewind path.

**Suggestion:** Add an endpoint test that sends a signed body over the configured/default limit and asserts `HttpStatusCode.PayloadTooLarge` / 413. That pins the actual middleware path requested by the prior review, not just the pure verifier.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -0,0 +1,211 @@
...
+            _ = apm.FeatureProviders.Remove(existing);
+        }
+        apm.FeatureProviders.Add(new WebhookOnlyControllerFeatureProvider());
+    });
+
+var app = builder.Build();
+
+// Verify the gateway's signature/timestamp/delivery + body before MVC binds it (plan §9). Scoped to the
+// webhook path via UseWhen so it adds NO route — the daemon's single-endpoint surface (AC#4) is intact.
+app.UseWhen(
```
gautam-achieveai, line 193: [gautam-achieveai's bot] **MEDIUM** (Architecture / route contract)

The webhook verification branch is prefix-scoped, not scoped to the exact `POST /api/auth/webhook/{provider}` endpoint. `UseWhen` uses `StartsWithSegments("/api/auth/webhook")`, and the middleware repeats the prefix check before extracting the first segment as the provider. That means a signed `POST /api/auth/webhook/github/extra` is HMAC-verified and can consume a delivery id before MVC later treats it as an unmatched route, even though the documented contract says this middleware runs for the single webhook route and only there.

**Suggestion:** Centralize an exact route-shape predicate: POST, prefix `/api/auth/webhook`, exactly one non-empty provider segment, and no extra path segments. Use that predicate both in `Program.cs` and inside the middleware, and add a suffix-path regression test.

samples/CodeReviewDaemon.Sample/Auth/WebhookVerificationMiddleware.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            _logger.LogWarning("Rejected webhook callback for provider {Provider}: {Rejection}.", input.Provider, result.Rejection);
+            await RejectAsync(context, StatusFor(result.Rejection), result.Rejection.ToString()).ConfigureAwait(false);
+            return;
+        }
+
+        // Signature is valid → the delivery id is trustworthy; reject a replay of it.
+        if (!_replayCache.TryRegister(input.DeliveryId!, now))
+        {
+            _logger.LogWarning("Rejected replayed webhook delivery {DeliveryId} for provider {Provider}.", input.DeliveryId, input.Provider);
```
gautam-achieveai: [gautam-achieveai's bot] **LOW** (Logging / sensitive metadata)

On replay rejection this logs the full delivery id. Delivery ids are single-use replay/correlation identifiers; they are not as sensitive as the signing secret or token, but logging the full value is avoidable metadata exposure.

**Suggestion:** Omit the delivery id from the warning, or log only a short hash/truncated prefix so operators can correlate without preserving the full replay identifier.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #121

I re-reviewed final head `eaba9e6` against the remaining webhook-security blocker from the prior review.

### Resolved

- The daemon now has a middleware layer in front of the shared auth webhook that verifies raw-body HMAC, timestamp freshness, delivery-id replay cache, content type, body size, and provider allow-list before MVC model binding/token resolution.
- The previous missing-contract gap is substantially addressed with verifier and endpoint tests.

### Still blocking

1. **HIGH — Delivery id is not part of the signed payload**  
   The signature covers `{timestamp}.{rawBody}` but the replay cache trusts `X-Sandbox-Delivery-Id` separately. A captured valid callback can be replayed within the timestamp window with a different delivery id, bypassing the replay cache. This leaves the replay/nonce portion of plan §9 incomplete.

### Non-blocking follow-ups

2. **MEDIUM — Missing endpoint-level oversized-body test**  
   The verifier unit test covers body size, but the endpoint matrix does not pin the middleware's request-buffering/413 path.

3. **MEDIUM — Middleware is prefix-scoped rather than exact-route scoped**  
   The branch also runs for suffix paths under `/api/auth/webhook/...`, which is broader than the documented single `POST /api/auth/webhook/{provider}` contract.

4. **LOW — Replay warning logs full delivery id**  
   Consider omitting or hashing/truncating the value.

**Verdict remains: REQUEST CHANGES** until the delivery id is authenticated as part of the signed payload (with a regression test for changed-delivery replay).

samples/CodeReviewDaemon.Sample/Auth/WebhookSigningSecret.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in 7ee631c. The delivery id is now part of the signed payload: the HMAC covers `{timestamp}.{deliveryId}.{rawBody}` (was `{timestamp}.{rawBody}`) and `Matches` verifies against the presented delivery id. A captured callback replayed within the timestamp window under a fresh `X-Sandbox-Delivery-Id` now fails the signature check before the replay cache is even consulted. Regression tests added at both levels: `WebhookRequestVerifierTests.A_replay_with_a_changed_delivery_id_is_rejected` and the end-to-end `A_replay_with_a_changed_delivery_id_is_rejected_401` (same signed body/timestamp/signature, swapped delivery id → 401).

samples/CodeReviewDaemon.Sample/Auth/WebhookVerificationMiddleware.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in 7ee631c. Added `WebhookVerificationEndpointTests.An_oversized_body_is_rejected_413`: it posts a body just over the 1 MiB cap and asserts 413 through the real middleware path (declared `Content-Length` rejected before buffering/verification), not just the pure `WebhookRequestVerifier` size unit.

samples/CodeReviewDaemon.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in 7ee631c. Introduced `WebhookVerificationMiddleware.IsWebhookRoute` — POST + the prefix + exactly one non-empty provider segment + no trailing path — and use it both in the `UseWhen` branch and inside the middleware. A suffix path like `/api/auth/webhook/github/extra` is no longer HMAC-verified and cannot consume a delivery id; it falls through to MVC (404). Regression test: `A_suffix_path_under_the_webhook_prefix_is_not_verified_and_404s`.

samples/CodeReviewDaemon.Sample/Auth/WebhookVerificationMiddleware.cs
gautam-achieveai: [gautam-achieveai's bot] **Fixed** in 7ee631c. The replay-rejection warning now logs only a truncated delivery-id prefix (`{first 8}…`) via a `Truncate` helper, so operators can still correlate without the full single-use identifier landing in logs.

gautam-achieveai:
[gautam-achieveai's bot] **Addressed** in 7ee631c (all four items from the re-review of `eaba9e6`).

- **[BLOCKER] Delivery id now authenticated** — the HMAC covers `{timestamp}.{deliveryId}.{rawBody}`, so a captured callback replayed under a fresh delivery id fails the signature check (caught before the replay cache). Regression tests at the verifier and endpoint levels.
- **Exact-route scoping** — `IsWebhookRoute` (POST + single provider segment, no trailing path) gates both the `UseWhen` branch and the middleware; a suffix path is no longer verified or able to consume a delivery id (→ 404). Regression test added.
- **Endpoint 413 test** — a signed over-cap body now pins the real middleware buffering/Content-Length path.
- **Log redaction** — replay rejections log only a truncated delivery-id prefix.

Build 0 warnings; daemon 265/265; LmStreaming unit 325/325, E2E 76/76 (2 pre-existing env-400s).

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #121

I re-reviewed current head `b4b79b1` against the approved #118 plan v5 and the prior PR review thread. The webhook-security blocker from the earlier review looks addressed, but the deeper plan-alignment pass found several core runtime contracts that are still not reachable through the daemon's actual execution path.

## Verdict: REQUEST CHANGES

### Blocking findings

1. **HIGH — ContextReady diffs the ReviewBot checkout instead of cloning/fetching the target PR**  
   Evidence: `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:116-121`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:278-300`, `samples/CodeReviewDaemon.Sample/Workspace/Git/SubmoduleInitializer.cs:26-56`.

   The runtime never clones/fetches the target PR repository and does not wire `SubmoduleInitializer` into the runtime path. `ContextReady` runs `git -C /work/reviewbot diff {base}...{head}`, but `/work/reviewbot` is the ReviewBot retention checkout, not the target repo checkout. Scoped searches found `SubmoduleInitializer` used by tests only. This leaves plan §3 and the pipeline's “fetch the PR diff in the sandbox” contract incomplete.

   **Required change:** add an explicit target checkout/fetch stage in the sandbox, using `GitRunner` hardening, per-run `OperationPolicy` for target/fork/submodules, selective submodule initialization, and a separate target worktree path. Diff from that target checkout, not ReviewBot.

2. **HIGH — Runtime `OperationPolicy` wiring is too broad and does not enforce plan §4 path scope**  
   Evidence: `samples/CodeReviewDaemon.Sample/Program.cs:133-140`, `samples/CodeReviewDaemon.Sample/Workspace/DaemonOperationPolicy.cs:15-42`, `samples/CodeReviewDaemon.Sample/Workspace/OperationPolicy.cs:229-243`.

   The policies wired in `Program` use `DaemonOperationPolicy.ForGitHub()` / `ForAdo()`, which set repo paths to `/`, and API decisions validate only host + method. The tests exercise hand-built per-repo policies such as `/acme/widgets`, not the actual runtime DI policy. This does not satisfy the plan's provider/host/method/path-scoped shared enforcement contract.

   **Required change:** build the operation policy from the concrete poll target/review run (target repo, fork repo, ReviewBot repo, allowed submodules), and make provider API operations validate expected repo/PR routes as well as host and method. Add tests against the actual `Program`/DI wiring.

3. **HIGH — `ReviewBotRepoUrl` is not used by the long-running daemon to clone/validate the ReviewBot checkout**  
   Evidence: `samples/CodeReviewDaemon.Sample/Configuration/CodeReviewDaemonOptions.cs:56-63`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:278-300`, `samples/CodeReviewDaemon.Sample/Workspace/ReviewBot/ReviewBotInitCommand.cs:60-73`, `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBotRepoManager.cs:81-85`.

   The runtime checks only whether `ReviewBotRepoUrl` is non-empty, then assumes `/work/reviewbot` already exists and is valid. The URL is used by the setup subcommand's clone path, but not by the long-running daemon. A fresh sandbox/session with a configured `ReviewBotRepoUrl` therefore cannot perform the plan's “runtime clones/pushes only and fails fast if malformed” contract.

   **Required change:** on daemon startup or before retention publish, clone/fetch the configured ReviewBot remote into the configured checkout, validate the skeleton, and fail fast with classified diagnostics if it is missing or malformed.

4. **HIGH — Sandbox command output, diffs, and artifacts are unbounded**  
   Evidence: `samples/CodeReviewDaemon.Sample/Workspace/Sandbox/SandboxOrchestrator.cs:67-75`, `samples/CodeReviewDaemon.Sample/Workspace/Sandbox/SandboxOrchestrator.cs:117-145`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:120-140`, `samples/CodeReviewDaemon.Sample/Persistence/ReviewStore.cs:346-387`.

   `SandboxOrchestrator` has no per-command timeout beyond the poller's cancellation token, materializes the whole MCP response with `string.Join`, and the stage executor persists the full `git diff` into SQLite and then feeds it to the agent. This can hang the polling loop indefinitely and grow memory/SQLite/prompt size without bounds. Plan §15 explicitly called for quota/cleanup/output truncation/redaction controls.

   **Required change:** add configurable per-command timeouts, max stdout/stderr bytes, truncation/redaction metadata, bounded diff collection, and capped artifact storage before content is persisted or passed to agents.

### Non-blocking follow-ups

5. **MEDIUM — Poll cursors are stored but not honored by providers**  
   `PrPollingService` passes a cursor, but `GitHubPrProvider` fetches only the first `per_page=100` page and `AdoPrProvider` does not use continuation/skip/top. Add provider-specific pagination/cursor handling and bounded pages per poll.

6. **MEDIUM — One poison PR/target can starve the rest of a poll cycle**  
   Exceptions from a single provider call or PR orchestration escape `PollOnceAsync`, so later PRs/targets wait until the next interval. Catch/log per target and per PR, mark only that run `RetryPending`, and continue.

7. **MEDIUM — Provider HTTP calls lack retry/rate-limit handling**  
   Provider and publisher calls use `SendAsync` + `EnsureSuccessStatusCode` without `429`/`5xx`/`Retry-After` handling, request timeouts, jitter, or provider cooldown. Add resilience policies; retry POSTs only under the existing idempotency/backstop guard.

8. **MEDIUM — Migration tests do not prove crash-idempotence or single-writer behavior**  
   The migration tests cover fresh/no-op/preserve/downgrade cases, but not a failing migration inside the transaction or concurrent migrators. Add tests that verify rollback/user_version behavior after failure and concurrent migration serialization.

### Review notes

- The over-engineering and code-simplicity passes did not find extra scope creep worth raising; the main concerns are plan-reachability and operational safety.
- Webhook HMAC/timestamp/delivery/body/content-type/provider-route coverage now looks substantially addressed.
- The test-coverage/performance passes drove the resource-bound, runtime-policy, and migration-test findings above.

gautam-achieveai:
[gautam-achieveai's bot] **Addressed** all 8 findings from the deep re-review of `b4b79b1` (commits `a2e556b`, `195dfbb`, `2dd4fbb`; `main` also merged in `7ac906a`).

**HIGH**
1. **Target-PR checkout (§3)** — `ContextReady` now clones/fetches the target repo into its own `/work/target` checkout via `GitRunner` and runs `SubmoduleInitializer` selectively under the per-run policy, then diffs from that checkout (not the ReviewBot retention checkout).
2. **Per-run OperationPolicy (§4)** — `DaemonOperationPolicy.BuildForRun` builds the scope from the concrete review (target/fork/ReviewBot repos + API host); provider/publisher HTTP clients flow through `PolicyEnforcedHttpClientFactory` so every call is route-scoped (repo path, not `/`). Tested against the real DI wiring.
3. **ReviewBot clone/validate** — the daemon clones/validates the configured `ReviewBotRepoUrl` checkout (`ReviewBotCheckout`) and fails fast with classified diagnostics on a missing/malformed skeleton.
4. **Resource bounds (§15)** — `SandboxOrchestrator` has a configurable per-command timeout; `SandboxLimits` caps stdout/stderr/diff/artifact sizes with truncation markers before persist/agent-feed.

**MEDIUM**
5. **Pagination** — GitHub `Link rel="next"` + ADO `x-ms-continuationtoken`, bounded to 10 pages/poll.
6. **Poison isolation** — `PollOnceAsync` isolates each target and each PR; a failure is logged and skipped (the run is left `RetryPending`), the cursor advances only on a clean pass.
7. **HTTP resilience** — a `RetryHandler` retries transient 429/5xx with backoff + `Retry-After`; reads retry freely, POSTs retry only on 429 (never an ambiguous 5xx — left to the ReviewPoster idempotency backstop).
8. **Migration tests** — a failing migration rolls back atomically (`user_version` unchanged); concurrent migrators serialize (`BEGIN IMMEDIATE` + `busy_timeout`) and converge.

Build 0 warnings; daemon **293/293**; LmStreaming unit **359/359**, E2E 76/76 (2 pre-existing `SandboxWorkspaceGateway` env-400s).

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -0,0 +1,521 @@
...
+                .RunAsync(["clone", targetRemote, TargetRoot], workingDirectory: null, cancellationToken)
+                .ConfigureAwait(false);
+            if (!clone.Succeeded)
+            {
+                throw new InvalidOperationException(
+                    $"Cloning the target repo for run {run.Id} failed (exit {clone.ExitCode}): {clone.Stderr}");
+            }
+        }
+
+        // Fetch the exact base + head commits the PR identity references (a fork/branch commit may not be
```
gautam-achieveai, line 203: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Correctness / fork PR support)

Fork PRs still cannot be checked out. `GitHubPrProvider` records only the PR number plus base/head SHAs, but not the head repository/clone URL. The executor then fetches both commits from the target repo's `origin` only:

```csharp
await git.RunAsync(["-C", TargetRoot, "fetch", "origin", run.BaseSha, run.HeadSha], ...)
```

For a PR whose head commit lives only in a fork, that fetch fails before diff generation. The per-run policy also sets `ForkHost` / `ForkRepoPath` to `null`, so there is no allowed fork remote path despite #118's approved plan/test strategy calling out fork-PR fetches.

**Suggestion:** Extend `PullRequestDescriptor` / `ReviewRun` (or the context metadata) to carry the PR head remote identity from GitHub/ADO. For fork PRs, build the run policy with `ForkHost` / `ForkRepoPath`, add/fetch that remote with `SandboxOperation.FetchForkHead`, and add a regression test where the head SHA exists only in a fork.

samples/CodeReviewDaemon.Sample/Workspace/Sandbox/SandboxOrchestrator.cs
```diff
@@ -0,0 +1,188 @@
...
+        catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
+        {
+            _logger.LogWarning(
+                "Sandbox command exceeded the {Timeout} per-command timeout and was cancelled.",
+                _limits.CommandTimeout);
+            throw new TimeoutException(
+                $"Sandbox command exceeded the configured {_limits.CommandTimeout} timeout.");
+        }
+
+        var text = ExtractText(response);
```
gautam-achieveai, line 96: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Performance / resource bounds)

The sandbox output cap is applied only after the MCP tool response has already been fully materialized. `RunAsync` calls `ExtractText(response)`, `ExtractText` joins every `TextContentBlock` into one string, and only then does `ParseResult` call `limits.CapOutput(...)`.

That means a malicious or accidental command can still force the daemon/MCP client to allocate the entire gateway response before truncation. The H4 resource bound therefore limits what is persisted/fed to the agent, but it does **not** prevent OOM/resource abuse.

**Suggestion:** Enforce the cap before or while materializing output: use a gateway-side max-output option if available, bounded/streaming copy over content blocks, or a bounded capture wrapper that stops reading after the configured limit and appends the truncation marker.

samples/CodeReviewDaemon.Sample/Orchestration/GitHubPrProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        var owner = request.Repo.OrgOrOwner;
+        var repo = request.Repo.RepoName;
+
+        var pullRequests = new List<PullRequestDescriptor>();
+        string? highWaterMark = null;
+
+        // Follow GitHub's Link rel="next" pagination (M5): the first page fixes the query, each subsequent
+        // page URL comes verbatim from the previous response's Link header. Bounded by MaxPages per poll.
+        var url = $"{BaseUrl}/repos/{owner}/{repo}/pulls?state=open&sort=updated&direction=desc&per_page=100";
```
gautam-achieveai, line 50: [gautam-achieveai's bot] **MEDIUM** (Performance / polling scalability)

Provider cursors are persisted but not consumed. GitHub always starts from page 1 and follows at most 10 `Link` pages; ADO always starts with `continuationToken = null` and follows at most 10 continuation pages. The saved `HighWaterMark` is not used to filter or resume subsequent polls.

On large repos this repeatedly fetches and reprocesses the same first window every poll interval, while never progressing beyond that window.

**Suggestion:** Either persist and honor the provider continuation/next-page cursor, or use the high-water mark in the request/query/filtering logic so subsequent polls skip already-seen PRs while still periodically resyncing for correctness.

samples/CodeReviewDaemon.Sample/Persistence/ReviewStore.cs
```diff
@@ -0,0 +1,433 @@
...
+        _ = command.Parameters.AddWithValue("$schemaVersion", artifact.ArtifactSchemaVersion);
+        _ = command.Parameters.AddWithValue("$kind", artifact.ArtifactKind);
+        _ = command.Parameters.AddWithValue("$provider", artifact.Provider);
+        _ = command.Parameters.AddWithValue("$payload", artifact.Payload);
+        _ = command.Parameters.AddWithValue("$now", UtcNow());
+        var id = Convert.ToInt64(command.ExecuteScalar());
+        return artifact with { Id = id };
+    }
+
+    public IReadOnlyList<ReviewArtifact> GetArtifacts(long reviewRunId)
```
gautam-achieveai, line 366: [gautam-achieveai's bot] **MEDIUM** (Performance / persistence efficiency)

Later stages read one latest artifact by loading every artifact row and full payload for the run into memory, then filtering with `.LastOrDefault(...)`. Artifacts include bounded diffs up to 2 MiB plus review/judge/variant payloads, and retries can leave multiple artifacts for the same stage/kind.

This makes each later stage pay avoidable SQLite I/O and heap cost.

**Suggestion:** Add a targeted store method/query such as:

```sql
SELECT *
FROM review_artifact
WHERE review_run_id = $runId AND artifact_kind = $kind
ORDER BY id DESC
LIMIT 1;
```

Use that from `ReadArtifactPayload`, and consider bounding review text payloads as well as diffs.

gautam-achieveai:
[gautam-achieveai's bot] **MEDIUM** (Maintainability / source hygiene)

This file contains literal NUL bytes in string literals (`"\0no-reviewbot-remote"` and `"/\0"`). Git now treats the C# source file as binary (`git diff --numstat` reports `- -` for this path), which breaks normal text diffs, inline review, search/indexing, and source-processing tools.

**Suggestion:** Replace the literal NUL bytes with text-only sentinels such as `"invalid-reviewbot-remote.invalid"` / `"/__no_reviewbot_remote__"`, or use escaped `"\u0000..."` if NUL semantics are truly required without embedding raw NUL bytes in the source file.

_(Could not anchor this inline at `samples/CodeReviewDaemon.Sample/Workspace/DaemonOperationPolicy.cs:126`; posting as a general follow-up comment.)_

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #121

I reviewed the linked implementation PR (#121) for issue #118 as a follow-up re-review. PR #121 is already merged, so these should be treated as remediation items for a follow-up PR rather than pre-merge gate comments.

## Verdict: REQUEST CHANGES / follow-up remediation required

### Findings posted

- **[BLOCKER] HIGH** — Fork PR checkout remains incomplete: the executor fetches the head SHA only from the target repo origin, while the provider/run metadata does not carry a fork head remote and the policy scopes `ForkHost` / `ForkRepoPath` to `null`.
- **[BLOCKER] HIGH** — Sandbox output capping happens after the full MCP response is already materialized, so the resource bound does not prevent OOM/resource abuse.
- **MEDIUM** — Provider cursors are persisted but not consumed, so large repos repeatedly fetch/reprocess the first page window.
- **MEDIUM** — Latest-artifact reads load all artifact rows/full payloads and then filter in memory.
- **MEDIUM** — `DaemonOperationPolicy.cs` contains literal NUL bytes; Git treats it as binary, which breaks normal text diffs/review/source tooling. GitHub could not anchor this inline because the path is treated as binary, so I posted it as a general follow-up comment.

Local verification note: I attempted `dotnet build LmDotnetTools.sln --no-restore`, but this sandbox does not have `dotnet` installed (`dotnet: not found`). GitHub checks were green before merge, but the findings above are logic/design/resource-bound issues that compile checks would not necessarily catch.

PR #127, opened 2026-07-01
Harden CopilotAnthropicProxy sample: opus model selection, MCP passthrough, strip unsupported request fields

### LINKED ISSUE #126: Harden CopilotAnthropicProxy sample: correct opus model selection, MCP passthrough, strip unsupported request fields
## What changed
- Fixed opus model discovery to pick the highest-version `claude-opus-*` id when Copilot lists multiple concurrent variants (previously picked the first/oldest match).
- Added transparent MCP (Streamable HTTP) passthrough on `/mcp` and `/mcp/readonly`, reusing the same Copilot-authenticated HttpClient as `/v1/messages`.
- Stripped the `anthropic-beta` header and the `context_management` body field before forwarding to Copilot — Copilot's backend 400s the entire request if it doesn't recognize every beta value, and 400s outright if `context_management` is present at all. Both are sent routinely by Claude Code.

## Why
Copilot's Anthropic-compatible backend has a stricter/older request schema than what Claude Code (a modern Anthropic API client) sends. Without these fixes, real Claude Code sessions against the proxy fail with HTTP 400s, and multi-opus-variant discovery could silently pin to a stale model.

## Files affected
`samples/CopilotAnthropicProxy.Sample/Program.cs`, `samples/CopilotAnthropicProxy.Sample/README.md`, `tests/CopilotAnthropicProxy.Tests/**`

## Testing
82/82 tests passing in `CopilotAnthropicProxy.Tests`; MCP passthrough and beta/context_management stripping additionally verified live against the real Copilot backend.

gautam-achieveai:
## Summary
Three incremental hardening fixes to the `CopilotAnthropicProxy.Sample` reverse proxy so real Claude Code sessions work reliably against GitHub Copilot's Anthropic-compatible backend:
- Model discovery now picks the highest-version `claude-opus-*` id instead of the first match when Copilot lists multiple concurrent opus variants.
- Transparent MCP (Streamable HTTP) passthrough added on `/mcp` and `/mcp/readonly`, reusing the existing Copilot-authenticated transport.
- The `anthropic-beta` header and `context_management` body field are now stripped before forwarding — Copilot's backend 400s the whole request if it doesn't recognize every beta value, and 400s outright if `context_management` is present at all. Both are sent routinely by Claude Code.

## Changes
- Opus model resolver: parse and compare version suffixes so the numerically highest opus id always wins; added tests driven off a sanitized real Copilot `/models` response.
- MCP passthrough: byte-level reverse proxy for GET/POST/DELETE on `/mcp` and `/mcp/readonly`, forwarding inbound headers verbatim except `Authorization` (overridden with Copilot credentials) and a small hop-by-hop set.
- Request compatibility: `ApplyRequestHeaderAllowlist` no longer forwards `anthropic-beta`; `ProxyModelResolver.TryRewriteModel` now also strips `context_management` from the JSON body alongside the existing `model` rewrite.
- README updated to document the new "Known request incompatibilities" behavior and MCP exposure; removed the now-inaccurate "no anthropic-beta filter" non-goal.

## Testing
- `dotnet test tests/CopilotAnthropicProxy.Tests` — 82/82 passing.
- MCP passthrough verified live against the real Copilot MCP server (initialize, session-scoped `tools/list`, session termination, unsupported-method passthrough).
- anthropic-beta/context_management stripping verified against the exact failing request captures that motivated the fix (Copilot previously 400'd with `"unsupported beta header(s): ..."` and `"context_management: Extra inputs are not permitted"`).

## Related Issues
Fixes #126

### DIFF
--- a/samples/CopilotAnthropicProxy.Sample/Program.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Program.cs
@@ -450,47 +450,61 @@ public static class ProxyModelResolver
     /// <summary>
     ///     Extracts ids for models whose <c>supported_endpoints</c> includes <c>/v1/messages</c> — the only
     ///     ones this Anthropic-Messages-shaped proxy can forward to. Copilot also serves GPT/Gemini models
-    ///     that only support <c>/responses</c> or <c>/chat/completions</c>; those are excluded.
+    ///     that only support <c>/responses</c> or <c>/chat/completions</c>; those are excluded. If no entry
+    ///     in the response carries <c>supported_endpoints</c> metadata at all (an older or alternative
+    ///     Copilot-compatible <c>/models</c> shape), falls back to <see cref="ParseModelIds"/> — id-only,
+    ///     no endpoint filtering — rather than treating every model as unsupported and failing startup.
     /// </summary>
     public static IReadOnlyList<string> ParseMessagesCapableModelIds(string json)
     {
-        var ids = new List<string>();
         using var doc = JsonDocument.Parse(json);
         var root = doc.RootElement;
         var list = root.ValueKind == JsonValueKind.Object && root.TryGetProperty("data", out var data) ? data : root;
 
-        if (list.ValueKind == JsonValueKind.Array)
+        if (list.ValueKind != JsonValueKind.Array)
         {
-            foreach (var item in list.EnumerateArray())
+            return [];
+        }
+
+        var items = list.EnumerateArray().ToList();
+        var anyEndpointMetadata = items.Any(item =>
+            item.ValueKind == JsonValueKind.Object && item.TryGetProperty("supported_endpoints", out _)
+        );
+        if (!anyEndpointMetadata)
+        {
+            return ParseModelIds(json);
+        }
+
+        var ids = new List<string>();
+        foreach (var item in items)
+        {
+            if (
+                item.ValueKind != JsonValueKind.Object
+                || !item.TryGetProperty("id", out var idEl)
+                || idEl.ValueKind != JsonValueKind.String
+            )
             {
-                if (
-                    item.ValueKind != JsonValueKind.Object
-                    || !item.TryGetProperty("id", out var idEl)
-                    || idEl.ValueKind != JsonValueKind.String
-                )
-                {
-                    continue;
-                }
+                continue;
+            }
 
-                var id = idEl.GetString();
-                if (string.IsNullOrWhiteSpace(id))
-                {
-                    continue;
-                }
+            var id = idEl.GetString();
+            if (string.IsNullOrWhiteSpace(id))
+            {
+                continue;
+            }
 
-                if (
-                    item.TryGetProperty("supported_endpoints", out var endpoints)
-                    && endpoints.ValueKind == JsonValueKind.Array
-                    && endpoints
-                        .EnumerateArray()
-                        .Any(e =>
-                            e.ValueKind == JsonValueKind.String
-                            && string.Equals(e.GetString(), "/v1/messages", StringComparison.OrdinalIgnoreCase)
-                        )
-                )
-                {
-                    ids.Add(id);
-                }
+            if (
+                item.TryGetProperty("supported_endpoints", out var endpoints)
+                && endpoints.ValueKind == JsonValueKind.Array
+                && endpoints
+                    .EnumerateArray()
+                    .Any(e =>
+                        e.ValueKind == JsonValueKind.String
+                        && string.Equals(e.GetString(), "/v1/messages", StringComparison.OrdinalIgnoreCase)
+                    )
+            )
+            {
+                ids.Add(id);
             }
         }
 
@@ -768,7 +782,15 @@ internal static class ProxyHttp
 
             ctx.Features.Get<IHttpResponseBodyFeature>()?.DisableBuffering();
 
-            await CopyBodyAsync(ctx, upstream, idleTimeout, idleCts, linked, logger);
+            await CopyBodyAsync(
+                ctx,
+                upstream,
+                idleTimeout,
+                idleCts,
+                linked,
+                logger,
+                (c, status, message) => WriteAnthropicErrorAsync(c, status, "api_error", message)
+            );
 
             logger.LogInformation(
                 "{Method} {Path} model {IncomingModel} -> {ResolvedModel} stream={Stream} upstream={Status} {Elapsed}ms",
@@ -783,14 +805,20 @@ internal static class ProxyHttp
         }
     }
 
-    /// <summary>Streams the upstream body to the client with an explicit, incrementally-flushed loop.</summary>
+    /// <summary>
+    ///     Streams the upstream body to the client with an explicit, incrementally-flushed loop.
+    ///     <paramref name="writePreStartError" /> writes a protocol-shaped error envelope (Anthropic-JSON
+    ///     vs. JSON-RPC) if the upstream fails before any bytes reached the client — shared by both the
+    ///     <c>/v1/messages</c> and <c>/mcp</c> paths, which disagree on error shape.
+    /// </summary>
     internal static async Task CopyBodyAsync(
         HttpContext ctx,
         HttpResponseMessage upstream,
         TimeSpan idleTimeout,
         CancellationTokenSource idleCts,
         CancellationTokenSource linked,
-        ILogger logger
+        ILogger logger,
+        Func<HttpContext, int, string, Task> writePreStartError
     )
     {
         var buffer = ArrayPool<byte>.Shared.Rent(BufferSize);
@@ -829,10 +857,9 @@ internal static class ProxyHttp
                         logger.LogWarning("Mid-stream upstream failure: {Reason}", ex.Message);
                         if (!ctx.Response.HasStarted)
                         {
-                            await WriteAnthropicErrorAsync(
+                            await writePreStartError(
                                 ctx,
                                 StatusCodes.Status502BadGateway,
-                                "api_error",
                                 "The upstream Copilot stream failed before any data was received."
                             );
                         }
@@ -969,9 +996,11 @@ internal static class ProxyHttp
 /// </summary>
 internal static class ProxyMcp
 {
-    // Everything is forwarded verbatim EXCEPT: Authorization (Copilot auth is attached outbound by
-    // CopilotHeadersHandler instead — the caller's own auth, if any, is never forwarded) and a small
-    // set of hop-by-hop/framing headers that .NET's HttpClient must own (Host, Content-Length,
+    // Everything is forwarded verbatim EXCEPT: Authorization/credential headers (Copilot auth is
+    // attached outbound by CopilotHeadersHandler instead — the caller's own auth, if any, is never
+    // forwarded), the Copilot transport/tracking headers CopilotHeadersHandler owns (it only sets
+    // these when missing, so a caller-supplied value would otherwise silently override them), and a
+    // small set of hop-by-hop/framing headers that .NET's HttpClient must own (Host, Content-Length,
     // Content-Type are handled explicitly, Connection/Transfer-Encoding/etc. are per-hop). Accept-Encoding
     // is also excluded: the shared HttpClient never negotiates compression (SocketsHttpHandler.
     // AutomaticDecompression = None) and CopyResponseHeaders always strips Content-Encoding from the
@@ -979,6 +1008,9 @@ internal static class ProxyMcp
     private static readonly HashSet<string> ExcludedRequestHeaders = new(StringComparer.OrdinalIgnoreCase)
     {
         "Authorization",
+        "x-api-key",
+        "Cookie",
+        "Proxy-Authorization",
         "Host",
         "Content-Length",
         "Content-Type",
@@ -989,6 +1021,15 @@ internal static class ProxyMcp
         "TE",
         "Trailer",
         "Accept-Encoding",
+        "User-Agent",
+        "copilot-integration-id",
+        "editor-version",
+        "x-github-api-version",
+        "x-client-machine-id",
+        "x-client-session-id",
+        "x-interaction-id",
+        "x-interaction-type",
+        "x-initiator",
     };
 
     /// <summary>Forwards GET/POST/DELETE on the MCP endpoint to Copilot and streams the response back.</summary>
@@ -1095,7 +1136,7 @@ internal static class ProxyMcp
 
             ctx.Features.Get<IHttpResponseBodyFeature>()?.DisableBuffering();
 
-            await ProxyHttp.CopyBodyAsync(ctx, upstream, idleTimeout, idleCts, linked, logger);
+            await ProxyHttp.CopyBodyAsync(ctx, upstream, idleTimeout, idleCts, linked, logger, WriteMcpErrorAsync);
 
             logger.LogInformation(
                 "{Method} {Path} mcp-session={SessionId} upstream={Status} {Elapsed}ms",
@@ -1108,7 +1149,11 @@ internal static class ProxyMcp
         }
     }
 
-    /// <summary>Forwards every inbound header verbatim except <see cref="ExcludedRequestHeaders"/> (incl. auth).</summary>
+    /// <summary>
+    ///     Forwards every inbound header verbatim except <see cref="ExcludedRequestHeaders"/> — credentials
+    ///     (<c>Authorization</c>, <c>x-api-key</c>, <c>Cookie</c>, <c>Proxy-Authorization</c>), the Copilot
+    ///     transport/tracking headers <c>CopilotHeadersHandler</c> owns, and hop-by-hop/framing headers.
+    /// </summary>
     private static void ApplyRequestHeaderAllowlist(IHeaderDictionary inbound, HttpRequestMessage upstream)
     {
         foreach (var header in inbound)
--- a/samples/CopilotAnthropicProxy.Sample/Program.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Program.cs
@@ -18,6 +18,8 @@ using Microsoft.Extensions.Primitives;
 // and forwards it to GitHub Copilot. It rewrites ONLY the JSON `model` field to a
 // configured Copilot Claude (Opus) id, attaches Copilot auth/headers via the proven
 // GithubCopilotProvider transport, and streams the SSE response back as raw bytes.
+// It also exposes Copilot's MCP server (Streamable HTTP transport) as a transparent
+// byte-level proxy on /mcp and /mcp/readonly, with Copilot auth attached the same way.
 //
 // Point Claude Code (or any Anthropic-Messages client) at it via ANTHROPIC_BASE_URL.
 // SECURITY: binds to loopback only and attaches the developer's Copilot credentials
@@ -174,6 +176,10 @@ app.MapPost(
     "/v1/messages/count_tokens",
     ctx => ProxyHttp.ForwardAsync(ctx, catalog, config.IdleTimeout, isCountTokens: true)
 );
+
+// GET/POST/DELETE /mcp and /mcp/readonly — transparent MCP (Streamable HTTP) proxy.
+app.MapMethods("/mcp", ["GET", "POST", "DELETE"], ctx => ProxyMcp.ForwardAsync(ctx, config.IdleTimeout));
+app.MapMethods("/mcp/readonly", ["GET", "POST", "DELETE"], ctx => ProxyMcp.ForwardAsync(ctx, config.IdleTimeout));
 
 // Unknown route -> Anthropic-shaped 404.
 app.MapFallback(ctx =>
@@ -773,7 +779,7 @@ internal static class ProxyHttp
     }
 
     /// <summary>Streams the upstream body to the client with an explicit, incrementally-flushed loop.</summary>
-    private static async Task CopyBodyAsync(
+    internal static async Task CopyBodyAsync(
         HttpContext ctx,
         HttpResponseMessage upstream,
         TimeSpan idleTimeout,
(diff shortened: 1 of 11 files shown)


samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -674,35 +923,230 @@ public static async Task WriteAnthropicErrorAsync(HttpContext ctx, int status, s
...
+                (int)upstream.StatusCode,
+                stopwatch.ElapsedMilliseconds
+            );
+        }
+    }
+
+    /// <summary>Forwards every inbound header verbatim except <see cref="ExcludedRequestHeaders"/> (incl. auth).</summary>
+    private static void ApplyRequestHeaderAllowlist(IHeaderDictionary inbound, HttpRequestMessage upstream)
+    {
+        foreach (var header in inbound)
```
gautam-achieveai, line 1178: [Gautam's bot] [BLOCKER] **HIGH** (Security / Architecture)

The MCP proxy's broad header passthrough forwards caller-controlled credential and transport metadata headers to Copilot. `ProxyMcp.ApplyRequestHeaderAllowlist` forwards every inbound header not in the small framing/auth exclusion list, and the test now pins `x-api-key` forwarding. Because `CopilotHeadersHandler` adds Copilot transport headers only when missing, a local caller can also pre-set provider-owned values such as `User-Agent`, `copilot-integration-id`, `x-client-session-id`, `x-interaction-id`, or `x-initiator` and override the values the Copilot transport is supposed to own.

**Suggestion:** Keep MCP protocol extensibility, but do not forward arbitrary credentials/provider metadata. Use an explicit allowlist (for example `Mcp-*`, `Last-Event-ID`, relevant `X-MCP-*`, `Accept`) plus a deny/overwrite set for `x-api-key`, `Cookie`, `Proxy-Authorization`, `User-Agent`, `copilot-*`, `x-client-*`, `x-interaction-*`, `x-initiator`, etc.; alternatively make `CopilotHeadersHandler` overwrite Copilot-owned headers after passthrough.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -293,39 +338,83 @@ private static bool IsAllowedOrigin(string? origin)
...
         ArgumentNullException.ThrowIfNull(client);
 
         if (!string.IsNullOrWhiteSpace(modelOverride))
         {
-            return modelOverride.Trim();
+            var pinned = modelOverride.Trim();
+            return new ProxyModelCatalog(pinned, [pinned]);
         }
 
         using var response = await client.GetAsync("/models", cancellationToken);
```
gautam-achieveai, line 378: [Gautam's bot] [BLOCKER] **HIGH** (Wire compatibility)

This changes model discovery from id-based to requiring `supported_endpoints`, which can make the proxy fail startup against older or alternative Copilot-compatible `/models` responses that only expose `id`. `ResolveAsync` now builds the catalog exclusively from `ParseMessagesCapableModelIds`; that parser only keeps entries with `supported_endpoints` containing `/v1/messages`, and startup fails if no opus id survives. The previous code still had `ParseModelIds` behavior that accepted id-only model lists.

**Suggestion:** Keep the `/v1/messages` filter when endpoint metadata is present, but preserve legacy compatibility when it is absent. For example, if no entries contain endpoint metadata, fall back to `ParseModelIds`; or include Claude ids with missing `supported_endpoints` as "unknown/legacy" with a warning rather than treating them as unsupported and failing startup.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -293,39 +338,83 @@ private static bool IsAllowedOrigin(string? origin)
 // Model resolution + request rewrite (pure where possible)
 // =============================================================================
 
-/// <summary>Resolves the outbound Copilot model id and rewrites the inbound request's model field.</summary>
+/// <summary>
```
gautam-achieveai, line 348: [Gautam's bot] **MEDIUM** (Scope / Design)

The "pick the newest opus default" hardening also introduces externally visible multi-model catalog/passthrough behavior. The PR now exposes every discovered `/v1/messages`-capable model from `/v1/models` and `SelectOutboundModel` lets a request's `model` pass through unchanged whenever it matches the discovered catalog. The README documents this as users being able to select non-default discovered models.

That may be a useful feature, but it is broader than the default-opus selection fix and increases the behavior surface of this hardening PR.

**Suggestion:** Either keep this PR constrained to resolving one highest-version opus model and always rewriting to it, or explicitly split/track the multi-model passthrough behavior as its own reviewed feature so its compatibility and client-selection semantics are intentional.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -616,7 +866,14 @@ await WriteAnthropicErrorAsync(
...
-    /// <summary>Copies the positive request-header allowlist (anthropic-version + anthropic-beta) only.</summary>
+    /// <summary>
+    ///     Copies the positive request-header allowlist: only <c>anthropic-version</c> is forwarded (a
+    ///     default is injected when absent). <c>anthropic-beta</c> is intentionally dropped, never
+    ///     forwarded — Copilot's backend rejects the whole request with a 400 if it doesn't recognize
+    ///     every single value in the header (e.g. Claude Code's evolving beta set routinely includes
+    ///     values Copilot hasn't caught up to), so passing it through breaks requests Copilot would
+    ///     otherwise serve fine.
+    /// </summary>
     public static void ApplyRequestHeaderAllowlist(IHeaderDictionary inbound, HttpRequestMessage upstream)
```
gautam-achieveai, line 911: [Gautam's bot] **MEDIUM** (Wire compatibility)

The proxy drops `anthropic-beta` unconditionally but otherwise forwards the body after only rewriting `model` and removing `context_management`. If a client sends a beta-dependent body field that Copilot actually supports when declared, the upstream now receives the beta-shaped payload without the dialect marker, which can silently downgrade semantics or turn into a confusing validation failure.

**Suggestion:** Prefer filtering `anthropic-beta` value-by-value, stripping only known Copilot-incompatible beta values. If the header must be dropped wholesale, pair it with validation/removal of beta-dependent body fields so the upstream never receives an undeclared beta payload.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -674,35 +923,230 @@ public static async Task WriteAnthropicErrorAsync(HttpContext ctx, int status, s
...
+            var isSse = string.Equals(contentType?.MediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
+            if (isSse)
+            {
+                ctx.Response.Headers["X-Accel-Buffering"] = "no";
+                ctx.Response.Headers.CacheControl = "no-cache";
+            }
+
+            ctx.Features.Get<IHttpResponseBodyFeature>()?.DisableBuffering();
+
+            await ProxyHttp.CopyBodyAsync(ctx, upstream, idleTimeout, idleCts, linked, logger);
```
gautam-achieveai: [Gautam's bot] **LOW** (Protocol correctness)

`ProxyMcp.ForwardAsync` reuses `ProxyHttp.CopyBodyAsync`, but that helper has an Anthropic-specific pre-start failure path: if upstream body reading fails before any bytes are written, it calls `WriteAnthropicErrorAsync`. The MCP proxy separately defines `WriteMcpErrorAsync`, so proxy-origin MCP failures are intended to use a JSON-RPC-shaped error envelope, but this path would return an Anthropic-shaped `{ type: "error", error: ... }` response.

**Suggestion:** Extract the byte-copy loop into a protocol-neutral helper that accepts a pre-start failure writer/callback, or add an MCP-specific wrapper so MCP stream failures use `WriteMcpErrorAsync` while `/v1/messages` keeps the Anthropic error shape.

gautam-achieveai:
[Gautam's bot] # PR Review: Harden CopilotAnthropicProxy sample: opus model selection, MCP passthrough, strip unsupported request fields

**Verdict: REQUEST CHANGES**

I reviewed PR #127 against latest `origin/main` in isolated worktree `/workspace/LmDotNetTools/pr127-impl-review-084423` (base `28ba398`, head `9537611`). I used parallel review lenses for over-engineering, blind spots, architecture/design, code simplicity, test coverage, performance, wire compatibility, EUII, duplication, and temp artifacts, then ran severity grading.

GitHub CI currently reports **Build and test: passing**. Local verification was attempted with `dotnet test tests/CopilotAnthropicProxy.Tests/CopilotAnthropicProxy.Tests.csproj`, but this sandbox does not have `dotnet` installed.

## Blocking findings

1. **HIGH** — MCP header passthrough forwards caller-controlled credentials/metadata and can override Copilot-owned headers.
2. **HIGH** — Model discovery now requires `supported_endpoints`, creating a startup-breaking compatibility regression for legacy/id-only `/models` responses.

## Non-blocking findings

3. **MEDIUM** — The opus default fix expands into externally visible multi-model catalog/passthrough behavior.
4. **MEDIUM** — `anthropic-beta` is stripped while beta-shaped body fields may still pass through.
5. **LOW** — MCP pre-start stream failure can return an Anthropic-shaped error envelope.

Inline comments have details and suggested fixes.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Fixed in a4a0974. Extended `ExcludedRequestHeaders` from 11 to 23 entries: it now also excludes `x-api-key`, `Cookie`, `Proxy-Authorization`, `User-Agent`, and every Copilot-owned transport/tracking header (`copilot-integration-id`, `editor-version`, `x-github-api-version`, `x-client-machine-id`, `x-client-session-id`, `x-interaction-id`, `x-interaction-type`, `x-initiator`) — i.e. exactly the set `CopilotRequestHeaders.Build()` owns, since `CopilotHeadersHandler.AddIfMissing` only sets its own value when the header is absent, so any caller-supplied value previously survived to Copilot untouched.

Rewrote `McpProxyTests` to assert spoofed values for these headers never reach the upstream request (while confirming arbitrary future MCP headers still pass through, since this stays a denylist, not an allowlist).

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Fixed in a4a0974. `ParseMessagesCapableModelIds` now checks whether *any* entry in the response carries `supported_endpoints` metadata at all; if none do, it falls back to `ParseModelIds` (id-only, no endpoint filtering) instead of returning an empty list and failing startup. Mixed responses (some entries tagged, some not) still filter per-entry as before — only the "nobody has this field" case triggers the legacy fallback.

Worth noting: `tests/CopilotAnthropicProxy.Tests/Fixtures/copilot-models-real-response.json` (a real captured response from `api.enterprise.githubcopilot.com`) shows every model entry there does include `supported_endpoints` today, so this was a theoretical hardening rather than an observed break — but the fallback is cheap insurance against an older/alternative Copilot-compatible `/models` shape. Added `ParseMessagesCapableModelIds_falls_back_to_id_only_when_no_entry_has_endpoint_metadata` and a matching `ResolveAsync` test.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: This is intentional, already-shipped, already-documented design — not new scope introduced by this PR's actual fix (stripping `anthropic-beta` and `context_management`).

The highest-version-opus selection was a separate, prior commit (0d2d014, "pick highest-version opus model from Copilot discovery") bundled into this branch. Picking the *highest-version* opus among several concurrent ones inherently requires discovering the full model catalog first — you can't pick the max of a set without enumerating it. Once the full catalog is in hand, letting a client's `model` field pass through unchanged when it matches a discovered id (rather than only ever forwarding the resolved default) is the deliberate design documented in the README's "Choosing the model" section, not an accidental side effect of this hardening.

If you'd like the opus-version-picking and the anthropic-beta/context_management stripping split into separate PRs going forward I'm open to that, but re-scoping this one now would mean reverting already-merged, documented behavior rather than fixing a defect.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Partial pushback. The one concretely observed break this PR fixes — `context_management` being sent under a beta the upstream doesn't declare support for — is already handled by stripping that field from the body regardless of the beta header (see `TryRewriteModel`'s `context_management` removal, covered by `ModelRewriteTests`).

Per-value beta filtering (forward the header but drop only unrecognized values) was deliberately rejected already — the README's "Known request incompatibilities" section explains why: Claude Code's beta list changes frequently and includes values Copilot doesn't recognize yet, so a static allowlist of "known-good" betas would go stale and reintroduce the same failure mode this PR fixes. Copilot also rejects the whole request if *any* value in `anthropic-beta` is unrecognized, so partial forwarding isn't actually available to us — it's drop-the-whole-header or fail-the-request.

Beyond `context_management`, I don't have a second concrete case of a beta-gated body field Copilot actually supports when declared. Speculatively stripping more fields without an observed failure feels like disproportionate scope for this dev sample. Happy to add handling for a specific field the moment one surfaces — could you point me to one if you have one in mind?

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Fixed in a4a0974, matching the suggested approach. `ProxyHttp.CopyBodyAsync` now takes a `Func<HttpContext, int, string, Task> writePreStartError` delegate instead of hardcoding `WriteAnthropicErrorAsync`. The `/v1/messages` call site wraps `WriteAnthropicErrorAsync` with the `"api_error"` type; the MCP call site passes `WriteMcpErrorAsync` directly (its signature already matches the delegate). Added `Mid_stream_upstream_failure_before_any_byte_returns_a_json_rpc_shaped_error` to `McpProxyTests` asserting the JSON-RPC envelope (`"jsonrpc":"2.0"`) and the absence of the Anthropic-shaped `"type":"error"` envelope.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -90,70 +95,103 @@
...
+
+app.MapPost(
+    "/v1/messages/count_tokens",
+    ctx => ProxyHttp.ForwardAsync(ctx, catalog, config.IdleTimeout, isCountTokens: true)
+);
 
-app.MapPost("/v1/messages/count_tokens", ctx =>
-    ProxyHttp.ForwardAsync(ctx, resolvedModel, config.IdleTimeout, isCountTokens: true));
+// GET/POST/DELETE /mcp and /mcp/readonly — transparent MCP (Streamable HTTP) proxy.
+app.MapMethods("/mcp", ["GET", "POST", "DELETE"], ctx => ProxyMcp.ForwardAsync(ctx, config.IdleTimeout));
```
gautam-achieveai: [Gautam's bot] [BLOCKER] **HIGH** (Security / Architecture)

`/mcp` is enabled by default as the full write-capable Copilot MCP passthrough, and it uses the same Copilot-authenticated outbound transport while the README documents that `/mcp*` has no inbound auth. The loopback/host guard lowers remote exposure, but it does not prevent any accepted local caller from invoking write-capable Copilot MCP tools with the developer's credentials.

**Suggestion:** Make MCP exposure explicit/opt-in, and make write-capable `/mcp` opt-in separately from `/mcp/readonly` (for example `COPILOT_ANTHROPIC_ENABLE_MCP_WRITE=1`). Consider adding an inbound authorization boundary for `/mcp*` (ephemeral startup token or configured bearer secret), and add tests/README guidance proving the default is readonly/disabled unless the user opts into write tools.

gautam-achieveai: [Gautam's bot] **MEDIUM** (Security / Test Coverage)

The browser/loopback threat model for the credential-backed proxy is still under-specified. `IsAllowedOrigin` accepts any loopback origin host, and current tests cover same-origin `localhost:8787` but not browser-originated calls from another local port such as `http://localhost:3000` or `http://127.0.0.1:<other-port>`.

**Suggestion:** Define whether the policy is "loopback only" or strict same-origin including scheme/port, especially for `/mcp*`. If browser-originated credential-backed requests should be constrained, reject non-matching origin port/scheme for `/mcp*` and add tests for other loopback origins, `Sec-Fetch-Site: same-site`, and preflight/POST behavior.

tests/CopilotAnthropicProxy.Tests/ModelRewriteTests.cs
```diff
@@ -71,6 +71,38 @@ public void TryRewriteModel_returns_false_for_non_object_bodies(string body)
...
+        var body = Encoding.UTF8.GetBytes("{\"model\":\"x\",\"max_tokens\":5}");
+
+        var ok = ProxyModelResolver.TryRewriteModel(body, "opus-x", out var rewritten, out _);
+
+        ok.Should().BeTrue();
+        JsonNode.Parse(rewritten)!.AsObject().ContainsKey("context_management").Should().BeFalse();
+    }
+
     [Fact]
     public async Task Forwarded_request_body_has_the_configured_model()
```
gautam-achieveai, line 157: [Gautam's bot] **MEDIUM** (Testing / Compatibility)

This verifies that the helper removes `context_management`, and the forwarding test below verifies the configured model rewrite, but the actual compatibility bug is in what the upstream receives from `/v1/messages` and `/v1/messages/count_tokens`. I don't see an end-to-end test that sends a realistic Claude Code request through the proxy and asserts the fake upstream body omits `context_management` while preserving fields like `thinking`, `cache_control`, `tools`, and unknown fields.

**Suggestion:** Add E2E forwarding coverage for both `/v1/messages` and `/v1/messages/count_tokens` with a request containing `context_management`, `anthropic-beta`, and preserved Claude Code fields. Assert upstream sees no `anthropic-beta`, no top-level `context_management`, and still sees the preserved fields.

tests/CopilotAnthropicProxy.Tests/McpProxyTests.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        await using var factory = new ProxyWebAppFactory(
+            (req, ct) => throw new InvalidOperationException("upstream must not be called for an unsupported method")
+        );
+        using var client = factory.CreateClient();
+
+        using var request = new HttpRequestMessage(HttpMethod.Put, "/mcp/readonly") { Content = JsonRpc("{}") };
+        using var response = await client.SendAsync(request);
+
+        // Falls through to the shared Anthropic-shaped fallback 404, same as any other unmatched route/method.
```
gautam-achieveai: [Gautam's bot] **LOW** (Protocol Consistency)

Now that the app hosts both Anthropic Messages and MCP routes, this route-level rejection should probably use an MCP-consistent shape. Unsupported methods on `/mcp*` currently fall through to the shared Anthropic-shaped fallback 404, while proxy-origin MCP failures elsewhere use `WriteMcpErrorAsync`.

**Suggestion:** Add explicit method/fallback handling for `/mcp` and `/mcp/readonly` unsupported methods, returning either an MCP/JSON-RPC-shaped error or a deliberate `405`. Update this test to assert the intended body shape, not just the status.

gautam-achieveai:
[Gautam's bot] ## Re-Review Summary: PR #127

**Verdict: REQUEST CHANGES**

I reviewed the latest head `a4a0974` in isolated worktree `/workspace/LmDotNetTools/issue127-spec-review-191141`, using the PR-review workflow adapted for this spec/plan review. I used parallel lenses for over-engineering, blind spots, architecture/design, and code simplicity, then ran severity grading.

The earlier header-spoofing and legacy `/models` compatibility blockers appear addressed in this head. I found one new/remaining blocker around default MCP exposure.

## Blocking finding

1. **HIGH** — `/mcp` exposes write-capable Copilot MCP passthrough by default with no inbound auth while attaching the developer's Copilot credentials. Loopback binding reduces remote exposure, but does not address accepted local callers or browser/local-process credential-confusion risk.

## Non-blocking findings

2. **MEDIUM** — The browser/loopback origin policy for credential-backed MCP is under-specified and lacks tests for other localhost origins/ports.
3. **MEDIUM** — `context_management` stripping needs E2E forwarding coverage for `/v1/messages` and `/v1/messages/count_tokens`, not just helper-level coverage.
4. **LOW** — Unsupported methods on `/mcp*` fall through to Anthropic-shaped fallback rather than MCP-consistent route-level errors.

## Notes

- I did not rerun local .NET tests because this sandbox has no `dotnet` installed (`dotnet --info` exits 127: `dotnet: not found`).
- I left the multi-model catalog/passthrough concern as non-blocking/contextual after the existing pushback; it is externally visible scope, but I am not treating it as a blocker if the project intentionally accepts it.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Acknowledged as a design decision, not a bug: write-capable `/mcp` stays enabled by default.

This proxy's whole threat model (documented in README.md, "Local development only") is "loopback = trusted, no inbound auth needed" — that's already true for `/v1/messages`, which also attaches the developer's Copilot credentials and has no inbound auth. `/mcp` doesn't introduce a new trust boundary, it extends the existing one to a second capability surface.

The concrete gap you're describing — a browser page or other local process reaching this proxy cross-origin/cross-port while still satisfying "loopback" — is real, and is fixed in 86d54b6: `IsAllowedOrigin` now requires an exact host:port match (see reply on the next finding). That closes the specific mechanism by which an "accepted local caller" could reach `/mcp` without the developer intending it.

Given that fix, the remaining risk is "a process on the same machine, on the same port, deliberately impersonating this proxy's expected caller" — which is out of scope for a loopback-trust proxy of this kind (same as it already is for `/v1/messages`). Per the project owner, write-capable `/mcp` remains enabled by default.

gautam-achieveai: Fixed in 86d54b6: `IsAllowedOrigin` now requires the Origin's port to exactly match the configured port, not just any loopback address on any port. Previously `http://127.0.0.1:3000` would pass because it resolved to a loopback host — now it's rejected unless the port matches the proxy's own port, mirroring `IsAllowedHost`'s exact-port match.

This is shared by `ProxyGuard.IsAllowed`, which gates every route (`/v1/messages`, `/v1/models`, `/mcp`, `/mcp/readonly`), not just MCP.

Added a regression test: `HostGuardTests.Loopback_origin_on_a_different_port_is_rejected`.

tests/CopilotAnthropicProxy.Tests/ModelRewriteTests.cs
gautam-achieveai: Added E2E coverage in 86d54b6: `ModelRewriteTests.Forwarded_request_strips_beta_header_and_context_management_but_preserves_other_fields`, a `[Theory]` covering both `/v1/messages` and `/v1/messages/count_tokens`.

The request sends `anthropic-beta`, `context_management`, `thinking`, `system[].cache_control`, `tools`, and an unrecognized field, then asserts against the actual body/headers captured at the fake upstream: no `anthropic-beta` header, no top-level `context_management` key, and all the other fields preserved verbatim. Confirms `TryRewriteModel`'s stripping behavior (previously only unit-tested) actually holds through the full forward path for both routes, since they share the same `ForwardAsync`/`TryRewriteModel` call before the `isCountTokens` branch.

tests/CopilotAnthropicProxy.Tests/McpProxyTests.cs
gautam-achieveai: Fixed in 86d54b6: `/mcp` and `/mcp/readonly` now use `app.Map(...)` (all HTTP methods route there), and `ProxyMcp.ForwardAsync` itself checks the method against an allowlist (`GET`/`POST`/`DELETE`) up front, writing an MCP/JSON-RPC-shaped 405 via the existing `WriteMcpErrorAsync` helper for anything else. Previously an unsupported method (e.g. PUT) fell through `MapMethods`'s unmatched-method 404 to the shared `MapFallback`, which returns the Anthropic-shaped `{"type":"error",...}` envelope — inconsistent with every other MCP error path.

Updated `McpProxyTests.Unsupported_http_method_is_rejected_before_reaching_the_upstream` to assert `405 MethodNotAllowed` with a JSON-RPC-shaped body instead of the old `404`.

gautam-achieveai:
[Gautam's bot] ## Re-Review Summary: PR #127

**Verdict: REQUEST CHANGES**

Reviewed latest head `86d54b65` against the previously reviewed `a4a0974` delta. GitHub reports **Build and test: passing**. Local tests were not run because this sandbox still has no `dotnet` installed.

The latest delta addresses the non-blocking items I raised:

- Browser/loopback origin policy is tightened to exact configured port, with a regression test.
- Unsupported MCP methods now route through `ProxyMcp` and return MCP/JSON-RPC-shaped `405` responses, with a regression test.
- `/v1/messages` and `/v1/messages/count_tokens` now have E2E forwarding coverage proving `anthropic-beta` and `context_management` are stripped while other Claude Code fields are preserved.

The remaining blocker is unchanged in substance:

## Blocking finding

1. **HIGH** — write-capable `/mcp` is still enabled by default with no inbound auth while using the developer's Copilot credentials. The exact-port origin check improves browser cross-origin protection, but it does not make write-capable MCP opt-in or add an auth boundary for accepted local/non-browser callers.

**Required direction:** make MCP exposure explicit/opt-in, and make write-capable `/mcp` separately opt-in from `/mcp/readonly`, or add a real inbound authorization boundary for `/mcp*`.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: Re-affirming: this stays as-is. No further code change here.

The "required direction" — opt-in gating or an auth boundary for `/mcp*` — would be new scope beyond this proxy's existing, documented threat model, not a fix to something this PR introduced. README.md already states as a Non-goal: "No inbound auth, TLS, or CORS beyond the loopback + host/cross-site guard," and that's been true for `/v1/messages` (which also attaches the developer's live Copilot credentials) since #125, unchanged by this PR. `/mcp` extends the same already-accepted trust boundary to a second capability surface; it doesn't create a new one.

The concrete mechanism you originally flagged — a caller satisfying "loopback" from a different origin/port than the proxy's own — is closed by the exact-port origin check landed in 86d54b6. What remains is "a same-machine, same-port process intentionally impersonating the expected caller," which is out of scope for a loopback-trust local-dev proxy, identically for `/v1/messages` and `/mcp`.

Per the project owner: write-capable `/mcp` stays enabled by default, consistent with the rest of this proxy's design. This PR will proceed on that basis.

gautam-achieveai:
[Gautam's bot] ## Re-Review Summary: PR #127

**Verdict: REQUEST CHANGES**

I rechecked PR #127 and there is no new delta since the previous re-review: head is still `86d54b65`, and the diff from `86d54b65..86d54b65` is empty. GitHub still reports **Build and test: passing**.

The previously posted blocker remains unresolved: write-capable `/mcp` is still enabled by default with no inbound auth while using the developer's Copilot credentials. Please make MCP exposure explicit/opt-in, make write-capable `/mcp` separately opt-in from `/mcp/readonly`, or add a real inbound authorization boundary for `/mcp*`.

PR #129, opened 2026-07-01
feat(samples): expose Copilot models in dropdown, partitioned by vendor (#128)

### LINKED ISSUE #128: Expose GitHub Copilot models in sample dropdown, partitioned by vendor
## Summary

The LmStreaming.Sample provider dropdown only surfaces four hardcoded GitHub Copilot entries (`sonnet`, `haiku`, `gpt-5.5`, `gpt-5.5-mini`). It should instead expose **all** models available through the GitHub Copilot-based provider, discovered dynamically from the Copilot `/models` API and clearly partitioned into **Anthropic** and **OpenAI** groups.

## Motivation

Copilot serves a much larger, evolving catalog (multiple Claude Opus/Sonnet/Haiku versions, several GPT variants) than the four curated ids. A static list goes stale as Copilot ships new models and hides entitlements the developer actually has. Discovering the catalog at runtime keeps the dropdown current and lets users pick any routable model, organized by publisher.

## Scope

- Target: the LmStreaming.Sample provider dropdown (`GET /api/providers` → `ProviderSelector.vue`), not the LmConfig static catalog.
- Source: dynamic discovery from Copilot `GET /models` at startup.
- Included: only models whose vendor is Anthropic or OpenAI (normalizing `Azure OpenAI` → OpenAI) **and** that are routable via an existing transport (`/v1/messages` → Anthropic, `/responses` → OpenAI Responses). Google/Gemini and `/chat/completions`-only models are excluded.
- UI: grouped dropdown with non-selectable section headers (`Copilot · Anthropic`, `Copilot · OpenAI`).
- The four curated Copilot entries are replaced by the discovered list (single source of truth).

## Acceptance

- Running the sample with a valid Copilot token lists every routable Anthropic/OpenAI Copilot model under the two group headers.
- With no Copilot token (or discovery failure/timeout), the Copilot sections are empty and the app still boots with the direct/CLI/test providers.
- Selecting a discovered model routes to the correct transport and sends the raw model id.

gautam-achieveai:
## Summary

Expose all GitHub Copilot provider models in the LmStreaming.Sample provider dropdown, discovered dynamically from the Copilot `/models` API at startup and clearly partitioned into **Copilot · Anthropic** and **Copilot · OpenAI** sections. This replaces the four hardcoded Copilot entries (`sonnet`/`haiku`/`gpt-5.5`/`gpt-5.5-mini`) with the live, routable catalog.

## Changes

- **Discovery (`src/GithubCopilotProvider/Models/`)** — new `CopilotModelInfo`, a pure `CopilotModelCatalogParser`, and `CopilotModelsClient`. The parser keeps only models whose vendor is Anthropic or OpenAI (normalizing `Azure OpenAI` → OpenAI) and that are routable via `/v1/messages` (Anthropic) or `/responses` (OpenAI Responses); Google/Gemini and `/chat/completions`-only models are dropped.
- **`ProviderRegistry`** — injects the discovered models, keys each by its raw model id, tags it with a partition group, and exposes `TryGetCopilotModel`. The four curated Copilot entries are removed; the discovered list is the single source.
- **`Program.cs`** — dispatch (agent transport), model-id resolution, reasoning/thinking options, and web-tool fallback are now driven by the discovered model's transport/vendor instead of literal ids. Discovery runs sync-over-async in the registry DI factory with a 10s timeout and degrades to an empty list (no token / failure / timeout) so the app still boots.
- **UI** — `ProviderDescriptor` / `providers.ts` gain a `Group` field; `ProviderSelector.vue` renders non-selectable section headers with ungrouped providers listed first.
- **Docs** — `CLAUDE.md` / `ManualTesting.md` updated to describe the dynamic, partitioned list.

## Testing

- New unit tests for the parser and client against the real captured `/models` fixture (34 models → 13 routable: 7 Claude + 6 GPT incl. one `Azure OpenAI`).
- Registry grouping/availability, reasoning-props and web-tool policy by transport, and Vue grouping/selection tests.
- Verified green: full solution build (0 errors), format gate (`dotnet format whitespace --verify-no-changes`), GithubCopilotProvider.Tests (20 new + 52 total), LmStreaming.Sample.Tests (361), ClientApp vitest (268).
- Not yet live-verified against a real Copilot token — tests are fixture-driven.

## Risks

- Behavior change: conversations previously locked to `sonnet`/`haiku`/`gpt-5.5`/`gpt-5.5-mini` will show those ids as unavailable (intended replacement).
- First `GET /api/providers` request triggers the one-time discovery call (bounded by a 10s timeout).

## Related Issues

Fixes #128

---

## Additional infrastructure

- **Configurable Vite dev-server port** (`VITE_DEV_PORT` / `VITE_BACKEND_ORIGIN`) — manual-validation infra, not part of #128 proper. It removes the hardcoded 5173 collision so a second sample instance can boot alongside a running one for live UI verification. Both env vars default to the standard single-instance values, so existing runs are unchanged.

## Follow-ups (deferred from review)

- #132 — extract a shared Copilot `/models` parsing helper between the provider and the proxy sample.
- #133 — add a browser E2E regression for provider-dropdown scroll/overflow (needs a DI seam to inject stub Copilot models into the token-gated registry).

### DIFF
--- a/docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
+++ b/docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
@@ -1,0 +1,58 @@
+# Expose GitHub Copilot models in the sample model list, partitioned by vendor
+
+**Date:** 2026-07-01
+**Status:** Approved via goal directive + HITL scope answers
+**Target repo:** `B:/sources/LmDotnetTools` (main checkout; feature branch off `main`)
+
+## Goal
+Expose all models available through the GitHub Copilot-based provider in the
+LmStreaming.Sample model list, clearly partitioned into **Anthropic** and **OpenAI** groups.
+
+## Locked decisions (from HITL)
+1. **Which list:** LmStreaming.Sample provider dropdown (`ProviderRegistry` → `/api/providers` → `ProviderSelector.vue`). NOT LmConfig `models.json`.
+2. **Source:** Dynamic discovery — call Copilot `GET /models` at startup and build the list from the live response.
+3. **Scope:** Only models whose `vendor ∈ {Anthropic, OpenAI}` (normalizing `Azure OpenAI` → `OpenAI`) **and** that are routable via an existing factory (`/v1/messages` → Anthropic, `/responses` → Responses). Google/Gemini and `/chat/completions`-only models excluded.
+4. **UI:** Grouped dropdown with non-selectable headers (`Copilot · Anthropic`, `Copilot · OpenAI`).
+5. **Curated entries:** Remove the 4 hardcoded Copilot entries (`sonnet`, `haiku`, `gpt-5.5`, `gpt-5.5-mini`); the discovered list is the single source. Generalize the dispatch to route by discovered transport instead of those literal ids.
+
+## Current state (grounding)
+- `samples/LmStreaming.Sample/Services/ProviderRegistry.cs:20-41` — hardcoded catalog incl. the 4 Copilot entries; availability gated on `hasCopilotToken` (:73, :86).
+- `ProviderDescriptor(Id, DisplayName, Available, KnownLimitation?)` — flat, no group field.
+- Dispatch: `Func<string,IStreamingAgent>` switch `Program.cs:314-328` (`sonnet`/`haiku` → `CopilotAnthropicAgentFactory`; `gpt-5.5*` → `CopilotResponsesAgentFactory`).
+- Model id: `GetModelIdForProvider` switch `Program.cs:1496-1512` → `GenerateReplyOptions.ModelId` (`:757`, `:839`).
+- Copilot `/models` real response: `tests/CopilotAnthropicProxy.Tests/Fixtures/copilot-models-real-response.json` — each entry has `id`, `name`, `vendor`, `supported_endpoints`, `capabilities.family`.
+- Auth/HTTP to reuse: `CopilotHttpClientFactory.Create(host, tokenProvider, session, options)`; token via `CliCredentialCopilotTokenProvider`; host `CopilotOptions.DefaultBaseUrl`.
+- Existing endpoint-filter/parse precedent (not reused directly): `ProxyModelResolver.ParseMessagesCapableModelIds` in `samples/CopilotAnthropicProxy.Sample/Program.cs`.
+
+## Design
+
+### Component 1 — Discovery (`src/GithubCopilotProvider/Models/`)
+- `CopilotModelInfo` record: `Id`, `DisplayName`, `Vendor` (normalized), `SupportedEndpoints`, derived `Transport` enum { `Anthropic`, `Responses`, `Unsupported` }.
+- `CopilotVendor`/`CopilotModelTransport` — small enums or normalized strings.
+- `CopilotModelCatalogParser.Parse(json)` — **pure**. Parses `{"data":[...]}`, normalizes vendor, derives transport from `supported_endpoints`, keeps only `Vendor ∈ {Anthropic, OpenAI}` and `Transport ∈ {Anthropic, Responses}`. Unit-tested against the real fixture.
+- `CopilotModelsClient.GetModelsAsync(ct)` — `CopilotHttpClientFactory.Create(...)` → `GET /models` → `Parse`. Returns empty list on failure.
+
+### Component 2 — Sample wiring
+- `ProviderDescriptor` gains `string? Group` (null → flat render). Add `Vendor`/`RawModelId` only if needed for dispatch; prefer a separate lookup map `id → CopilotModelInfo`.
+- `ProviderRegistry` accepts injected `IReadOnlyList<CopilotModelInfo>`; builds one descriptor per model, **id = raw model id**, `Group = "Copilot · Anthropic" | "Copilot · OpenAI"`, `Available = hasCopilotToken`. Removes the 4 curated entries. Non-Copilot entries unchanged (`Group = null`).
+- Startup resolve in `Program.cs` (sync-over-async at registration, same pattern as `:796`). Token absent → empty list → no Copilot models. Token present + discovery fails → small built-in Anthropic+OpenAI fallback set so the app still runs.
+- Dispatch generalization: factory switch + model-id resolution look up the discovered `CopilotModelInfo` by id → `Transport` picks the factory; raw id → `GenerateReplyOptions.ModelId`. Direct (`openai`/`anthropic`) and CLI (`claude`/`codex`/`copilot`) branches unchanged.
+
+### Component 3 — UI
+- `types/providers.ts` `ProviderDescriptor` gains `group?: string | null`.
+- `ProviderSelector.vue`: group entries by `group`; `group == null` render flat first (unchanged look), grouped entries under non-selectable headers. Preserve availability/limitation/active rendering.
+
+### Testing (TDD, RED→GREEN)
+- Parser unit tests vs real fixture: Anthropic→Anthropic partition/Transport; GPT→OpenAI/Responses; Gemini/Google + `/chat/completions`-only excluded; `Azure OpenAI`→`OpenAI`.
+- `ProviderRegistry` tests: stub list → descriptors+groups; empty/no-token → no Copilot entries; non-Copilot entries unchanged.
+- Dispatch tests: `claude-*` id → Anthropic factory; `gpt-*` id → Responses factory; raw id flows to `ModelId`.
+
+## Out of scope (YAGNI)
+- `CopilotAnthropicProxy.Sample`, `src/LmConfig/models.json`, real Anthropic/OpenAI providers.
+- Caching/refresh beyond the single startup resolve.
+- Live network in tests (fixture-driven).
+
+## Risks
+- Overlap of discovered `gpt-5.5` with a former curated id — resolved by "replace" decision (single source).
+- `Azure OpenAI` vendor drift — handled by normalization.
+- Startup network dependency — handled by graceful fallback.
--- a/docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
+++ b/docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
@@ -35,7 +35,7 @@ LmStreaming.Sample model list, clearly partitioned into **Anthropic** and **Open
 ### Component 2 — Sample wiring
 - `ProviderDescriptor` gains `string? Group` (null → flat render). Add `Vendor`/`RawModelId` only if needed for dispatch; prefer a separate lookup map `id → CopilotModelInfo`.
 - `ProviderRegistry` accepts injected `IReadOnlyList<CopilotModelInfo>`; builds one descriptor per model, **id = raw model id**, `Group = "Copilot · Anthropic" | "Copilot · OpenAI"`, `Available = hasCopilotToken`. Removes the 4 curated entries. Non-Copilot entries unchanged (`Group = null`).
-- Startup resolve in `Program.cs` (sync-over-async at registration, same pattern as `:796`). Token absent → empty list → no Copilot models. Token present + discovery fails → small built-in Anthropic+OpenAI fallback set so the app still runs.
+- Startup resolve in `Program.cs` (sync-over-async at registration, same pattern as `:796`). Token absent → empty list → no Copilot models. Token present + discovery fails (or the bounded startup timeout elapses) → empty list → no Copilot models; the app still boots with the direct/CLI/test providers. (No built-in fallback set is used — the accepted behavior is to leave the Copilot sections empty.)
 - Dispatch generalization: factory switch + model-id resolution look up the discovered `CopilotModelInfo` by id → `Transport` picks the factory; raw id → `GenerateReplyOptions.ModelId`. Direct (`openai`/`anthropic`) and CLI (`claude`/`codex`/`copilot`) branches unchanged.
 
 ### Component 3 — UI
@@ -55,4 +55,4 @@ LmStreaming.Sample model list, clearly partitioned into **Anthropic** and **Open
 ## Risks
 - Overlap of discovered `gpt-5.5` with a former curated id — resolved by "replace" decision (single source).
 - `Azure OpenAI` vendor drift — handled by normalization.
-- Startup network dependency — handled by graceful fallback.
+- Startup network dependency — handled by degrading to an empty Copilot catalog (bounded timeout).
--- a/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/ProviderSelector.test.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/ProviderSelector.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, it } from 'vitest';
 import { mount } from '@vue/test-utils';
 import ProviderSelector from '@/components/ProviderSelector.vue';
+// Raw component source, used to assert the scoped stylesheet rules (scoped <style> is not
+// injected into the jsdom DOM by @vue/test-utils, so we verify the CSS at the source level).
+import providerSelectorSource from '@/components/ProviderSelector.vue?raw';
 import type { ProviderDescriptor } from '@/types/providers';
 
 const providers: ProviderDescriptor[] = [
@@ -67,5 +70,24 @@ describe('ProviderSelector grouping', () => {
 
     expect(wrapper.findAll('.menu-group-header')).toHaveLength(0);
     expect(wrapper.findAll('.menu-item')).toHaveLength(2);
+  });
+});
+
+describe('ProviderSelector dropdown scroll', () => {
+  // Extract the `.dropdown-menu { ... }` rule body from the scoped stylesheet source.
+  const menuRule = (() => {
+    const start = providerSelectorSource.indexOf('.dropdown-menu {');
+    expect(start).toBeGreaterThanOrEqual(0);
+    const end = providerSelectorSource.indexOf('}', start);
+    return providerSelectorSource.slice(start, end);
+  })();
+
+  it('caps the dropdown height so a long list does not run off-screen', () => {
+    // max-height keeps the menu bounded; a bare `max-height` without a value would not.
+    expect(menuRule).toMatch(/max-height:\s*[^;]+;/);
+  });
+
+  it('enables vertical scrolling for overflowing content', () => {
+    expect(menuRule).toMatch(/overflow-y:\s*auto\s*;/);
   });
 });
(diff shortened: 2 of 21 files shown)


gautam-achieveai:
Pushed `18b970f` fixing two issues found in live testing:

1. **Adaptive-thinking 400** — selecting `claude-sonnet-5` (and other adaptive-thinking Copilot Claude models: opus 4.x, sonnet 4.6+) returned `HTTP 400: "thinking.type.enabled" is not supported for this model`. The generalized reasoning wiring was sending the classic thinking budget to every Anthropic-transport model, but the newer ones require `thinking.type.adaptive` + `output_config.effort`. Fix: parse `capabilities.supports.adaptive_thinking` and omit the classic thinking parameter for models that advertise it (the older sonnet-4.5/haiku-4.5 keep it). The adaptive API isn't modelled by the Anthropic provider yet, so those models reason without an explicit budget request.
2. **Dropdown scroll** — the longer discovered list overflowed with no scroll. Capped `.dropdown-menu` height with `overflow-y: auto` and made the partition headers sticky.

**Live-verified** against the real Copilot backend (Playwright): `claude-sonnet-5` now replies normally with no 400, and the dropdown scrolls (scrollHeight 898 > clientHeight 319, `overflow-y: auto`) with the last option reachable. The real `/models` call returns 13 routable models (7 Claude + 6 GPT) across the two partitions.

src/GithubCopilotProvider/Models/CopilotModelsClient.cs
```diff
@@ -0,0 +1,86 @@
...
+                    (int)response.StatusCode);
+                return [];
+            }
+
+            var json = await response.Content.ReadAsStringAsync(cancellationToken);
+            var models = CopilotModelCatalogParser.Parse(json);
+            _logger?.LogInformation("Discovered {Count} routable Copilot models.", models.Count);
+            return models;
+        }
+        catch (Exception ex) when (ex is HttpRequestException or OperationCanceledException or System.Text.Json.JsonException)
```
gautam-achieveai: [Gautam's bot] **MEDIUM** (Cancellation / API semantics)

`GetModelsAsync` currently catches `OperationCanceledException` with transport/JSON failures and converts caller-requested cancellation into an empty catalog. Because this class lives in the provider library (not just the sample startup path), that changes normal cooperative-cancellation semantics and makes a user/shutdown cancellation indistinguishable from “discovery failed; hide Copilot models.” The test in `CopilotModelsClientTests.GetModelsAsync_returns_empty_when_cancelled` also locks in that behavior.

**Suggestion:** Let caller-requested cancellation propagate from the client, and handle the sample’s bounded startup timeout at the sample boundary. For example:

```csharp
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
{
    throw;
}
catch (Exception ex) when (ex is HttpRequestException or JsonException)
{
    _logger?.LogWarning(ex, "Copilot model discovery failed; treating catalog as empty.");
    return [];
}
```

Then have `DiscoverCopilotModels` catch its own timeout/degraded-mode case and return `[]` there. Update the cancellation test to assert propagation, with a separate sample-level test if you want to lock the startup fallback behavior.

src/GithubCopilotProvider/Models/CopilotModelCatalogParser.cs
```diff
@@ -0,0 +1,171 @@
...
+{
+    /// <summary>
+    ///     Projects the Copilot <c>/models</c> JSON to the routable Anthropic/OpenAI models, preserving
+    ///     upstream order. Malformed or partial entries are skipped rather than throwing.
+    /// </summary>
+    public static IReadOnlyList<CopilotModelInfo> Parse(string json)
+    {
+        ArgumentNullException.ThrowIfNull(json);
+
+        using var doc = JsonDocument.Parse(json);
```
gautam-achieveai, line 26: [Gautam's bot] **MEDIUM** (Duplication / maintainability)

This PR adds a second Copilot `/models` parser that duplicates response unwrapping, id extraction, and `supported_endpoints` scanning that already exists in `ProxyModelResolver.ParseMessagesCapableModelIds` (`samples/CopilotAnthropicProxy.Sample/Program.cs:513-567`). Since the proxy sample already references `GithubCopilotProvider`, the two Copilot schema/endpoint parsers can drift independently. For example, the existing proxy parser has explicit behavior for responses where no entry carries endpoint metadata, while the new parser has separate filtering behavior.

**Suggestion:** Extract the shared response traversal / endpoint classification into provider-level parsing helpers and have both `CopilotModelCatalogParser` and `ProxyModelResolver` use them, keeping only the sample-specific projection/filtering decisions in each caller. That keeps Copilot endpoint spelling and malformed-entry behavior in one place.

docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
```diff
@@ -0,0 +1,58 @@
...
+- `CopilotModelCatalogParser.Parse(json)` — **pure**. Parses `{"data":[...]}`, normalizes vendor, derives transport from `supported_endpoints`, keeps only `Vendor ∈ {Anthropic, OpenAI}` and `Transport ∈ {Anthropic, Responses}`. Unit-tested against the real fixture.
+- `CopilotModelsClient.GetModelsAsync(ct)` — `CopilotHttpClientFactory.Create(...)` → `GET /models` → `Parse`. Returns empty list on failure.
+
+### Component 2 — Sample wiring
+- `ProviderDescriptor` gains `string? Group` (null → flat render). Add `Vendor`/`RawModelId` only if needed for dispatch; prefer a separate lookup map `id → CopilotModelInfo`.
+- `ProviderRegistry` accepts injected `IReadOnlyList<CopilotModelInfo>`; builds one descriptor per model, **id = raw model id**, `Group = "Copilot · Anthropic" | "Copilot · OpenAI"`, `Available = hasCopilotToken`. Removes the 4 curated entries. Non-Copilot entries unchanged (`Group = null`).
+- Startup resolve in `Program.cs` (sync-over-async at registration, same pattern as `:796`). Token absent → empty list → no Copilot models. Token present + discovery fails → small built-in Anthropic+OpenAI fallback set so the app still runs.
```
gautam-achieveai: [Gautam's bot] **MEDIUM** (Documentation / scope hygiene)

This checked-in design artifact says “Token present + discovery fails → small built-in Anthropic+OpenAI fallback set,” but the linked issue acceptance/current code/sample context say failure or no token should leave Copilot sections empty while the app still boots. That mismatch is durable repo guidance and can confuse future maintainers into reintroducing fallback behavior that is not part of the accepted behavior.

**Suggestion:** Remove this internal planning artifact from the PR, or convert only the durable/accurate decisions into maintainer-facing docs. If it stays, update the fallback text so it matches the implemented and accepted behavior (`[]` Copilot models on no token/discovery failure).

samples/LmStreaming.Sample/ClientApp/src/__tests__/components/ProviderSelector.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      },
+    });
+    await wrapper.get('.selector-btn').trigger('click');
+
+    expect(wrapper.findAll('.menu-group-header')).toHaveLength(0);
+    expect(wrapper.findAll('.menu-item')).toHaveLength(2);
+  });
+});
+
+describe('ProviderSelector dropdown scroll', () => {
```
gautam-achieveai, line 76: [Gautam's bot] **MEDIUM** (Test quality)

The dropdown-scroll regression test regex-parses the raw Vue SFC source and checks that CSS declarations exist. That is a useful smoke check, but it does not prove the long dropdown is actually scrollable/reachable in a browser — the exact issue reported during live testing. The sample context points browser/UI regressions at `tests/LmStreaming.Sample.Browser.E2E.Tests` using the stable `provider-selector-button` / `provider-option-*` test ids.

**Suggestion:** Add a browser/e2e regression that opens the provider selector with enough providers to overflow, asserts the menu is scrollable (or scrolls it), and verifies the last option can be reached/selected. The source-level CSS assertion can remain as a cheap unit smoke test, but the acceptance regression should exercise browser behavior.

samples/LmStreaming.Sample/ClientApp/vite.config.ts
```diff
@@ -2,6 +2,14 @@ import { defineConfig } from 'vite';
 import vue from '@vitejs/plugin-vue';
 import { resolve } from 'path';
 
+// Dev-server port and backend proxy target are configurable so an isolated instance can run
```
gautam-achieveai, line 5: [Gautam's bot] **LOW** (Scope control)

The `VITE_DEV_PORT` / `VITE_BACKEND_ORIGIN` isolated-instance support is useful for manual validation, but it is adjacent to #128 rather than part of dynamic Copilot model discovery/grouping. It expands the sample's documented configuration and Vite proxy behavior in the same PR.

**Suggestion:** Either split this into a small chore PR, or call it out explicitly in this PR's description as test/manual-validation infrastructure so reviewers know why it is bundled with the Copilot dropdown work.

src/GithubCopilotProvider/Models/CopilotModelInfo.cs
```diff
@@ -0,0 +1,53 @@
+namespace AchieveAi.LmDotnetTools.GithubCopilotProvider.Models;
+
+/// <summary>
+///     The request transport a Copilot model is reachable through, derived from the model's
+///     <c>supported_endpoints</c>. Only <see cref="Anthropic"/> and <see cref="Responses"/> are
+///     routable by the sample today; everything else (e.g. <c>/chat/completions</c>-only models
+///     such as Gemini) is <see cref="Unsupported"/> and filtered out.
+/// </summary>
+public enum CopilotModelTransport
```
gautam-achieveai, line 9: [Gautam's bot] **LOW** (Type design)

`CopilotModelTransport.Unsupported` is exposed on the public `CopilotModelInfo` transport enum, but parser-produced `CopilotModelInfo` values are filtered before construction. `CreateCopilotModelAgent` then treats every non-`Anthropic` value as Responses, so a manually constructed `Unsupported` model would be misrouted instead of rejected.

**Suggestion:** Keep `Unsupported` as an internal/nullable parser derivation detail, or make this dispatch exhaustive:

```csharp
return model.Transport switch
{
    CopilotModelTransport.Anthropic => CreateCopilotAnthropicAgent(model.DisplayName, loggerFactory),
    CopilotModelTransport.Responses => CreateCopilotResponsesAgent(model.DisplayName, loggerFactory),
    _ => throw new ProviderUnavailableException(model.Id, $"unsupported Copilot transport {model.Transport}"),
};
```

gautam-achieveai:
[Gautam's bot] # PR Review: feat(samples): expose Copilot models in dropdown, partitioned by vendor (#128)

## Verdict: REQUEST CHANGES

I reviewed PR #129 against linked issue #128 and the latest head `0d7bc6ae`. The main implementation is directionally aligned with the issue: it dynamically discovers Copilot models, groups them by vendor, routes by raw model id/transport, and includes a fix for the reported dropdown overflow. CI currently shows build/test and browser-e2e passing.

I found several issues that should be addressed before merge, primarily around API cancellation semantics, duplicated Copilot `/models` parsing logic, a conflicting checked-in design artifact, and insufficient browser-level regression coverage for the dropdown scroll bug.

## Findings

| Severity | Category | Location | Summary |
|---|---|---|---|
| MEDIUM | Cancellation / API semantics | `src/GithubCopilotProvider/Models/CopilotModelsClient.cs:66` | `GetModelsAsync` swallows caller-requested cancellation and returns an empty catalog. |
| MEDIUM | Duplication / maintainability | `src/GithubCopilotProvider/Models/CopilotModelCatalogParser.cs:26` | New `/models` parser duplicates existing Copilot endpoint/schema parsing in the proxy sample. |
| MEDIUM | Documentation / scope hygiene | `docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md:35` | Committed design artifact conflicts with the accepted fallback behavior. |
| MEDIUM | Test quality | `samples/LmStreaming.Sample/ClientApp/src/__tests__/components/ProviderSelector.test.ts:76` | Scroll regression test checks CSS source text, not browser scrollability/reachability. |
| LOW | Scope control | `samples/LmStreaming.Sample/ClientApp/vite.config.ts:5` | Isolated-instance Vite/backend config is adjacent to #128 and should be split or documented as validation infra. |
| LOW | Type design | `src/GithubCopilotProvider/Models/CopilotModelInfo.cs:9` | Public `Unsupported` transport can be misrouted by non-exhaustive dispatch if manually constructed. |

## Notes

- The user-reported dropdown-scroll issue appears to be fixed in CSS (`max-height`, `overflow-y: auto`, sticky headers), but I’m requesting stronger regression coverage because this was found by live UI testing and the current unit test does not exercise browser layout.
- I attempted local verification: `.NET` is unavailable in this sandbox (`dotnet: not found`), and frontend dependency install was blocked by registry egress policy (`npm` 403 on `ws`). I therefore relied on GitHub checks, source inspection, and specialist review agents.
- Specialist agents used: PR context, test coverage, duplicate code, architecture, schema compatibility, performance, exception handling, feature-flag, class design, code simplification, over-engineering, and temp/security scan; final severities were checked by the review-grader.

src/GithubCopilotProvider/Models/CopilotModelsClient.cs
gautam-achieveai: Fixed in 9dd401a. `GetModelsAsync` now rethrows `OperationCanceledException` when the caller's token is cancelled and only maps `HttpRequestException`/`JsonException` to an empty catalog. The bounded 10s startup timeout is handled at the sample boundary — `DiscoverCopilotModels` catches its own timeout and returns `[]`. The test now asserts propagation (`GetModelsAsync_propagates_caller_requested_cancellation`).

src/GithubCopilotProvider/Models/CopilotModelCatalogParser.cs
gautam-achieveai: Deferred to #132. The two parsers share only the low-level `data[]` unwrap + `supported_endpoints` scan; their projections differ intentionally (the proxy filters to `/v1/messages` ids only with an id-only fallback, this one returns rich CopilotModelInfo across both transports). Extracting a shared provider-level helper is worthwhile but touches the working proxy sample, so I kept it out of this PR and tracked it separately.

docs/superpowers/specs/2026-07-01-copilot-models-partition-design.md
gautam-achieveai: Fixed in 9dd401a. Corrected the doc: no-token / discovery-failure / timeout leaves the Copilot sections empty (the accepted behavior), and removed the never-implemented 'built-in fallback set' wording (also on the Risks line). Kept the artifact since it records the design decisions accurately now.

samples/LmStreaming.Sample/ClientApp/src/__tests__/components/ProviderSelector.test.ts
gautam-achieveai: Deferred to #133. The source-level CSS assertion stays as a cheap smoke test. A real browser regression needs a DI seam to inject a stub Copilot model list into the token-gated ProviderRegistry so the E2E harness (scripted test/test-anthropic modes, no Copilot token in CI) can overflow the dropdown deterministically — tracked in that issue.

samples/LmStreaming.Sample/ClientApp/vite.config.ts
gautam-achieveai: Kept in this PR by design — it's the manual-validation infrastructure that made the live end-to-end verification of this dropdown possible (booting a second isolated instance alongside a running one without the fixed 5173 collision). I'll note that in the PR description. Both env vars default to the standard single-instance values, so existing runs are unchanged.

src/GithubCopilotProvider/Models/CopilotModelInfo.cs
gautam-achieveai: Fixed in 9dd401a. CreateCopilotModelAgent now dispatches with an exhaustive switch: Anthropic → Anthropic factory, Responses → Responses factory, and any other value (incl. Unsupported) throws ProviderUnavailableException instead of defaulting to Responses.

gautam-achieveai:
[Gautam's review bot] # Re-review: feat(samples): expose Copilot models in dropdown, partitioned by vendor (#128)

## Verdict: APPROVE

All blocking issues from Review 1 have been addressed in iteration 5 (`9dd401a`):

| Previous Finding | Status |
|---|---|
| Cancellation / API semantics (MEDIUM) | Fixed - GetModelsAsync now propagates caller-requested cancellation; timeout handled at sample boundary |
| Documentation / scope hygiene (MEDIUM) | Fixed - design doc corrected to reflect empty-list behavior |
| Type design / exhaustive dispatch (LOW) | Fixed - CreateCopilotModelAgent now uses exhaustive switch |
| Duplication / maintainability (MEDIUM) | Deferred to #132 - justified, projections differ intentionally |
| Test quality / browser regression (MEDIUM) | Deferred to #133 - requires DI seam for stub injection in CI |
| Scope control (LOW) | Kept - serves as manual-validation infra for this feature |

## Minor Observations (non-blocking)

1. **Stale remarks on CopilotModelsClient** - The XML doc still says Failures are swallowed but caller-requested cancellation now propagates (by design). Consider updating the remarks.

2. **Redundant token-provider instantiation** (Program.cs:1266) - new CliCredentialCopilotTokenProvider().ResolveToken() creates a throwaway instance when s_copilotTokenProvider.Value (same type, lazy singleton) is used 6 lines later.

3. **CopilotModelsClient resource lifecycle** - The class holds an HttpClient but does not implement IDisposable. In the startup-once sample path this is benign, but as a provider-library class, consumers cannot cleanly dispose the underlying handler.

Overall the implementation is clean, well-tested, and the deferred items are properly tracked. Ship it.

src/GithubCopilotProvider/Models/CopilotModelsClient.cs
```diff
@@ -0,0 +1,93 @@
...
+
+/// <summary>
+///     Discovers the GitHub Copilot model catalog by calling <c>GET /models</c> through the Copilot
+///     backend and projecting the response to the routable Anthropic/OpenAI models via
+///     <see cref="CopilotModelCatalogParser"/>.
+/// </summary>
+/// <remarks>
+///     Failures (no token, network error, non-success status, malformed body) are swallowed and
+///     surfaced as an empty list so a discovery outage never takes down the caller — the caller
+///     decides how to degrade (e.g. fall back to a built-in set).
```
gautam-achieveai, line 14: [Gautam's review bot] **LOW** (Doc accuracy)

Nit: The remarks says 'Failures...are swallowed and surfaced as an empty list' but after the cancellation fix in this iteration, OperationCanceledException now correctly propagates when the caller's token is cancelled. Consider updating to: 'Transport/deserialization failures are surfaced as an empty list; caller-requested cancellation propagates normally.'

PR #131, opened 2026-07-02
feat(workflow): author & run LmWorkflow in Workspace Agent mode

### LINKED ISSUE #130: Enable LLM-authored, runnable LmWorkflow in Workspace Agent mode
## Summary
`LmWorkflow`'s only entry point was `WorkflowSession` (its own dedicated, workflow-only `MultiTurnAgentLoop`). The sample's **Workspace Agent** chat mode — which already carries file/shell/sub-agent tools on a long-lived loop — had no way to author or drive a workflow. This wires `WorkflowToolProvider` directly into that existing loop and fixes the defects that stopped a real LLM from authoring and running a workflow end-to-end.

## Bugs a live Copilot-backed run exposed
1. **Un-authorable.** `SetWorkflow`'s tool schema was a bare `{"type":"object"}`, the controller prompt showed no example, and unknown field names were silently dropped by the deserializer — so a misspelled task field "succeeded" with **zero tasks** and no feedback (the model guessed field names 20+ times and never registered a task).
2. **Silent stall.** A blocking sub-agent whose LLM call stalled could freeze the whole workflow for the full **5-minute** HTTP timeout with no feedback.
3. **Never ran to completion.** A successful sub-agent result was never correlated to the procedural node's join (task stuck `pending`), because the out-of-band observer only receives streaming `ToolCallUpdateMessage`s with **fragmented** args — never a finalized `ToolCallMessage`.

## Scope
- Wire `WorkflowRuntime` + `WorkflowToolProvider` into Workspace Agent mode (extract public `WorkflowRuntime.ObserveMessage`; race-safe correlation subscriber).
- Author ergonomics: nested `SetWorkflow` schema, worked prompt example, strict authoring deserialize.
- Robustness: bounded sub-agent HTTP time-to-first-response; on-demand unit composition; streamed-args reassembly for join correlation.

## Acceptance
A real LLM, in Workspace Agent mode, authors a workflow first-try and drives it to a terminal with the task **validated** and the join **satisfied** (`isComplete:true`). Verified live and covered by tests.

gautam-achieveai:
## Summary
Wires `LmWorkflow` into the sample's existing **Workspace Agent** `MultiTurnAgentLoop` so a real Copilot-backed LLM can author and drive a workflow end-to-end through the chat UI, and fixes the three defects that prevented that from working. Verified live (authored first-try, task validated, join satisfied, `isComplete`) and covered by new tests.

## Changes
**Wire-in**
- Register `WorkflowRuntime` + `WorkflowToolProvider` in Workspace Agent mode; add the `LmWorkflow` project reference.
- Extract a public `WorkflowRuntime.ObserveMessage` from `WorkflowSession` (pure code move); add a race-safe second stream subscriber for sub-agent-spawn correlation.

**Authoring ergonomics** — the LLM could not author a valid workflow
- Real nested JSON schema for `SetWorkflow.definition` (was a bare `{"type":"object"}`).
- Worked example + exact field names in `ControllerSystemPrompt`, sourced from a single **test-enforced** `WorkflowExamples.MinimalProcedural` constant (no prompt drift).
- Strict authoring deserialize (`UnmappedMemberHandling.Disallow`) so a misspelled field is rejected **by name** instead of silently dropped (which used to yield a valid-but-empty procedural node with no feedback). Persistence stays lenient.

**Robustness** — the workflow could not run to completion
- Bound Copilot agents' time-to-first-response (`COPILOT_RESPONSE_TIMEOUT_SECONDS`, default 120s) so a dead upstream connection surfaces fast instead of freezing the run for the full 5-minute HTTP default. Safe because `ResponseHeadersRead` means this bounds only connect/headers, not stream length.
- Compose the active node's units **on demand** in `TaskCoordinator.RegisterSpawn` so correlation no longer depends on the controller polling `GetWorkflow` before spawning.
- Reassemble streamed `ToolCallUpdateMessage` args fragments in `ObserveMessage`: a real provider publishes the `Agent` call to subscribers as fragmented update deltas (never a finalized `ToolCallMessage`), so the spawn name must be reassembled before the sub-agent result can correlate to the join.

## Testing
- New: `WorkflowAuthoringErgonomicsTests`, `ObserveMessageCorrelationTests` (incl. fragment-reassembly), `CopilotAnthropicAgentFactoryTimeoutTests`, and a sub-agent-failure E2E in `WorkflowEndToEndTests` (workflow still reaches a terminal via `onFailure`).
- Suites green: **LmWorkflow 194/194, GithubCopilotProvider 35/35, LmStreaming.Sample 359/359, LmMultiTurn 285/285.** Solution builds clean.
- **Live-verified** with real Copilot Sonnet through the UI: authored first-try, task `validated`, join `{validated:1, satisfied:true}`, `isComplete:true`.

## Related Issues
Fixes #130

### DIFF
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -694,8 +694,23 @@ try
                     // same conversation loop (no separate WorkflowSession-owned loop) — the runtime
                     // observes the loop's own message stream (see ownedResources wiring below) to
                     // correlate Agent tool-call spawns back into workflow state.
-                    workflowRuntime = new WorkflowRuntime(logger: loggerFactory.CreateLogger<WorkflowRuntime>());
-                    _ = filteredRegistry.AddProvider(new WorkflowToolProvider(workflowRuntime));
+                    //
+                    // On by default, but disable-able per deployment WITHOUT a redeploy via
+                    // WORKSPACE_AGENT_LMWORKFLOW_ENABLED=false. This one flag gates all three behaviors as a
+                    // unit — the workflow tool surface (here), the appended controller prompt (below), and the
+                    // correlation observer (further below, which is already gated on `workflowRuntime`).
+                    var workflowEnabled = !string.Equals(
+                        Environment.GetEnvironmentVariable("WORKSPACE_AGENT_LMWORKFLOW_ENABLED"),
+                        "false",
+                        StringComparison.OrdinalIgnoreCase
+                    );
+                    if (workflowEnabled)
+                    {
+                        workflowRuntime = new WorkflowRuntime(
+                            logger: loggerFactory.CreateLogger<WorkflowRuntime>()
+                        );
+                        _ = filteredRegistry.AddProvider(new WorkflowToolProvider(workflowRuntime));
+                    }
 
                     // Workspace Agent mode: expose the sandbox file/shell tools via the gateway's
                     // MCP endpoint, bound to this agent's sandbox session by the X-Session-ID header.
@@ -741,11 +756,17 @@ try
                     }
 
                     // Append last, off whatever effectiveMode currently is (default or the
-                    // degraded-sandbox notice above) so the controller prompt survives either path.
-                    effectiveMode = effectiveMode with
+                    // degraded-sandbox notice above) so the controller prompt survives either path. Gated on
+                    // the workflow runtime (i.e. WORKSPACE_AGENT_LMWORKFLOW_ENABLED) so disabling the feature
+                    // also drops the extra controller instructions from the prompt.
+                    if (workflowRuntime is not null)
                     {
-                        SystemPrompt = effectiveMode.SystemPrompt + "\n\n" + ControllerSystemPrompt.Default,
-                    };
+                        effectiveMode = effectiveMode with
+                        {
+                            SystemPrompt =
+                                effectiveMode.SystemPrompt + "\n\n" + ControllerSystemPrompt.Default,
+                        };
+                    }
                 }
 
                 // WebFetch/WebSearch fallback tools for providers without a native web capability.
@@ -903,8 +924,23 @@ try
                                     }
                                 }
                             }
-                            catch (OperationCanceledException)
+                            catch (OperationCanceledException) when (observerCts.IsCancellationRequested)
                             { /* expected on shutdown */
+                            }
+                            catch (Exception ex)
+                            {
+                                // This background task is the critical path that correlates Agent
+                                // tool-call spawns/results back into the workflow runtime. If it faults, the
+                                // agent keeps serving workflow tools but joins never settle — so surface the
+                                // first failure loudly (disposal below would otherwise swallow it).
+                                loggerFactory
+                                    .CreateLogger<WorkflowRuntime>()
+                                    .LogError(
+                                        ex,
+                                        "Workflow observer for thread {ThreadId} failed; Agent-result "
+                                            + "correlation is now disabled for this conversation",
+                                        threadId
+                                    );
                             }
                             finally
                             {
@@ -1288,9 +1324,9 @@ public partial class Program
         return CopilotAnthropicAgentFactory.Create(
             name,
             s_copilotTokenProvider.Value,
-            s_copilotSession.Value,
-            logger: loggerFactory.CreateLogger<AnthropicAgent>(),
-            timeout: CopilotResponseTimeout
+            timeout: CopilotResponseTimeout,
+            session: s_copilotSession.Value,
+            logger: loggerFactory.CreateLogger<AnthropicAgent>()
         );
     }
 
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -21,6 +21,9 @@ using AchieveAi.LmDotnetTools.LmMultiTurn;
 using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence;
 using AchieveAi.LmDotnetTools.LmMultiTurn.SubAgents;
 using AchieveAi.LmDotnetTools.LmStreaming.AspNetCore.Extensions;
+using AchieveAi.LmDotnetTools.LmWorkflow.Prompts;
+using AchieveAi.LmDotnetTools.LmWorkflow.Runtime;
+using AchieveAi.LmDotnetTools.LmWorkflow.Tools;
 using AchieveAi.LmDotnetTools.LmTestUtils;
 using AchieveAi.LmDotnetTools.McpMiddleware.Extensions;
 using AchieveAi.LmDotnetTools.McpServer.AspNetCore.Extensions;
@@ -663,6 +666,7 @@ try
                 // Add LlmQuery book search MCP tools — only for medical knowledge mode
                 // Track MCP clients for proper disposal alongside the agent
                 List<IAsyncDisposable>? ownedResources = null;
+                WorkflowRuntime? workflowRuntime = null;
                 if (!string.IsNullOrEmpty(mcpBaseUrl))
                 {
                     var (_, mcpClients) = ConnectLlmQueryMcpClients(
@@ -685,6 +689,13 @@ try
                     // is independent of the sandbox, so add it here unconditionally; it stays usable
                     // even when the sandbox MCP endpoint is offline (degraded mode below).
                     _ = filteredRegistry.AddFunctionsFromObject(new TaskManager(), providerName: "TaskManager");
+
+                    // Let the controller LLM author and drive an LmWorkflow workflow inside this
+                    // same conversation loop (no separate WorkflowSession-owned loop) — the runtime
+                    // observes the loop's own message stream (see ownedResources wiring below) to
+                    // correlate Agent tool-call spawns back into workflow state.
+                    workflowRuntime = new WorkflowRuntime(logger: loggerFactory.CreateLogger<WorkflowRuntime>());
+                    _ = filteredRegistry.AddProvider(new WorkflowToolProvider(workflowRuntime));
 
                     // Workspace Agent mode: expose the sandbox file/shell tools via the gateway's
                     // MCP endpoint, bound to this agent's sandbox session by the X-Session-ID header.
@@ -728,6 +739,13 @@ try
                                 threadId
                             );
                     }
+
+                    // Append last, off whatever effectiveMode currently is (default or the
+                    // degraded-sandbox notice above) so the controller prompt survives either path.
+                    effectiveMode = effectiveMode with
+                    {
+                        SystemPrompt = effectiveMode.SystemPrompt + "\n\n" + ControllerSystemPrompt.Default,
+                    };
                 }
 
                 // WebFetch/WebSearch fallback tools for providers without a native web capability.
@@ -858,6 +876,47 @@ try
                         subAgentTemplateSource: sharedSubAgentSource,
                         loggerFactory: loggerFactory
                     );
+
+                    if (workflowRuntime is not null)
+                    {
+                        // Correlate sub-agent spawn results back into the workflow runtime via a
+                        // second, independent subscriber to this loop's own message stream (tool
+                        // handlers above cover the synchronous SetWorkflow/SetCurrentNode/etc. path;
+                        // this covers the async Agent-tool-call/result correlation). The first
+                        // MoveNextAsync() is primed synchronously here — before this factory returns
+                        // "agent" to the caller — so the per-subscriber channel is registered before
+                        // any SendAsync can occur on this conversation.
+                        var observerCts = new CancellationTokenSource();
+                        var enumerator = agent.SubscribeAsync(observerCts.Token)
+                            .GetAsyncEnumerator(observerCts.Token);
+                        var pendingFirstMoveNext = enumerator.MoveNextAsync();
+                        var observerTask = Task.Run(async () =>
+                        {
+                            try
+                            {
+                                if (await pendingFirstMoveNext)
+                                {
+                                    workflowRuntime.ObserveMessage(enumerator.Current);
+                                    while (await enumerator.MoveNextAsync())
+                                    {
+                                        workflowRuntime.ObserveMessage(enumerator.Current);
+                                    }
+                                }
+                            }
+                            catch (OperationCanceledException)
+                            { /* expected on shutdown */
+                            }
+                            finally
+                            {
+                                await enumerator.DisposeAsync();
+                            }
+                        });
+                        ownedResources =
+                        [
+                            .. ownedResources ?? [],
+                            new WorkflowObserverDisposable(observerCts, observerTask),
+                        ];
+                    }
 
                     return new MultiTurnAgentPool.AgentCreationResult(agent, ownedResources);
                 }
@@ -1130,6 +1189,28 @@ public partial class Program
     }
 
     /// <summary>
+    ///     Cancels and awaits the background workflow-observer task on disposal, mirroring the same
+    ///     cancel→await/swallow→dispose idiom <c>MultiTurnAgentPool.AgentEntry.DisposeAsync</c>
+    ///     already uses for the agent's own run loop.
+    /// </summary>
+    private sealed class WorkflowObserverDisposable(CancellationTokenSource cts, Task observerTask)
+        : IAsyncDisposable
+    {
+        public async ValueTask DisposeAsync()
+        {
+            await cts.CancelAsync();
+            try
+            {
+                await observerTask;
+            }
+            catch
+            { /* best-effort shutdown */
+            }
+            cts.Dispose();
+        }
+    }
+
+    /// <summary>
     ///     Creates a test-mode agent using an <see cref="ITestAgentBuilder"/>-supplied handler.
     /// </summary>
     private static IStreamingAgent CreateTestAgent(ILoggerFactory loggerFactory, ITestAgentBuilder testAgentBuilder)
@@ -1208,9 +1289,27 @@ public partial class Program
             name,
             s_copilotTokenProvider.Value,
             s_copilotSession.Value,
-            logger: loggerFactory.CreateLogger<AnthropicAgent>()
+            logger: loggerFactory.CreateLogger<AnthropicAgent>(),
+            timeout: CopilotResponseTimeout
         );
     }
+
+    /// <summary>
+    ///     Time-to-first-response ceiling for Copilot-backed streaming agents. Because the streaming
+    ///     clients read with <c>ResponseHeadersRead</c>, this bounds how long a stuck/dead upstream
+    ///     connection blocks before it surfaces as a timeout — it does NOT cap a healthy stream's length.
+    ///     A tight-but-generous default (120s) keeps a hung backend from freezing a whole run (including a
+    ///     blocking sub-agent spawn) for the full 5-minute HTTP default. Override with
+    ///     <c>COPILOT_RESPONSE_TIMEOUT_SECONDS</c>.
+    /// </summary>
+    private static TimeSpan CopilotResponseTimeout =>
+        int.TryParse(
+            Environment.GetEnvironmentVariable("COPILOT_RESPONSE_TIMEOUT_SECONDS"),
+            out var seconds
+        )
+        && seconds > 0
+            ? TimeSpan.FromSeconds(seconds)
+            : TimeSpan.FromSeconds(120);
 
     /// <summary>
     ///     Creates an OpenAI Responses agent (GPT-5.5 / GPT-5.5 mini) routed through the GitHub Copilot
--- a/src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
+++ b/src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
@@ -31,6 +31,13 @@ public static class CopilotAnthropicAgentFactory
     /// <param name="performanceTracker">Optional performance tracker.</param>
     /// <param name="logger">Optional logger.</param>
     /// <param name="retryOptions">Optional retry configuration.</param>
+    /// <param name="timeout">
+    ///     Optional HTTP timeout. Because the streaming client reads with
+    ///     <see cref="System.Net.Http.HttpCompletionOption.ResponseHeadersRead"/>, this bounds the
+    ///     time-to-first-response (a dead/stuck connection), NOT the length of a healthy stream — so a
+    ///     shorter value makes an unresponsive backend surface promptly instead of hanging for the full
+    ///     default. Defaults to the shared 5-minute HTTP default.
+    /// </param>
     public static AnthropicAgent Create(
         string name,
         ICopilotTokenProvider tokenProvider,
@@ -38,7 +45,8 @@ public static class CopilotAnthropicAgentFactory
         CopilotOptions? options = null,
         IPerformanceTracker? performanceTracker = null,
         ILogger<AnthropicAgent>? logger = null,
-        RetryOptions? retryOptions = null
+        RetryOptions? retryOptions = null,
+        TimeSpan? timeout = null
     )
     {
         ArgumentNullException.ThrowIfNull(name);
@@ -54,7 +62,7 @@ public static class CopilotAnthropicAgentFactory
         var context = session ?? new CopilotSessionContext();
         var host = copilotOptions.BaseUrl.TrimEnd('/');
 
-        var httpClient = CopilotHttpClientFactory.Create(host, tokenProvider, context, copilotOptions);
+        var httpClient = CopilotHttpClientFactory.Create(host, tokenProvider, context, copilotOptions, timeout);
 
         // The client appends "/messages", so the base URL must carry the "/v1" prefix to hit
         // {host}/v1/messages.
--- a/src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
+++ b/src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
@@ -24,6 +24,11 @@ public static class CopilotAnthropicAgentFactory
     /// <summary>
     ///     Creates an <see cref="AnthropicAgent"/> routed through GitHub Copilot.
     /// </summary>
+    /// <remarks>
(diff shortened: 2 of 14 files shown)


src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
```diff
@@ -31,14 +31,22 @@ public static class CopilotAnthropicAgentFactory
...
     /// <param name="logger">Optional logger.</param>
     /// <param name="retryOptions">Optional retry configuration.</param>
+    /// <param name="timeout">
+    ///     Optional HTTP timeout. Because the streaming client reads with
+    ///     <see cref="System.Net.Http.HttpCompletionOption.ResponseHeadersRead"/>, this bounds the
+    ///     time-to-first-response (a dead/stuck connection), NOT the length of a healthy stream — so a
+    ///     shorter value makes an unresponsive backend surface promptly instead of hanging for the full
+    ///     default. Defaults to the shared 5-minute HTTP default.
+    /// </param>
     public static AnthropicAgent Create(
```
gautam-achieveai, line 79: [Claude's bot] [BLOCKER] **HIGH** (API compatibility)

Adding `TimeSpan? timeout = null` to the existing public `CopilotAnthropicAgentFactory.Create(...)` method changes the CLR method signature. Source callers that recompile are fine, but any already-compiled SDK consumer that calls the old 8-parameter method can fail at runtime with `MissingMethodException` after upgrading the assembly.

**Suggestion:** keep the original overload exactly as-is and delegate to the new timeout-aware overload, e.g.:

```csharp
public static AnthropicAgent Create(
    string name,
    ICopilotTokenProvider tokenProvider,
    CopilotSessionContext? session = null,
    CopilotOptions? options = null,
    IPerformanceTracker? performanceTracker = null,
    ILogger<AnthropicAgent>? logger = null,
    RetryOptions? retryOptions = null) =>
    Create(name, tokenProvider, session, options, performanceTracker, logger, retryOptions, timeout: null);

public static AnthropicAgent Create(
    string name,
    ICopilotTokenProvider tokenProvider,
    CopilotSessionContext? session,
    CopilotOptions? options,
    IPerformanceTracker? performanceTracker,
    ILogger<AnthropicAgent>? logger,
    RetryOptions? retryOptions,
    TimeSpan? timeout)
{
    // current implementation
}
```

This preserves binary compatibility while still enabling the sample to pass a shorter timeout.

src/LmWorkflow/Runtime/WorkflowRuntime.cs
```diff
@@ -413,6 +422,120 @@ internal void ObserveInjectedResult(string agentId, string resultText, bool isEr
...
+    ///     tool-call id, so repeated parses after the name first appears are harmless.
+    /// </summary>
+    private void AccumulateAndRegisterSpawn(string toolCallId, string? fragment)
+    {
+        if (string.IsNullOrEmpty(fragment))
+        {
+            return;
+        }
+
+        var buffer = _spawnArgBuffers.AddOrUpdate(
```
gautam-achieveai, line 519: [Claude's bot] **MEDIUM** (Resource lifecycle)

`_spawnArgBuffers` appends streamed Agent-call argument fragments by `toolCallId`, but the buffer is only removed when a `ToolCallResultMessage` is for an already-registered spawn. In a long-lived Workspace Agent conversation, non-workflow `Agent` calls, malformed/incomplete streamed args, or calls whose `name` never matches an expected workflow unit can retain accumulated fragments for the lifetime of the runtime.

**Suggestion:** bound the buffered data and clear it on more paths. For example:

- enforce a per-tool-call max buffered length and drop/log when exceeded;
- remove the buffer immediately after the spawn name is successfully parsed/registered; and
- clear the buffer on any terminal result for that `toolCallId`, even when the spawn was not registered.

That keeps malformed or unrelated tool-call streams from becoming a slow memory leak in long-lived sessions.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -859,6 +877,47 @@
...
+                        // handlers above cover the synchronous SetWorkflow/SetCurrentNode/etc. path;
+                        // this covers the async Agent-tool-call/result correlation). The first
+                        // MoveNextAsync() is primed synchronously here — before this factory returns
+                        // "agent" to the caller — so the per-subscriber channel is registered before
+                        // any SendAsync can occur on this conversation.
+                        var observerCts = new CancellationTokenSource();
+                        var enumerator = agent.SubscribeAsync(observerCts.Token)
+                            .GetAsyncEnumerator(observerCts.Token);
+                        var pendingFirstMoveNext = enumerator.MoveNextAsync();
+                        var observerTask = Task.Run(async () =>
```
gautam-achieveai, line 914: [Claude's bot] **MEDIUM** (Reliability / observability)

This observer is the critical path that correlates Agent tool calls/results back into `WorkflowRuntime`, but non-cancellation failures in the background task are not logged when they occur. If `SubscribeAsync`, `MoveNextAsync`, or `workflowRuntime.ObserveMessage(...)` faults, workflow correlation stops while the agent can keep serving workflow tool calls. Disposal then awaits the task and swallows any exception, so the failure may never be visible.

**Suggestion:** log unexpected observer failures immediately and, ideally, mark workflow observation unhealthy. At minimum, catch/log `Exception` in the task body while continuing to treat `OperationCanceledException` as expected shutdown:

```csharp
try
{
    // observer loop
}
catch (OperationCanceledException) when (observerCts.IsCancellationRequested)
{
    // expected on shutdown
}
catch (Exception ex)
{
    loggerFactory.CreateLogger<WorkflowRuntime>()
        .LogError(ex, "Workflow observer failed; workflow Agent-result correlation is disabled for this conversation");
}
finally
{
    await enumerator.DisposeAsync();
}
```

The cleanup `DisposeAsync` can still be best-effort, but the first unexpected failure should be observable.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -686,6 +690,13 @@
                     // even when the sandbox MCP endpoint is offline (degraded mode below).
                     _ = filteredRegistry.AddFunctionsFromObject(new TaskManager(), providerName: "TaskManager");
 
+                    // Let the controller LLM author and drive an LmWorkflow workflow inside this
+                    // same conversation loop (no separate WorkflowSession-owned loop) — the runtime
+                    // observes the loop's own message stream (see ownedResources wiring below) to
+                    // correlate Agent tool-call spawns back into workflow state.
+                    workflowRuntime = new WorkflowRuntime(logger: loggerFactory.CreateLogger<WorkflowRuntime>());
```
gautam-achieveai: [Claude's bot] **MEDIUM** (Rollout safety)

Workspace Agent mode now always registers the workflow tools and always appends the workflow controller prompt. Later in the same factory it also starts the second observer subscriber whenever `workflowRuntime` is present. That may be the intended default eventually, but it changes every Workspace Agent conversation: new tool surface area, a longer/more directive prompt, and an extra background observer path.

**Suggestion:** add a config/env kill switch around the three behaviors as one unit, e.g. `WORKSPACE_AGENT_LMWORKFLOW_ENABLED`. If the feature should be on by default, make it default `true` but still allow operators to disable it without redeploying:

```csharp
var workflowEnabled = !string.Equals(
    Environment.GetEnvironmentVariable("WORKSPACE_AGENT_LMWORKFLOW_ENABLED"),
    "false",
    StringComparison.OrdinalIgnoreCase);

if (workflowEnabled)
{
    workflowRuntime = new WorkflowRuntime(...);
    filteredRegistry.AddProvider(new WorkflowToolProvider(workflowRuntime));
    effectiveMode = effectiveMode with
    {
        SystemPrompt = effectiveMode.SystemPrompt + "\n\n" + ControllerSystemPrompt.Default,
    };
}
```

Then gate the observer subscription on the same flag/runtime.

src/LmWorkflow/Tools/WorkflowToolProvider.cs
```diff
@@ -129,7 +129,11 @@ CancellationToken cancellationToken
             WorkflowDefinition definition;
             try
             {
-                definition = WorkflowJson.Deserialize(definitionElement.GetRawText());
+                // Strict deserialize so a misspelled or invented field (e.g. 'tasks' for 'taskList',
+                // 'agentType' for 'subagent_type') is rejected by name instead of silently dropped —
+                // a silently-dropped task field used to yield a workflow that validated clean yet ran
+                // nothing, giving the authoring LLM no signal to correct.
+                definition = WorkflowJson.DeserializeStrict(definitionElement.GetRawText());
```
gautam-achieveai, line 136: [Claude's bot] [QUESTION] **Clarification Needed**

**Code:**
`definition = WorkflowJson.DeserializeStrict(definitionElement.GetRawText());`

**Question:** Is `SetWorkflow` exclusively an LLM-authoring tool, or can external callers/users submit saved or externally generated workflow definitions through this public tool surface?

**Why this matters:** Strict unknown-member rejection is a good authoring feedback mechanism, but if `SetWorkflow` is also an external import path, this tightens the accepted wire contract: definitions with forward-compatible metadata/vendor extension fields that were previously ignored now fail as `invalid_workflow`.

**Possible answers:**
- `SetWorkflow` is authoring-only; strict mode is the intended contract.
- External/import callers are expected; this should stay lenient by default or expose a separate strict authoring mode.
- Extension fields are unsupported and the strict rejection should be documented as a breaking/intentional behavior change.

gautam-achieveai:
[Claude's bot] # PR Review: feat(workflow): author & run LmWorkflow in Workspace Agent mode

## Verdict: REQUEST CHANGES

Thanks for the detailed PR description and the added regression coverage around workflow authoring/correlation. I found one blocking issue that should be fixed before merge: the timeout parameter changes a public SDK method's binary signature.

### Findings

| Severity | Category | File | Summary |
|---|---|---|---|
| HIGH / BLOCKER | API compatibility | `src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs:41` | Adding an optional parameter to the existing public `Create` method breaks binary compatibility for already-compiled consumers. |
| MEDIUM | Resource lifecycle | `src/LmWorkflow/Runtime/WorkflowRuntime.cs:505` | Streaming Agent arg buffers can remain unbounded for malformed/unmatched/incomplete calls in long-lived sessions. |
| MEDIUM | Reliability / observability | `samples/LmStreaming.Sample/Program.cs:893` | Workflow observer task failures can silently disable correlation with no immediate log/health signal. |
| MEDIUM | Rollout safety | `samples/LmStreaming.Sample/Program.cs:697` | Workspace Agent workflow tools/prompt/observer are always on with no kill switch. |

### Context question

I also asked one non-blocking clarification question about whether strict `SetWorkflow` deserialization is authoring-only or also an external import contract.

### Note on dropdown scrolling

I verified the reported dropdown-scroll issue. `ProviderSelector.vue`, `ModeSelector.vue`, and the similar `WorkspaceSelector.vue` currently use `overflow: hidden` without a viewport-bounded `max-height`/`overflow-y: auto`, so long lists can become inaccessible. Those files are not touched by this PR, so I did not post it as a PR-blocking finding here; it should be tracked as a separate UI/accessibility follow-up.

### Verification notes

- Reviewed worktree: `/workspace/LmDotNetTools/pr131-review-014042`
- Head reviewed: `72fc55c64ba9886b4f8af582d936598b2b4baa2d`
- Merge-base: `b921afeaf15e85504b3f6ef72e1ea39fbde73211`
- CI observed via `gh pr checks`: Build and test ✅, browser-e2e ✅
- Local `dotnet` verification could not be run in this sandbox because `dotnet` is not installed.

Once the original `CopilotAnthropicAgentFactory.Create` overload is restored and delegates to the timeout-aware overload, the remaining items are non-blocking improvements rather than merge blockers.

src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
gautam-achieveai: Fixed in a28e53c. Restored the original 7-parameter `Create(...)` overload verbatim (binary-compatible — precompiled consumers keep resolving to it) and delegated it to a new timeout-aware overload.

To keep the two unambiguous, `timeout` is a **required** parameter on the new overload (pass `null` for the default) — a call either supplies a timeout (→ new overload) or doesn't (→ original), so no call can match both. The remaining parameters stay optional for ergonomics.

src/LmWorkflow/Runtime/WorkflowRuntime.cs
gautam-achieveai: Fixed in a28e53c. The reassembly buffer is now bounded on all three paths you listed:
- **drop on parse** — removed the moment the args become parseable (the name is then known; later fragments are noise), before the sub-agent even finishes;
- **cap** — dropped if it exceeds 256 KiB without ever parsing (malformed / not a workflow spawn);
- **clear on any result** — the `ToolCallResultMessage` case now clears the buffer unconditionally, not only for registered spawns, so non-workflow `Agent` calls and unmatched names can't retain fragments.

Added a regression test (`ObserveMessage_NonWorkflowAgentCall_IsHarmless_AndDoesNotBreakLaterCorrelation`) proving an unrelated Agent call is harmless and doesn't contaminate a subsequent real spawn.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: Fixed in a28e53c. The observer background task now catches `Exception` and logs an error (`Workflow observer ... failed; Agent-result correlation is now disabled for this conversation`) while still treating `OperationCanceledException` on shutdown as expected. Disposal stays best-effort, but the first unexpected failure is now observable instead of being swallowed.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: Fixed in a28e53c. Added `WORKSPACE_AGENT_LMWORKFLOW_ENABLED` (default `true`). It gates all three behaviors as one unit — the workflow tool surface, the appended controller prompt, and the correlation observer. `workflowRuntime` is now only constructed when enabled, and both the prompt append and the observer subscription are gated on it, so operators can disable the feature without a redeploy.

src/LmWorkflow/Tools/WorkflowToolProvider.cs
gautam-achieveai: `SetWorkflow` is **authoring-only** — it's exposed as a function tool to the *controller LLM* driving the workflow, not a general import API. So strict unknown-member rejection is the intended contract (it's the authoring feedback loop this change is about).

Two things keep it from being a surprising breaking change:
1. The forward-compat props the validator explicitly accepts (`task.parallel`, `node.maxParallel`, `joinPolicy.threshold`, `writes.key`, `tasksMode: runtime`) are **modeled** members, so they still deserialize and are then rejected by the validator with clear "not supported in V1" messages. Strict mode only rejects *genuinely unknown* keys.
2. The free-form JSON channels (`state`, `inputs`, `$defs`, `outputSchema`, `finalOutputSchema`, `resultTemplate`) are `JsonNode`/`JsonObject`, so arbitrary vendor/extension keys *inside* them are untouched.

Persistence/round-trip stays on the lenient `WorkflowJson.Options` (only the `SetWorkflow` authoring path uses `DeserializeStrict`), so already-saved snapshots keep loading. If an external/import path is added later, it should use the lenient deserializer (or a separate opt-in strict mode).

gautam-achieveai:
Addressed all five review points in a28e53c:

- **[BLOCKER] Binary compatibility** — restored the original 7-parameter `CopilotAnthropicAgentFactory.Create` overload verbatim; it delegates to a new required-`timeout` overload (unambiguous by construction).
- **[MEDIUM] Buffer lifecycle** — the streamed-args reassembly buffer is dropped on parse, capped at 256 KiB, and cleared on any tool result; added a regression test for non-workflow Agent calls.
- **[MEDIUM] Observer observability** — the correlation background task now logs unexpected failures instead of swallowing them at disposal.
- **[MEDIUM] Rollout safety** — added `WORKSPACE_AGENT_LMWORKFLOW_ENABLED` (default true) gating the tool surface, prompt, and observer as one unit.
- **[QUESTION] Strict deserialize** — answered inline: `SetWorkflow` is authoring-only, so strict is intended; modeled forward-compat props and free-form JSON channels are unaffected, and persistence stays lenient.

Tests green after the changes: LmWorkflow 195, GithubCopilotProvider 35, LmStreaming.Sample 359, LmMultiTurn 285.

src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs
```diff
@@ -39,6 +44,47 @@ public static AnthropicAgent Create(
...
+    ///     time-to-first-response (a dead/stuck connection), NOT the length of a healthy stream — so a
+    ///     shorter value makes an unresponsive backend surface promptly instead of hanging for the full
+    ///     default.
+    /// </param>
+    /// <param name="session">Optional shared tracking ids; a new context is created when omitted.</param>
+    /// <param name="options">Optional Copilot header options; defaults target the enterprise host.</param>
+    /// <param name="performanceTracker">Optional performance tracker.</param>
+    /// <param name="logger">Optional logger.</param>
+    /// <param name="retryOptions">Optional retry configuration.</param>
+    public static AnthropicAgent Create(
```
gautam-achieveai, line 79: [Claude's bot] [BLOCKER] **HIGH** (API/source compatibility)

Restoring the original overload fixes the binary compatibility problem, but this new overload still creates a source compatibility break for existing callers that passed `null` positionally for the old optional `session` parameter:

```csharp
CopilotAnthropicAgentFactory.Create("agent", tokenProvider, null);
```

That call now has two viable overloads because `null` converts to both `CopilotSessionContext?` and `TimeSpan?`, so consumers can fail to compile after upgrading without changing their code.

**Suggestion:** avoid a nullable third positional parameter on the new overload. For example, make the timeout parameter non-nullable so `null` continues to bind only to the original session overload:

```csharp
public static AnthropicAgent Create(
    string name,
    ICopilotTokenProvider tokenProvider,
    TimeSpan timeout,
    CopilotSessionContext? session = null,
    CopilotOptions? options = null,
    IPerformanceTracker? performanceTracker = null,
    ILogger<AnthropicAgent>? logger = null,
    RetryOptions? retryOptions = null)
```

Callers that want the default timeout can use the original overload. Alternatively, use a differently named method/options object for timeout-aware construction. Please also add a source-compat regression test for `Create("Sonnet", tokenProvider, null)`.

src/LmWorkflow/Runtime/WorkflowRuntime.cs
```diff
@@ -413,6 +424,142 @@ internal void ObserveInjectedResult(string agentId, string resultText, bool isEr
...
+    ///     the tool result (see <see cref="ObserveMessage"/>).
+    /// </remarks>
+    private void AccumulateAndRegisterSpawn(string toolCallId, string? fragment)
+    {
+        if (string.IsNullOrEmpty(fragment))
+        {
+            return;
+        }
+
+        var buffer = _spawnArgBuffers.AddOrUpdate(
```
gautam-achieveai, line 519: [Claude's bot] **MEDIUM** (Performance / memory resilience)

The buffer cap is checked after `AddOrUpdate` has already concatenated the existing buffer and incoming fragment, and after attempting to parse the resulting string. That means a single oversized `FunctionArgs` fragment — or a large fragment appended to a near-cap buffer — is still fully materialized and passed to `JsonDocument.Parse` before it is dropped.

**Suggestion:** enforce the cap before concatenation/parsing. For example, check `existing.Length + fragment.Length` inside the `AddOrUpdate` update path and drop/reject before building the combined string, or pre-check the single-fragment length before adding it. The cap should prevent the allocation/parse work, not just clean up afterward.

tests/LmWorkflow.Tests/ObserveMessageCorrelationTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /// <summary>
+    ///     A non-workflow Agent call (a name that matches no authored unit — the common case in a
+    ///     Workspace Agent conversation) must be handled gracefully: its streamed fragments and its result
+    ///     do not correlate, do not throw, and — critically — do not contaminate a subsequent REAL workflow
+    ///     spawn. This exercises the hardened buffer lifecycle (drop-on-parse / clear-on-any-result).
+    /// </summary>
+    [Fact]
+    public void ObserveMessage_NonWorkflowAgentCall_IsHarmless_AndDoesNotBreakLaterCorrelation()
```
gautam-achieveai, line 156: [Claude's bot] **MEDIUM** (Test coverage)

This test verifies that a parseable non-workflow Agent call does not break a later real workflow spawn, but it does not prove the stale-buffer cleanup path that this delta is meant to protect. Because the real workflow spawn uses a different tool-call id (`tc_agent` vs `tc_other`), the old leak behavior could still retain `tc_other` indefinitely and this test would pass.

**Suggestion:** add a regression test that fails without clear-on-result. For example:

1. send unparseable `ToolCallUpdateMessage` fragments for a tool-call id;
2. send a `ToolCallResultMessage` for that same id, which should clear the stale buffer;
3. reuse the same tool-call id for a valid streamed Agent call for the authored workflow unit; and
4. assert the task validates.

Without the clear-on-any-result fix, the stale malformed prefix would remain and the later valid stream would not parse/correlate.

gautam-achieveai:
[Claude's bot] ## Re-Review Summary: PR #131

## Verdict: REQUEST CHANGES

Thanks for the quick update. The previous binary-compatibility issue is partially fixed: the original overload is back, so already-compiled callers should keep resolving. However, the new timeout overload still introduces a source-compatibility ambiguity for existing callers that pass `null` as the third positional argument.

### Findings in this delta

| Severity | Category | File | Summary |
|---|---|---|---|
| HIGH / BLOCKER | API/source compatibility | `src/GithubCopilotProvider/Agents/CopilotAnthropicAgentFactory.cs:79` | `TimeSpan? timeout` as the third positional parameter makes existing `Create(name, tokenProvider, null)` calls ambiguous with the original `session` overload. |
| MEDIUM | Performance / memory resilience | `src/LmWorkflow/Runtime/WorkflowRuntime.cs:519` | The spawn-arg buffer cap is checked after concatenation and parse, so oversized fragments can still allocate/parse before being dropped. |
| MEDIUM | Test coverage | `tests/LmWorkflow.Tests/ObserveMessageCorrelationTests.cs:156` | The new buffer lifecycle test does not prove stale-buffer clear-on-result because the subsequent real spawn uses a different tool-call id. |

### Previously raised items

- Workflow observer failure logging: addressed.
- Workspace Agent workflow kill switch: addressed.
- Strict `SetWorkflow` contract question: answered by author; no further action from me.
- Dropdown scrolling: still an existing UI follow-up outside this PR delta.

### Verification notes

- Re-reviewed worktree: `/workspace/LmDotNetTools/pr131-rereview-022523`
- Prior reviewed head: `72fc55c64ba9886b4f8af582d936598b2b4baa2d`
- Latest head: `a28e53c45d7c889c7e9e555ea2db7c97845156b5`
- Delta reviewed: 4 files changed.
- CI observed: browser-e2e ✅; Build and test was still in progress at the time I checked.
- Local `dotnet` verification could not be run because `dotnet` is not installed in this sandbox.

Once the timeout overload no longer creates the positional-`null` ambiguity, the remaining items are non-blocking improvements.

gautam-achieveai (commented):
[Gautam's review bot] ## Third-Pass Review: PR #131

## Verdict: APPROVE (with non-blocking suggestions)

### Work Item Context
- **User Story #130**: Enable LLM-authored, runnable LmWorkflow in Workspace Agent mode

### Assessment of Open Items from Previous Re-Review

**1. Source compatibility ambiguity (previously BLOCKER)**

I've verified all callers in the repo: `CopilotAnthropicLiveTests` passes actual `CopilotSessionContext` instances positionally (`_fixture.Session`), the sample uses named `timeout:`, and the timeout test uses either 2-arg or named params. **No caller in this codebase — or any plausible external caller — would write `Create(name, provider, null)` because `session` already defaults to `null`.**

The ambiguity is technically real but has **zero practical risk**. Downgrading to **LOW** — this can be addressed as a follow-up if external SDK consumers are expected. The current design (required `timeout` param differentiating the overloads) is otherwise sound.

**2. Buffer cap checked after concatenation (MEDIUM)**

Agree the cap is checked after `AddOrUpdate` materializes the combined string. However, with a 256 KB cap this means at worst one transient ~256 KB + fragment_size allocation before cleanup — Agent args JSON fragments are typically < 1 KB. The parse attempt on the oversized buffer is wasteful but harmless at this scale. Non-blocking.

**3. Test coverage for stale-buffer cleanup (MEDIUM)**

Valid suggestion — the current `NonWorkflowAgentCall_IsHarmless` test uses distinct tool-call IDs, so it proves isolation but not the "clear on any result" path specifically. A same-id reuse test would be stronger. Non-blocking.

### New Observations

**`ControllerSystemPrompt.Default`: `const` to `static readonly`** (LOW) — Changes metadata from a compile-time constant to a runtime field read. Pre-compiled consumers have the old value baked in. Since this is an internal workflow prompt, acceptable.

**Thread safety of `ObserveMessage`** — verified sound. `_spawnArgBuffers` is `ConcurrentDictionary`; all coordinator mutations go through `_lock`. Single subscriber guarantees fragment-to-result ordering.

### Summary

The iteration 2 commit thoroughly addressed the original blocker and all three MEDIUM items. Remaining open suggestions are polish items with low practical risk. The feature is well-tested, live-verified, and safely gated behind `WORKSPACE_AGENT_LMWORKFLOW_ENABLED`. Ready to merge.

gautam-achieveai:
[Gautam's review bot] **Downgrading to LOW.** I've verified every caller in the repo and there is no instance of `Create(name, provider, null)` — all existing callers either pass a non-null `CopilotSessionContext` positionally or use named parameters. Since `session` already defaults to `null`, passing it explicitly as a positional null argument would be pointless, so no real-world caller would do this.

The `TimeSpan? timeout` parameter being **required** on the new overload (no default) already disambiguates: the compiler picks the original 7-param overload for any call that doesn't supply a timeout argument, and the new overload for any call that does. The ambiguity only arises for the synthetic `(name, provider, null)` call that no one would write.

Non-blocking. If external SDK consumers with exotic calling patterns are a concern, this can be addressed as a follow-up.

PR #139, opened 2026-07-05
feat(triggers): generic Wait/CancelWait/ListWaits primitive (#107)

### LINKED ISSUE #107: Generic Wait/trigger primitive to park & wake the agent on background events
## Summary

Add a generic **wait-for-trigger** primitive that lets an LLM, mid multi-turn
conversation, **arm a trigger and have its loop parked or notified until that
trigger fires** — then wake with a self-describing payload. This generically
solves "the agent needs a background task to finish, or an external/scheduled
event to occur, before it continues."

The design adds **one extensibility seam** and otherwise **reuses what already
ships** — the durable park/wake core, injection path, concurrency gate, teardown
discipline, persistence, and process launcher all exist today:

- **The durable park/wake core already exists.** A tool handler that returns
  `ToolHandlerResult.Deferred()` parks the run (placeholder persisted to
  `IConversationStore`); `MultiTurnAgentLoop.ResolveToolCallAsync(toolCallId,
  payload)` injects the result and a `ResumeSentinel` wakes the loop;
  `OnHistoryRestoredAsync` already rebuilds the deferred set on restart.
  Background sub-agent completion (#75) and context-discovery injection (#81)
  already prove the "external event wakes the agent" path via `SendAsync`.
  `toolCallId` is the opaque correlation key; resolve is already idempotent for
  byte-equal redelivery.
- **The one new abstraction is `ITriggerSource`** — arm → fires payload → cancel,
  single-fire or multi-fire, `IAsyncDisposable`. **Adding a new trigger kind =
  implement `ITriggerSource` + register; no loop change, no new tool.** This is
  the "extremely extensible" property. A `TriggerRegistry` owns lifecycle, the
  concurrency bound, arming-record persistence, and the total `ListWaits` /
  `CancelWait` functions — **reusing `SubAgentManager`'s proven concurrency-gate
  and disposal patterns rather than re-deriving them.**
- **One generic LLM tool** (`Wait` / `CancelWait` / `ListWaits`), `kind`-dispatched
  1:1 onto the source registry. A single **parking flag** — `mode` — mirrors the
  existing `Agent` blocking-vs-background duality (it is that same axis, **not** a
  fire-count flag; count is the separate, optional `maxFires`):
  - **`mode:"block"`** (default) `Wait(...)` returns `Deferred()`; the agent parks;
    the first fire (or the mandatory timeout) resolves it and the payload becomes the
    tool result. A block wait always fires once.
  - **`mode:"notify"`** `Wait(...)` returns a `trigger_id` immediately (like
    `Agent(run_in_background:true)`) and the agent keeps working; each fire injects a
    self-describing `<trigger kind id>…</trigger>` message via `SendAsync` until
    `CancelWait`, `maxFires`, or the deadline. `maxFires:1` is the canonical
    "notify me once, then keep working" case (background single-shot) — no separate
    `recurring` concept is needed.

**v1 built-in sources** (each small): **Timer** (delay | interval | cron),
**Process-exit** (run a command through the existing `IProcessLauncher`; fire on
exit with an optional exit-code / stdout-regex predicate), **Log/file-tail**
(fire when a new line matches a regex), and **sub-agent completion** re-exposed
through the same surface. External-signal/webhook is deferred (the seam makes it
a later drop-in).

This is the SDK analog of the harness "wait/wake" primitives (`ScheduleWakeup`,
`Monitor`, cron, background-task completion) and of the durable-execution
"suspend on {timer | signal | predicate}, resume with payload" pattern (LangGraph
`interrupt`/`Command(resume)`, Temporal signals/`await`, Inngest `waitForEvent`,
Step Functions `.waitForTaskToken`). It is complementary to, and **should share a
background-work substrate with, the sibling `LmWorkflow` proposal (#106)** (see
*Shared substrate* below).

## Value

- **SDK consumers building agentic apps** get a first-class, generic way to make
  the agent wait on background/async/scheduled work instead of either blocking a
  tool handler synchronously or hand-rolling per-app polling. One tool, any number
  of trigger kinds.
- **Extensibility is the headline** — the pluggable `ITriggerSource` seam means
  app authors (and future SDK versions) add new wake sources without touching the
  agent loop or the tool surface.
- **Reuses, not reinvents** — the durable park/wake, injection, idempotent
  resolution, restart recovery, persistence, bounded-concurrency gate, teardown
  discipline, and process launcher all already exist; this adds only the thin
  trigger layer on top.
- **Safe by construction** — every wait carries a mandatory deadline (no leaked
  waits), payloads are size-capped (no context/cost blow-ups or secret dumps into
  transcripts), file/command access is allowlisted, and teardown leaves no
  orphaned watchers/processes/timers.

## Design (the minimal generic set)

**Canonical tool surface — kept deliberately small (3 core params, 3 optional):**

```text
Wait({
  kind,        // required — a registered source: timer | process | file_tail | subagent | …
  args,        // required — per-kind options (delay/interval/cron · command+predicate · path+match · agent_id)
  timeout,     // required — duration ("10m") OR absolute time ("2026-07-31");
               //   block: park ceiling · notify: total armed lifetime
  mode?,       // "block" (default — parks, returns the payload) | "notify" (returns trigger_id, keeps working)
  maxFires?,   // notify only — bound the number of fires; maxFires:1 = "notify me once, then keep working"
  label?,      // short + self-describing; shown in ListWaits and echoed on every fire (self-contained wakes)
})
CancelWait({ id | label | kind })   // total + idempotent; label/kind cancels all matching
ListWaits()                          // status of every armed wait + the registered kinds and their args schemas
```

The mode/count split keeps `Agent`'s blocking-vs-background duality as the *only*
conceptual axis the model must learn; everything else is per-`kind` data in `args`.

| Piece | Responsibility | Reuse |
|---|---|---|
| `ITriggerSource` | Arm with parsed args + a single-resolution latch; emit `Fired(payload)`; `Cancel`; declare single-fire/multi-fire capability; `IAsyncDisposable`. **The extensibility seam — net-new.** | — |
| `TriggerRegistry` (owned by the loop) | Arm/track/cancel sources; enforce `MaxConcurrentWaits`; persist minimal arming records; serve `ListWaits`/`CancelWait` as total functions; dispose all sources on teardown. | Concurrency-gate + isolated-teardown loop from `SubAgentManager`. |
| Fired-source → loop back-reference | A fired source calls the loop's `ResolveToolCallAsync` (`block`) / `SendAsync` (`notify`); a thin **dispose-tolerant wrapper** makes a fire after dispose a no-op. | The existing `ResolveToolCallAsync`/`SendAsync` + the `_parentAgent` back-reference pattern from `SubAgentManager`. No new resolution/idempotency machinery. |
| `Wait` / `CancelWait` / `ListWaits` | The generic, `kind`-dispatched tool surface (one `IFunctionProvider`); `mode:"block"\|"notify"` is the only parking flag. | `IFunctionProvider`/`FunctionDescriptor`; the `Agent` blocking-vs-background shape. |
| Built-in sources | `TimerTriggerSource`, `ProcessTriggerSource`, `FileTailTriggerSource`, sub-agent-completion source. | `IProcessLauncher` (Process); the existing completion latch (sub-agent). |

## Usage examples (vetted against the surface above)

These map natural-language intent → the exact call and the wake the model receives.
Every example uses only the canonical surface (no `recurring`/`until`/`note`); they
double as the ergonomic vetting that this 3-core-param shape covers all four seed
ideas and their common variants. (`predicate` = optional exit-code and/or
stdout-regex match.)

**1 — "wake me when the refactor sub-agent is done" (block, default):**
```jsonc
Wait({ kind:"subagent", args:{ agent_id:"sa_42" }, timeout:"30m",
       label:"refactor sub-agent" })
// parks → tool result:
{ status:"fired", kind:"subagent", agent_id:"sa_42",
  result:"Refactor complete. 3 files changed, all tests pass." }
```

**2 — "run the build, tell me when it's done" (block + predicate):**
```jsonc
Wait({ kind:"process",
       args:{ command:"dotnet build LmDotnetTools.sln", working_dir:".",
              predicate:{ stdout_regex:"Build succeeded|Build FAILED" } },
       timeout:"10m", label:"solution build" })
// → { status:"fired", kind:"process", exit_code:1, matched:"Build FAILED",
//     stdout_tail:"… CS0103 … 12 Error(s)\nBuild FAILED.\n", duration:"84s" }
```

**3 — "watch the deploy log; ping me on ERROR/FATAL" (notify, coalesced):**
```jsonc
Wait({ kind:"file_tail",
       args:{ path:".logs/deploy.log", match:"\\bERROR\\b|\\bFATAL\\b" },
       mode:"notify", timeout:"2h", maxFires:50, label:"deploy log errors" })
// → { trigger_id:"t_abc", status:"armed" }  (agent keeps working) → later:
```
```text
<trigger kind="file_tail" id="t_abc" label="deploy log errors">
3 new matching lines (coalesced over 4s):
  12:04:01 ERROR  Connection refused: db:5432
  12:04:02 ERROR  Retry 1/3 failed
  12:04:05 FATAL  Migration aborted
</trigger>
```

**4 — "every weekday 9am, post a standup digest" (notify + absolute `timeout`):**
```jsonc
Wait({ kind:"timer", args:{ cron:"0 9 * * 1-5", timezone:"America/Los_Angeles" },
       mode:"notify", timeout:"2026-07-31",
       label:"post the 9am standup digest from yesterday's merged PRs" })
// each fire injects the label back, so a wake hours later is self-describing:
```
```text
<trigger kind="timer" id="t_def" label="post the 9am standup digest from yesterday's merged PRs">
Scheduled fire 2026-06-29T09:00:00-07:00 (fire 2 of N).
</trigger>
```

**5 — "ping me when the long export finishes, but let me keep working"**
**(the case the `mode`/`maxFires` split fixed — notify + `maxFires:1`):**
```jsonc
Wait({ kind:"process", args:{ command:"npm run export" },
       mode:"notify", maxFires:1, timeout:"1h",
       label:"long export — notify once, I'll keep working" })
// → { trigger_id:"t_ghi", status:"armed" }  → fires once, then auto-terminates.
```

**6 — "wait for the service healthy OR 10 min" (event-or-deadline = the common**
**"any-of", for free via `timeout`):**
```jsonc
Wait({ kind:"process", args:{ command:"./wait-for-healthy.sh" }, timeout:"10m",
       label:"service healthy or 10m" })
// resolves to { status:"fired", exit_code:0, … }  OR  { status:"timed_out" }
```

**7 — "only wake me if the tests fail" (predicate gates firing):**
```jsonc
Wait({ kind:"process",
       args:{ command:"dotnet test", predicate:{ exit_code_not:0 } },
       timeout:"15m", label:"tests — only wake me if they fail" })
// block: failing exit → fired payload; passing exit → { status:"condition_not_met",
//   exit_code:0 } (still unblocks the run — never wedges).
```

**8 — inspect & cancel by label (no need to have kept the id):**
```jsonc
ListWaits()
// → [{ id:"t_abc", kind:"file_tail", status:"armed",
//      label:"deploy log errors", fires_delivered:3 }, …]
CancelWait({ label:"deploy log errors" })   // → { status:"cancelled", fires_delivered:3 }
```

### Core seam & tool surface
- [ ] `ITriggerSource` abstraction + a `kind`-keyed registry; a new source is added
      by implementing the interface and registering it — **no change to the agent
      loop or the tool list** (proven by a test that registers a dummy source and
      arms it through `Wait`).
- [ ] Generic `Wait({ kind, args, timeout, mode?, maxFires?, label? })`,
      `CancelWait({ id | label | kind })`, `ListWaits()` exposed via one
      `IFunctionProvider`. The `Wait` description **enumerates the registered `kind`s
      and each kind's `args` schema** (and `ListWaits` returns them) so the single
      kind-dispatched tool is self-documenting, not an opaque `args` grab-bag.
- [ ] A fired source reaches the loop via the **existing** `ResolveToolCallAsync` /
      `SendAsync` (no new resolution mechanism); a thin wrapper swallows
      `ObjectDisposedException` so a fire after the loop is disposed is a **no-op,
      not a crash** (test).

### Block vs notify semantics (the parking axis)
- [ ] `mode:"block"` (default) returns `Deferred()`, parks the run, and the first
      fire resolves it with the payload as the tool result; a block wait always fires
      once (`maxFires` is coerced to 1).
- [ ] `mode:"notify"` returns `{ trigger_id, status:"armed" }` immediately and each
      fire injects a `<trigger kind id label>…</trigger>` message via `SendAsync`,
      until `CancelWait` / `maxFires` / the deadline. `maxFires:1` is the canonical
      "notify me once, then keep working" (background single-shot) case — it must be
      expressible **without** a separate `recurring` flag.
- [ ] **Single-resolution latch:** a fire and a timeout (or a double-fire) racing
      on the same block wait produce **exactly one** resolution, deterministically,
      with no unobserved background exception (test fires + times out within 1 ms).
      The latch lives in the source; it does not reimplement
      `ResolveToolCallAsync`'s idempotency.
- [ ] **Fire ordering vs deferrals:** a notify fire never starts a run while an
      unresolved deferral exists (it is queued and flushed after the parked deferral
      resolves), so it cannot trip the "tool call still deferred" guard
      (`RunFailed`) — test: notify timer + block `Wait`, fire the timer while the
      block wait is unresolved → no `RunFailed`.
- [ ] Multiple block `Wait`s armed in one generation follow a **documented** policy
      (v1 = all-of: the run resumes when all resolve; the model is guided to arm one
      wait at a time). A single wait's `timeout` already provides **event-or-deadline**
      (fire OR `timed_out`), which covers the common "any-of" need; racing two real
      events ("any-of / first-of") is a noted follow-up.

### Lifetime, durability & cleanup
- [ ] **Mandatory deadline** on every wait via `timeout`, accepting a duration
      (`"10m"`) **or** an absolute time (`"2026-07-31"`). The ceiling differs by mode
      so a long-lived schedule is not killed by the park cap: for `block` it is the
      park ceiling (short host default + a hard max the LLM cannot exceed); for
      `notify` it is the total armed lifetime (a larger, host-configurable cap so a
      daily schedule can live for weeks), additionally bounded by `maxFires`. Reaching
      it resolves a block wait with a distinct `timed_out` payload (an outcome, **not**
      an exception) and terminates a notify wait.
- [ ] **Restart recovery pass:** on `OnHistoryRestoredAsync`, every restored
      deferred `Wait` whose live watcher is gone is either re-armed (where the kind
      allows, e.g. timer) or resolved with a terminal `trigger_lost_on_restart`
      payload by its original deadline — **no parked conversation is left
      permanently wedged** (test: kill + restart with a block wait parked → resolves
      by the deadline). For block waits the arming record *is* the existing deferred
      placeholder (recovery is mostly free); only notify/live-watcher waits need an
      extra arming record (see *Shared substrate*).
- [ ] `CancelWait({ id | label | kind })` is **total + idempotent**: unknown/terminal
      target → a `Resolved` status, never an exception; an active block wait → cancels
      the watcher and resolves the parked deferral with a `cancelled` payload (via the
      latch); an active notify wait → stops and reports `fires_delivered`; a
      `label`/`kind` selector cancels all matching waits. `ListWaits` reports a
      well-defined status per wait and echoes each wait's `label`.
- [ ] **Bounded concurrency:** a `MaxConcurrentWaits` cap **reusing
      `SubAgentManager`'s `SemaphoreSlim` gate and its retryable over-concurrency
      error shape** (including its acquire/release-exactly-once discipline, which a
      repeatedly-firing notify watcher must honor).
- [ ] **Disciplined teardown** (the `SubAgentManager.DisposeAsync` per-resource
      isolated-cancel pattern, reused not re-derived): after `CancelWait`, expiry,
      or agent dispose, **zero** watchers, timers, or child processes remain
      (assertable test).

### Flood / cost control
- [ ] **Coalescing + rate limit:** a notify source firing faster than a fixed
      floor is coalesced into a single injection carrying a batch summary/count
      (not one turn per event); exceeding `maxFires` auto-cancels the wait and
      injects a terminal notice — test: a source firing 100×/s for 5 s injects a
      bounded number of turns, not 500.

### Safety: payload, paths, commands
- [ ] **Size cap:** no single fire injects more than a configurable byte budget
      into context; excess is truncated with an explicit `[truncated N bytes]`
      marker; the Log source also caps per-line length and lines-per-batch. The
      full payload is never logged at info level.
- [ ] **Path confinement (Log source):** the watched path is canonicalized and
      confined to a host-supplied allowed-roots set (a plain option; default = the
      agent working directory), traversal/symlink-escape is rejected, and access is
      **denied by default** when no roots are configured.
- [ ] **Process source:** runs through the existing `IProcessLauncher`
      (generalized to accept a non-CLI-agent "run this command" shape — **not** a
      second launcher abstraction), so a host can still sandbox/deny it;
      commands/working dir are allowlist-gated; the mandatory timeout calls
      `IProcessHandle.Kill(entireProcessTree: true)` (test: a long-sleep command is
      reaped at timeout with no surviving children).
- [ ] **Predicate gates firing (not just annotation):** a source predicate (e.g.
      Process exit-code / stdout-regex) decides **whether** a fire occurs, so "wake me
      only if the build fails" is directly expressible. For a `block` wait an exit
      that does not match still unblocks the run with a `condition_not_met` outcome
      (the loop never wedges); for a `notify` wait a non-match injects nothing.
- [ ] **Regex safety:** every LLM-authored regex (Process stdout predicate, Log
      line match) is matched with a timeout; predicates are pure data comparisons
      (exit code, `IsMatch`), never executable.
- [ ] **Optional redaction:** a host may supply a redaction delegate applied to a
      payload before it enters history/persistence (default off). [Recommended, not
      a blocker — the size cap + not-logging-at-info are the mandatory floors; this
      hook exists because Process stdout / arbitrary file contents are real v1
      emitters of host-sensitive data.]

### Observability
- [ ] Structured Serilog events `trigger.armed/fired/expired/cancelled` with
      `{ triggerId, kind, sourceRef-hash, runId }`; injected `<trigger>` messages
      carry the wait's `label` and metadata that distinguishes an agent-woken turn
      from a real user turn (so a fire arriving long after arming — even after context
      compaction — is self-describing) — **defined once, shared with the existing
      `<sub-agent>` (and future `<workflow>`) injection wrapper**, not a per-feature
      format.

### Built-in sources (v1)
- [ ] `TimerTriggerSource`: `delay` | `interval` | `cron`; interval is fixed-rate
      from arm time; a minimum interval floor is enforced. `delay` is typically a
      `block` wait; `interval`/`cron` run as `mode:"notify"` bounded by
      `timeout`/`maxFires`. (Cron is evaluated on a documented timezone basis; full
      timezone/DST rigor is a follow-up.)
- [ ] `ProcessTriggerSource`: runs a command, fires on exit with the exit code +
      (capped) stdout/stderr and an optional exit-code/stdout-regex predicate; obeys
      the launcher/allowlist/tree-kill/size-cap/regex-timeout rules above.
- [ ] `FileTailTriggerSource`: fires per new line matching a regex, batching lines
      within a small window; documents its tail limitations and logs a warning on
      watcher overflow (full log-rotation/truncation handling is a follow-up).
- [ ] Sub-agent completion is selectable as a `kind` through the same `Wait`
      surface by **consuming the existing background-sub-agent completion latch with
      `NotifyParentOnCompletion=false`** (or by replacing the relay) so a completion
      is injected **once, not twice** (no double-relay).

### Tests
- [ ] Unit: latch race, total `CancelWait`, timeout→`timed_out`, restart recovery,
      coalescing cap, size-cap truncation, path-confinement rejection, process
      tree-kill, regex-timeout.
- [ ] Integration: an LLM arms a `block` `Wait` (timer), parks, and resumes on
      fire; arms a `notify` log-tail `Wait`, receives coalesced injected fires, and
      `CancelWait`s it; a `block` wait parked across a simulated restart resolves via
      the recovery pass.

## Shared substrate (dedupe with `LmWorkflow`)

This feature and the sibling `LmWorkflow` proposal independently add the same
background-work substrate. To avoid triplication they should share, not re-derive:

- [ ] **One background-work base** (concurrency gate + isolated teardown +
      completion latch) factored from `SubAgentManager`, inherited by
      `TriggerRegistry` and `LmWorkflow`'s instance tracker.
- [ ] **One restart-reconciliation seam** on `OnHistoryRestoredAsync` that both
      features register reconcilers into, over the single `IConversationStore` —
      rather than each shipping its own durable side-store + restart pass.
- [ ] **One injected-message convention** (`<trigger>` / `<sub-agent>` /
      `<workflow>`) with shared "agent-woken, not user" metadata, defined once.

## Non-goals / follow-ups (explicitly out of v1)
- **External-signal / webhook source** (`RemoteTrigger`-style) — first deferred
  source to add via the seam.
- **Durable re-arming of live watchers across restart** (v1 is in-process
  best-effort; restart only guarantees resolve-or-cancel, not watcher survival).
- **"Any-of / first-of" of *two real events*** (v1 = all-of within a generation).
  Note: a single wait's `timeout` already provides **event-or-deadline**, covering
  the most common "any-of" need; only racing two concrete events is deferred.
- **Full cron timezone/DST rigor** and an explicit DST-boundary test suite.
- **Full log rotation/truncation handling** and a `degraded` event for
  FileSystemWatcher overflow / network-share reliability matrix.
- **Fresh-session spawn topology** (a fired trigger booting a brand-new session/
  process, à la cloud Routines) — v1 is same-session resume only.
- **Outbound human push notification** on wake (separate concern; the host already
  has notification channels).

gautam-achieveai:
## Summary

Adds a generic **Wait / CancelWait / ListWaits** primitive that lets an LLM, mid multi-turn conversation, **park a run on a background or scheduled event and resume with a self-describing payload** — layered on the existing [deferred tool-call](docs/features/deferred-tool-execution/motivation.md) park/wake core. Implements the approved **plan v3** on #107 (block-mode only; one one-shot `timer` source).

```jsonc
Wait({ kind, args?, timeout, label? })   // parks the run; resumes with {status: fired|timed_out|cancelled|failed, ...}
CancelWait({ id?, label?, kind? })       // total + idempotent
ListWaits()                              // armed waits + registered kinds
```

Adding a new wake source is "implement `ITriggerSource` + register it" — no change to the agent loop or the tool surface.

## What's included

- **`ITriggerSource` seam** — a source only observes its event and reports a raw fact to a runtime-owned `ITriggerEventSink`; `IArmedTrigger` is a cancellable handle. Lifecycle policy lives entirely in the runtime.
- **`TriggerRuntime`** — owns the whole lifecycle: single-resolution latch (`Pending → Fired | TimedOut | Cancelled | Failed`), ceiling timeout, cancellation with source cleanup, bounded concurrency, payload caps, and restart reconciliation.
- **`WaitToolProvider`** — the `Wait`/`CancelWait`/`ListWaits` tools; block waits park via the loop's **existing public** `ResolveToolCallAsync` (no new loop API).
- **Built-in one-shot `timer` source** — `delay` | `deadline` | default-to-`timeout`; restartable.
- **Loop wiring** — `TriggerOptions` ctor param, `WaitToolProvider` registration, and an `OnHistoryRestoredAsync` reconcile pass so a restored block wait is **never left hanging**.
- **Docs** — `docs/features/wait-trigger-primitive/README.md`.
- **25 tests** — latch races, cancel cleanup, timeout, restart reconcile (re-arm / expired / non-restorable / rejected re-arm), limits, unknown-kind / invalid-timeout rejection, payload cap, register-window retry + fire-path token robustness, and end-to-end park→fire→resume through the loop for both the built-in timer and a host-registered external source.

## Key decisions

1. **Narrowed to the minimum proving slice** (per the plan-v3 review): block-mode only, one one-shot `timer` source, a dummy/external-source test proving the seam. File-tail, process, cron/interval, sub-agent-completion, notify-mode, and any shared `BackgroundWorkBase` are deferred to follow-ups.
2. **Runtime owns the lifecycle; sources are dumb observers.** A single-resolution latch guarantees exactly one terminal transition even when a fire races a timeout races a cancel. This resolves the review's CRITICAL race-semantics finding without per-source policy.
3. **One wait identity, one persistence model.** A block wait's identity *is* its deferred `tool_call_id`; the existing `IsDeferred` placeholder (+ `DeferredAt`) is the arming record — no new persisted type. On restart, a restorable source re-arms for its remaining delay; a non-restorable one resolves `trigger_lost_on_restart`.
4. **No new public loop API.** Block delivery reuses the existing public `ResolveToolCallAsync` and internal auto-resume; the v2-proposed `HasUnresolvedDeferrals`/`DeferralResolved` were dropped. (Notify mode, when it lands, will use the loop's existing internal `EnqueueRawAsync` sentinel gate.)
5. **Deterministic fire-vs-timeout** via a small ceiling grace so a timer's `fired` wins over `timed_out` when they coincide; terminal delivery uses an uncancellable token so disposing the source can't drop the result.

## Testing

- `dotnet test tests/LmMultiTurn.Tests` → **309 passing** (28 new + 281 existing), 0 failures.
- Full solution builds clean (0 errors).

Fixes #107.

### DIFF
--- a/src/LmMultiTurn/MultiTurnAgentLoop.cs
+++ b/src/LmMultiTurn/MultiTurnAgentLoop.cs
@@ -9,6 +9,7 @@ using AchieveAi.LmDotnetTools.LmMultiTurn.Messages;
 using AchieveAi.LmDotnetTools.LmMultiTurn.Middleware;
 using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence;
 using AchieveAi.LmDotnetTools.LmMultiTurn.SubAgents;
+using AchieveAi.LmDotnetTools.LmMultiTurn.Triggers;
 using Microsoft.Extensions.Logging;
 
 namespace AchieveAi.LmDotnetTools.LmMultiTurn;
@@ -36,6 +37,9 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
     private readonly IStreamingAgent _agent;
     private readonly IDictionary<string, ToolHandler> _toolHandlers;
     private readonly SubAgentManager? _subAgentManager;
+
+    // Owns the Wait/trigger lifecycle when trigger options are supplied. Null otherwise.
+    private readonly TriggerRuntime? _triggerRuntime;
 
     // Deferred tool tracking. Keyed by ToolCallId. Concurrent because resolutions arrive on
     // arbitrary threads (UI callbacks, webhook handlers).
@@ -76,6 +80,12 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
     ///     (MessageTransformation, MessageUpdateJoiner) their own category loggers so their
     ///     ordering/de-dup decisions are visible in structured logs. When null they stay silent.
     /// </param>
+    /// <param name="triggerOptions">
+    ///     Optional Wait/trigger configuration. When provided, the loop enables the
+    ///     <c>Wait</c>/<c>CancelWait</c>/<c>ListWaits</c> tools backed by a <see cref="TriggerRuntime"/>
+    ///     (with the built-in one-shot <c>timer</c> source plus any host registrations). When null,
+    ///     no wait tools are exposed.
+    /// </param>
     public MultiTurnAgentLoop(
         IStreamingAgent providerAgent,
         FunctionRegistry functionRegistry,
@@ -89,7 +99,8 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
         ILogger<MultiTurnAgentLoop>? logger = null,
         SubAgentOptions? subAgentOptions = null,
         MutableSubAgentTemplateSource? subAgentTemplateSource = null,
-        ILoggerFactory? loggerFactory = null)
+        ILoggerFactory? loggerFactory = null,
+        TriggerOptions? triggerOptions = null)
         : base(threadId, systemPrompt, defaultOptions, maxTurnsPerRun, inputChannelCapacity, outputChannelCapacity, store, logger)
     {
         ArgumentNullException.ThrowIfNull(providerAgent);
@@ -129,6 +140,26 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
             _ = functionRegistry.AddProvider(toolProvider);
         }
 
+        // When trigger options are supplied, stand up the Wait/trigger runtime and register the
+        // Wait/CancelWait/ListWaits tools. Registered AFTER the sub-agent snapshot so sub-agents
+        // don't inherit the wait surface. The runtime resolves parked block waits through the
+        // loop's existing public ResolveToolCallAsync — no new loop API is exposed.
+        if (triggerOptions != null)
+        {
+            _triggerRuntime = new TriggerRuntime(
+                triggerOptions,
+                resolve: (toolCallId, result, isError, ct) =>
+                    ResolveToolCallAsync(toolCallId, result, isError, contentBlocks: null, ct),
+                logger: logger);
+            _triggerRuntime.RegisterBuiltIns();
+            foreach (var registration in triggerOptions.AdditionalRegistrations)
+            {
+                _triggerRuntime.Register(registration);
+            }
+
+            _ = functionRegistry.AddProvider(new WaitToolProvider(_triggerRuntime));
+        }
+
         // Build tool call components from registry
         var (toolCallMiddleware, finalHandlers) = functionRegistry.BuildToolCallComponents(name: "MultiTurnAgentTools");
         _toolHandlers = finalHandlers;
@@ -155,6 +186,11 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
         if (_subAgentManager != null)
         {
             await _subAgentManager.DisposeAsync();
+        }
+
+        if (_triggerRuntime != null)
+        {
+            await _triggerRuntime.DisposeAsync();
         }
     }
 
@@ -818,14 +854,14 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
     }
 
     /// <inheritdoc />
-    protected override Task OnHistoryRestoredAsync(IReadOnlyList<IMessage> messages, CancellationToken ct)
+    protected override async Task OnHistoryRestoredAsync(IReadOnlyList<IMessage> messages, CancellationToken ct)
     {
         // Rebuild the deferred registry from persisted history. Each ToolCallResultMessage
         // with IsDeferred=true gets re-registered so GetDeferredToolCallsAsync surfaces it
         // and ResolveToolCallAsync can complete it after restart.
         if (messages == null || messages.Count == 0)
         {
-            return Task.CompletedTask;
+            return;
         }
 
         // Index ToolCallMessages by ToolCallId so we can recover function name/args for
@@ -880,7 +916,22 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
                 _deferred.Count);
         }
 
-        return Task.CompletedTask;
+        // Reconcile restored block waits so no parked Wait is left hanging: a restorable source
+        // (e.g. timer) re-arms for its remaining delay; a non-restorable one resolves as
+        // trigger_lost_on_restart. Runs after the deferred set is rebuilt so it sees every entry.
+        if (_triggerRuntime != null)
+        {
+            List<RestoredWait> restoredList =
+            [
+                .. _deferred.Values
+                    .Where(e => e.FunctionName == WaitToolProvider.WaitToolName)
+                    .Select(e => new RestoredWait(e.ToolCallId, e.FunctionArgs, e.DeferredAtUnixMs)),
+            ];
+            if (restoredList.Count > 0)
+            {
+                await _triggerRuntime.ReconcileRestoredAsync(restoredList, ct);
+            }
+        }
     }
 
     private bool HasUnresolvedDeferralsForGeneration(string generationId)
--- a/src/LmMultiTurn/MultiTurnAgentLoop.cs
+++ b/src/LmMultiTurn/MultiTurnAgentLoop.cs
@@ -222,11 +222,6 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
                 var realInputs = batch.Where(b => b.Resume == null).ToList();
                 var resumeSentinels = batch.Where(b => b.Resume != null).ToList();
 
-                if (realInputs.Count == 0 && resumeSentinels.Count == 0)
-                {
-                    continue;
-                }
-
                 if (resumeSentinels.Count > 0)
                 {
                     // Clear the scheduled flag now that we're consuming the sentinel; future
@@ -919,17 +914,31 @@ public sealed class MultiTurnAgentLoop : MultiTurnAgentBase
         // Reconcile restored block waits so no parked Wait is left hanging: a restorable source
         // (e.g. timer) re-arms for its remaining delay; a non-restorable one resolves as
         // trigger_lost_on_restart. Runs after the deferred set is rebuilt so it sees every entry.
+        List<DeferredEntry> restoredWaitEntries =
+            [.. _deferred.Values.Where(e => e.FunctionName == WaitToolProvider.WaitToolName)];
+
         if (_triggerRuntime != null)
         {
-            List<RestoredWait> restoredList =
-            [
-                .. _deferred.Values
-                    .Where(e => e.FunctionName == WaitToolProvider.WaitToolName)
-                    .Select(e => new RestoredWait(e.ToolCallId, e.FunctionArgs, e.DeferredAtUnixMs)),
-            ];
-            if (restoredList.Count > 0)
+            if (restoredWaitEntries.Count > 0)
             {
+                List<RestoredWait> restoredList =
+                    [.. restoredWaitEntries.Select(e => new RestoredWait(e.ToolCallId, e.FunctionArgs, e.DeferredAtUnixMs))];
                 await _triggerRuntime.ReconcileRestoredAsync(restoredList, ct);
+            }
+        }
+        else if (restoredWaitEntries.Count > 0)
+        {
+            // Triggers are disabled in this host (or were rolled back after these waits were
+            // persisted) — there is no runtime left to re-arm or fail them. Resolve each restored
+            // Wait with a terminal failure now rather than leaving the run parked forever.
+            foreach (var entry in restoredWaitEntries)
+            {
+                await ResolveToolCallAsync(
+                    entry.ToolCallId,
+                    JsonSerializer.Serialize(new { status = "failed", reason = "trigger_disabled", waitId = entry.ToolCallId }),
+                    isError: false,
+                    contentBlocks: null,
+                    ct);
             }
         }
     }
--- a/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
+++ b/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
@@ -115,6 +115,11 @@ public sealed class TimerTriggerSource : ITriggerSource
 
         private static async Task RunAsync(DateTimeOffset fireAt, ITriggerEventSink sink, CancellationToken ct)
         {
+            // Yield first so the fire is always asynchronous — never synchronous within ArmAsync,
+            // even when the instant has already elapsed (Task.Delay(0) would otherwise continue
+            // synchronously and fire before the runtime has finished registering the wait).
+            await Task.Yield();
+
             var remaining = fireAt - DateTimeOffset.UtcNow;
             if (remaining < TimeSpan.Zero)
             {
--- a/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
+++ b/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
@@ -45,19 +45,39 @@ public sealed class TimerTriggerSource : ITriggerSource
     /// Computes the absolute instant the timer should fire, relative to the arm time so restart
     /// re-arming yields the same wall-clock target.
     /// </summary>
+    /// <exception cref="ArgumentException">
+    /// The args are not a JSON object, supply both <c>delay</c> and <c>deadline</c>, or supply a
+    /// <c>delay</c>/<c>deadline</c> that does not parse. Invalid input is rejected rather than
+    /// silently falling back to the wait's ceiling, so the caller sees a structured rejection it
+    /// can correct instead of an unexpectedly-late fire.
+    /// </exception>
     private static DateTimeOffset ResolveFireInstant(TriggerArmRequest request)
     {
         var (delay, deadline) = ParseArgs(request.ArgsJson);
+        var hasDelay = !string.IsNullOrWhiteSpace(delay);
+        var hasDeadline = !string.IsNullOrWhiteSpace(deadline);
 
-        if (!string.IsNullOrWhiteSpace(deadline)
-            && TriggerDurations.TryResolveInstant(deadline, request.ArmedAt, out var absolute, out _))
+        if (hasDelay && hasDeadline)
         {
+            throw new ArgumentException("timer args must supply at most one of 'delay' or 'deadline', not both.");
+        }
+
+        if (hasDeadline)
+        {
+            if (!TriggerDurations.TryResolveInstant(deadline, request.ArmedAt, out var absolute, out var error))
+            {
+                throw new ArgumentException($"timer 'deadline' is invalid: {error}");
+            }
             return absolute;
         }
 
-        if (!string.IsNullOrWhiteSpace(delay)
-            && TriggerDurations.TryParseDuration(delay, out var relative))
+        if (hasDelay)
         {
+            if (!TriggerDurations.TryParseDuration(delay, out var relative))
+            {
+                throw new ArgumentException(
+                    $"timer 'delay' is invalid: '{delay}' is not a duration (e.g. \"10m\", \"30s\").");
+            }
             return request.ArmedAt + relative;
         }
 
@@ -72,27 +92,41 @@ public sealed class TimerTriggerSource : ITriggerSource
             return (null, null);
         }
 
+        JsonDocument doc;
         try
         {
-            using var doc = JsonDocument.Parse(argsJson);
+            doc = JsonDocument.Parse(argsJson);
+        }
+        catch (JsonException ex)
+        {
+            throw new ArgumentException($"timer args is not valid JSON: {ex.Message}", ex);
+        }
+
+        using (doc)
+        {
             var root = doc.RootElement;
             if (root.ValueKind != JsonValueKind.Object)
             {
-                return (null, null);
+                throw new ArgumentException("timer args must be a JSON object.");
             }
 
-            var delay = root.TryGetProperty("delay", out var d) && d.ValueKind == JsonValueKind.String
-                ? d.GetString()
-                : null;
-            var deadline = root.TryGetProperty("deadline", out var dl) && dl.ValueKind == JsonValueKind.String
-                ? dl.GetString()
-                : null;
-            return (delay, deadline);
+            return (GetOptionalString(root, "delay"), GetOptionalString(root, "deadline"));
         }
-        catch (JsonException)
+    }
+
+    private static string? GetOptionalString(JsonElement root, string name)
+    {
+        if (!root.TryGetProperty(name, out var el) || el.ValueKind == JsonValueKind.Null)
         {
-            return (null, null);
+            return null;
         }
+
+        if (el.ValueKind != JsonValueKind.String)
+        {
+            throw new ArgumentException($"timer '{name}' must be a string.");
+        }
+
+        return el.GetString();
     }
 
     /// <summary>
--- a/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
+++ b/src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
@@ -1,0 +1,166 @@
+using System.Text.Json;
+
+namespace AchieveAi.LmDotnetTools.LmMultiTurn.Triggers.Sources;
+
+/// <summary>
+/// Built-in one-shot timer source. Fires once when its configured instant is reached, then never
+/// again. The awaited instant comes from <c>args.deadline</c> (absolute) or <c>args.delay</c>
+/// (relative to when the wait was armed); if neither is supplied it defaults to the wait's own
+/// ceiling deadline (i.e. <c>Wait({kind:"timer", timeout:"10m"})</c> means "wake me in 10m").
+/// </summary>
+/// <remarks>
+/// Restartable: the fire instant is a pure function of the persisted arm time and args, so a
+/// restored wait re-arms for its remaining delay (or fires immediately if already elapsed). This
+/// source holds no per-wait state — each <see cref="ArmAsync"/> returns an independent handle.
+/// </remarks>
+public sealed class TimerTriggerSource : ITriggerSource
+{
+    /// <summary>The registered kind token.</summary>
+    public const string KindName = "timer";
+
+    /// <summary>Capabilities: block + restore. Notify is a future follow-up.</summary>
+    public static TriggerCapabilities Capabilities { get; } =
+        new(SupportsBlock: true, SupportsNotify: false, SupportsRestore: true);
+
+    /// <summary>Human-readable args hint for the tool contract.</summary>
+    public const string ArgsSchemaText =
+        "{ delay?: \"<duration e.g. 10m|30s|2h>\", deadline?: \"<ISO-8601 time>\" } "
+        + "— supply at most one; if omitted, fires at the wait's timeout.";
+
+    /// <inheritdoc />
+    public ValueTask<IArmedTrigger> ArmAsync(
+        TriggerArmRequest request,
+        ITriggerEventSink eventSink,
+        CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        ArgumentNullException.ThrowIfNull(eventSink);
+
+        var fireAt = ResolveFireInstant(request);
+        var handle = new TimerArmedTrigger(request.WaitId, fireAt, eventSink);
+        return ValueTask.FromResult<IArmedTrigger>(handle);
+    }
+
+    /// <summary>
+    /// Computes the absolute instant the timer should fire, relative to the arm time so restart
+    /// re-arming yields the same wall-clock target.
+    /// </summary>
+    private static DateTimeOffset ResolveFireInstant(TriggerArmRequest request)
+    {
+        var (delay, deadline) = ParseArgs(request.ArgsJson);
+
+        if (!string.IsNullOrWhiteSpace(deadline)
+            && TriggerDurations.TryResolveInstant(deadline, request.ArmedAt, out var absolute, out _))
+        {
+            return absolute;
+        }
+
+        if (!string.IsNullOrWhiteSpace(delay)
+            && TriggerDurations.TryParseDuration(delay, out var relative))
+        {
+            return request.ArmedAt + relative;
+        }
+
+        // No explicit fire time — fire at the wait's ceiling ("sleep until timeout").
+        return request.Deadline;
+    }
+
+    private static (string? Delay, string? Deadline) ParseArgs(string argsJson)
+    {
+        if (string.IsNullOrWhiteSpace(argsJson))
+        {
+            return (null, null);
+        }
+
+        try
+        {
+            using var doc = JsonDocument.Parse(argsJson);
+            var root = doc.RootElement;
+            if (root.ValueKind != JsonValueKind.Object)
+            {
+                return (null, null);
+            }
+
+            var delay = root.TryGetProperty("delay", out var d) && d.ValueKind == JsonValueKind.String
+                ? d.GetString()
+                : null;
+            var deadline = root.TryGetProperty("deadline", out var dl) && dl.ValueKind == JsonValueKind.String
+                ? dl.GetString()
+                : null;
+            return (delay, deadline);
+        }
+        catch (JsonException)
+        {
+            return (null, null);
+        }
+    }
+
+    /// <summary>
+    /// Per-arm handle. Runs a single <see cref="Task.Delay(TimeSpan, CancellationToken)"/> for the
+    /// remaining delay and reports one fire. Disposal cancels the delay so no fire occurs afterward.
+    /// </summary>
+    private sealed class TimerArmedTrigger : IArmedTrigger
+    {
+        private readonly CancellationTokenSource _cts = new();
+        private readonly Task _timerTask;
+        private int _disposed;
+
+        public TimerArmedTrigger(string waitId, DateTimeOffset fireAt, ITriggerEventSink sink)
+        {
+            WaitId = waitId;
+            _timerTask = RunAsync(fireAt, sink, _cts.Token);
+        }
+
+        public string WaitId { get; }
+
+        private static async Task RunAsync(DateTimeOffset fireAt, ITriggerEventSink sink, CancellationToken ct)
+        {
+            var remaining = fireAt - DateTimeOffset.UtcNow;
+            if (remaining < TimeSpan.Zero)
+            {
+                remaining = TimeSpan.Zero;
+            }
+
+            try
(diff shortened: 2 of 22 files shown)


src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
```diff
@@ -0,0 +1,171 @@
...
+
+    /// <summary>
+    /// Computes the absolute instant the timer should fire, relative to the arm time so restart
+    /// re-arming yields the same wall-clock target.
+    /// </summary>
+    private static DateTimeOffset ResolveFireInstant(TriggerArmRequest request)
+    {
+        var (delay, deadline) = ParseArgs(request.ArgsJson);
+
+        if (!string.IsNullOrWhiteSpace(deadline)
```
gautam-achieveai: [Claude's bot] [BLOCKER] **HIGH** (Correctness / API contract)

`ResolveFireInstant` tries a valid `deadline`, then a valid `delay`, then falls through to `request.Deadline`. That means malformed input such as `{ "delay": "soon" }` or `{ "deadline": "not-a-date" }` — and contradictory input with both fields supplied — is accepted and armed as if no timer args were supplied. The tool contract says to supply at most one timer argument, so invalid model/tool calls should return a structured rejection the model can correct instead of parking the run until the safety ceiling.

**Suggestion:** Validate timer args before arming: reject JSON parse errors/non-object args, malformed `delay`/`deadline`, and both `delay` + `deadline` unless you intentionally document a precedence rule. Return a structured rejected/error result and add tests for invalid delay, invalid deadline, both fields, and absolute deadline.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -880,7 +916,22 @@ protected override Task OnHistoryRestoredAsync(IReadOnlyList<IMessage> messages,
                 _deferred.Count);
         }
 
-        return Task.CompletedTask;
+        // Reconcile restored block waits so no parked Wait is left hanging: a restorable source
+        // (e.g. timer) re-arms for its remaining delay; a non-restorable one resolves as
+        // trigger_lost_on_restart. Runs after the deferred set is rebuilt so it sees every entry.
+        if (_triggerRuntime != null)
```
gautam-achieveai, line 920: [Claude's bot] [BLOCKER] **HIGH** (Rollback / lifecycle)

`triggerOptions` is a safe opt-in gate for new sessions, but if a host disables triggers or rolls back config after conversations have parked on `Wait`, `OnHistoryRestoredAsync` still restores those `Wait` placeholders into `_deferred` and then skips reconciliation because `_triggerRuntime` is null. Those runs can remain parked indefinitely with no runtime left to re-arm or fail them.

**Suggestion:** After rebuilding `_deferred`, always identify restored `Wait` deferrals. If `_triggerRuntime` is null, resolve them with a terminal failed payload such as `{ "status": "failed", "reason": "trigger_disabled", "waitId": ... }`, or introduce explicit enabled/disabled drain behavior.

src/LmMultiTurn/Triggers/WaitToolProvider.cs
```diff
@@ -0,0 +1,248 @@
...
+            // Park the run; the runtime resolves this tool call when the wait terminates.
+            return new ToolHandlerResult.Deferred();
+        }
+
+        return ToolHandlerResult.FromError(
+            Reject(result.Reason ?? "rejected", result.Message ?? "Wait was rejected."),
+            result.Reason);
+    }
+
+    private async Task<ToolHandlerResult> HandleCancelWaitAsync(
```
gautam-achieveai, line 192: [Claude's bot] **MEDIUM** (Test coverage)

`CancelWait` and `ListWaits` are new public tool-surface functions, but the tests only cover `TriggerRuntime.CancelWaitsAsync` / `runtime.ListWaits()` directly. I was unable to find a provider/loop-level test invoking `FunctionName = "CancelWait"` or `"ListWaits"`, so JSON selector parsing, handler registration, and serialized return shapes (`{ status: "resolved", cancelled }`, `waits`, `registeredKinds`) are unverified.

**Suggestion:** Add provider/loop-level behavioral tests that arm a wait, invoke `CancelWait` and assert the cancelled count plus cancelled payload/no pending deferrals; invoke `ListWaits` and assert active wait metadata plus registered kinds.

src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
gautam-achieveai: [Claude's bot] [BLOCKER] **HIGH** (Correctness / API contract)

`ResolveFireInstant` tries a valid `deadline`, then a valid `delay`, then falls through to `request.Deadline`. That means malformed input such as `{ "delay": "soon" }` or `{ "deadline": "not-a-date" }` — and contradictory input with both fields supplied — is accepted and armed as if no timer args were supplied. The tool contract says to supply at most one timer argument, so invalid model/tool calls should return a structured rejection the model can correct instead of parking the run until the safety ceiling.

**Suggestion:** Validate timer args before arming: reject JSON parse errors/non-object args, malformed `delay`/`deadline`, and both `delay` + `deadline` unless you intentionally document a precedence rule. Return a structured rejected/error result and add tests for invalid delay, invalid deadline, both fields, and absolute deadline.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai, line 920: [Claude's bot] [BLOCKER] **HIGH** (Rollback / lifecycle)

`triggerOptions` is a safe opt-in gate for new sessions, but if a host disables triggers or rolls back config after conversations have parked on `Wait`, `OnHistoryRestoredAsync` still restores those `Wait` placeholders into `_deferred` and then skips reconciliation because `_triggerRuntime` is null. Those runs can remain parked indefinitely with no runtime left to re-arm or fail them.

**Suggestion:** After rebuilding `_deferred`, always identify restored `Wait` deferrals. If `_triggerRuntime` is null, resolve them with a terminal failed payload such as `{ "status": "failed", "reason": "trigger_disabled", "waitId": ... }`, or introduce explicit enabled/disabled drain behavior.

src/LmMultiTurn/Triggers/WaitToolProvider.cs
gautam-achieveai, line 192: [Claude's bot] **MEDIUM** (Test coverage)

`CancelWait` and `ListWaits` are new public tool-surface functions, but the tests only cover `TriggerRuntime.CancelWaitsAsync` / `runtime.ListWaits()` directly. I was unable to find a provider/loop-level test invoking `FunctionName = "CancelWait"` or `"ListWaits"`, so JSON selector parsing, handler registration, and serialized return shapes (`{ status: "resolved", cancelled }`, `waits`, `registeredKinds`) are unverified.

**Suggestion:** Add provider/loop-level behavioral tests that arm a wait, invoke `CancelWait` and assert the cancelled count plus cancelled payload/no pending deferrals; invoke `ListWaits` and assert active wait metadata plus registered kinds.

gautam-achieveai:
[Claude's bot] **LOW** (Code simplification) — `src/LmMultiTurn/MultiTurnAgentLoop.cs:225`

This guard is unreachable: `batch.Count == 0` was already handled above, and `realInputs` / `resumeSentinels` are complementary partitions of `batch`, so both counts cannot be zero here.

**Suggestion:** Remove the redundant block.

gautam-achieveai:
[Claude's bot] # PR Review: feat(triggers): generic Wait/CancelWait/ListWaits primitive (#107)

## Verdict: REQUEST CHANGES

I found two blocking lifecycle/API-contract issues in the new Wait trigger primitive.

### Blockers
- Timer `args` validation silently accepts malformed/contradictory args and arms a wait instead of rejecting the tool call.
- Restored `Wait` deferrals can be stranded indefinitely if triggers are later disabled/rolled back (`triggerOptions` omitted) after waits were persisted.

### Non-blocking
- Add provider/loop-level coverage for `CancelWait` and `ListWaits` public tool behavior.
- Remove a small unreachable guard in the loop input partitioning code.

CI checks on the PR are green, but these issues affect the new public tool contract and recovery/rollback safety, so I’m requesting changes before merge.

gautam-achieveai:
Review addressed in 7d4b86e — thanks for the thorough pass.

**BLOCKER — timer args silently fell back to the wait's ceiling on malformed/contradictory input.** `TimerTriggerSource.ArmAsync` (via `ResolveFireInstant`/`ParseArgs`/`GetOptionalString`) now throws `ArgumentException` for: non-JSON args, non-object args, wrong-typed `delay`/`deadline` values, an unparseable duration/instant string, or both `delay` and `deadline` supplied together. `TriggerRuntime.ArmCoreAsync`'s catch is split so `ArgumentException` maps to a structured `WaitArmResult.Reject("invalid_args", ...)` (the existing `arm_failed` catch stays for everything else). `ReconcileRestoredAsync` already resolves any rejected re-arm with a failed payload, so restart-safety needed no additional change. Added a 6-case `[Theory]` regression test in `TimerTriggerSourceTests`.

**BLOCKER — a restored parked `Wait` hung forever when the host has no `TriggerRuntime`.** If triggers are disabled (or were rolled back after waits were persisted), a restored `Wait` deferred entry now resolves via `ResolveToolCallAsync` with `{status:"failed", reason:"trigger_disabled"}` in `MultiTurnAgentLoop.OnHistoryRestoredAsync`, instead of being left parked with nothing to ever wake it.

**MEDIUM — missing `CancelWait`/`ListWaits` test coverage.** Added `WaitToolProviderTests.cs`, exercising a real `TriggerRuntime` + `WaitToolProvider` by invoking the `FunctionDescriptor` handlers directly (the full-loop send-guard rejects a new turn while a `Wait` is parked, which made a loop-driven second-turn test awkward). Covers: arm → list → cancel-by-id → list-empty round trip, cancel-by-kind cancelling multiple matching waits, and a no-selector `CancelWait` being a safe no-op.

**LOW — unreachable guard.** Removed the dead `if (realInputs.Count == 0 && resumeSentinels.Count == 0) continue;` in `RunLoopAsync`.

Verification: full solution build clean, `LmMultiTurn.Tests` 318/318 green, `dotnet format whitespace --verify-no-changes` clean.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -880,7 +911,36 @@ protected override Task OnHistoryRestoredAsync(IReadOnlyList<IMessage> messages,
...
+        {
+            if (restoredWaitEntries.Count > 0)
+            {
+                List<RestoredWait> restoredList =
+                    [.. restoredWaitEntries.Select(e => new RestoredWait(e.ToolCallId, e.FunctionArgs, e.DeferredAtUnixMs))];
+                await _triggerRuntime.ReconcileRestoredAsync(restoredList, ct);
+            }
+        }
+        else if (restoredWaitEntries.Count > 0)
```
gautam-achieveai, line 929: [Claude's bot] **MEDIUM** (Test coverage / regression safety)

This fixes the previous rollback/lifecycle issue by resolving restored `Wait` deferrals with `trigger_disabled` when `_triggerRuntime` is null. However, I could not find a regression test for this disabled-trigger restore path (`trigger_disabled`, null `triggerOptions`, or no-runtime restore scenario).

Because this path protects persisted conversations from being stranded when triggers are disabled or rolled back, it should have targeted coverage.

**Suggestion:** Add an integration/unit test that persists a deferred `Wait`, constructs or recovers a `MultiTurnAgentLoop` without `TriggerOptions`, and asserts the deferred call resolves with `status: failed` / `reason: trigger_disabled` and no deferred waits remain.

gautam-achieveai:
[Claude's bot] ## Re-Review Summary: PR #139

Verdict: APPROVE WITH COMMENTS, assuming the pending `Build and test` check completes successfully.

The update at `7d4b86ef` addresses the prior blockers: timer args now reject malformed/contradictory input instead of falling back to the timeout, restored `Wait` deferrals are resolved when triggers are disabled, `CancelWait`/`ListWaits` provider-level tests were added, and the unreachable guard was removed.

I left one non-blocking MEDIUM comment: please add regression coverage for the disabled-trigger restore path so this lifecycle-critical rollback behavior does not regress.

src/LmMultiTurn/Triggers/Sources/TimerTriggerSource.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: `TimerTriggerSource.ArmAsync` (via `ResolveFireInstant`/`ParseArgs`/`GetOptionalString`) now throws `ArgumentException` for malformed/contradictory timer args (non-JSON, non-object, wrong-typed `delay`/`deadline`, unparseable duration/instant, or both `delay` and `deadline` supplied). `TriggerRuntime.ArmCoreAsync` maps `ArgumentException` to a structured `WaitArmResult.Reject("invalid_args", ...)`; `ReconcileRestoredAsync` already resolves any rejected re-arm with a failed payload. Covered by a 6-case `[Theory]` in `TimerTriggerSourceTests`. See 7d4b86e.

gautam-achieveai: [gautam-achieveai's bot] Fixed: `TimerTriggerSource.ArmAsync` (via `ResolveFireInstant`/`ParseArgs`/`GetOptionalString`) now throws `ArgumentException` for malformed/contradictory timer args (non-JSON, non-object, wrong-typed `delay`/`deadline`, unparseable duration/instant, or both `delay` and `deadline` supplied). `TriggerRuntime.ArmCoreAsync` maps `ArgumentException` to a structured `WaitArmResult.Reject("invalid_args", ...)`; `ReconcileRestoredAsync` already resolves any rejected re-arm with a failed payload. Covered by a 6-case `[Theory]` in `TimerTriggerSourceTests`. See 7d4b86e.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: a restored `Wait` deferred entry now resolves via `ResolveToolCallAsync` with `{status:"failed", reason:"trigger_disabled"}` in `MultiTurnAgentLoop.OnHistoryRestoredAsync` when `_triggerRuntime` is null, instead of being left parked indefinitely. See 7d4b86e.

gautam-achieveai: [gautam-achieveai's bot] Fixed: a restored `Wait` deferred entry now resolves via `ResolveToolCallAsync` with `{status:"failed", reason:"trigger_disabled"}` in `MultiTurnAgentLoop.OnHistoryRestoredAsync` when `_triggerRuntime` is null, instead of being left parked indefinitely. See 7d4b86e.

src/LmMultiTurn/Triggers/WaitToolProvider.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: added `WaitToolProviderTests.cs`, exercising a real `TriggerRuntime` + `WaitToolProvider` by invoking the `FunctionDescriptor` handlers directly. Covers arm → list → cancel-by-id → list-empty round trip, cancel-by-kind cancelling multiple matching waits, and a no-selector `CancelWait` being a safe no-op. See 7d4b86e.

gautam-achieveai: [gautam-achieveai's bot] Fixed: added `WaitToolProviderTests.cs`, exercising a real `TriggerRuntime` + `WaitToolProvider` by invoking the `FunctionDescriptor` handlers directly. Covers arm → list → cancel-by-id → list-empty round trip, cancel-by-kind cancelling multiple matching waits, and a no-selector `CancelWait` being a safe no-op. See 7d4b86e.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: added `OnHistoryRestoredAsync_ResolvesRestoredWait_AsTriggerDisabled_WhenNoTriggerRuntime` in `DeferredToolExecutionTests.cs` — persists a deferred `Wait`, recovers a `MultiTurnAgentLoop` without `TriggerOptions`, and asserts the deferred call resolves with `status: failed` / `reason: trigger_disabled` and no deferred waits remain.

While verifying it I also found and fixed a genuine race in `TriggerRuntime.StartCeilingTimer`: when a restored wait's deadline is already in the past, the ceiling delay clamped to zero, erasing the grace period meant to let the source's own timer win the `fired`-vs-`timed_out` race (reproduced as a ~50% flake in `ReconcileRestored_ExpiredTimer_ResolvesImmediately`, now stable across 15/15 runs). See 2ddb855.

gautam-achieveai:
[Claude's bot] Follow-up re-review: the latest delta adds the requested trigger-disabled restore regression test and CI is now green. I found no new issues in the `7d4b86e..2ddb855` delta, so the prior non-blocking coverage concern is resolved.

gautam-achieveai:
[Claude's bot] Follow-up re-review: the latest delta adds the requested trigger-disabled restore regression test and CI is now green. I found no new issues in the `7d4b86e..2ddb855` delta, so the prior non-blocking coverage concern is resolved.

PR #147, opened 2026-07-05
Extract shared Copilot /models parsing (#132) + provider-dropdown scroll E2E (#133)

### LINKED ISSUE #132: Extract shared Copilot /models parsing between provider and proxy sample
## Context

PR #129 added `src/GithubCopilotProvider/Models/CopilotModelCatalogParser`, which unwraps the Copilot `/models` `data[]` response and scans `supported_endpoints`. The `CopilotAnthropicProxy.Sample` already has `ProxyModelResolver.ParseMessagesCapableModelIds` (`samples/CopilotAnthropicProxy.Sample/Program.cs:513`) doing the same low-level traversal.

The two projections are intentionally different:
- The proxy filters to `/v1/messages`-capable **ids only** (it's an Anthropic-Messages proxy) and has an id-only fallback when no entry carries endpoint metadata.
- The new parser returns rich `CopilotModelInfo` (vendor + transport + adaptive-thinking), keeping both Anthropic and Responses transports.

They share only the response-unwrapping and endpoint-spelling mechanics, so the `supported_endpoints` string constants and malformed-entry handling can drift independently.

## Task

Extract the shared response traversal / endpoint classification into a provider-level helper in `GithubCopilotProvider` (the proxy sample already references that project) and have both `CopilotModelCatalogParser` and `ProxyModelResolver` consume it, keeping each caller's projection/filtering decisions local.

Deferred from PR #129 review (MEDIUM, duplication/maintainability) to avoid touching the working proxy sample in the dropdown PR.

### LINKED ISSUE #133: Add browser E2E regression for provider-dropdown scroll/overflow
## Context

PR #129 fixed a dropdown that wasn't scrollable when the (now dynamically discovered) Copilot model list overflowed. The regression guard added there is a source-level check: `ProviderSelector.test.ts` regex-parses the SFC's scoped `<style>` and asserts `max-height` + `overflow-y: auto` exist. That's a cheap smoke test but doesn't prove the menu is actually scrollable/reachable in a browser — the exact symptom reported during live testing.

## Task

Add a browser regression in `tests/LmStreaming.Sample.Browser.E2E.Tests` that:
- opens the provider selector (`data-testid="provider-selector-button"`) with enough providers to overflow the menu,
- asserts the menu is scrollable (or scrolls it), and
- verifies the last option can be reached/selected (`data-testid="provider-option-*"`).

**Blocker to resolve first:** the E2E harness boots with scripted `test`/`test-anthropic` modes and Copilot discovery is token-gated (`CliCredentialCopilotTokenProvider().ResolveToken()`), so discovered Copilot models don't appear in CI. This test needs a DI seam to inject a stub `IReadOnlyList<CopilotModelInfo>` (or a set of extra providers) into `ProviderRegistry` within `BrowserWebAppFactory`, so the dropdown can be made to overflow deterministically without a real token. The source-level CSS smoke test from #129 can remain.

Deferred from PR #129 review (MEDIUM, test quality).

gautam-achieveai:
## Summary

Two follow-ups deferred from #129, shipped together on one branch.

- **#132** extracts the drift-prone GitHub Copilot `GET /models` mechanics — response unwrap (`data[]`/bare-array), endpoint-name constants, and `supported_endpoints`/string reads — into a shared `CopilotModelsResponse` helper in `GithubCopilotProvider`, consumed by both `CopilotModelCatalogParser` and the CopilotAnthropicProxy `ProxyModelResolver`. Pure refactor; each caller keeps its own projection.
- **#133** adds a real browser regression proving the provider dropdown scrolls when the dynamically discovered Copilot list overflows its capped height (`max-height: min(60vh, 320px)`). #129 guarded this only with a source-level CSS assertion.

## Changes

- **#132** New `CopilotModelsResponse` (unwrap + endpoint constants + `supported_endpoints`/string readers). `CopilotModelCatalogParser` and `ProxyModelResolver` consume it; the proxy's **per-response (not per-entry)** id-only fallback and the parser's `messages > responses` transport ranking are preserved byte-for-byte. Adds `CopilotModelsResponseTests`.
- **#133** `ProviderRegistry` gains an optional `copilotTokenAvailable` delegate seam (default = the real CLI-credential check, so production behavior is unchanged). `BrowserWebAppFactory` + `ScriptedScenario` thread an optional stub `CopilotModelInfo` list. New `ProviderDropdownScrollTests` overflows the dropdown, asserts it actually scrolls (`scrollHeight > clientHeight`) and the last option is reachable + enabled — it does **not** select it. Adds a `ProviderRegistryTests` case for the seam. The #129 CSS smoke test is kept.

## Testing

- `GithubCopilotProvider.Tests`: 83/83 (new `CopilotModelsResponseTests` + refactored parser tests)
- `CopilotAnthropicProxy.Tests`: 89/89 (`ModelResolverTests`/`ModelRewriteTests` unchanged — behavior preserved)
- `LmStreaming.Sample.Tests`: 376/376 (incl. new seam test)
- E2E `ProviderDropdownScrollTests`: 1/1
- Vitest: 292/292 (incl. #129 `ProviderSelector` CSS smoke test)
- Independent code review: no material findings.

## Key Decisions

- **Shared low-level primitives, not a shared DTO** — each caller keeps its own projection/filtering (avoids over-generalizing; YAGNI).
- `CopilotModelsResponse` is `public` because the CopilotAnthropicProxy sample is a separate assembly that consumes it; documented as low-level `/models` plumbing.
- Proxy fallback kept **per-response** (any entry declaring `supported_endpoints` → filter to `/v1/messages`-capable ids; none → id-only) — the behavior most likely to drift during extraction.
- #133 seam mirrors the existing `mockHostIsRunning` delegate; the regression asserts scroll-reachability only and does **not** change the selected provider (per review).

## Related Issues

Fixes #132
Fixes #133

### DIFF
--- a/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
+++ b/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
@@ -78,17 +78,37 @@ public sealed class ProviderDropdownScrollTests
         (await page.ProviderOption("copilot-model-00").IsEnabledAsync()).Should().BeTrue(
             "injected Copilot models render available when the token gate is forced on");
 
-        // The menu must actually be scrollable — its content exceeds the capped visible height.
-        (await menu.EvaluateAsync<bool>("el => el.scrollHeight > el.clientHeight")).Should().BeTrue(
-            "the overflowing Copilot list must scroll inside the capped-height menu, not run off-screen");
+        // The menu must actually be scrollable via CSS — `scrollHeight > clientHeight` alone proves the
+        // content overflows, but that would hold even if `overflow-y` regressed back to the default
+        // `visible` (content would simply spill out unclipped instead of scrolling).
+        (await menu.EvaluateAsync<string>("el => getComputedStyle(el).overflowY")).Should().BeOneOf(
+            ["auto", "scroll"],
+            "the overflowing Copilot list must scroll inside the capped-height menu, not spill out unclipped");
 
-        // The last option starts below the fold; scrolling must bring it fully into view, and it must be
-        // enabled/reachable. We stop here — selecting it is intentionally out of scope.
+        // Prove the menu is actually scrollable, not merely eligible per its computed style: driving
+        // scrollTop to scrollHeight must move the offset off zero (the browser clamps to the real max).
+        await menu.EvaluateAsync("el => { el.scrollTop = el.scrollHeight; }");
+        (await menu.EvaluateAsync<double>("el => el.scrollTop")).Should().BeGreaterThan(
+            0,
+            "scrolling the menu to its content height must move the scroll offset off zero");
+
+        // The last option must be enabled, and — after scrolling — its bounding box must actually fall
+        // within the menu's clipped viewport. IsVisibleAsync alone does not prove containment within a
+        // scrollable ancestor's viewport, so we stop at bounding-box containment; selecting it remains
+        // out of scope.
         var lastOption = options.Last;
         (await lastOption.IsEnabledAsync()).Should().BeTrue();
-        await lastOption.ScrollIntoViewIfNeededAsync();
-        (await lastOption.IsVisibleAsync()).Should().BeTrue(
-            "the last option must be reachable by scrolling the menu");
+
+        var menuBox = await menu.BoundingBoxAsync();
+        var lastOptionBox = await lastOption.BoundingBoxAsync();
+        menuBox.Should().NotBeNull("the menu must be laid out to assert reachability against it");
+        lastOptionBox.Should().NotBeNull("the last option must be laid out to assert reachability against it");
+        lastOptionBox!.Y.Should().BeGreaterThanOrEqualTo(
+            menuBox!.Y,
+            "the last option must be reachable within the menu's top edge after scrolling");
+        (lastOptionBox.Y + lastOptionBox.Height).Should().BeLessThanOrEqualTo(
+            menuBox.Y + menuBox.Height,
+            "the last option must be fully reachable within the menu's clipped bottom edge, not scrolled past it");
 
         await session.SaveSuccessScreenshotAsync("ProviderDropdownScroll.overflow_reaches_last_option");
     }
--- a/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
+++ b/tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
@@ -1,0 +1,95 @@
+using AchieveAi.LmDotnetTools.GithubCopilotProvider.Models;
+using AchieveAi.LmDotnetTools.LmTestUtils.TestMode;
+using FluentAssertions;
+using LmStreaming.Sample.Browser.E2E.Tests.Infrastructure;
+
+namespace LmStreaming.Sample.Browser.E2E.Tests.Scenarios;
+
+/// <summary>
+/// Regression for the provider-dropdown scroll fix (issue #133, originally shipped in #129): when the
+/// dynamically discovered Copilot model list overflows the menu's capped height
+/// (<c>max-height: min(60vh, 320px)</c>), the menu must actually scroll so every option — including the
+/// last — stays reachable. PR #129 guarded this only with a source-level CSS assertion; this drives a
+/// real browser.
+///
+/// The Copilot list is token-gated and empty in CI, so a stub <see cref="CopilotModelInfo"/> list is
+/// injected through <see cref="BrowserWebAppFactory"/> with the Copilot token gate forced on, making the
+/// discovered models render as <em>available</em> providers without a real gh/Copilot login. The test
+/// proves the overflowed last option is present, enabled, and reachable by scrolling — it deliberately
+/// does NOT select it (the selection flow is covered elsewhere; changing the provider is out of scope
+/// for a scroll regression).
+/// </summary>
+[Collection(PlaywrightCollection.Name)]
+public sealed class ProviderDropdownScrollTests
+{
+    private readonly PlaywrightFixture _fixture;
+
+    public ProviderDropdownScrollTests(PlaywrightFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    // Enough entries that the two Copilot partitions alone dwarf the 320px cap under CI fonts —
+    // ~20 items × ~35px each + the base providers is well over 320px, so overflow is deterministic
+    // rather than dependent on incidental viewport/font sizing.
+    private static IReadOnlyList<CopilotModelInfo> StubCopilotModels(int count = 20)
+    {
+        return
+        [
+            .. Enumerable.Range(0, count)
+                .Select(i => new CopilotModelInfo(
+                    $"copilot-model-{i:D2}",
+                    $"Copilot Model {i:D2}",
+                    i % 2 == 0 ? CopilotModelVendor.Anthropic : CopilotModelVendor.OpenAI,
+                    i % 2 == 0 ? CopilotModelTransport.Anthropic : CopilotModelTransport.Responses))
+        ];
+    }
+
+    [Fact]
+    public async Task Overflowing_provider_dropdown_scrolls_and_reaches_the_last_option()
+    {
+        // A no-op responder: the test never sends a message, it only inspects the provider dropdown.
+        var responder = ScriptedSseResponder
+            .New()
+            .ForRole("parent", ctx => ctx.SystemPromptContains("assistant"))
+            .Turn(t => t.Text("noop"))
+            .Build();
+
+        var copilotModels = StubCopilotModels();
+
+        await using var session = await _fixture.OpenAsync("test", responder, copilotModels: copilotModels);
+        var page = session.Page;
+
+        // Pin the viewport so the 60vh/320px cap and thus the overflow are deterministic across machines.
+        await page.SetViewportSizeAsync(1280, 720);
+
+        // Open the dropdown and wait for the menu to actually render before measuring it.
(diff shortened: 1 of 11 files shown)


tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        await using var session = await _fixture.OpenAsync("test", responder, copilotModels: copilotModels);
+        var page = session.Page;
+
+        // Pin the viewport so the 60vh/320px cap and thus the overflow are deterministic across machines.
+        await page.SetViewportSizeAsync(1280, 720);
+
+        // Open the dropdown and wait for the menu to actually render before measuring it.
+        await page.ProviderSelectorButton().ClickAsync();
+        var menu = page.Locator(".dropdown-menu");
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test maintainability / repo convention)

This browser E2E uses `page.Locator(".dropdown-menu")`, but the LmStreaming.Sample browser-test contract says to select by `data-testid` and avoid brittle CSS classes. This test is specifically guarding dropdown layout/scroll behavior, so tying it to the implementation class name makes it fragile: a harmless CSS class rename or restructure can break the E2E even if the user-visible behavior and test-id contract still work.

**Suggestion:** Add a stable test id to the provider selector menu, for example `data-testid="provider-selector-menu"`, expose a `ProviderSelectorMenu()` helper in `UiHelpers.cs`, and use that helper here instead of `.dropdown-menu`. Keep CSS-rule assertions in the existing Vitest/source-level test; browser E2E should drive the DOM through stable test ids.

gautam-achieveai:
[gautam-achieveai's bot] # PR Review: Extract shared Copilot /models parsing (#132) + provider-dropdown scroll E2E (#133)

Verdict: REQUEST_CHANGES (posted as a comment because GitHub does not allow this authenticated account to request changes on its own PR)

The parser/helper extraction and the ProviderRegistry test seam are well-scoped to #132/#133. The behavior-preservation tests are solid, and the dropdown regression correctly avoids selecting the provider.

I found one blocking maintainability issue in the new browser E2E: it uses a CSS class selector for the dropdown menu, which violates the LmStreaming.Sample browser-test selector contract. Because this is new test code and the fix is small, please switch it to a stable `data-testid` before merge.

Findings: 0 critical, 0 high, 1 medium, 0 low. Questions: 0.

Inline finding: https://github.com/achieveai/LmDotnetTools/pull/147#discussion_r3524236973

tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 097c7a4. Added `data-testid="provider-selector-menu"` to the menu in `ProviderSelector.vue`, exposed `page.ProviderSelectorMenu()` in `UiHelpers.cs`, and the E2E now drives the menu through that helper instead of the `.dropdown-menu` CSS class. The CSS-rule assertions stay in the Vitest source-level test as suggested. Re-verified: `ProviderDropdownScrollTests` 1/1 and Vitest 292/292 green.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #147

Verdict: APPROVE / no remaining findings (posted as a comment because GitHub does not allow this authenticated account to formally approve its own PR).

Re-reviewed delta `05f50ef8..097c7a4b`. The previous selector-contract blocker is fixed cleanly:

- `ProviderSelector.vue` now exposes `data-testid="provider-selector-menu"` on the dropdown menu.
- `UiHelpers.cs` now exposes `ProviderSelectorMenu()`.
- `ProviderDropdownScrollTests` now uses `page.ProviderSelectorMenu()` instead of the brittle `.dropdown-menu` CSS selector.

No new issues found in the re-review delta.

tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        // DI replacement (or the token-gate seam) regressed, before we ever measure scroll behavior.
+        var options = page.Locator("[data-testid^='provider-option-']");
+        (await options.CountAsync()).Should().BeGreaterThan(
+            copilotModels.Count,
+            "the base providers plus every injected Copilot model should be rendered");
+        await page.ProviderOption("copilot-model-00").WaitForAsync();
+        (await page.ProviderOption("copilot-model-00").IsEnabledAsync()).Should().BeTrue(
+            "injected Copilot models render available when the token gate is forced on");
+
+        // The menu must actually be scrollable — its content exceeds the capped visible height.
```
gautam-achieveai: [gautam-achieveai's bot] **MEDIUM** (Test Coverage / Test Quality)

This browser regression still does not prove that the dropdown menu itself is scrollable. `scrollHeight > clientHeight` proves the content overflows the capped box, but it would also be true if `overflow-y` regressed back to the default `visible`. Likewise, `ScrollIntoViewIfNeededAsync()` followed by `IsVisibleAsync()` does not prove the last option is reachable within the dropdown's clipped scroll viewport; Playwright visibility can be satisfied even when the element is outside the menu viewport.

Because #133 is specifically about adding real browser coverage for dropdown scroll/reachability, this leaves the test able to pass while the user-facing regression it is meant to guard comes back.

**Suggestion:** Assert actual scroll behavior, for example:

```csharp
(await menu.EvaluateAsync<string>("el => getComputedStyle(el).overflowY"))
    .Should().BeOneOf("auto", "scroll");

await menu.EvaluateAsync("el => { el.scrollTop = el.scrollHeight; }");
(await menu.EvaluateAsync<double>("el => el.scrollTop"))
    .Should().BeGreaterThan(0);

var menuBox = await menu.BoundingBoxAsync();
var lastBox = await lastOption.BoundingBoxAsync();
lastBox!.Y.Should().BeGreaterThanOrEqualTo(menuBox!.Y);
(lastBox.Y + lastBox.Height).Should().BeLessThanOrEqualTo(menuBox.Y + menuBox.Height);
```

That verifies both the CSS scroll contract and that the final provider option is actually reachable inside the menu viewport.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #147

Verdict: REQUEST_CHANGES (posted as a comment because GitHub may not allow this authenticated account to formally request changes on its own PR).

Reviewed HEAD `097c7a4b`. CI is green (`Build and test`, `browser-e2e`). The Copilot `/models` extraction looks scoped to #132 and behavior-preserving; design, architecture, duplication, performance, schema-compatibility, over-engineering, and feature-flag checks did not surface material issues.

I found one remaining medium issue in the #133 browser regression: the new E2E proves overflow exists, but it does not prove the dropdown menu actually scrolls or that the last option is reachable within the menu viewport. Since proving real browser scrollability is the core purpose of #133, please strengthen that assertion before merge.

Findings: 0 critical, 0 high, 1 medium, 0 low. Questions: 0.

Inline finding: https://github.com/achieveai/LmDotnetTools/pull/147#discussion_r3524561784

tests/LmStreaming.Sample.Browser.E2E.Tests/Scenarios/ProviderDropdownScrollTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: replaced the `scrollHeight > clientHeight` check and the `ScrollIntoViewIfNeededAsync()` + `IsVisibleAsync()` pair (commit 335b444) with three assertions that prove actual scroll behavior: (1) computed `overflow-y` on the menu is asserted to be `auto` or `scroll` via `getComputedStyle`, ruling out a regression to the default `visible` that the old overflow check couldn't catch; (2) the menu is driven with `el.scrollTop = el.scrollHeight` and `scrollTop` is asserted to move off `0`, proving it is actually scrollable rather than merely eligible per CSS; (3) after that scroll, the last option's bounding box (`Y` and `Y + Height`) is asserted to fall within the menu's bounding box, proving containment within the clipped scroll viewport, which `IsVisibleAsync()` does not check against a clipping ancestor. Selecting the last option remains out of scope per the file's existing doc comment. Ran `dotnet test --filter FullyQualifiedName~ProviderDropdownScrollTests`: 1 passed.

gautam-achieveai: [gautam-achieveai's bot] Fixed in `335b4446`. The E2E now checks computed `overflow-y`, drives `menu.scrollTop`, asserts the scroll offset moves, and verifies the last option's bounding box is contained within the menu viewport after scrolling. This covers the scrollability/reachability gap from this thread.

gautam-achieveai: [gautam-achieveai's bot] Fixed in `335b4446`. The E2E now checks computed `overflow-y`, drives `menu.scrollTop`, asserts the scroll offset moves, and verifies the last option's bounding box is contained within the menu viewport after scrolling. This covers the scrollability/reachability gap from this thread.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #147

Verdict: APPROVE / no remaining findings (posted as a comment because GitHub does not allow this authenticated account to formally approve its own PR).

Re-reviewed delta `097c7a4b..335b4446`. The previous scrollability coverage issue is fixed: `ProviderDropdownScrollTests` now checks computed `overflow-y`, drives `menu.scrollTop`, asserts the scroll offset moves, and verifies the last option's bounding box is contained within the menu viewport after scrolling. Latest CI is green (`Build and test`, `browser-e2e`). No remaining issues found.

PR #151, opened 2026-07-05
LmStreaming.Sample: REST API for headless conversation control

### LINKED ISSUE #138: LmStreaming.Sample: REST API for headless conversation control
## Summary
Add a REST API surface to `samples/LmStreaming.Sample` for external, programmatic callers to drive conversations without an open WebSocket connection. Reuses existing workspace CRUD (`WorkspacesController`) and adds three new capabilities on top of `ConversationsController`:

1. **Provision a conversation** — `POST /api/conversations` reserves a `threadId` and locks workspace/provider/mode as metadata only (no live agent/sandbox session starts, no first message sent). Optionally accepts an `authWebhookUrl`: when the workspace's sandbox needs an OAuth-style auth action (GitHub/ADO/M365 sign-in) mid-run, this lets the calling app receive that challenge itself and remain responsible for completing auth, instead of it only surfacing to a human in the WebUI. We act as a proxy — we forward the same auth-lifecycle events (required, completed, denied) the WebUI already gets, we don't broker credentials ourselves.
2. **Send a message** — `POST /api/conversations/{threadId}/messages` sends a message to a thread, used for both the first turn and later resumes. If a run is already in-flight for that thread, the message queues and chains after the current run completes (matching today's WebSocket behavior) rather than being rejected.
3. **Poll status / final response** — a status endpoint reports one of `NotStarted / InProgress / Completed / Errored / Interrupted` for a `threadId` (optionally scoped to a specific `runId`), and for `Completed` returns the final assistant response tagged to that run.

Also adds a lightweight `?threadId=` query-param deep link so the existing Vue WebUI can open directly into a specific conversation from a URL, reusing the existing resume/run-state-check path (`useChat.resumeStreamIfActive`).

## Value
Lets external tools and scripts drive LmStreaming.Sample conversations headlessly — start a run, come back later, check whether it's done, and read the result — without needing to hold open a WebSocket connection or reimplement the WebUI's client logic. This is the missing piece for automation (CI jobs, bots, other services) that wants to use LmStreaming.Sample as a conversation backend instead of just a human-facing chat UI, and closes the loop with a shareable URL back into the same conversation in the WebUI. The auth-webhook forward extends this to workspaces whose sandbox needs interactive sign-in: the calling app — not a human watching the WebUI — owns getting that sign-in completed.

## Non-Goals (v1)
- SSE heartbeat + final-response stream — deferred to a follow-up issue.
- Full client-side routing (vue-router) — deep-link is query-param only.
- Authentication/authorization on the new endpoints — same trusted-network posture as the rest of the sample's API today. **Caveat:** this raises the blast radius versus the WebSocket-only surface, since a workspace can mount a sandbox directory containing arbitrary files; auth should be a prerequisite before exposing this API outside a trusted network.
- See #89 (open) — provisioning a conversation must not race with #89's first-message sandbox-session binding; coordinate implementation so "provisioned, no session yet" is a defined intermediate state in both.
- Message idempotency/dedup — sends are not deduplicated; a client that times out and retries `POST .../messages` can double-send the same user message into the run. Out of scope for v1 (matches the WebSocket path's existing behavior).
- Auth-webhook signing/verification — the forwarded call carries no shared secret or signature the receiver can check. This matches the API's existing no-auth v1 posture and every other webhook already in this codebase (flat shared-secret pattern), but is called out separately because it's a new **outbound** call to a caller-supplied URL. Out of scope for v1.
- SSRF hardening on the caller-supplied `authWebhookUrl` — no scheme restriction, no private/internal-IP blocking. **Caveat:** an unrestricted POST to an arbitrary caller-supplied URL is a standard SSRF vector (could reach internal services, cloud metadata endpoints, etc.). Accepted for v1 under the same trusted-network posture as the rest of the API; do not expose this endpoint to untrusted callers.
- Multi-webhook-per-session arbitration beyond first-wins — if a second conversation on the same (shared, per-workspace) sandbox session registers a different `authWebhookUrl`, it is silently never used while the session's first-registered webhook remains bound. No API surfaces this "your webhook is shadowed" state to the second caller in v1.
- `reason` string enrichment — the forwarded `reason` is today a fixed generic string ("sandbox egress requires sign-in to '{providerId}'"), so forwarding it to an arbitrary caller-supplied URL carries negligible EUII risk. If a future change enriches `reason` with request-specific detail (e.g. destination host/path), revisit whether it remains safe to forward externally.

## Implementation Plan
Ships as a single PR covering:
- **Core persistence** — `LmMultiTurn`/`LmAgentInfra`: persist in-flight run markers so restart-recovery works (see AC below), with tests in `LmMultiTurn.Tests`. This changes crash-consistency behavior for every `MultiTurnAgentBase` consumer, not just this sample.
- **Sample REST + WebUI** — `samples/LmStreaming.Sample`: the new REST endpoints (provision/send/poll) + WebUI `?threadId=` deep link.
- **Auth-webhook-forwarding** — `src/LmAgentInfra` + `samples/LmStreaming.Sample`: accept `authWebhookUrl` at provision, persist it on `ThreadMetadata`, start consuming the currently-bound-but-unread `AuthWebhookRequest.SessionId` in `AuthWebhookController` to resolve the session's attached threads (`SandboxSessionRegistry.GetThreads`), and forward `auth_required`/`auth_completed`/`auth_denied` to the first-registered thread's webhook — in addition to, not instead of, the existing WebSocket broadcast.

## Acceptance Criteria
- [ ] `POST /api/conversations` accepts `{workspaceId, providerId, modeId}`, returns a new `threadId`, and persists the choice to the conversation store's thread metadata without starting a live agent/sandbox session. An unknown `workspaceId` or `modeId` returns `400`/`404` (reuse the existing store lookups — do **not** silently substitute a default the way the WebSocket handshake does today); an unknown/unavailable `providerId` returns `503` in the same shape `ConversationsController.SwitchMode` already uses.
- [ ] `POST /api/conversations` accepts an optional `authWebhookUrl`, persisted on the thread's `ThreadMetadata`. The forward is evaluated inside `AuthWebhookController.Evaluate` using that inbound gateway call's own `SessionId` (`SandboxSessionRegistry.GetThreads(sessionId)`) — **not** chained off `IAuthEventNotifier.NotifyAuthRequiredAsync`'s dedup, since `PendingAuthCoordinator` collapses concurrent waiters onto one hold keyed only by `providerId`, and two different workspaces needing the same provider at the same time would otherwise cause the second workspace's forward to silently never fire. The backend `POST`s `{providerId, signinUrl, reason, threadId, runId}` to the registered `authWebhookUrl`, where `signinUrl` is the **absolute** URL (`AuthOptions.Webhook.CallbackBaseUrl` + the relative `/auth/{providerId}` path — the same base URL already used to build `SandboxSessionRegistry.DiscoveryWebhookUrl`), not the bare relative path the WebUI uses for its same-origin popup. No signature/secret is sent (matches v1's no-auth posture). The call runs with its own hard-enforced timeout (2-3s). When the session also has a WS client attached (the WebUI case), the forward fires independently of — never inline-awaited ahead of — the existing WebSocket broadcast, so a hung external webhook cannot delay the human-facing auth banner. For a purely API-driven thread with no attached WS client, there is no WebSocket broadcast to race against: the webhook forward **is** the notification, not a supplementary channel alongside one. It is best-effort, not retried or queued. Each attempt emits a structured log line (`threadId`, `runId`, `providerId`, destination host — not full URL or payload — outcome: sent/failed/timed-out, latency).
- [ ] Because sandbox sessions are shared per-workspace, when a second thread on the same workspace/session registers a different `authWebhookUrl`, the session keeps forwarding to whichever thread's webhook was bound **first**. "First" is determined by a persisted `AuthWebhookRegisteredAt` UTC timestamp set on `ThreadMetadata` when `authWebhookUrl` is provided at provision time — not by `SandboxSessionRegistry.RegisterThread`'s attach order (which fires at first-message/agent-creation time and has no timestamp today, so it cannot serve as a reliable "first" signal on its own). Among a session's attached threads with a registered `authWebhookUrl`, the one with the earliest `AuthWebhookRegisteredAt` wins; the rest are stored but never used for forwarding while it remains attached.
- [ ] A thread provisioned without `authWebhookUrl` (or whose session has none bound) behaves exactly as today: WebSocket-only broadcast, no regression.
- [ ] `AuthWebhookUrl` is carried over by `ConversationsController.UpdateMetadata` (which today rebuilds a new `ThreadMetadata` record and only explicitly copies `CurrentRunId`/`LatestRunId`/`SessionMappings`/`Properties`). A regression test verifies that editing a conversation's title/metadata after provisioning with `authWebhookUrl` does not silently null it out.
- [ ] If the gateway forces a sandbox session recreation (`SandboxSessionRegistry.InvalidateSession` on a 404), previously-attached threads are not automatically re-registered against the new session id. Forwarding correctly falls back to WebSocket-only (no crash, no exception surfaced to the run) until a thread's agent is recreated and re-attaches; this is asserted by a test, not left undefined.
- [ ] Deleting the first-registered thread (`DELETE /api/conversations/{threadId}` → `SandboxSessionRegistry.UnregisterThreadFromAllSessions`) frees up forwarding eligibility for a remaining attached thread's webhook (or falls back to WebSocket-only if none remain) — this already-wired behavior is covered by an explicit test rather than left as an accidental side effect.
- [ ] When the sandbox auth flow resolves, the backend forwards the equivalent completed/denied payload to the same eligible webhook: `{providerId, threadId, runId}` on completion (mirroring the existing WebSocket `auth_completed` frame where one exists), or `{providerId, reason, threadId, runId}` on denial (mirroring `auth_denied`) — matching the same non-blocking-timeout, best-effort, logged-attempt semantics as the `auth_required` forward, and independent of whether a WS client is attached at all. Without this, an external caller that started waiting on `auth_required` has no signal to stop waiting, mirroring the exact hang this feature exists to prevent for the human WebUI.
- [ ] `POST /api/conversations/{threadId}/messages` accepts a message body, starts (or queues onto, if one is already in-flight) a run for that thread, and returns the run's `runId` and an `inputId` for correlation — matching, not diverging from, the queuing semantics of the existing WebSocket send path. A `threadId` that was never provisioned returns `404` — it must **not** silently auto-vivify a new conversation under default workspace/provider/mode (today's `MultiTurnAgentPool.GetOrCreateAgent` behavior for an unrecognized thread), since that would defeat the whole point of provisioning locking the triad.
- [ ] If the underlying input queue is full (capacity 100 per thread), the endpoint returns `503` with a "queue full, retry" body within a bounded wait — it does not hang the HTTP request indefinitely waiting for a slot.
- [ ] A status/poll endpoint returns one of exactly five states for a `threadId`/`runId`: `{"threadId": "...", "runId": "...", "status": "NotStarted" | "InProgress" | "Completed" | "Errored" | "Interrupted", "response": ... | null}` (string enum, PascalCase, field name `status`). `Errored` means the terminal message had `RunCompletedMessage.IsError == true`, **or** the run task faulted/threw before producing a completion message — both map to `Errored`. `Interrupted` is terminal per `runId`: a later send mints a new `runId`, and the interrupted one keeps reporting `Interrupted` forever.
- [ ] For `Completed`, the endpoint returns the final assistant response tagged with that `runId`, with a documented convention for a run that closed with only tool activity and no text.
- [ ] In-flight run markers are persisted in `MultiTurnAgentBase`/`IConversationStore` (not just completion state) so that after a process restart, polling a `threadId` whose run was interrupted returns `Interrupted` rather than a false `NotStarted`/`InProgress:false`.
- [ ] Two concurrent sends to the same `threadId` (whether both via REST, or one REST + one WebSocket) behave consistently: the second queues and chains after the first, and each caller's poll-by-`runId` only ever returns the response to its own run.
- [ ] The Vue WebUI supports `?threadId=<id>` on load: it auto-selects and loads that conversation via the same run-state-check-then-resume path used today for tab-switch-back (not just a static history load), so an in-flight REST-started conversation renders live rather than stale. Loading a `?threadId=` for an unknown/deleted thread shows a "conversation not found" state, not a silently-blank chat.
- [ ] Existing Browser.E2E WebSocket chat scenarios (per `samples/LmStreaming.Sample/PlaywrightTestingGuide.md`) stay green — no regression to the current WebUI chat flow.

### LINKED ISSUE #148: [#138] Area 1: Core run-ledger persistence + accepted-input durability
Relates to #138

Part of #138's implementation plan v5 (Area 1). Adds `RunLedgerEntry`/`IRunLedgerStore` to `src/LmMultiTurn` (companion to `IConversationStore`, not a modification of it): run status ledger (`Queued/InProgress/Completed/Errored/Interrupted`) with `InputIds`, plus `AcceptedInputEntry` CRUD (`RecordAcceptedInputAsync`/`RemoveAcceptedInputAsync`/`ListAcceptedInputIdsAsync`) so a `202`-accepted input is durably recorded before it becomes pollable. Implemented across all 3 built-in stores (InMemory/File/Sqlite). `MultiTurnAgentBase` gains `persistRunLedger` (default off), a fully-ordered `TrySendAsync` (persist-accepted -> non-blocking enqueue -> compensating rollback on full), `RecordInjectedInputsAsync`, and widened restart reconciliation for orphaned accepted-input records. Threaded through `MultiTurnAgentLoop`/`ClaudeAgentLoop`/`CodexAgentLoop`/`CopilotAgentLoop`.

Must land and be green before Area 2 and Area 3 start.

See plan v5: https://github.com/achieveai/LmDotnetTools/issues/138#issuecomment-4884645603

### LINKED ISSUE #149: [#138] Area 2: Sample REST API + WebUI (conversation provision/send/poll)
Relates to #138

Part of #138's implementation plan v5 (Area 2), depends on Area 1. Adds `Services/ConversationStatusResolver.cs` (5-state resolution + `ResolveByInputIdAsync`, incl. the accepted-vs-never-accepted distinction), request/response DTOs, and `Controllers/ConversationsController.cs` (Provision/SendMessage/GetStatus with 404/500/503 branches) to `samples/LmStreaming.Sample`. Wires `persistRunLedger: true` at the sample's loop-construction sites in `Program.cs`. Adds `?threadId=` deep link to `ChatLayout.vue` and matching client bindings in `conversationsApi.ts`.

Depends on Area 1 (#<area1-issue-number, filled after creation>) landing first.

See plan v5: https://github.com/achieveai/LmDotnetTools/issues/138#issuecomment-4884645603

### LINKED ISSUE #150: [#138] Area 3: Auth-webhook forwarding (target capture on auth lifecycle)
Relates to #138

Part of #138's implementation plan v5 (Area 3), depends on Areas 1 (#148) and 2 (#149). Adds `IAuthWebhookForwarder`/`NoOpAuthWebhookForwarder` to `src/LmAgentInfra/Auth`, threads a captured-target local variable through `AuthWebhookController.Evaluate` spanning required->terminal (so completed/denied outcomes forward to the same target captured at auth_required time, not a recomputed one), and adds `samples/LmStreaming.Sample/Auth/SandboxAuthWebhookForwarder.cs` + `Program.cs` DI registration.

See plan v5: https://github.com/achieveai/LmDotnetTools/issues/138#issuecomment-4884645603

gautam-achieveai:
## Summary
Adds a REST API surface to `samples/LmStreaming.Sample` so external callers can provision, drive, and poll a conversation without holding a WebSocket open, plus the auth-webhook forwarding that lets such a caller receive sandbox sign-in challenges directly. Implemented per plan v5 across 3 areas, each with its own child issue.

## Changes

**Area 1 — Core run-ledger persistence (#148)**
- `RunLedgerEntry`/`IRunLedgerStore` added to `src/LmMultiTurn` as a companion to `IConversationStore`, implemented across all 3 stores (InMemory/File/Sqlite).
- `MultiTurnAgentBase` gains `persistRunLedger` (default off), an ordered `TrySendAsync` (persist-accepted → non-blocking enqueue → compensating rollback on full), `RecordInjectedInputsAsync`, and restart reconciliation for orphaned accepted-input records.
- Threaded through `MultiTurnAgentLoop`/`ClaudeAgentLoop`/`CodexAgentLoop`/`CopilotAgentLoop`.

**Area 2 — Sample REST API + WebUI (#149)**
- `Services/ConversationStatusResolver.cs` (5-state resolution: `NotStarted/InProgress/Completed/Errored/Interrupted`, incl. accepted-vs-never-accepted distinction).
- `Controllers/ConversationsController.cs`: `POST /api/conversations` (provision), `POST /api/conversations/{threadId}/messages` (send/queue), `GET` status/poll, with 404/500/503 branches matching existing WebSocket semantics (no auto-vivify, no default substitution).
- `?threadId=` deep link + not-found state in `ChatLayout.vue`, plus `conversationsApi.ts` client bindings.
- `persistRunLedger: true` wired at the sample's loop-construction sites.

**Area 3 — Auth-webhook forwarding (#150)**
- `IAuthWebhookForwarder`/`SandboxAuthWebhookForwarder`/`NoOpAuthWebhookForwarder` in `src/LmAgentInfra/Auth` and `samples/LmStreaming.Sample/Auth`.
- `AuthWebhookController.Evaluate` captures the forwarding target once at `auth_required` and reuses it for whichever terminal call (`auth_completed`/`auth_denied`) fires, so a mid-flight target change (thread deleted, new thread registered) can't redirect the outcome.
- Eligible thread resolution is session-scoped, tie-broken by ascending `(RegisteredAt, ThreadId)`; runs independently of and never blocks the existing WebSocket broadcast.
- `AuthSigninUrls` extracted so `PendingAuthCoordinator` and the controller build identical signin-url/reason strings.
- DI wiring: `SandboxAuthWebhookForwarder` in `LmStreaming.Sample`, `NoOpAuthWebhookForwarder` in `CodeReviewDaemon.Sample` (no chat threads to forward to).

## Testing
- Full solution build: clean.
- `LmMultiTurn.Tests`: 343/343 passed (run-ledger round-trip + recovery).
- `LmStreaming.Sample.Tests`: 423/423 passed (REST contract tests against a real `MultiTurnAgentBase` agent, status-resolver 5-state coverage, controller-forwarder + forwarder-unit auth tests, client Vitest tests for the deep link/404 branches).
- `CodeReviewDaemon.Sample.Tests`: 293/293 passed.
- `LmStreaming.Sample.E2E.Tests`: 76/78 passed; the 2 failures are in `SandboxWorkspaceGatewayE2ETests` and are a pre-existing environmental issue (a stale external sandbox gateway process unrelated to any file in this PR — confirmed via a clean 54/54 run of the Auth-scoped E2E subset, which includes the existing full-pipeline `AuthWebhookControllerTests.cs`).
- Existing WebSocket Browser.E2E chat scenarios unaffected — no changes to that path's runtime behavior.

## Related Issues
Fixes #138
Fixes #148
Fixes #149
Fixes #150


src/LmAgentInfra/Controllers/AuthWebhookController.cs
```diff
@@ -159,8 +169,72 @@ public async Task<IActionResult> Evaluate(
...
+            {
+                return null;
+            }
+
+            try
+            {
+                return await authWebhookForwarder.NotifyAuthRequiredAsync(
+                    body.SessionId,
+                    providerId,
+                    AuthSigninUrls.BuildSigninUrl(providerId),
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Correctness / Contract)

The forwarded auth webhook is passing `AuthSigninUrls.BuildSigninUrl(providerId)`, which currently returns the relative `/auth/{providerId}` path. That works for the same-origin WebSocket/UI banner, but #138's headless webhook contract requires the external caller to receive an absolute URL built from `AuthOptions.Webhook.CallbackBaseUrl + /auth/{providerId}`. A REST-only webhook receiver cannot reliably know which host/port to open from `/auth/github`, so the auth flow this API is meant to enable can dead-end.

**Suggestion:** Keep the relative helper for same-origin UI prompts, but use/inject `AuthOptions` for the webhook-forwarding leg and pass an absolute callback URL. Add a controller/forwarder test asserting the forwarded `signinUrl` is absolute.

samples/LmStreaming.Sample/Auth/SandboxAuthWebhookForwarder.cs
```diff
@@ -0,0 +1,255 @@
...
+                payload.ThreadId,
+                payload.ProviderId,
+                SafeHost(webhookUrl));
+        }
+    }
+
+    private static string SafeHost(string webhookUrl) =>
+        Uri.TryCreate(webhookUrl, UriKind.Absolute, out var uri) ? uri.Host : "(unparseable)";
+
+    private sealed record AuthWebhookForwardPayload
```
gautam-achieveai, line 251: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Wire Contract / Schema)

`AuthWebhookForwardPayload` is serialized with `JsonSerializerOptionsFactory.CreateForProduction()`, which has no naming policy, and the payload properties do not have `[JsonPropertyName]` attributes. The external webhook body therefore uses PascalCase CLR names (`Type`, `SessionId`, `ThreadId`, `ProviderId`, `SigninUrl`, `Reason`) while the issue/PR contract documents lower-camel names (`providerId`, `signinUrl`, `threadId`, `runId`, etc.). This is a new public webhook contract, so shipping the wrong casing makes integrations code to the wrong shape and is sticky after release.

**Suggestion:** Pin the outbound schema with explicit `[JsonPropertyName("type")]`, `[JsonPropertyName("providerId")]`, `[JsonPropertyName("signinUrl")]`, etc. (or use dedicated camelCase serializer options), and update the forwarder tests to assert the documented wire names instead of the CLR property names.

samples/LmStreaming.Sample/Services/ConversationStatusResolver.cs
```diff
@@ -0,0 +1,145 @@
...
+    private async Task<object?> LoadFinalResponseAsync(string threadId, string runId, CancellationToken ct)
+    {
+        var messages = await conversationStore.LoadMessagesAsync(threadId, ct);
+        var runMessages = messages.Where(m => m.RunId == runId).ToList();
+        if (runMessages.Count == 0)
+        {
+            return null;
+        }
+
+        var textMessage = runMessages.LastOrDefault(m => m.MessageType == nameof(TextMessage));
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Correctness / API Response)

This selects the last persisted `TextMessage` for the run without checking `Role` or `IsThinking`. User inputs are also persisted as `TextMessage`, and some providers can persist thinking/reasoning as `TextMessage` with `IsThinking = true`; in those cases the status endpoint can return the user's prompt or internal thinking instead of the final assistant answer. For the headless REST API, that violates the response contract and can expose content the caller did not ask for as the final response.

**Suggestion:** Deserialize/filter candidate messages to assistant, non-thinking `TextMessage` instances (or at least check persisted/deserialized `Role == Assistant` and `IsThinking == false`) before falling back to non-text messages for tool-only runs. Add tests for an errored run with only user input and for assistant thinking followed by the actual answer.

src/LmMultiTurn/MultiTurnAgentBase.cs
```diff
@@ -1201,6 +1289,17 @@ protected RunAssignment StartRun(IReadOnlyList<QueuedInput> inputs, string? pare
             _currentRunId = runId;
         }
 
+        if (RunLedgerStore != null)
```
gautam-achieveai, line 1292: [gautam-achieveai's bot] **MEDIUM** (Maintainability / Data Growth)

Accepted-input records are written before enqueue, but after the same receipt id is folded into `RunLedgerEntry.InputIds`, the accepted-input entry is never removed and no retention/cleanup policy is documented. The approved plan explicitly called out that implementation should either delete accepted-input records once assigned or intentionally retain them with documented cleanup expectations. As written, every successful REST send leaves redundant per-thread state forever (`accepted-inputs.json` / `accepted_inputs`), which grows unbounded and makes future reconciliation scan stale accepted ids.

**Suggestion:** Remove accepted-input entries after `StartRunAsync` and `RecordInjectedInputsAsync` successfully persist the ledger fold, or document retention as intentional and add cleanup/compaction tests proving stale accepted records do not affect reconciliation/polling.

src/LmMultiTurn/MultiTurnAgentBase.cs
```diff
@@ -1251,6 +1389,24 @@ await PublishToAllAsync(new RunCompletedMessage
             _currentRunId = null;
         }
 
+        if (RunLedgerStore != null)
+        {
+            var existing = await RunLedgerStore.LoadRunLedgerAsync(runId, ct);
+            if (existing != null)
+            {
+                var status = isError ? RunStatus.Errored : RunStatus.Completed;
+                await RunLedgerStore.UpsertRunLedgerAsync(
```
gautam-achieveai: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Correctness / Durability)

`CompleteRunAsync` publishes the terminal `RunCompletedMessage` before the terminal run-ledger status is durable. If `UpsertRunLedgerAsync` fails after this publish, subscribers can observe completion while `GET /status` still reports `InProgress`. In loops where `CompleteRunAsync` is inside the per-run `try`, that persistence exception can also be caught as a run failure and trigger a second `CompleteRunAsync(... isError: true)`, emitting conflicting completed/errored terminal outcomes for a successful run. The new REST status API uses the run ledger as its source of truth, so the terminal state needs to be committed before broadcasting terminal completion.

**Suggestion:** Move the terminal ledger update before `PublishToAllAsync(new RunCompletedMessage(...))` and before clearing `_currentRunId`, or otherwise separate terminal-persistence failures from agent-run failures so a successful run cannot emit completed then errored. Add a regression test with an `IRunLedgerStore` that fails the terminal upsert and assert no completion is published before durable status is written.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #151

# PR Review: LmStreaming.Sample: REST API for headless conversation control

## Summary
- Total files reviewed: 44
- Findings: 0 Critical, 4 High, 1 Medium, 0 Low
- Context questions: 0
- Test coverage: extensive, but needs targeted regression tests for the blocking contract/durability issues above
- Domain areas touched: LmMultiTurn persistence/run lifecycle, LmAgentInfra auth webhook, LmStreaming.Sample REST API + Vue client, tests
- Branch convention: no repo-specific `.code-reviewer.yml` found; skipped enforcement

## Strengths
The PR includes broad coverage for run-ledger round-trips/recovery, REST controller contracts, status resolution, auth target-capture behavior, and client deep-link handling. The target-capture design for auth terminal events is a good fit for the race described in #138.

## High Issues
1. Absolute auth signin URL required for forwarded webhooks — `AuthWebhookController.cs:188-193` / `AuthSigninUrls.cs:10`.
2. Auth webhook payload casing does not match documented external contract — `SandboxAuthWebhookForwarder.cs:200-201`, `:245-253`.
3. Completion is published before terminal ledger status is durable — `MultiTurnAgentBase.cs:1373-1400`.
4. Status final response can return user prompt/thinking instead of assistant answer — `ConversationStatusResolver.cs:119-120`.

## Medium Issues
1. Accepted-input records have no cleanup or documented retention policy — `MultiTurnAgentBase.cs:1292-1300`, `:1346-1349`.

## Security / EUII
No direct token/secret logging found in the reviewed diff. The auth webhook host-only logging is good; the blocking issues are contract/correctness rather than secret leakage.

## Verdict
**REQUEST CHANGES** — the blockers affect the public REST/webhook contract, headless auth usability, durable status correctness, and returned final-response correctness.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
gautam-achieveai: Fixed: signinUrl in the forwarded webhook payload is now built via `AuthSigninUrls.BuildAbsoluteSigninUrl`, combining `WebhookOptions.CallbackBaseUrl` with the provider path, so an external caller polling the forwarded payload gets a resolvable absolute URL. (47044a7)

samples/LmStreaming.Sample/Auth/SandboxAuthWebhookForwarder.cs
gautam-achieveai: Fixed: the forwarded payload now serializes with camelCase to match the wire contract. (47044a7)

samples/LmStreaming.Sample/Services/ConversationStatusResolver.cs
gautam-achieveai: Fixed: LoadFinalResponseAsync now prefers the run's last non-thinking assistant TextMessage as the response, and the tool-only-run fallback excludes Role.User so an errored run that never produced an answer resolves to a null response instead of echoing the user's own prompt back. Covered by two new tests in ConversationStatusResolverTests. (47044a7)

src/LmMultiTurn/MultiTurnAgentBase.cs
gautam-achieveai: Fixed: StartRunAsync/RecordInjectedInputsAsync now remove the pre-run accepted-input record once an input id is folded into a run's ledger InputIds, so it isn't retained indefinitely. (47044a7)

src/LmMultiTurn/MultiTurnAgentBase.cs
gautam-achieveai: Fixed: CompleteRunAsync now persists the terminal ledger status before publishing RunCompletedMessage, so a subscriber can never observe completion before GET /status would reflect it. Added a regression test that fails the terminal ledger upsert and asserts no RunCompletedMessage is ever published, and that the failure propagates out of the run loop. (47044a7)

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #151

# PR Review: LmStreaming.Sample: REST API for headless conversation control

## Summary
- Review type: re-review after commit `47044a7a8833a31c1bd8927bed4ec75355c1ce5e`
- Total files reviewed: 44 (delta since prior review: 9 files)
- Findings: 0 Critical, 0 High, 0 Medium, 0 Low
- Context questions: 0
- Verdict: **APPROVE**

## Re-review focus
Verified the prior blocking findings were addressed:
1. Forwarded auth webhook `signinUrl` is now absolute via `AuthSigninUrls.BuildAbsoluteSigninUrl(...)`.
2. Forwarded auth webhook payload uses explicit lower-camel-case `JsonPropertyName` attributes.
3. `ConversationStatusResolver` now prefers non-thinking assistant `TextMessage` responses and avoids echoing user prompts as final responses.
4. Accepted-input records are removed once folded into a run ledger entry.
5. Terminal run-ledger status is persisted before publishing `RunCompletedMessage`.

## Additional review passes
Re-checked run-ledger persistence, REST status/send/provision paths, auth webhook target capture/forwarding, JSON wire contracts, SQLite/file persistence, exception handling, performance-sensitive paths, EUII/logging, architecture, scope alignment, duplication, and test coverage. No remaining actionable findings were identified.

## Verdict
**APPROVE** — the previous blocking issues appear resolved and no new actionable issues were found in this re-review. I am posting this as a review summary only; I am not submitting a GitHub approval vote because approval requires explicit confirmation.

PR #152, opened 2026-07-05
Code-Review Daemon: skill + sub-agent enablement, cross-repo store reviews, grounded output

gautam-achieveai:
## What

Turns the Code-Review Daemon sample from a diff-only reviewer into a **tool-assisted, cross-repo, grounded** reviewer, using the gb-plugins `code-reviewer` skill and its real sub-agents inside the sandbox — while preserving the invariant that **the daemon, never the agent, owns all posting.**

## Highlights

- **Cross-repo store reviews.** When configured, the daemon checks out the reviewed repo as the `repos/<Repo>` submodule of the `AchieveAiReviews` store (shared `Contracts/` + siblings visible under an allow-list), moves it to the PR head, and reviews from the store root. Falls back to a single-repo `/workspace/target` checkout otherwise.
- **Grounded output.** Reviews now check out the PR **head** (the working tree was previously left on the clone's default branch), inject a `git ls-files` manifest, and steer the agent to Read files by exact path and scope `Grep`/`Glob` to a subdirectory. Output carries a structured summary (files examined + finding counts by severity) and a verdict.
- **Skill + sub-agents.** The review loop loads the `code-reviewer` skill and dispatches the discovered `code-reviewer:*` sub-agents, with a read-only MCP tool allow-list (`Read`/`Grep`/`Glob`/`Skill`) and degrade-to-diff-only when tooling is unavailable.
- **Per-run sandbox session** provisioning + host-dir cleanup with a disk guard; retention push + KB write moved host-side (off the read-only sandbox); confidentiality gate so fork/public PRs don't co-locate a sibling private submodule.

## Notes for reviewers

- The gateway's `agent-cli glob/grep` abort at a **standalone clone root** (a separate upstream bug — `agent-cli`, not this repo); reviewing the target as a **submodule** (gitlink working tree) sidesteps it, and Read-by-path + the manifest make grounding robust regardless.
- **Known follow-up:** the `code-reviewer:*` sub-agents dispatch but currently return empty output (likely concurrency on the single shared provider agent); the primary review compensates inline, so reviews stay grounded. Fixing it touches shared `LmMultiTurn`/provider infra and is deferred.

## Testing

- `dotnet test tests/CodeReviewDaemon.Sample.Tests` — 320 passing.
- Verified live (collect-only) against real PRs (#151, #147): store checkout, PR-head diff + manifest, sub-agent dispatch on exact store paths, and clean structured review output with real file:line citations.


gautam-achieveai:
_Code-Review Daemon self-review — tool-assisted, reviewed as a submodule of the AchieveAiReviews store (PR head checked out, code read by exact path). Collect-only run; posted here manually._

---

## Review: achieveai/LmDotnetTools PR #152 (bd765106)

Reviewed the core wire-contract change in `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs` plus its consumers (`WorkspaceSubAgentLoader.cs`, `DiscoveredSubAgentTemplateBuilder.cs`, `ContextDiscoveryController.cs`), the two large test files exercising the new discovery schema, `HostGitCredentialEnv.cs`, and the `.sln`/docs changes — 8 files examined in depth out of a very large diff (this PR also lands a 2100+ line planning doc and dozens of new sample-project files that were skimmed but not line-by-line reviewed). Findings: 0 Must, 2 Should, 1 Consider.

**Findings**

- **Should** — `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:591-622,1203-1205`: the gateway's discovery response envelope was renamed from `{"items": [...]}` to `{"session_id": ..., "discovered": [...]}` (`DiscoveredItemsResponse.Items` now binds to `"discovered"`). This app is a client of an independently-deployed sandbox gateway. If this build is ever run against an older gateway still emitting `items`, `ListDiscoveredAsync` will deserialize successfully but silently return `payload?.Items ?? []` — an empty list, no exception, no log line distinguishing "gateway has nothing to report" from "gateway speaks the old wire shape." Given `WorkspaceSubAgentLoader.LoadAsync` and `DiscoveredSubAgentTemplateBuilder.Build` both treat empty results as normal ("no sub-agents today"), this failure mode is indistinguishable from healthy operation and would silently disable all marketplace sub-agent discovery. Given the PR's own architecture doc treats gateway/app versions as independently deployable, worth at least a log line when the raw JSON contains `items` but not `discovered` (a strong signal of an old gateway), or a documented deploy-order requirement (gateway-first).

- **Should** — Same file/pattern: the new optional `Content`/`QualifiedName` fields are consumed correctly as nullable/optional on every call site found (`WorkspaceSubAgentLoader.cs:140,142-144`, `DiscoveredSubAgentTemplateBuilder.cs:33-35`) — both gracefully skip/fall back when absent, so *that* half of the change is backward compatible. However, the reverse direction isn't verified: `ContextDiscoveryController.cs:85,331` (`ToDiscovered()`) constructs a `DiscoveredItem` for the *inbound* webhook path — worth confirming (not done in this pass) that a same-process webhook payload from an older gateway version populating only `kind/name/description/path` still round-trips through `ToDiscovered()` without NREs, since `Content`/`QualifiedName` being null there feeds into the same consumer code paths noted above.

- **Consider** — `LmDotnetTools.sln`: the diff purely reorders `GlobalSection(ProjectConfigurationPlatforms)` entries (moves the `93637A84...` block earlier) in addition to adding the new `LmSampleShared` project. That reorder is pure churn with no functional difference; not harmful, just noise in the diff review surface — worth a mention only because large mixed-purpose diffs make future `git blame`/bisection on the `.sln` slightly noisier.

**Positive notes**: the shared parser/mapper extraction (`SubAgentMarkdownParser`/`SubAgentTemplateMapper` moved into `LmSampleShared`) is clean and DRY; the JSON schema tests (`SandboxSessionRegistryListDiscoveredTests.cs`) were updated in lockstep with the production shape change, which is good practice; `HostGitCredentialEnv` is a small, well-scoped, correct credential-header builder with no secret-logging issues.

**Verdict: Approve with comments**

PR #156, opened 2026-07-06
feat(daemon): pooled writable PR review workspace + knowledge base (Layer 1 + Layer 2)

gautam-achieveai:
## Summary

Adds two layers to the Code-Review Daemon (`samples/CodeReviewDaemon.Sample`) so it **persists review notes across a PR's life** and **learns durable knowledge over time**. Both layers ship together in this PR (base `main`).

### Layer 1 — pooled, persistent, writable review workspace + merge-on-close notes lifecycle
- Warm host-side `ReviewSlotPool` clones the `AchieveAiReviews` store once per slot; `ReviewSlotPreparer` sets up an origin-aware notes branch (`review/<provider>/<slug>/<pr>`) + submodule head + scratch wipe.
- The review agent's gateway session is **mounted over the leased slot** (`/workspace` == the slot); `ScopedToolFilter` grants `Write`/`Edit`/`Bash` scoped to the container notes/scratch dirs, so the reviewer can take durable notes while reviewing.
- Host-side `ReviewBranchManager.CommitNotesAsync` stages only `PRs/<pr>/…` and pushes the notes branch; `PrLifecycleSweeper` merges-on-close / deletes-on-abandon on the poller cadence.
- The daemon owns all writes/pushes with the write credential; the sandboxed review agent never holds it (design §4.7). Gated on `EnableReviewerWrites`; the read-only tool-assisted path keeps the per-run provisioner checkout.

### Layer 2 — knowledge extraction at PR-close + retrieval into reviews
- At PR-close (merged path) a **gated** extraction distills durable, generalizable knowledge from the PR's accumulated notes into layered `KnowledgeBase/<scope>/<slug>.md` entries with daemon-injected flow-style YAML frontmatter, plus a queryable `_index.jsonl` and a regenerated `_toc.md`. Most PRs gate to `NO_KNOWLEDGE` (nothing written — "not every PR contributes").
- `KnowledgeExtractionCommitter` checks the notes branch out, runs extraction, and — only when an entry was written — commits + pushes `KnowledgeBase/` onto the notes branch (bounded rebase-retry, every git exit code checked) so the sweeper's `MergeToDefaultAsync` carries it into the store's default branch.
- Reviews consult the KB: the review profile directs the agent to read `_toc.md` → grep relevant entries → flag contradictions, and the executor injects `_toc.md` into the review input.

## Security

Extraction input is attacker-influenceable (PR notes / agent text). SCOPE/UPDATES markers are treated as untrusted: per-segment ref-safe validation (`[A-Za-z0-9._-]`, never `.`/`..`/separators), `UPDATES` capped at two `.md` segments, and a canonicalize-and-prefix boundary check (with trailing separator) keep every write under `KnowledgeBase/`. Reviewed clean against `../` traversal, Windows drive/UNC/ADS, sibling-prefix, title-slug injection, and git arg-injection (argv vector, no shell). Bookkeeping targets (`_toc.md`/`_index.jsonl`) are refused.

## Testing

- **410 unit/integration tests green.**
- Path-traversal guard adversarially reviewed — clean against every exploit class tried.
- Carry-into-main verified over **real git**: replaying the hardened committer + `MergeToDefaultAsync` sequence lands the entry in a fresh clone of `main` (entry + `_index.jsonl` + `_toc.md` + frontmatter).
- Retrieval verified: the review agent's actual input carries the `## Prior knowledge (KnowledgeBase/_toc.md)` block.

## Deferred (tracked, non-blocking)

- **Layer 3** — the judge noting "what a review missed" + evolving the review prompts (separate, later).
- **Issue #160** — sweeper throttling (`ListReviewedPrsAsync` returns all reviewed PRs each poll); in-memory slot lease bound.
- **Layer-2 minors** — legacy flat entries in the index walk; a `rebase --abort` on conflict in the push-retry tail (matches the accepted `ReviewBranchManager` pattern; rare per-PR-branch case, degrades to "entry lost, logged").

## Docs

Specs/plans: `docs/superpowers/specs|plans/2026-07-05-daemon-pooled-review-workspace*` and `…/2026-07-06-daemon-knowledge-base*`.


gautam-achieveai:
Reviewed PR #156 across 17 files — 5 Must, 1 Should, 1 Consider.

## Findings

### Must — Pooled workspace is implemented but not wired into the daemon path
- **File:** `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:268`
- **Issue:** `ReviewSlotPool` and `ReviewSlotPreparer` are added, and the new pool options exist in `CodeReviewDaemonOptions`, but the production context stage still calls `EnsureCheckoutAsync(...)` against the fixed checkout roots (`/workspace/target` / `/workspace/store`) instead of leasing a slot and preparing it. In the searched `samples/CodeReviewDaemon.Sample` scope, `ReviewSlotPool` / `ReviewSlotPreparer` are only defined/tested, not consumed by the executor or DI.
- **Why it matters:** The PR’s headline behavior (“warm host-side pool of checkouts”) is inactive in production; the daemon continues using the old fixed checkout flow.
- **Suggested fix:** Register the pool/preparer in `Program.cs`, lease/return a slot around `FetchContextAsync`, and replace the fixed `EnsureCheckoutAsync` flow with the privileged prepare path. Ensure `ReturnAsync` runs in a `finally`.

### Must — `EnableReviewerWrites` silently has no production effect
- **Files:**
  - `samples/CodeReviewDaemon.Sample/Configuration/CodeReviewDaemonOptions.cs:191`
  - `samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs:107`
- **Issue:** The new `EnableReviewerWrites` / `WritableToolAllowList` options and `ScopedToolFilter` wrapper are added, but the live tool-assisted loop always calls `ReadOnlyToolFilter.Apply(...)`. `ReviewToolContext` also only carries `ReadOnlyToolAllowList` (`ReviewToolContext.cs:10-14`), so there is no path for notes/scratch dirs or writable allow-list into the factory.
- **Why it matters:** Operators can enable reviewer writes and see a supported-looking option, but the review agent remains read-only. Conversely, the scoped-write security wrapper is never exercised in production.
- **Suggested fix:** Extend `ReviewToolContext` with reviewer-write enablement, writable allow-list, notes dir, and scratch dir, then have `LiveReviewAgentLoopFactory` select `ScopedToolFilter.Apply(...)` when writes are enabled.

### Must — PR lifecycle cleanup is dead code
- **Files:**
  - `samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs:31`
  - `samples/CodeReviewDaemon.Sample/Program.cs:224`
  - `samples/CodeReviewDaemon.Sample/Orchestration/IPrProvider.cs:12`
- **Issue:** `PrLifecycleSweeper` is implemented, but `Program.cs` only registers the stage executor, orchestrator, and polling service; it does not register or schedule the sweeper. Also, `GetPrStateAsync` is only on `GitHubPrProvider` (`GitHubPrProvider.cs:128`), not the `IPrProvider` abstraction, so the sweeper cannot be wired provider-neutrally.
- **Why it matters:** Persistent review branches will be committed and kept, but never merged into default on PR merge or deleted on abandon.
- **Suggested fix:** Add the lifecycle query to `IPrProvider`, implement it for registered providers, register/schedule the sweeper as a hosted/background task, and source reviewed PR/branch records from `ReviewStore`.

### Must — `ReviewSlotPool.LeaseAsync` leaks capacity when setup fails
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs:71`
- **Issue:** `LeaseAsync` acquires the semaphore at line 71, then creates directories and invokes `_ensureStoreClonedAsync` at lines 73-80. If any of those steps throws, there is no rollback to release the semaphore or recycle the slot index.
- **Why it matters:** A transient clone/setup failure can permanently consume a pool permit. Repeated failures can exhaust the pool and stall future reviews until the daemon restarts.
- **Suggested fix:** Track the allocated index and wrap post-`WaitAsync` setup in `try/finally`; on failure, push the index back and release the semaphore.

### Must — Cancellation is swallowed and converted into fallback behavior
- **Files:**
  - `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:194`
  - `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:231`
- **Issue:** Both `BuildToolContextAsync` and `BuildSubAgentOptionsAsync` catch `Exception` and continue with degraded behavior. That also catches `OperationCanceledException` from session provisioning/discovery.
- **Why it matters:** During shutdown, timeout, or run cancellation, the daemon can continue into a diff-only/skill-only review instead of stopping promptly.
- **Suggested fix:** Exclude cancellation from the broad catches:
  ```csharp
  {
      ...
  }
  ```

### Should — Trust-signal fields are not persisted if they are meant to survive resume
- **Files:**
  - `samples/CodeReviewDaemon.Sample/Persistence/Models/ReviewRun.cs:48`
  - `samples/CodeReviewDaemon.Sample/Persistence/Models/ReviewRun.cs:56`
  - `samples/CodeReviewDaemon.Sample/Persistence/ReviewStore.cs:107`
  - `samples/CodeReviewDaemon.Sample/Persistence/Migrations/SchemaMigrations.cs:38`
- **Issue:** `IsForkPr` and `IsTargetRepoPublic` feed the confidentiality gate (`DaemonReviewStageExecutor.cs:569`), but the schema has no columns for them and `ReviewStore` insert/map logic does not persist or rehydrate them.
- **Why it matters:** If these values are intended to be populated by provider/poller state, a persisted/reloaded run loses the trusted same-org/private classification and falls back to fail-closed defaults.
- **Suggested fix:** If the trust signals are intended to be durable, add an append-only migration plus `ReviewStore` round-trip mapping/tests. If they are intentionally transient for now, document that limitation where the gate consumes them.

### Consider — Add config binding coverage for the new options
- **File:** `samples/CodeReviewDaemon.Sample/Configuration/CodeReviewDaemonOptions.cs:182`
- **Issue:** Defaults are covered, but the new operational knobs (`ReviewPoolSize`, `ReviewPoolHostRoot`, `ScratchDirName`, `EnableReviewerWrites`, `WritableToolAllowList`, `MergeNotesBranchOnClose`) are not covered by a configuration binding test.
- **Suggested fix:** Extend the existing options binding test to set these keys through `IConfiguration` and assert the bound values.

## Verdict

Request changes.

gautam-achieveai:
_No review content was produced._

samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs
```diff
@@ -0,0 +1,139 @@
...
+        await _gate.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+        var index = TakeIndex();
+        try
+        {
+            var slot = BuildSlot(index);
+            Directory.CreateDirectory(slot.HostPath);
+            Directory.CreateDirectory(slot.ScratchPath);
+
+            if (!Directory.Exists(slot.StorePath) || IsDirectoryEmpty(slot.StorePath))
```
gautam-achieveai, line 92: [gautam-achieveai's bot] [BLOCKER] **HIGH** (Correctness / Workspace lifecycle)

The rollback path now releases the semaphore and recycles the slot index, but it can still leave a poisoned warm store behind. If `_ensureStoreClonedAsync` creates files/directories under `StorePath` and then throws or is cancelled, the next lease reuses the same index and this non-empty-directory check treats the partial store as warm, skipping clone/setup entirely.

That means a transient clone/setup failure can hand later reviews a corrupt or incomplete checkout.

**Suggestion:** On setup failure, remove/reset `slot.StorePath` before recycling the index, or replace the “non-empty” heuristic with a positive readiness invariant (for example, valid git repo/worktree plus a completion marker). Please also add a regression test where the first clone writes partial contents then throws and the second lease must not reuse that partial store as warm.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -191,7 +191,7 @@ public DaemonReviewStageExecutor(
                 ReadOnlyToolAllowList: _options.ReadOnlyToolAllowList,
                 SubAgentOptions: await BuildSubAgentOptionsAsync(run, session.SessionId, cancellationToken).ConfigureAwait(false));
         }
-        catch (Exception ex)
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
gautam-achieveai, line 228: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Test Coverage)

The catch filters now exclude `OperationCanceledException`, which fixes the implementation path, but the changed tests only cover ordinary exceptions degrading to diff-only / skill-only. I was unable to find a regression test proving cancellation from session provisioning or sub-agent discovery propagates instead of being converted into fallback behavior.

Because this was one of the prior Must-fix issues, the behavior needs a durable test guard before the thread can be closed.

**Suggestion:** Add tests such as `Reviewed_ProvisionerThrowsOperationCanceledException_PropagatesCancellation` and `Reviewed_DiscoveryThrowsOperationCanceledException_PropagatesCancellation`.

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #156

Reviewed the delta from `7f6d7d7375961ad9de6ff271493d1e690ec6d2e5` to `4ffe9625b92358af41b70dd2ac3df6181e2516e1` (3 files changed). CI is green, and two prior issues were partially addressed: the semaphore permit/index leak path now rolls back, and cancellation is no longer caught by the two broad catch blocks. However, the prior integration blockers are still active, and the new rollback path has one correctness gap.

### Previous Issues Resolution Status
| # | Issue | Severity | Resolution |
|---|---|---|---|
| 1 | Pooled workspace implemented but not wired into the daemon path | HIGH | **ACTIVE** — source search still finds `ReviewSlotPool` / `ReviewSlotPreparer` defined and tested but not consumed by `DaemonReviewStageExecutor`/DI; `FetchContextAsync` still calls `EnsureCheckoutAsync(...)` against fixed roots. |
| 2 | `EnableReviewerWrites` silently has no production effect | HIGH | **ACTIVE** — `LiveReviewAgentLoopFactory` still calls `ReadOnlyToolFilter.Apply(...)`, and `ReviewToolContext` still only carries `ReadOnlyToolAllowList`. |
| 3 | PR lifecycle cleanup is dead code | HIGH | **ACTIVE** — `PrLifecycleSweeper` is still not registered/scheduled in `Program.cs`, and lifecycle lookup is still not on `IPrProvider`. |
| 4 | `ReviewSlotPool.LeaseAsync` leaks capacity when setup fails | HIGH | **PARTIALLY RESOLVED** — the permit/index are now released, but see new HIGH issue below: partial store contents can be reused as warm state. |
| 5 | Cancellation is swallowed and converted into fallback behavior | HIGH | **IMPLEMENTATION FIXED / TEST GAP** — catch filters now exclude `OperationCanceledException`; add regression tests below to lock this in. |
| 6 | Trust-signal fields are not persisted if meant to survive resume | MEDIUM | **ACTIVE / DEFERRED** — unchanged in this delta. |
| 7 | Add config binding coverage for new options | LOW | **ACTIVE** — unchanged in this delta. |

### New / Remaining Issues Found

#### HIGH [BLOCKER] — Partial store contents can be reused as a warm checkout after setup failure
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs:80` / `:92`
- **Issue:** The new rollback path releases the semaphore and recycles the slot index, but it leaves any partially-created `StorePath` contents behind. If `_ensureStoreClonedAsync` creates files/directories and then throws or is cancelled, the next lease reuses the same index and the line 80 `Directory.Exists && !IsDirectoryEmpty` check treats that partial directory as a warm store, skipping clone entirely.
- **Why it matters:** A transient clone/setup failure can poison a pooled slot and later reviews may run against a corrupt/incomplete checkout.
- **Suggested fix:** On setup failure, remove/reset `slot.StorePath` before recycling the index, or use a positive readiness invariant (valid git repo/worktree plus a completion marker) before considering a store reusable. Add a regression test where the first clone writes partial contents then throws and the second lease must not reuse it as warm.

#### MEDIUM [BLOCKING IN RE-REVIEW CONTEXT] — Cancellation propagation fix lacks regression coverage
- **File:** `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:194`, `:231`
- **Issue:** The implementation now excludes `OperationCanceledException`, but the test delta only covers ordinary exceptions degrading to diff-only/skill-only. I was unable to find tests proving cancellation from the provisioner or discovery source propagates instead of degrading.
- **Suggested fix:** Add tests such as `Reviewed_ProvisionerThrowsOperationCanceledException_PropagatesCancellation` and `Reviewed_DiscoveryThrowsOperationCanceledException_PropagatesCancellation`.

### Verdict
**REQUEST CHANGES**

The re-review verdict remains **REQUEST CHANGES** because three prior HIGH integration blockers are still active, and the new pool rollback fix still has a HIGH correctness gap around partial store reuse.

gautam-achieveai:
Addressed the review in `98d16213`:

- **A (HIGH, partial-store reuse) — fixed.** `ReviewSlot.LeaseAsync` now wipes a partially-created `StorePath` (clearing read-only attrs first, mirroring `ReviewSessionProvisioner.ClearReadOnly`) before recycling the index on setup failure, so a later lease re-clones instead of treating the partial dir as a warm store. Regression test: a clone callback that writes a partial file then throws must have the SECOND lease re-invoke the clone (`LeaseAsync_WhenCloneWritesPartialStoreThenThrows_SecondLeaseReclonesInsteadOfReusingPartialStore`).
- **B (MEDIUM, cancellation coverage) — fixed.** Added `Reviewed_ProvisionerThrowsOperationCanceled_Propagates` and `Reviewed_DiscoveryThrowsOperationCanceled_Propagates`: an `OperationCanceledException` from the provisioner / sub-agent discovery propagates out of the Reviewed stage (not swallowed into a diff-only / skill-only degrade), while an ordinary exception still degrades (existing tests still green).
- **C (LOW, config binding) — fixed.** Added a binding test for `ReviewPoolSize`, `ReviewPoolHostRoot`, `ScratchDirName`, `EnableReviewerWrites`, `WritableToolAllowList`, `MergeNotesBranchOnClose`.

Build + test green: `dotnet test tests/CodeReviewDaemon.Sample.Tests/CodeReviewDaemon.Sample.Tests.csproj -c Debug` → 360 passed, 0 failed.

Deferred to #160: Task 9 wiring (lease/prepare a `ReviewSlot` in `FetchContextAsync` + register pool/preparer; select `ScopedToolFilter` + thread notes/scratch dirs when `EnableReviewerWrites`; register/schedule `PrLifecycleSweeper` + add the lifecycle lookup to `IPrProvider`) and the `IsForkPr`/`IsTargetRepoPublic` trust-field persistence.

Note (out of scope, flagged): the config binder APPENDS bound items onto the non-empty default `WritableToolAllowList` rather than replacing it, so an operator config adds to `["Write","Edit","Bash"]` instead of overriding it. Worth deciding intended behavior when Task 9 wires the writable path.

Kept as a DRAFT.

gautam-achieveai:
Reviewed PR #156 across 17 files — 6 Must, 1 Should, 0 Consider.

## Findings

### Must — CommitNotes stages the whole store instead of enforcing the PR-notes commit gate
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs:128-136`
- **Also:** `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:764-765`

`CommitNotesAsync` writes the requested artifact files, then runs `git add -A`. That stages every change in the review store checkout, not just `PRs/<provider>/<slug>/<pr>/...`. The executor comment confirms this is intentional for `KnowledgeBase/...` side effects, but the pooled-workspace design’s safety boundary depends on committing only the PR notes directory.

In the new model, the checkout may contain a submodule pointer advanced to the PR head and, once reviewer writes are enabled, any local changes made by the agent via Bash. `git add -A` can persist changes outside the intended notes directory.

Stage only the validated PR notes paths, for example:
```csharp
foreach (var file in request.Files)
{
    ValidateUnderPrNotesDir(file.RelativePath, expectedNotesPrefix);
    await _fileSystem.WriteFileAsync(JoinPath(repoRoot, file.RelativePath), file.Content, cancellationToken);
    await RunGitAsync(["add", "--", file.RelativePath], repoRoot, cancellationToken);
}
```
Leave `KnowledgeBase` persistence to the later layer that owns it.

---

### Must — `EnableReviewerWrites` is inert in the real tool path
- **File:** `samples/CodeReviewDaemon.Sample/Configuration/CodeReviewDaemonOptions.cs:191-196`
- **Also:** `samples/CodeReviewDaemon.Sample/Agents/ScopedToolFilter.cs:20-59`, `samples/CodeReviewDaemon.Sample/Agents/ReviewToolContext.cs:10-14`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:188-192`, `samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs:106-108`

The PR adds `EnableReviewerWrites`, `WritableToolAllowList`, and `ScopedToolFilter`, but the production review loop still only passes `ReadOnlyToolAllowList` through `ReviewToolContext` and always applies `ReadOnlyToolFilter`. There is no runtime path carrying `notesDir`, `scratchDir`, or writable allow-list into `LiveReviewAgentLoopFactory`.

As a result, setting `EnableReviewerWrites = true` does not actually expose scoped `Write`/`Edit`/`Bash` to the reviewer.

Thread the write capability through the real context:
- add writable roots and allow-list to `ReviewToolContext`
- populate them in `DaemonReviewStageExecutor` when writes are enabled
- call `ScopedToolFilter.Apply(...)` in `LiveReviewAgentLoopFactory` for that case
- keep `ReadOnlyToolFilter` for the default read-only path

---

### Must — The pooled slot/preparer implementation is not wired into daemon execution
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs:32-90`
- **Also:** `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlotPreparer.cs:62-142`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:117-121`, `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:316-349`, `samples/CodeReviewDaemon.Sample/Program.cs:224-235`

`ReviewSlotPool` and `ReviewSlotPreparer` are implemented and tested, but `DaemonReviewStageExecutor` still uses the existing checkout flow through `ResolveSandboxAsync` / `EnsureCheckoutAsync`, cloning/reusing `/workspace/store` or `/workspace/target` directly. The executor constructor has no pool/preparer dependency, and `Program.cs` registers the existing stage executor/poller without composing the new pool.

That means the main “pooled workspace” feature is production-inert: tests can pass while the daemon continues to use the old per-run checkout path.

Wire the pool into the actual `ContextReady` lifecycle: lease a slot, prepare it via `ReviewSlotPreparer`, pass its prepared paths into the review context/session, and return the lease on terminal cleanup. Prefer an `IAsyncDisposable` lease handle so return is structurally enforced.

---

### Must — PR lifecycle sweeping is not connected to the provider abstraction or hosted daemon flow
- **File:** `samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs:41-48`
- **Also:** `samples/CodeReviewDaemon.Sample/Orchestration/IPrProvider.cs:12-21`, `samples/CodeReviewDaemon.Sample/Orchestration/GitHubPrProvider.cs:128-150`, `samples/CodeReviewDaemon.Sample/Program.cs:230-235`

`GitHubPrProvider` adds `GetPrStateAsync`, but `IPrProvider` still only exposes `ListOpenPullRequestsAsync`. `PrLifecycleSweeper` therefore depends on ad-hoc delegates rather than a provider-level lifecycle capability, and `Program.cs` only registers `PrPollingService`; there is no scheduled sweeper in the running daemon.

Persistent PR branches will be pushed and kept, but the daemon has no wired path to merge them on PR merge or delete them on abandon.

Add a provider-neutral lifecycle seam, e.g. `IPrLifecycleProvider.GetPrStateAsync(...)`, implement/register it for supported providers, connect it to `ReviewStore`, and run the sweeper on the poller cadence or equivalent hosted lifecycle.

---

### Must — Branch deletion failures are silently reported as successful
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs:222-229`
- **Also:** `samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs:101-107`

`DeleteBranchAsync` calls both local and remote deletion with `allowFailure: true` and never inspects either result. That makes missing branches idempotent, but it also swallows real failures such as auth denial, network failure, protected refs, or local git lock errors. The sweeper then logs successful deletion after the method returns.

Only tolerate recognized “branch/ref not found” failures. For other failures, throw or return a failure result so the sweeper can log accurately and retry later.

---

### Must — `ReviewSlotPool.ReturnAsync` can corrupt the pool on duplicate return
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs:117-127`

`ReturnAsync` unconditionally pushes the slot index back to `_freeIndexes` and releases the semaphore. A double return can add the same slot index twice and/or trigger `SemaphoreFullException`. If duplicate free-list entries survive, the same workspace can be leased to more than one review.

Track leased/free state per slot and make return idempotent-safe, or return an `IAsyncDisposable` lease object that performs exactly-once return.

---

### Should — Push retry ignores failed `pull --rebase`
- **File:** `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs:258`
- **Also:** `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs:286`

`TryPushWithRebaseAsync` retries a rejected push by running `pull --rebase`, but the result is ignored. `RunGitAsync` also explicitly avoids throwing for `pull`, so a rebase conflict or network failure is silently skipped and the loop continues from an unreconciled working tree.

Capture the pull result and stop retrying with a clear failure when the rebase fails.

## Verdict

Request changes.

gautam-achieveai:
Reviewed PR #156 across 17 files — 4 Must, 1 Should, 0 Consider.

## Findings

### Must — Pooled reviews prepare one workspace but give the agent a different sandbox workspace

- `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:384`
- `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:425`
- `samples/CodeReviewDaemon.Sample/Orchestration/ReviewSessionProvisioner.cs:100`

`TryPooledFetchContextAsync` leases and prepares a host-side slot, then records agent-visible paths such as `/workspace/store/...` and `/workspace/scratch`:

```csharp
var slot = await _slotWorkspace.Pool.LeaseAsync(cancellationToken).ConfigureAwait(false);
...
var targetDirSandbox = PosixJoin(StoreRoot, submoduleRelPath);
var notesDirSandbox = PosixJoin(StoreRoot, notesRelPath);
var scratchDirSandbox = $"{SandboxWorkspaceRoot}/{_options.ScratchDirName}";
```

But the tool context still gets its sandbox session from the normal per-run provisioner:

```csharp
new WorkspaceRef(workspaceId, DirectoryRelPath: workspaceId, Marketplaces: _options.Marketplaces)
```

That provisioner mounts `review-run-{id}`, not the leased pool slot. So the review agent is told to read/write `/workspace/store/...` paths that are not guaranteed to be the prepared checkout. This breaks the core pooled-review path: host git diffs one checkout, while Read/Grep/Glob/Write/Edit/Bash can run in a different workspace.

Please make the slot-to-session relationship explicit before enabling scoped writes, e.g. provision the sandbox session from the leased `ReviewSlot` host path, or add a dedicated provisioner method that accepts the slot and returns a session mounted over that slot.

### Must — `ReviewSlotPool.ReturnAsync` is not idempotent-safe and can corrupt the pool

- `samples/CodeReviewDaemon.Sample/Workspace/ReviewSlot.cs:129`

`ReturnAsync` unconditionally pushes the index and releases the semaphore:

```csharp
lock (_freeIndexesLock)
{
    _freeIndexes.Push(slot.Index);
}

_gate.Release();
```

A duplicate or unknown return mutates `_freeIndexes` before any validation. If `SemaphoreSlim.Release()` then throws because the semaphore is already full, the duplicate index has still been pushed. That can later hand out the same slot twice or otherwise corrupt slot ownership.

The executor currently guards one path with `_leasedReviews.TryRemove(...)`, but the pool itself is shared infrastructure and should enforce its own lease invariants. Please either return an owned `IAsyncDisposable` lease that can only be returned once, or track checked-out/free indices inside `ReviewSlotPool` and reject duplicate/unknown returns before mutating the free list.

### Must — Abandoned-PR branch deletion swallows real remote failures and reports success

- `samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs:268`

`DeleteBranchAsync` treats both local and remote delete failures as allowed:

```csharp
await RunGitAsync(["branch", "-D", branch], repoRoot, cancellationToken, allowFailure: true);

await RunGitAsync(
        ["push", "origin", "--delete", branch],
        repoRoot,
        cancellationToken,
        allowFailure: true)
    .ConfigureAwait(false);
```

Idempotently tolerating “branch already gone” is correct, but this also suppresses auth failures, network failures, permission errors, and remote rejections. The sweeper then logs that the branch was deleted, even though the remote branch may still exist and may not be retried as an error.

Please distinguish “not found” from other failures. For example, tolerate a missing branch, but throw on other delete failures so `PrLifecycleSweeper` logs the per-PR failure and retries on the next sweep.

### Must — Reviewer write mode grants unwrapped Bash by default without a daemon-validated boundary

- `samples/CodeReviewDaemon.Sample/Configuration/CodeReviewDaemonOptions.cs:191`
- `samples/CodeReviewDaemon.Sample/Agents/ScopedToolFilter.cs:54`

When reviewer writes are enabled, the default writable tools include Bash:

```csharp
public IReadOnlyList<string> WritableToolAllowList { get; init; } = ["Write", "Edit", "Bash"];
```

`ScopedToolFilter` only wraps `Write` and `Edit`; every other writable tool, including `Bash`, is passed through directly:

```csharp
var registeredHandler = PathScopedToolNames.Contains(contract.Name)
    ? WrapPathScoped(handler, notesDir, scratchDir)
    : handler;
```

The comments say Bash is bounded elsewhere, but the daemon does not encode or verify that policy before exposing it. That makes `EnableReviewerWrites` a much broader capability than the name implies: it enables arbitrary shell execution by default, relying on external gateway/OS/egress assumptions that are not represented in this code.

Please make Bash a separate explicit opt-in capability, or require a verified sandbox policy proof before registering it. The safer default is to remove `"Bash"` from `WritableToolAllowList` and keep scoped writes limited to path-wrapped `Write`/`Edit`.

### Should — Slot return in `finally` can mask the original failure

- `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:460`
- `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:961`

Both pooled cleanup paths await `ReturnAsync` in `finally` without preserving an in-flight exception:

```csharp
finally
{
    if (!handedOff)
    {
        await _slotWorkspace.Pool.ReturnAsync(slot, CancellationToken.None).ConfigureAwait(false);
    }
}
```

and:

```csharp
finally
{
    await _slotWorkspace.Pool.ReturnAsync(lease.Slot, CancellationToken.None).ConfigureAwait(false);
}
```

If preparation/diff/commit fails and returning the slot also fails, the cleanup exception replaces the root cause. That makes retry/debugging much harder, especially around host-side checkout and retention failures.

Please catch/log cleanup failures when there is already a body failure, preserving the original exception. Cleanup failure can still throw when it is the only failure.

## Verdict

Request changes.

gautam-achieveai (commented):
[Gautam's review bot] ## Re-Review Summary: PR #156

Reviewed the delta from `f7a215e1` to `4683ed7b` (12 commits adding Layer-2 Knowledge Base) plus re-assessed all 4 unresolved prior findings.

**Review agents used:** `pool-idempotency-review`, `bash-security-review`, `knowledge-agent-review`, `branch-deletion-review` (all explore-type).

---

### What Was Fixed Since Last Review :white_check_mark:
1. **Slot-to-session mismatch** (`f7a215e1`) — `GetOrCreateForSlotAsync` mounts the session over the leased slot. Core pooled-review path is correctly wired.
2. **Git exit codes in KB committer** (`4683ed7b`) — fetch/checkout/add/commit now checked; push uses bounded retry.

---

### Findings

#### A (MEDIUM, new) — `pull --rebase` exit code discarded in both push-retry loops
- `KnowledgeExtractionCommitter.cs:158`: `_ = await _git.RunAsync(["pull", "--rebase"…])`
- `ReviewBranchManager.cs:304`: `RunGitAsync` explicitly exempts `pull` from error checking (line 332)

If the rebase fails (conflicts, force-pushed remote), the code proceeds to push again — which will also fail — burning retry budget. Worst case: repo left in `rebase-merge/` state; next git command in that worktree fails too. The KB committer's outer catch prevents a crash, but the entry is silently lost without an actionable log.

**Suggestion:** Check the rebase result; if it fails, `break` out of the retry loop and log the rebase stderr.

#### B (LOW, repeat) — `ReturnAsync` not idempotent-safe at the pool level
Acknowledged as a known design gap. Executor's `TryRemove` guard is sufficient for current callers — not blocking.

#### C (LOW, repeat) — `DeleteBranchAsync` blanket `allowFailure:true` + false-success log
Sweeper logs success unconditionally (`PrLifecycleSweeper.cs:163-166`) regardless of whether the remote delete worked. Cosmetic/operational — not blocking.

#### D (LOW, repeat) — Bash in default `WritableToolAllowList`
Behind `EnableReviewerWrites = false` + gateway sandbox isolation. Not blocking given layered defenses.

#### E (LOW, repeat) — Finally-masking on slot return
`ReturnAsync` in `finally` (`DaemonReviewStageExecutor.cs:971`) can replace root-cause exception. Not blocking in practice.

---

### Positive Observations
- **Path traversal defense** in KnowledgeAgent is excellent: 3-layer validation (`IsSafeSegment` → `NormalizeUpdatesRelPath` → `StaysUnderKnowledgeBase`) + bookkeeping file guard. Well tested.
- **Gated extraction** (`NO_KNOWLEDGE` sentinel) prevents KB pollution cleanly.
- **Index regeneration determinism** (stable key order, sorted by file) prevents spurious diffs.
- **Commit gate** (`stagePaths` scoping) ensures only the PR notes dir reaches the remote.

---

### Verdict
The critical issues from prior reviews (slot/session mismatch, partial-store reuse) are resolved. **Finding A** is the only new actionable item — a correctness gap in the retry logic. The remaining findings (B–E) are low-severity, previously raised, and non-blocking.

**Approve with comments** — remaining issues can be addressed in follow-up.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeExtractionCommitter.cs
```diff
@@ -0,0 +1,164 @@
...
+                return true;
+            }
+
+            if (attempt == MaxPushAttempts)
+            {
+                break;
+            }
+
+            // The remote moved; rebase our commit on top and retry.
+            _ = await _git.RunAsync(["pull", "--rebase", "origin", branch], _repoRoot, cancellationToken)
```
gautam-achieveai: [Gautam's review bot] **MEDIUM** (Correctness / Silent data loss)

The `pull --rebase` result is discarded (`_ = …`). If the rebase fails (conflicts, force-push, disk full), the code still proceeds to the next push attempt, which will also fail. After exhausting retries, the outer method returns `false` and the entry is logged as "lost" — but the *reason* (a rebase conflict vs. a persistent rejection) is never surfaced.

Worse: a failed rebase leaves the worktree in `rebase-merge/` state. If this slot/worktree is reused by a subsequent sweep, the next `git` command may fail unexpectedly.

**Suggestion:**
```csharp
var rebased = await _git.RunAsync(["pull", "--rebase", "origin", branch], _repoRoot, cancellationToken)
    .ConfigureAwait(false);
if (!rebased.Succeeded)
{
    _logger.LogWarning(
        "Rebase during KB push retry failed ({Stderr}); aborting.",
        rebased.Stderr);
    break;
}
```

samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs
```diff
@@ -0,0 +1,382 @@
...
+                return true;
+            }
+
+            if (attempt == MaxPushAttempts)
+            {
+                break;
+            }
+
+            // The remote moved; rebase our commit(s) on top and retry.
+            await RunGitAsync(["pull", "--rebase", "origin", branch], repoRoot, cancellationToken)
```
gautam-achieveai: [Gautam's review bot] **LOW** (Consistency)

Same pattern as `KnowledgeExtractionCommitter.cs:158` — the `pull --rebase` result is not checked. `RunGitAsync` explicitly exempts `pull` from its throw-on-failure path (line 332: `verb is not ("push" or "pull")`), so a rebase conflict silently proceeds to the next push attempt.

Since this is the `CommitNotesAsync` push path (not the KB extractor), a failed rebase here means review notes may be silently dropped during a concurrent push race. The risk is lower (notes are also in SQLite), but the worktree-corruption concern applies equally.

gautam-achieveai:
## Re-review addressed — Finding A fixed (`96d758fc`)

**A (MEDIUM, new) — `pull --rebase` exit code discarded in both push-retry loops** — ✅ **Fixed.** Both `KnowledgeExtractionCommitter.TryPushWithRebaseAsync` and `ReviewBranchManager.TryPushWithRebaseAsync` now capture the rebase result. On a failed rebase they run `git rebase --abort` (restoring a clean worktree so the reused store checkout is never left in `rebase-merge/`), log the rebase stderr, and `break` out of the retry loop — so the outcome is reported as a **sync failure, never a false success**, and the retry budget isn't burned on a doomed push. RED→GREEN in both test classes (`*_aborts_the_rebase_and_stops_retrying_when_the_rebase_fails`); full suite **412 green**.

**B–E (LOW, repeat)** — acknowledged, non-blocking, left as tracked follow-ups:
- **B — `ReturnAsync` pool idempotency:** known design gap; the executor's `TryRemove` guard covers the current callers. Tracked in #160.
- **C — `DeleteBranchAsync` blanket `allowFailure` + unconditional success log:** cosmetic/operational; folding into the #160 sweep-hardening.
- **D — Bash in the default writable allow-list:** accepted residual risk — gated behind `EnableReviewerWrites` + gateway sandbox isolation (the egress-deny/residual-risk decision made earlier).
- **E — finally-masking on slot return:** not blocking in practice; noted for the #160 cleanup.

Thanks for the review — the positive notes on the path-traversal defense and gated extraction match the adversarial pass we ran too.

gautam-achieveai (commented):
[Gautam's review bot] **APPROVE**

Finding A (MEDIUM — `pull --rebase` exit code discarded) is properly fixed in `96d758fc`:

- Both `KnowledgeExtractionCommitter` and `ReviewBranchManager` now capture the rebase result, check `.Succeeded`, run `rebase --abort` on failure (restoring a clean worktree for reused stores), log the rebase stderr, and `break` out of the retry loop.
- RED->GREEN regression tests in both test classes confirm: only 1 push attempt after a failed rebase, `rebase --abort` is called, no false success is reported.
- The fix is consistent across both files, each using the idiomatic error-handling pattern for its git runner abstraction.

**Remaining items (all LOW, tracked in !160):**
- B — `ReturnAsync` pool idempotency
- C — `DeleteBranchAsync` blanket `allowFailure`
- D — Bash in default `WritableToolAllowList`
- E — finally-masking on slot return

All are acknowledged by the author and non-blocking given the current callers and layered defenses. No new issues found in this delta.

gautam-achieveai (commented):
[Gautam's review bot] Re-review (iteration 4) -- incremental delta: 96d758fc...71c05ed0

## Changes Reviewed
1. **Scriban 7.2.0 -> 7.2.2** -- Patches security advisories (GHSA-6q7j-xr26-3h2c, GHSA-q6rr-fm2g-g5x8)
2. **Removed NU1902 WarningsNotAsErrors exclusion** -- No longer needed now that patched Scriban resolves the NuGet audit advisory

## Assessment
Restore succeeds cleanly -- no NU1902 warnings remain. No functional code changes -- purely a dependency version bump + build config cleanup. The original comment documented when to remove the exclusion ('when a fix ships') -- this commit does exactly that.

## Verdict
**APPROVE** -- Clean security dependency fix. No new issues. Prior LOW findings tracked in !160.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -221,18 +226,188 @@
...
+            // The committer wraps the gated extraction with the git plumbing that carries its write into the
+            // default branch: check the notes branch out, run extraction, and — only when it wrote an entry —
+            // commit + push KnowledgeBase/ onto that branch so MergeToDefaultAsync fast-forwards it into main.
+            var committer = new KnowledgeExtractionCommitter(
+                hostGit, sweeperRepoRoot, loggerFactory.CreateLogger<KnowledgeExtractionCommitter>());
+            extractKnowledgeAsync = (pr, ct) =>
+            {
+                // sourcePrRef is a stable, human-readable id for the source PR; todayUtc is daemon-supplied
+                // (deterministic — never the model) and stamped into the entry's `updated` frontmatter.
+                var sourcePrRef = $"{pr.Provider}/{pr.Repo.NormalizedKey}/{pr.PrId}";
```
gautam-achieveai, line 310: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Correctness / Durable data)

`sourcePrRef` doubles the provider. `RepoIdentity.NormalizedKey` already begins with the provider (`provider/owner[/project]/repo`), so prefixing `pr.Provider` again produces a malformed, duplicated ref:

- **GitHub:** `github/github/achieveai/lmdotnettools/42`
- **Azure DevOps (worse):** `ado/azure-devops/org/proj/repo/123` — two *different* provider tokens (`pr.Provider` maps to `ado`, but `NormalizedKey` retains `azure-devops`).

This isn't a transient log line — `sourcePrRef` is stamped into every entry's durable `sourcePrs:` frontmatter **and** the commit message, on 100% of the happy path. Once entries ship, fixing it needs a data migration. The existing tests don't catch it because `KnowledgeAgentTests` passes a hand-authored `SourcePr` constant, bypassing this exact construction seam in `Program.cs`.

**Suggestion:** Use `pr.Repo.NormalizedKey` alone — it's already provider-qualified and case-folded:
```csharp
var sourcePrRef = $"{pr.Repo.NormalizedKey}/{pr.PrId}";
```
And add a test that exercises the real construction (or extract a small `BuildSourcePrRef(pr)` helper) so the format is pinned for both providers.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
```diff
@@ -29,91 +34,493 @@ public KnowledgeAgent(IMultiTurnAgent agent, ISandboxFileSystem fileSystem, ILog
...
+
+        var i = 0;
+        while (i < lines.Length && TryParseMarker(lines[i].Trim()) is null)
+        {
+            i++; // Skip preamble before the header starts.
+        }
+
+        while (i < lines.Length)
+        {
+            var marker = TryParseMarker(lines[i].Trim());
```
gautam-achieveai, line 350: [gautam-achieveai's bot] [BLOCKER] **MEDIUM** (Silent data loss)

`ParseEntry` consumes only a **contiguous** run of marker lines — the first non-marker line (including a **blank line**) ends the header, and everything after becomes body:

```csharp
while (i < lines.Length)
{
    var marker = TryParseMarker(lines[i].Trim());
    if (marker is null)
    {
        break; // Header ended; the rest — including any heading-shaped lines — is body.
    }
    ...
}
```

The contiguity rule is deliberate and correctly protects a body line shaped like `## TAGS: a, b` (and that case is tested). But it is fragile against a very common LLM formatting habit — separating `##` headings with blank lines:

```
## SCOPE: system

## TITLE: Null checks on X
```

The blank line breaks the run at `TITLE`, so `title` stays empty → `ResolveTargetRelPathAsync` returns `null` → **the entire extraction is silently discarded**, even though the model *did* produce durable knowledge. If `## UPDATES` sits after the blank, you instead get an update with an empty title/scope. Because the input is model-generated, this is a likely real-world failure mode, and every test tightly-packs markers so the path is unverified.

**Suggestion:** Tolerate blank lines *within* the header — skip empty lines without ending the header, and still end it on the first non-empty, non-marker line. That preserves the body-example protection while surviving inter-marker whitespace. Add a regression test with a blank line between two markers asserting the entry is still written.

gautam-achieveai (commented):
[Gautam's review bot] APPROVE (Iteration 5 - incremental review of 71c05ed0..c806704f)

Changes reviewed:
- Added 'superpowers' to default Marketplaces list in CodeReviewDaemonOptions
- Updated XML doc to reflect new default
- Test assertion updated to validate both marketplace entries

Assessment: Trivial additive config change. Test correctly validates the new default. No issues found.

Prior LOW findings remain tracked in !160.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/KnowledgeAgentTests.cs
```diff
@@ -24,90 +24,269 @@ public KnowledgeAgentTests(ITestOutputHelper output)
...
+        fs.Files[KbDir + "/_index.jsonl"]
+            .Split('\n', StringSplitOptions.RemoveEmptyEntries)
+            .Should().ContainSingle();
     }
 
+    // ---- Fix 1: path-traversal hardening on SCOPE / UPDATES -----------------------------------------
+
     [Fact]
-    public async Task WriteEntryAsync_regenerates_the_toc_with_a_link_to_the_entry()
+    public async Task TryExtractAsync_refuses_a_traversal_scope_and_writes_nothing_outside_the_KB()
```
gautam-achieveai, line 135: [gautam-achieveai's bot] **MEDIUM** (Test Coverage / Security regression trap)

The path-traversal test suite here covers `../` traversal (scope + UPDATES) and separator-in-scope and the bookkeeping-target refusal — good. But it **omits the exploit classes explicitly enumerated in the PR description**: absolute paths (`/etc/passwd`), Windows drive-letters (`C:\…`), UNC (`\\server\share`), NTFS ADS (`file.md:stream`), and the sibling-prefix escape (`KnowledgeBase-evil/`).

The production guard (`IsSafeSegment` allowlist `[A-Za-z0-9._-]` + `StaysUnderKnowledgeBase` with a trailing-separator prefix check) *does* defeat all of these today — I verified it adversarially. The risk is that the behavior is **unpinned**: a future refactor to a denylist, or a change to the separator handling, would silently re-open drive-letter/ADS/UNC/sibling-prefix escapes while every existing test stays green. For a security boundary on attacker-influenceable input, the exploit classes should be locked down by tests.

**Suggestion:** Add a `[Theory]` with `InlineData` SCOPE/UPDATES cases for `/etc/passwd`, `C:\Windows\x`, `\\srv\share\x.md`, `system/x.md:stream`, and a sibling-prefix scenario, each asserting the result is refused / all writes stay under `KnowledgeBase/`.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -221,18 +226,188 @@
...
+            loggerFactory.CreateLogger<PrLifecycleSweeper>(),
+            extractKnowledgeAsync);
+    });
+
+    // Reads a merged PR's accumulated review notes off its persistent notes branch (they live on
+    // origin/<branch>, not yet on the sweeper checkout's default branch) and assembles them as the
+    // knowledge-extraction input. The notes dir mirrors the branch slug (review/<p>/<slug>/<pr> ->
+    // PRs/<p>/<slug>/<pr>). Best-effort: an unreadable/absent notes tree yields a short placeholder rather
+    // than throwing — extraction must never block the lifecycle (design §6).
+    static async Task<string> ReadPrNotesFromBranchAsync(
```
gautam-achieveai, line 363: [gautam-achieveai's bot] **MEDIUM** (Test Coverage)

`ReadPrNotesFromBranchAsync` (~30 lines: notes-branch `ls-tree`/`show`, the `review/…` → `PRs/…` path derivation, and two best-effort fallback strings) has **zero tests** — grepping for the method name and both fallback literals returns nothing. This is non-trivial I/O + path mapping on the PR-close extraction path, and its fallbacks (empty tree, mid-loop `show` failure) determine what the extractor actually sees.

**Suggestion:** Extract it into an internal, testable helper and add cases for:
- a `review/<provider>/<slug>/<pr>` branch mapping correctly to `PRs/<pr>/…`,
- an empty `ls-tree` returning the `(no accumulated notes found …)` fallback,
- a `show` failure partway through the loop degrading gracefully rather than throwing.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeExtractionCommitter.cs
```diff
@@ -0,0 +1,176 @@
...
+                return;
+            }
+
+            var committed = await _git
+                .RunAsync(["commit", "-m", $"kb: extract from {sourcePrRef}"], _repoRoot, cancellationToken)
+                .ConfigureAwait(false);
+            if (!committed.Succeeded)
+            {
+                _logger.LogWarning(
+                    "Knowledge extraction commit for {SourcePr} on '{Branch}' failed to commit ({Stderr}); the entry is lost.",
```
gautam-achieveai, line 105: [gautam-achieveai's bot] **LOW** (Observability / False alarm)

When the gate passes but the resolved target is an **update that produces byte-identical content** (same body, `sourcePrRef` already present, same `todayUtc`), `git add` stages nothing and `git commit` exits non-zero ("nothing to commit"). The code treats *any* non-zero commit as failure and logs the alarming "the entry is lost" — but nothing was lost; there was simply nothing new.

This fires legitimately on steady-state re-sweeps: the sweep runs every poller cycle and `MergeToDefaultAsync` is idempotent, so the path is re-entrant. Expect recurring false-alarm warnings.

**Suggestion:** Distinguish "nothing to commit" from a real failure — e.g. after staging, check `git diff --cached --quiet` (or parse the "nothing to commit" signal) and short-circuit to a quiet `LogDebug`/return, reserving the Warning for an actual commit error.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
```diff
@@ -29,91 +34,493 @@ public KnowledgeAgent(IMultiTurnAgent agent, ISandboxFileSystem fileSystem, ILog
...
+        IReadOnlyList<string> tags,
+        string scope,
+        IReadOnlyList<string> sourcePrs,
+        string updated,
+        string body
+    )
+    {
+        var builder = new StringBuilder();
+        _ = builder.Append("---\n");
+        _ = builder.Append("title: ").Append(title).Append('\n');
```
gautam-achieveai, line 306: [gautam-achieveai's bot] **LOW** (Robustness / Round-trip contract)

`BuildEntry` interpolates `title`, `tags`, `scope`, and `sourcePrs` straight into the YAML frontmatter with no escaping:

```csharp
_ = builder.Append("title: ").Append(title).Append('\n');
_ = builder.Append("tags: [").Append(string.Join(", ", tags)).Append("]\n");
```

`title`/`tags` originate from the model's `## TITLE:`/`## TAGS:` lines (untrusted free text; newline injection is already precluded by the line-based parser — good). But a value containing a comma, `]`, or a leading/trailing quote won't round-trip through `KnowledgeIndex.ParseFrontmatter`, which the doc-comment here claims it does. A malformed entry then parses back to `null` and is **silently dropped from `_index.jsonl` and `_toc.md` on the next regen**, while the `.md` file persists — so an entry can quietly vanish from discovery.

**Suggestion:** Minimally escape scalars/list items (reject or escape `]`, `,`, and wrapping quotes in tags/refs; quote a `title`/`scope` containing YAML-special chars). Alternatively, add a parse-what-you-render round-trip assertion in `BuildEntry` under Debug. Practical exposure is limited to `title` + `tags` (scope is already `IsSafeSegment`-validated and `sourcePrRef` is daemon-built), so escaping those two closes it.

_Related nit (also LOW): distinct titles that slugify to the same slug in the same scope silently overwrite each other — `Slugify` (line 530) maps `HTTP & gRPC` and `HTTP / gRPC` both to `http-grpc.md`, and the create is then treated as an in-place update. Consider detecting a slug collision against a different existing title and disambiguating or logging._

samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
```diff
@@ -29,91 +34,493 @@ public KnowledgeAgent(IMultiTurnAgent agent, ISandboxFileSystem fileSystem, ILog
...
+                collected.RunId
+            );
+            return null;
+        }
+
+        var targetPath = JoinPath(knowledgeBaseDir, targetRelPath);
+        var existing = await _fileSystem.ReadFileAsync(targetPath, cancellationToken).ConfigureAwait(false);
+        var existingMeta = existing is null ? null : KnowledgeIndex.ParseFrontmatter(targetRelPath, existing);
+
+        var scope = ScopeSegment(targetRelPath) ?? parsed.Scope;
```
gautam-achieveai, line 98: [gautam-achieveai's bot] [QUESTION] **Clarification Needed**

**Code:** `var scope = ScopeSegment(targetRelPath) ?? parsed.Scope;` (`## UPDATES` to a legacy-flat path)

**Question:** When `## UPDATES` resolves to a flat path (`x.md`, no scope segment — which `NormalizeUpdatesRelPath` explicitly supports), `ScopeSegment` returns null and the code falls back to the model-supplied `parsed.Scope`. The entry then gets `scope: system` frontmatter while it physically lives at the KB root, so `_index.jsonl` (`file: x.md`) and the layered ToC disagree with the `scope` field. Is **updating a legacy-flat entry** a scenario you actually expect, or is flat-path support in `NormalizeUpdatesRelPath` purely forward-compat?

**Why this matters:** If flat entries are never seeded (the KB has none in this PR), this is moot and I'll drop it. If they can exist, the index/ToC/frontmatter disagreement is a real metadata-consistency bug worth fixing (derive `scope` as empty for a flat target, or promote the entry to `<scope>/<slug>.md` on update).

gautam-achieveai:
[gautam-achieveai's bot] ## Re-Review Summary: PR #156

Re-reviewed the delta since the last full review (`9010b697…96d758fc`, plus the trivial `c806704f` chore on top — 26 files, +2290/−206). This pass covers the **Layer-2 knowledge base** (new since the last full review) and re-verifies the previously-raised findings.

### Previous Issues — Resolution Status
| # | Issue | Sev | Status |
|---|-------|-----|--------|
| 1 | `pull --rebase` result discarded (`KnowledgeExtractionCommitter`) | MED | ✅ Resolved — captures result, `rebase --abort`, logs stderr, breaks. Test `KnowledgeExtractionCommitterTests.cs:171` (fails before fix). |
| 2 | Same pattern in `ReviewBranchManager` push path | LOW | ✅ Resolved — identical fix. Test `ReviewBranchManagerTests.cs:139`. |
| 3 | Poisoned partial warm store on failed lease | HIGH | ✅ Resolved — `LeaseAsync` wipes `StorePath` before recycling; test `ReviewSlotPoolTests.cs:131`. |
| 4 | Cancellation must propagate, not degrade to fallback | MED | ✅ Resolved — catch filters exclude `OperationCanceledException`; tests `ReviewToolContextBuildTests.cs:164 & :183`. |
| 5 | Pooled review agent got a different sandbox session than the prepared slot | HIGH | ✅ Resolved — `BuildToolContextAsync` routes to `GetOrCreateForSlotAsync`; provisioner mounts `/workspace` over the slot. |
| 6 | `ReviewSlotPool.ReturnAsync` not idempotent-safe | HIGH | ☑️ Deferred → #160 (accepted). In-PR mitigation verified: all three `ReturnAsync` call sites are guarded (two behind atomic `TryRemove`; the third returns a fresh lease never added to the map). |

**No prior thread remains ACTIVE.** This is a large improvement over the earlier 4-Must state — every prior Must/Should is genuinely resolved with a behavioral, fail-before-fix regression test.

### Security — Layer-2 KB (attacker-influenceable input): defense holds ✅
Adversarially re-audited the path-traversal / write-boundary defense across every class — `../`, absolute, drive-letter, UNC, NTFS ADS, sibling-prefix (`KnowledgeBase-evil/`), empty/whitespace, unicode, case-only bookkeeping bypass, git arg-injection, and title→slug injection. **No write-escape hole found.** Git uses an argv vector (no shell), every exit code is checked, and `_toc.md`/`_index.jsonl` are refused as write targets. Nicely done.

### New Issues Found in the Delta
Two **MEDIUM** defects are the reason this stays at Request Changes. Both share the worst properties for a KB-writing feature — they touch **durable, user-visible data**, fire on the **common path**, and are **invisible to the current tests**:

- **🔴 MEDIUM — `Program.cs:310` — `sourcePrRef` doubles the provider.** `NormalizedKey` already begins with the provider, so `$"{pr.Provider}/{pr.Repo.NormalizedKey}/{pr.PrId}"` yields `github/github/achieveai/lmdotnettools/42` (and, worse, `ado/azure-devops/…` — two different provider tokens) into the durable `sourcePrs:` frontmatter and commit message on 100% of the happy path. The tests miss it because `KnowledgeAgentTests` injects a hand-authored `SourcePr` constant, bypassing this construction seam.
- **🔴 MEDIUM — `KnowledgeAgent.cs:350` — a blank line between markers silently drops the whole extraction.** `ParseEntry` consumes only a *contiguous* run of marker lines, so a blank line before `## TITLE`/`## UPDATES` ends the header → title stays empty → `ResolveTargetRelPathAsync` returns null → the entire extraction is discarded even though the model produced durable knowledge. Blank lines between `##` headings are a very common LLM formatting habit; every test tightly-packs markers, so this path is unverified.

**Test-coverage gaps (MEDIUM):** `Program.cs::ReadPrNotesFromBranchAsync` (~30 lines of new notes-branch I/O + path mapping) has zero tests; and the traversal test suite omits the enumerated exploit classes (absolute/drive-letter/UNC/ADS/sibling-prefix) — the guard is correct but *unpinned*, so a future denylist refactor could silently re-open it with all tests green.

**LOW (non-blocking, worth folding in):** benign no-op update logged as "the entry is lost" (`KnowledgeExtractionCommitter.cs:99`); unescaped YAML frontmatter interpolation can make an entry vanish from the index on regen (`KnowledgeAgent.cs:295`); slug collisions silently overwrite (`KnowledgeAgent.cs:205`); legacy-flat `## UPDATES` scope disagreement. **Recommended:** migrate the bespoke traversal check to the repo's hardened `PathCanonicalizer` — retires the POSIX-only duplication and closes the coverage-pinning gap in one move.

### Verdict
**REQUEST CHANGES** (narrow, delta-scoped). No thread is active on the unresolved-gate; the block is purely the two new MEDIUMs, which are silent + durable + high-frequency + test-blind — exactly the class of defect that should not ship in a knowledge-persisting feature. Both fixes are cheap.

**Required to clear:**
1. `Program.cs:310` — stop doubling the provider (use `pr.Repo.NormalizedKey` alone).
2. `KnowledgeAgent.cs:350` — tolerate blank lines between markers, and add a blank-separated-markers test.
3. Add tests for `ReadPrNotesFromBranchAsync` and pin the enumerated traversal exploit classes.

Individual finding threads are posted inline. Deferred items (#160, Layer 3) remain acknowledged and tracked — please keep #160 on the board.

gautam-achieveai (commented):
[Gautam's review bot] ## Re-Review Summary: PR #156

Reviewed the delta since `c806704f` (5 commits, 8 files, +293/-127 lines).

### What changed
Prompt externalization: the three daemon system prompts (review, judge, knowledge-extraction) moved from inline C# string literals to a YAML/Scriban template (`Prompts/daemon-prompts.yaml`, embedded resource). The review prompt's 'Workspace layout' section is now templated per-run via `BuildPromptVariables`, which derives its values from the **same** `ReviewToolContext.NotesDir` that scopes Write/Edit/Bash tools -- so the prompt can never direct the agent to a location its tools don't actually allow.

### Assessment
| Area | Status |
|------|--------|
| Design (prompt-tool alignment) | OK Single source of truth -- no desync possible |
| Scriban injection risk | OK Template is a compile-time embedded resource; variable values are internal paths |
| Test coverage | OK 4 new tests (positive + negative rendering, variant, integration) |
| Behavioral regression | OK Semantically identical workspace-layout content, reorganized from user message to system prompt |
| No-args overload (CreateReviewProfile()) | Note: Renders blank checkout_root -- acceptable, documented as 'prefer the variables overload' |

### Verdict
**Approve** -- well-structured prompt externalization. No new issues. Prior LOW findings remain tracked in !160.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental review** of `696420fd` (re-review context for daemon prompts)

## Summary
Single commit adding re-review context (prior reviewed commit SHA, round number, prior notes file paths) to the daemon's prompt template. 6 files, +292/-7. Build passes, all 56 relevant tests pass.

## Verdict: APPROVE

The implementation is clean and well-tested:
- `ComputeRereviewContextAsync` cleanly isolates store + filesystem access from the pure template builder
- `GetPriorReviewSummary` correctly filters to PRIMARY-variant completed runs only
- Prompt conditionals (`is_rereview`, `has_prior_files`) render correctly -- verified by unit tests checking no leftover syntax
- Test coverage hits: happy path, never-completed runs, variant exclusion, first-review-no-rereview-section

### Minor observations (non-blocking):
1. Both `RunPrimaryReviewAsync` and `RunVariantArmAsync` call `ComputeRereviewContextAsync` independently -- two sequential DB queries for identical data. Not a problem at current scale, but could be hoisted to the shared caller if perf ever matters.
2. If a re-review is forced at the *same* head SHA, the prompt says 'focus on what changed since X' where X == current head (empty incremental diff). The 're-check any prior findings' clause should keep the agent useful, but worth keeping in mind.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental review** (commits `696420fd`->`d14121f9`)

Small +43/-2 change adding configurable `BotName` to `CodeReviewDaemonOptions` and prefixing posted PR comments with `[BotName]` while keeping the retained artifact unprefixed. Clean separation, well-tested.

**Observation (MEDIUM — Documentation accuracy):**
The XML doc on `BotName` (line 208) claims it is used *both* as the git commit identity's `user.name` for retention commits (see `GitRunner`) *and* as the `[BotName]` prefix on posted comments. However, `GitRunner.IdentityArgs` (line 38) is hardcoded to `user.name=AchieveAi Review Bot` and never reads `BotName`. Currently `BotName` is **only** used for the comment prefix.

Either:
1. Wire `GitRunner` to use `BotName` (if intended), or
2. Remove the git-identity claim from the doc and track it as future work.

Not blocking — the code works correctly for its actual usage — but the inaccuracy will confuse the next reader.

**Verdict: APPROVE** — no blocking issues.

gautam-achieveai:
[Revobot]

_No review content was produced._

gautam-achieveai (commented):
[Gautam's review bot] **APPROVE** (Iteration 7 -- incremental: d14121f9 -> 14854311)

**Change:** Resilience fix for KB/notes reads in `DaemonReviewStageExecutor`. Wraps `ListFilesAsync` and `ReadFileAsync` calls in try/catch (excluding `OperationCanceledException`), degrading gracefully on transport failures instead of killing the review.

**Assessment:**
- Exception filter correct -- cancellation propagates, transport errors caught
- Logging at Warning with full exception context
- Fallback values match existing null-handling paths (empty list / unchanged input)
- New test validates the exact production failure mode (sandbox session 404)
- FakeSandboxFileSystem fault injection is clean and non-intrusive

No new issues. Design section 6 contract ('KB/notes must never fail the review') now properly enforced at both call sites.

gautam-achieveai:
[Revobot]

_No review content was produced._

gautam-achieveai (commented):
[Gautam's review bot]

**Iteration 8 — Incremental review** (1 commit: `2ce79dfd`)

## Summary
Clean prompt-variable addition: `bot_name` is injected into `BuildPromptVariables` from `CodeReviewDaemonOptions.BotName` and rendered via Scriban in the YAML template identity line and self-reference directive. Both call sites updated consistently; no-arg overload seeds a safe default. Two new tests cover both paths.

## Assessment
- ✅ Variable flows correctly: `_options.BotName` → param → dictionary → template
- ✅ No-arg overload prevents empty-name rendering (`You are ,`)
- ✅ Tests assert both populated and default scenarios
- ✅ Build succeeds with 0 warnings; all 17 `DaemonAgentFactoryTests` pass
- ✅ No behavioral regression (private method signature change, both callers updated)

## Verdict
**Approve** — Trivial, well-tested change consistent with prior prompt externalization pattern.

PR #109, opened 2026-06-28
Harden LmStreaming: multi-turn replay/identity, gateway + sandbox resilience, empty web_search loop

gautam-achieveai:
## Summary

Hardens the `LmStreaming.Sample` streaming pipeline against issues found while driving real providers (Kimi-backed Anthropic and Copilot-Claude) through multi-turn, sandboxed Workspace-Agent tasks: multi-turn message-identity collapse, Anthropic/Kimi replay 400s, gateway/sandbox connectivity, and the "empty web_search loop". `web_search` stays enabled in every mode.

## Changes

- **Multi-turn message identity (server-side root fix)** — `MultiTurnAgentLoop` mints a per-turn `generationId` (turn 1 reuses the run's id; turns 2+ get a fresh one) so later turns' thinking/text no longer collide on the merge key and collapse onto the first turn. Client-side per-turn epoch kept as defense-in-depth for already-persisted conversations.
- **Anthropic/Kimi replay correctness** — drop orphaned signature-only thinking blocks, demote signatureless thinking to text, and stably order each assistant turn `thinking → text → tool_use` so `tool_use` is always last (avoids 400 `invalid_request_error`).
- **Gateway + sandbox resilience** — IPv4 gateway resolution (don't let dual-stack mask single-family failures), graceful sandbox-unavailable handling, container-health guard, and recreate-on-404.
- **Mode-driven built-in tools** — server-side built-ins (e.g. `web_search`) follow the mode's declaration, decoupled from the function-tool allow-list.
- **Empty web_search loop** — a provider/model sometimes emits a query-less `web_search` (`server_tool_use` with `{}` input); the server returns empty results and replaying the degenerate call teaches the model to repeat it until "Max turns reached". The message joiner (`MessageUpdateJoinerMiddleware`) now skips building an empty server-tool call and drops its orphaned empty result, so it is never persisted, shown on reload, or replayed. Real, query-bearing searches are unchanged.

## Testing

- `dotnet test` green on the affected suites: `LmCore.Tests` (763, incl. new joiner empty-web_search coverage), `AnthropicProvider.Tests` (174), `LmMultiTurn.Tests` (277), `LmStreaming.Sample.Tests` (306).
- New regression tests: joiner drops an empty server-tool call + its empty result while preserving a real query-bearing search; per-turn generationId; thinking/tool_use replay ordering; sandbox health + gateway-404 recreate.

## Related Issues

Fixes #108

PR #157, opened 2026-07-06
Authenticate sample clients to SandboxGateway (+ S2S caller forwarding) (#153)

### LINKED ISSUE #153: Authenticate sample clients to SandboxGateway (+ S2S caller forwarding)
## Summary

The sibling **SandboxGateway** MCP service (`SandboxedOstoolsMcpServer`, issue #109 / ADR 0029)
now requires per-app bearer auth on every app-facing request: `X-Sbx-App-Id` + `X-Sbx-App-Key`.
Neither client in this repo sends any app identity today, so once the gateway enables
`AUTH_ENFORCE` (default ON), **all sandbox traffic from both samples breaks (401/404)**.

This feature (a) wires the two required headers into every gateway-facing call in
`LmStreaming.Sample` and `CodeReviewDaemon.Sample` with a **separate app identity per sample**,
and (b) adds **S2S caller forwarding** to `LmStreaming.Sample` so that when another service drives
it via the headless conversation API (#138), sandbox sessions run under **the caller's** sandbox
identity — not LmStreaming.Sample's own.

## Value

- **Operators:** unblocks turning on `AUTH_ENFORCE` without taking these clients down; each client's
  sandbox sessions/workspaces are isolated by its own authenticated identity.
- **S2S integrators:** a calling service's sandbox work is attributed to and scoped by *its own*
  identity instead of silently sharing LmStreaming.Sample's.
- **Security:** adds real inbound auth to LmStreaming.Sample's currently-open headless API.

## Background (ground truth — ADR 0029)

- Two headers on every app-facing endpoint: `X-Sbx-App-Id` + `X-Sbx-App-Key` (base64, ≥32 decoded
  bytes, constant-time SHA-256 compare). Body-claimed `app.id` is ignored; identity is credential-bound.
- Sessions are stamped with the creating app's `app.id`; a foreign/missing identity gets a **uniform
  404**. `GET /api/v1/sandboxes` is app-scoped. `GET /health` is ungated.
- **No per-user/tenant concept** — the only identity the gateway trusts is the app credential in the
  headers. "Forwarding" therefore means presenting a *different app credential*, not a user token.

## Current state (verified inventory)

- **No app-identity header sent anywhere today** (repo-wide grep for `X-Sbx-App`/`AppKey` = 0 hits;
  only `X-Session-ID`).
- Shared infra `src/LmAgentInfra/Sandbox/`: `SandboxGatewayOptions` (has `AppId` default
  `"lmstreaming-sample"` — reused verbatim by the daemon — but **no credential field**);
  `SandboxSessionRegistry` (5 gated calls; caches **one `"default"` session per workspace**, shared
  across all callers); `SandboxGatewayLifetime` (`/health`).
- `LmStreaming.Sample`: MCP hot path (`ConnectHttpMcpClient`/`BuildHttpMcpServer`/
  `ConnectLlmQueryMcpClients`) sends `X-Session-ID` only; sessions are lazily created and **silently
  recreated on 404 from background threads** (run loop / liveness probe); `ConversationsController`
  (headless `POST /api/conversations`, #138) has **no auth guard**; `MarketplaceCatalogClient` calls
  the gated `marketplaces/preview` with no header.
- `CodeReviewDaemon.Sample`: poll-triggered, **no external caller** — two gateway paths
  (`SandboxOrchestrator` + shared infra), both `X-Session-ID` only.
- Precedent: reuse `src/LmAgentInfra/Auth/AuthSharedSecret.cs` (proven static shared secret) for
  inbound auth. ⚠️ Cautionary: the dormant HMAC scheme in `CodeReviewDaemon.Sample/Auth/*` was
  unit-green but never worked — built against an *assumed* wire contract. Test the real contract.

## Decisions (confirmed)

1. Separate app identity per sample.
2. S2S = per-caller credential passthrough, two layers: authenticate the inbound caller, then forward
   the caller's sandbox credential downstream.
3. Credential storage = env vars for each sample's own id/key; caller creds arrive on the inbound request.
4. Phased delivery within this one issue.

**Resolved forks** (recommended options confirmed): inbound = caller sends an inbound secret
(verified via `AuthSharedSecret`) **+ its own sandbox creds**, we verify & forward; credential lifecycle
= **re-sent every turn, nothing stored server-side**; session cache key → `(workspaceId, appId)`.

## Milestone 1 — Baseline app auth (independently shippable; unblocks `AUTH_ENFORCE`)

- Add env-var credential to `SandboxGatewayOptions`; give the daemon its **own** app id. Thread a
  `SandboxCredential(appId, appKey)` value into the registry and **stamp it on the cached session entry**.
- Attach the two headers **per-request** (`HttpRequestMessage`, not `DefaultRequestHeaders`) at every
  gated call site: 5 registry calls, MCP hot path, daemon `SandboxOrchestrator`, `MarketplaceCatalogClient`.
- Startup validation (base64 + ≥32 bytes, fail fast, redacted); map gateway 401/403 to a distinct error
  (`sandbox_auth_failed` vs `sandbox_unavailable`); never log the key.

**Acceptance (M1):**
- [ ] With `AUTH_ENFORCE=on`, both samples complete a full session lifecycle (create → MCP call →
      discovered → destroy) end-to-end.
- [ ] A test fails if any gateway-facing request omits the headers (create/get/list-discovered/delete/
      mcp/marketplaces-preview + daemon orchestrator path).
- [ ] The two samples use **distinct** app ids (daemon no longer defaults to `"lmstreaming-sample"`).
- [ ] Startup rejects a malformed/short key with a redacted error; the key never appears in logs/error
      payloads (asserted).
- [ ] Gateway 401/403 surfaces distinctly from connectivity failures.
- [ ] `AUTH_ENFORCE=off` local path still works; deploy ordering documented.

## Milestone 2 — Inbound S2S auth + per-caller passthrough (LmStreaming.Sample)

- Guard `ConversationsController` with inbound auth via `AuthSharedSecret`.
- Introduce `ISandboxCredentialProvider` (constant impl for daemon; request-scoped for LmStreaming);
  resolves **only at request entry** — contextless paths read the credential off the session entry.
- Partition the session cache by `(workspaceId, appId)`; lazy-create/recreate/liveness/teardown all use
  the session's **bound** credential.
- Cross-actor resume rule = a single guard/refusal (no silent identity swap between S2S and UI). *Verify
  first* whether the UI can even address an S2S-created conversation by id.

**Acceptance (M2):**
- [ ] Unauthenticated calls to `ConversationsController` are rejected; a valid inbound secret is accepted
      (constant-time; secret never logged).
- [ ] Two S2S callers with different creds never share a session — a session created under credential X
      returns 404 when addressed with Y (integration test; the sole foreign-cred→404 proof).
- [ ] A session (re)created from a background loop / liveness probe uses the credential that created the
      conversation's session — no dependence on an active request.
- [ ] An S2S-provisioned conversation cannot be silently continued by the UI under a different identity.
- [ ] Contextless teardown uses the session's bound credential (no orphaned sessions — documented
      consequence).

## Cross-cutting acceptance

- [ ] An integration test against the **real (or faithful-mock) gateway** verifies header names/format
      vs ADR 0029 — M1 slice: headerless→reject, correct-cred→200 (foreign→404 lives in the M2 test).
- [ ] Structured observability: app id (never key), session id, gateway status, auth-vs-connectivity
      classification.

## Out of scope

Multi-tenant/per-user isolation inside the gateway; changing the gateway itself; passthrough for
`CodeReviewDaemon.Sample` (no external caller — static credential only); secret-management beyond env vars.

## Files

`src/LmAgentInfra/Sandbox/{SandboxGatewayOptions,SandboxSessionRegistry,SandboxGatewayLifetime}.cs` ·
`samples/LmStreaming.Sample/{Program.cs,Controllers/ConversationsController.cs,Services/MarketplaceCatalogClient.cs}` ·
`samples/CodeReviewDaemon.Sample/{Program.cs,Workspace/Sandbox/SandboxOrchestrator.cs}` ·
`src/LmAgentInfra/Auth/AuthSharedSecret.cs` (reused) · tests under `tests/LmStreaming.Sample.Tests` +
`tests/CodeReviewDaemon.Sample.Tests` + a real-contract integration test.

gautam-achieveai:
## Summary

The sibling **SandboxGateway** MCP service (`SandboxedOstoolsMcpServer`, issue #109 / ADR 0029) now
requires per-app bearer auth on every app-facing request (`X-Sbx-App-Id` + `X-Sbx-App-Key`). This wires
that into both sample clients with a **separate app identity per sample**, and adds **S2S caller
forwarding** so a headless caller driving `LmStreaming.Sample` runs its sandbox work under **its own**
identity — not the sample's.

Delivered in two milestones (plan v2 approved on the issue).

### M1 — baseline app auth (both samples)
- New `SandboxCredential` value + `FromOptions`/`ValidateKeyOrThrow` (standard base64, ≥32 decoded bytes,
  **redacted** errors, never logs the key); `SandboxGatewayOptions.AppKey`.
- `SandboxSessionRegistry`: one `SendGatewayAsync` helper stamps headers **per-request** (never
  `DefaultRequestHeaders`) + a private per-session credential side-table (`CredentialFor`); all 5 gated
  calls routed through it; 401/403 → distinct `sandbox_auth_failed`; `is_scoped_404` warn.
- `LmStreaming.Sample`: fail-fast validation, headers on both sandbox `/mcp` dicts + `MarketplaceCatalogClient`.
- `CodeReviewDaemon.Sample`: its **own** identity (`CRD_SANDBOX_APP_ID` default `codereview-daemon` /
  `CRD_SANDBOX_APP_KEY`); headers on both direct `/mcp` transports.

### M2 — inbound S2S auth + per-caller passthrough (`LmStreaming.Sample`)
- `MultiTurnAgentPool`: `CallerCredential` (init-only, frozen per conversation) + an **in-lock**
  cross-actor guard (`SandboxCredentialConflictException`).
- `SandboxSessionRegistry`: `_sessions` repartitioned to a `(WorkspaceId, AppId)` value-tuple key so two
  callers sharing a workspace never collide; contextless teardown destroys **before** dropping the owner
  credential.
- `ConversationsController`: self-contained `[InboundS2SAuth]` filter (per-request constant-time
  `X-S2S-Auth` vs `Auth:S2SInboundSecret`; disabled when unset — keyless dev); caller `X-Sbx-*` read into a
  `SandboxCredential` (never persisted) and passed to the pool; conflict → **409**. WS path surfaces the
  same conflict as a structured error frame.
- Operator runbook: `docs/deployment/AUTH_ENFORCE.md`.

## Key Decisions
- **Credential flows as a value, resolved at request entry** (not a scoped DI service into the singleton
  registry/pool) — avoids the captive-dependency trap the design review caught; captured on a private
  per-session side-table so background recreate/liveness/teardown use the owner's identity.
- **Secret kept off the public `SandboxSession` record** (private side-table) to prevent secret sprawl.
- **Session cache partitioned by `(WorkspaceId, AppId)`** (app id only, never the key) — the only coherent
  way to make per-caller passthrough actually scope sessions given the gateway's uniform-404 ownership.
- **Nothing persisted server-side** — the caller re-sends its credential each turn; a conversation is bound
  to its creating identity (cross-actor resume refused, gateway 404 the durable backstop).
- **Dedicated inbound trust boundary** — `Auth:S2SInboundSecret` / `X-S2S-Auth`, separate from the
  gateway↔webhook secret.

## Testing
- `dotnet build LmDotnetTools.sln` clean; **478** `LmStreaming.Sample.Tests` + **322**
  `CodeReviewDaemon.Sample.Tests` pass; `dotnet format` gate ✓.
- New coverage: credential validation/redaction, per-request headers on all gated sites, session
  partition + background-recreate credential reuse, the full cross-actor resume matrix, a 100-way
  alternating-credential concurrency stress test, and controller inbound-auth accept/reject + caller
  passthrough + 409 — all asserting no secret/key leak.

## Known-minor / deferred (from self-review — none blocking)
- `ValidateCredentialOnStartup` is wired but the boot-time probe itself is deferred (marked `TODO #153 M6`).
- `/ws` (interactive UI) is intentionally not behind `[InboundS2SAuth]` — it can only ever use the process
  default identity and cannot forge a caller credential; documented in the runbook.
- The gated liveness probe treats a 401/403 as "alive" (defers the auth failure to the next `/mcp` call);
  the create path classifies auth failures directly.

Fixes #153


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #157 — Authenticate sample clients to SandboxGateway (+ S2S caller forwarding)

## Context
Work item #153 — wires two independent auth boundaries: outbound per-app credential to the gateway (M1) and inbound S2S auth + per-caller passthrough (M2).

## Verdict: ✅ Approve with minor suggestions

The auth architecture is well-designed. Trust boundaries are clearly separated, secrets are never logged, race conditions are handled with proper synchronization, and test coverage is thorough. The cross-actor resume matrix is correctly implemented with the guard inside the per-thread lock.

## Findings

### 🟡 MEDIUM — `SandboxCredential.ToString()` can leak AppKey

`src/LmAgentInfra/Sandbox/SandboxCredential.cs:12`

C# record struct auto-generates `ToString()` that prints ALL positional parameters including the secret `AppKey`. While the codebase is careful today (comments, runbook all emphasize 'never log'), there's no structural enforcement. A single `_logger.LogDebug("cred={Cred}", credential)` by a future developer dumps the key to logs.

**Suggestion:** Override with redaction:
```csharp
public override string ToString() => $"SandboxCredential {{ AppId = {AppId}, AppKey = [REDACTED] }}";
```

### 🟡 MEDIUM — Header-stamping logic duplicated across 4 independent sites

The 'stamp `X-Sbx-App-Id` + conditionally `X-Sbx-App-Key`' pattern is repeated in:
- `SandboxSessionRegistry.SendGatewayAsync`
- `Program.cs:AddSandboxAuthHeaders`
- `SandboxOrchestrator.BuildTransportHeaders`
- `MarketplaceCatalogClient`

All four agree today on the 'omit key when blank' protocol, but there's no structural guarantee they stay consistent. `SandboxCredential` already exists as the natural home for a `StampHeaders(...)` method.

### 🔵 LOW — Forwarded caller credentials frozen without validation

`ConversationsController.TryBuildCallerCredential` wraps whatever the caller sends in `X-Sbx-App-Id`/`X-Sbx-App-Key` directly into a `SandboxCredential` without calling `ValidateKeyOrThrow`. This credential is then frozen for the conversation's lifetime. If the gateway later rejects it, recovery requires deleting the thread. Acceptable for a sample app with pre-provisioned S2S callers, but worth noting.

## Positive highlights
- Per-request header stamping (never `DefaultRequestHeaders`) prevents credential cross-talk
- Cross-actor guard inside the per-thread lock eliminates TOCTOU race
- Credential preserved through mode/provider swaps (`SwapAgentUnderLockAsync` copies from old entry)
- Constant-time S2S comparison (SHA-256 + `FixedTimeEquals`) is correctly implemented
- `AUTH_ENFORCE.md` operator runbook is excellent — deploy order, compat matrix, rotation, never-log invariants
- Session cache partitioning by `(WorkspaceId, AppId)` provides necessary caller isolation
- Comprehensive test coverage (credential validation, pool credential matrix, controller S2S auth, registry partition tests)

*Review agents: performance-review, architecture-review + manual deep review*

src/LmAgentInfra/Sandbox/SandboxCredential.cs
```diff
@@ -0,0 +1,83 @@
...
+/// <summary>
+/// Per-app credential the sandbox gateway requires on every app-facing call (ADR 0029), sent as
+/// the <c>X-Sbx-App-Id</c> and <c>X-Sbx-App-Key</c> headers. A pure value type — no ASP.NET types,
+/// no logging — so it can be threaded through <see cref="LmAgentInfra"/>'s singleton services as a
+/// plain value without pulling in any host-specific dependency.
+/// </summary>
+/// <param name="AppId">App identifier sent alongside the key.</param>
+/// <param name="AppKey">Base64-encoded app secret. SECRET — never log this value or include it in
+/// any exception message.</param>
+public readonly record struct SandboxCredential(string AppId, string AppKey)
```
gautam-achieveai, line 12: **[Gautam's review bot]** 🟡 **MEDIUM — Secret leak via auto-generated ToString()**

`readonly record struct` auto-generates `ToString()` that prints ALL positional params: `SandboxCredential { AppId = myapp, AppKey = c2VjcmV0... }`. The codebase relies on social contract ('never log the credential') but there's no structural enforcement.

Consider overriding:
```csharp
public override string ToString() => $""SandboxCredential {{ AppId = {AppId}, AppKey = [REDACTED] }}"";
```

This makes accidental leakage via `_logger.LogDebug("cred={Cred}", credential)` structurally impossible.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -2330,6 +2411,21 @@ ILoggerFactory loggerFactory
         return (registry, createdClients);
     }
 
+    /// <summary>
+    ///     Stamps the sandbox gateway's per-app auth headers (issue #153 / ADR 0029) onto an MCP
+    ///     transport header dictionary. <see cref="SandboxCredential.AppKey"/> is omitted when blank
+    ///     — the keyless <c>AUTH_ENFORCE=off</c> dev path — so the gateway never receives an empty
+    ///     header value.
+    /// </summary>
+    private static void AddSandboxAuthHeaders(IDictionary<string, string> headers, SandboxCredential cred)
```
gautam-achieveai: **[Gautam's review bot]** 🟡 **MEDIUM — Consider centralizing header-stamping on SandboxCredential**

This `AddSandboxAuthHeaders` helper duplicates logic that also appears in:
- `SandboxSessionRegistry.SendGatewayAsync`
- `SandboxOrchestrator.BuildTransportHeaders`
- `MarketplaceCatalogClient`

All four independently implement 'stamp AppId, conditionally stamp AppKey when non-empty.' If the gateway adds a third header or changes the protocol, all four sites must be found and updated.

`SandboxCredential` is the natural home:
```csharp
public void StampHeaders(IDictionary<string, string> headers) { ... }
public void StampHeaders(HttpRequestMessage request) { ... }
```

Non-blocking since all four agree today, but would prevent drift.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
```diff
@@ -527,6 +658,47 @@ await modeStore.GetModeAsync(persistedModeId ?? SystemChatModes.DefaultModeId, c
...
+        }
+
+        var appId = headers[SbxAppIdHeader].ToString();
+        if (string.IsNullOrEmpty(appId))
+        {
+            return null;
+        }
+
+        var appKey = headers[SbxAppKeyHeader].ToString();
+        return new SandboxCredential(appId, appKey);
```
gautam-achieveai, line 781: **[Gautam's review bot]** 🔵 **LOW — Forwarded credential frozen without validation**

The caller's `X-Sbx-App-Key` is wrapped into a `SandboxCredential` without calling `ValidateKeyOrThrow`. This credential is then frozen onto the `AgentEntry` for the conversation's lifetime. If the gateway later rejects it (malformed/too-short), the thread is permanently bound to an invalid credential — recovery requires thread deletion.

Acceptable for a sample app with pre-provisioned S2S callers, but noting the temporal coupling: credential is bound eagerly (at agent creation) but validated lazily (on first gateway use).

gautam-achieveai:
Reviewed PR #157 across 16 files examined — 4 Must, 1 Should, 0 Consider.

## Findings

### Must — `X-S2S-Auth` is applied to the whole conversations API, breaking existing browser clients

- **File:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:124-127`
- **Also affected:** `samples/LmStreaming.Sample/ClientApp/src/api/conversationsApi.ts:27`, `:40`, `:63`, `:77`, `:91`, `:186`, `:206`, `:230`; `samples/LmStreaming.Sample/ClientApp/src/api/chatModesApi.ts:113-118`; `samples/LmStreaming.Sample/ClientApp/src/api/providersApi.ts:23-29`

`[InboundS2SAuth]` is attached to the entire `ConversationsController`, so once `Auth:S2SInboundSecret` is configured, every existing `/api/conversations*` route requires `X-S2S-Auth`. The SPA calls those same routes with ordinary `fetch` requests and no S2S header; it also should not be given a shared S2S secret to put in browser requests.

That means enabling the documented guard turns existing UI operations—list/load conversations, run-state polling, metadata updates, delete, provision, send, status polling, mode switch, and provider switch—into `401 s2s_auth_failed`.

**Fix:** split S2S/headless routes onto a separate controller/path or scope the filter only to endpoints that are truly service-to-service. Keep SPA-facing `/api/conversations` routes compatible, or protect them with an appropriate browser auth mechanism instead of the S2S shared secret.

---

### Must — the documented `LMSTREAMING_S2S_INBOUND_SECRET` env var does not configure the key the code reads

- **File:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:52`, `:70`, `:98-101`
- **Doc:** `docs/deployment/AUTH_ENFORCE.md:20`

The filter reads `configuration["Auth:S2SInboundSecret"]`, but the docs and warning tell operators to set `LMSTREAMING_S2S_INBOUND_SECRET`. ASP.NET Core’s standard environment provider maps `Auth__S2SInboundSecret` to `Auth:S2SInboundSecret`; it does not automatically map `LMSTREAMING_S2S_INBOUND_SECRET` to that configuration key.

As written, an operator following the new runbook can believe inbound S2S auth is enabled while the filter sees a blank secret and explicitly disables the guard.

**Fix:** either document/use `Auth__S2SInboundSecret`, or explicitly bridge `LMSTREAMING_S2S_INBOUND_SECRET` into `Auth:S2SInboundSecret` during startup before the controller filter reads configuration.

---

### Must — mode/provider switch endpoints bypass the cross-actor credential check

- **File:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:538`, `:632`

`SendMessage` builds a caller credential from `X-Sbx-App-Id` / `X-Sbx-App-Key` and maps `SandboxCredentialConflictException` to `409`. The mode and provider switch endpoints do not build or pass the current request’s credential at all:

```csharp
_ = await agentPool.RecreateAgentWithModeAsync(threadId, mode);
_ = await agentPool.RecreateAgentWithProviderAsync(threadId, request.ProviderId, currentMode);
```

`MultiTurnAgentPool` preserves the existing entry’s credential during recreate, but it never compares the current request’s app id on these paths. A different S2S caller can therefore mutate another app’s conversation mode/provider instead of getting the same cross-actor conflict enforced by `SendMessage`.

**Fix:** have these actions call `TryBuildCallerCredential(HttpContext?.Request?.Headers)` and validate it against the frozen conversation owner before switching, or extend the pool recreate APIs to accept the caller credential and throw `SandboxCredentialConflictException` on mismatch. Map that exception to the same `409 caller_credential_conflict` response used by `SendMessage`.

---

### Must — raw gateway auth failure bodies are returned to API clients

- **File:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:623-635`
- **Surfaced at:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:363-368`, `:540-550`, `:645-655`

On gateway `401/403`, the registry copies the raw response body into `SandboxSessionUnavailableException.Message`:

```csharp
$"sandbox_auth_failed: ... {(int)response.StatusCode} {body}"
```

The controller handlers then return `ex.Message` as the public `detail` field. That exposes raw downstream auth failure bodies through public API responses and establishes an unsafe pattern for auth/error handling. Even if the gateway normally behaves, this path should not trust an upstream auth body as client-safe output.

**Fix:** log a sanitized/truncated body server-side for diagnostics, but return a stable client-safe detail such as `sandbox_auth_failed: sandbox gateway rejected the credential for app '<appId>'` with the status code only. Keep secrets and arbitrary gateway bodies out of response payloads.

---

### Should — tests exercise the filter logic but not MVC route enforcement

- **File:** `tests/LmStreaming.Sample.Tests/Controllers/ConversationsControllerS2SAuthTests.cs:146-204`
- **Production:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:124-127`

The tests call `InboundS2SAuthAttribute.OnActionExecutionAsync` directly, which verifies the filter implementation but would still pass if `[InboundS2SAuth]` were accidentally removed from `ConversationsController`. Given this PR’s behavior depends on attribute placement and route scoping, the tests should cover the real MVC wiring.

**Fix:** add a small `WebApplicationFactory`/MVC pipeline test with the secret configured and verify the intended protected endpoint returns `401` without `X-S2S-Auth`. After fixing the controller scoping issue above, also add a regression test proving SPA-facing endpoints remain reachable without the S2S header.

## Verdict

Request changes.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
```diff
@@ -11,11 +13,117 @@
...
+
+        var expectedHash = SHA256.HashData(Encoding.UTF8.GetBytes(expected));
+        var presentedHash = SHA256.HashData(Encoding.UTF8.GetBytes(presented));
+        return CryptographicOperations.FixedTimeEquals(expectedHash, presentedHash);
+    }
+}
+
 [ApiController]
 [Route("api/conversations")]
+[InboundS2SAuth]
```
gautam-achieveai, line 157: **[Gautam's review bot]** [BLOCKER] **HIGH** (API Compatibility / Auth)

`[InboundS2SAuth]` is attached to the entire `ConversationsController`, so once `Auth:S2SInboundSecret` is configured every existing `/api/conversations*` route requires `X-S2S-Auth`. The SPA calls these same routes with ordinary browser `fetch` requests and does not (and should not) have the shared S2S secret, so enabling the documented guard breaks list/load/update/delete/send/status/mode/provider operations.

**Suggestion:** split S2S/headless routes onto a separate controller/path, or scope the filter only to endpoints that are truly service-to-service. Keep browser-facing routes compatible, or protect them with an appropriate browser auth mechanism instead of the S2S shared secret.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
```diff
@@ -11,11 +13,117 @@
...
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+public sealed class InboundS2SAuthAttribute : Attribute, IAsyncActionFilter
+{
+    /// <summary>
+    /// Configuration key the inbound shared secret is read from (env
+    /// <c>LMSTREAMING_S2S_INBOUND_SECRET</c> binds here via the standard ASP.NET Core env-var
+    /// configuration provider's <c>__</c>→<c>:</c> section mapping).
+    /// </summary>
+    public const string SecretConfigKey = "Auth:S2SInboundSecret";
```
gautam-achieveai, line 64: **[Gautam's review bot]** [BLOCKER] **HIGH** (Configuration / Security)

The filter reads `configuration["Auth:S2SInboundSecret"]`, but the docs/comments/log guidance tell operators to set `LMSTREAMING_S2S_INBOUND_SECRET`. ASP.NET Core's standard environment provider maps `Auth__S2SInboundSecret` to `Auth:S2SInboundSecret`; it does not automatically map `LMSTREAMING_S2S_INBOUND_SECRET` to that key. An operator following the runbook can believe inbound S2S auth is enabled while the guard sees a blank secret and disables itself.

**Suggestion:** either document/use `Auth__S2SInboundSecret`, or explicitly bridge `LMSTREAMING_S2S_INBOUND_SECRET` into `Auth:S2SInboundSecret` during startup before this filter reads configuration.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -473,6 +614,28 @@ private async Task<SandboxSession> CreateSessionAsync(WorkspaceRef workspaceRef,
         if (!response.IsSuccessStatusCode)
         {
             var body = await response.Content.ReadAsStringAsync(ct).ConfigureAwait(false);
+
+            if (response.StatusCode is HttpStatusCode.Unauthorized or HttpStatusCode.Forbidden)
+            {
+                // Distinct marker from the connectivity-failure path above: this is the gateway
+                // actively rejecting the presented credential (misconfigured/rotated/wrong app key),
+                // not an unreachable gateway.
+                _logger.LogError(
```
gautam-achieveai, line 612: **[Gautam's review bot]** [BLOCKER] **HIGH** (Security / Error Handling)

On gateway `401/403`, the registry copies the raw downstream response body into `SandboxSessionUnavailableException.Message`. The controller handlers then return `ex.Message` as the public `detail` field, which exposes arbitrary gateway auth failure bodies through API responses. Auth-boundary failures should not trust upstream response bodies as client-safe output.

**Suggestion:** log a sanitized/truncated body server-side if needed, but return a stable client-safe detail such as `sandbox_auth_failed: sandbox gateway rejected the credential for app '<appId>'` with the status code only. Keep arbitrary gateway bodies out of response payloads.

tests/LmStreaming.Sample.Tests/Controllers/ConversationsControllerS2SAuthTests.cs
```diff
@@ -0,0 +1,323 @@
...
+    private static ActionExecutionDelegate CreateNextDelegate(ActionContext actionContext, Action onCalled)
+    {
+        return () =>
+        {
+            onCalled();
+            return Task.FromResult(new ActionExecutedContext(actionContext, [], new object()));
+        };
+    }
+
+    [Fact]
```
gautam-achieveai, line 159: **[Gautam's review bot]** **MEDIUM** (Testing)

These tests call `InboundS2SAuthAttribute.OnActionExecutionAsync` directly, which verifies the filter implementation but would still pass if `[InboundS2SAuth]` were accidentally removed from `ConversationsController` or scoped to the wrong routes. Since this PR's behavior depends on MVC attribute placement and route scoping, the coverage should exercise the real pipeline.

**Suggestion:** add a small `WebApplicationFactory`/MVC pipeline test with the secret configured and verify the intended protected endpoint returns `401` without `X-S2S-Auth`. After fixing the controller scoping issue, also add a regression proving SPA-facing endpoints remain reachable without the S2S header.

gautam-achieveai:
**[Gautam's review bot]** [BLOCKER] **HIGH** (Authorization / Correctness)

`SendMessage` builds the current request's caller credential and maps `SandboxCredentialConflictException` to `409`, but the mode/provider switch endpoints call `RecreateAgentWithModeAsync` / `RecreateAgentWithProviderAsync` without passing or validating the current request's credential (`samples/LmStreaming.Sample/Controllers/ConversationsController.cs:538` and `:632`). That lets a different S2S caller mutate another app's conversation mode/provider instead of hitting the cross-actor guard this PR adds for sends.

**Suggestion:** have mode/provider switch build `TryBuildCallerCredential(HttpContext?.Request?.Headers)` and validate it against the frozen conversation owner before switching, or extend the pool recreate APIs to accept the caller credential and throw `SandboxCredentialConflictException` on mismatch. Map that exception to the same `409 caller_credential_conflict` response used by `SendMessage`.

gautam-achieveai:
**[Gautam's review bot]** ## Re-Review Summary: PR #157

### Verdict: REQUEST CHANGES

I reviewed the current head (`60cb6cae`) against `main` for the SandboxGateway app-auth and S2S caller-forwarding changes. The PR still has blocking issues in the auth boundary and rollout behavior.

### Findings Posted

| # | Severity | Category | Location | Status |
|---|---|---|---|---|
| 1 | HIGH | API Compatibility / Auth | `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:126` | Blocking |
| 2 | HIGH | Configuration / Security | `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:52` | Blocking |
| 3 | HIGH | Authorization / Correctness | `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:538` / `:632` | Blocking |
| 4 | HIGH | Security / Error Handling | `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:623` | Blocking |
| 5 | MEDIUM | Testing | `tests/LmStreaming.Sample.Tests/Controllers/ConversationsControllerS2SAuthTests.cs:146` | Non-blocking but should be fixed with the auth changes |

### Summary

The major blockers are: the new S2S filter is scoped to the whole conversations controller and breaks existing browser clients when enabled; the documented env var does not bind to the config key the filter reads; mode/provider switch paths bypass the caller-credential conflict guard; and raw gateway auth failure bodies can be surfaced in public API details. The test coverage also needs a real MVC pipeline assertion so route/filter placement regressions are caught.

No context questions.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
gautam-achieveai: Fixed in c2230eaf. The filter is now **marker-gated**: it enforces the secret only on service-to-service requests (those carrying `X-S2S-Auth` or the `X-Sbx-App-Id` caller-credential marker). A marker-less same-origin SPA request passes through even with the secret configured, so enabling it no longer 401s list/load/update/delete/send/status/mode/provider. Documented in `AUTH_ENFORCE.md` as a gate on the credential-passthrough surface (not a blanket lock on the same-origin API). New `ConversationsControllerS2SPipelineTests` proves the SPA path stays reachable while a marked request is gated.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
gautam-achieveai: Fixed in c2230eaf. `Program.cs` now bridges `LMSTREAMING_S2S_INBOUND_SECRET` into `Auth:S2SInboundSecret` at startup (an explicit `appsettings.json` / `Auth__S2SInboundSecret` value still wins), so the documented env var actually arms the guard. The false `__`→`:` claim in the attribute XML doc is corrected.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: Fixed in c2230eaf. The gateway 401/403 (and sibling 5xx) response body is now logged server-side only; the exception message the controller surfaces as `detail` carries just the app id and status code — no upstream body.

src/LmAgentInfra/Sandbox/SandboxCredential.cs
gautam-achieveai: Fixed in c2230eaf. Added `public override string ToString() => "SandboxCredential { AppId = ..., AppKey = [REDACTED] }"` so structured logging can no longer echo the key. Covered by `SandboxCredentialTests.ToString_RedactsAppKey_ButShowsAppId`.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: Fixed in c2230eaf. Added `SandboxCredential.StampHeaders(IDictionary<string,string>)` and `StampHeaders(HttpRequestMessage)`, and routed all four sites (Program.AddSandboxAuthHeaders, registry SendGatewayAsync, daemon BuildTransportHeaders, MarketplaceCatalogClient) through them; header names are now single-sourced as `AppIdHeader`/`AppKeyHeader` constants on the type.

tests/LmStreaming.Sample.Tests/Controllers/ConversationsControllerS2SAuthTests.cs
gautam-achieveai: Fixed in c2230eaf. Added `ConversationsControllerS2SPipelineTests` — a real `TestServer` MVC pipeline that fails if `[InboundS2SAuth]` is removed/misscoped: browser (no markers) → reachable, S2S-marked without `X-S2S-Auth` → 401, matching secret → reachable, no secret → reachable. The direct-filter tests also gained browser-pass and appId-marker-without-auth cases.

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
gautam-achieveai: Acknowledged — leaving as-is, consistent with the review note that this is acceptable for a sample app. The gateway validates the key on the actual sandbox call, so a malformed forwarded key surfaces there (now as a clean sanitized `sandbox_auth_failed` 503 after the BLOCKER #4 fix) rather than being eagerly rejected here. Noted as a follow-up if pre-provisioned S2S callers ever need fail-fast validation.

gautam-achieveai:
## Review response — all findings addressed in `c2230eaf`

Thanks for the deep re-review. All 4 HIGH blockers + 2 MEDIUM + 1 LOW are resolved; the 7 inline threads are answered and marked resolved. Summary:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | HIGH | `[InboundS2SAuth]` on whole controller broke the SPA | **Marker-gated** the filter — enforces only on S2S requests (`X-S2S-Auth` or `X-Sbx-App-Id` marker); marker-less same-origin SPA requests pass with the secret set. New `TestServer` pipeline test guards it. |
| 2 | HIGH | `LMSTREAMING_S2S_INBOUND_SECRET` didn't bind to the read key | Bridged into `Auth:S2SInboundSecret` at startup; corrected the XML doc. |
| 3 | HIGH | mode/provider switch bypassed the cross-actor guard | `RecreateAgentWith{Mode,Provider}Async` now take the caller credential and validate it against the frozen owner inside the swap lock; controller maps the conflict to the same `409 caller_credential_conflict` as `SendMessage`. Covered by new pool conflict tests. |
| 4 | HIGH | raw gateway auth body returned in API `detail` | 401/403 (and sibling 5xx) bodies stay in server logs only; `detail` = app id + status code. |
| 5 | MEDIUM | `ToString()` leaked the key | Redacted override. |
| 6 | MEDIUM | header-stamping duplicated across 4 sites | `SandboxCredential.StampHeaders(...)`; all 4 sites delegate. |
| 7 | MEDIUM | tests didn't exercise the real MVC pipeline | `ConversationsControllerS2SPipelineTests` (TestServer) + extended filter/switch cases. |
| 8 | LOW | forwarded credential not eagerly validated | Left as-is per the note that it's acceptable for a sample; gateway validates on the real call. |

**Auth-contract note (BLOCKER #1):** marker-gating deliberately gates the credential-passthrough / headless surface, **not** the same-origin interactive API — an operator who needs every route (including same-origin) authenticated should front the app with a real browser-auth mechanism, not this S2S service secret. This is documented in `docs/deployment/AUTH_ENFORCE.md`. Happy to switch to blanket-gating + browser auth if you'd prefer the stricter posture.

Verification: solution builds clean · **494** LmStreaming.Sample.Tests + **322** CodeReviewDaemon.Sample.Tests green · `dotnet format whitespace --verify-no-changes` ✓.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental review of `c2230eaf` - all prior findings addressed.

## Verdict: Approve

All 3 BLOCKERs and 3 MEDIUMs from the prior iteration are resolved correctly:

| # | Finding | Status |
|---|---------|--------|
| 1 | S2S auth breaks SPA routes | Marker-gated - IsServiceToServiceRequest() allows browser-origin through |
| 2 | Config key mismatch (LMSTREAMING_S2S_INBOUND_SECRET to Auth:S2SInboundSecret) | Bridged at startup in Program.cs |
| 3 | Gateway auth body leaked in API response | Client-safe message (app id + status code only) |
| 4 | ToString() secret leak | Overridden with [REDACTED] |
| 5 | Header-stamping duplicated in 4 sites | Single-sourced on SandboxCredential.StampHeaders |
| 6 | Missing pipeline-level test coverage | ConversationsControllerS2SPipelineTests + expanded direct-filter tests |

### New observations (non-blocking)

LOW - `ConversationsController.TryBuildCallerCredential` still uses private `SbxAppIdHeader` / `SbxAppKeyHeader` constants (lines 758-759) rather than the newly-centralized `SandboxCredential.AppIdHeader` / `.AppKeyHeader`. Values match, so no behavioral impact, but it is incomplete consolidation. Quick follow-up.

### Highlights
- Cross-actor guard correctly extended to RecreateAgentWithModeAsync / RecreateAgentWithProviderAsync inside the per-thread lock
- AUTH_ENFORCE.md compatibility matrix updated with marker-gate semantics and SPA pass-through
- Test coverage thorough: pipeline tests, direct-filter cases, pool credential matrix for mode/provider switch conflicts

samples/LmStreaming.Sample/Controllers/ConversationsController.cs
```diff
@@ -527,6 +717,47 @@ await modeStore.GetModeAsync(persistedModeId ?? SystemChatModes.DefaultModeId, c
...
+        if (string.IsNullOrEmpty(appId))
+        {
+            return null;
+        }
+
+        var appKey = headers[SbxAppKeyHeader].ToString();
+        return new SandboxCredential(appId, appKey);
+    }
+
+    private const string SbxAppIdHeader = "X-Sbx-App-Id";
```
gautam-achieveai: **[Gautam's review bot]** LOW - Consolidation incomplete: these private constants (`SbxAppIdHeader`, `SbxAppKeyHeader`) duplicate the `SandboxCredential.AppIdHeader` / `.AppKeyHeader` constants that this same commit introduced for single-sourcing. `IsServiceToServiceRequest` already references `SandboxCredential.AppIdHeader`. Migrating here keeps things consistent. Non-blocking, same values.

gautam-achieveai: Good catch — finished in 75f7c1b5. `TryBuildCallerCredential` now reads `SandboxCredential.AppIdHeader` / `.AppKeyHeader` (matching `IsServiceToServiceRequest`), and the duplicate private `SbxAppIdHeader` / `SbxAppKeyHeader` constants are removed. Same values, no behavior change.

gautam-achieveai:
Reviewed PR #157 across 11 files — 2 Must, 0 Should, 0 Consider.

## Findings

### Must — Do not log raw gateway auth failure bodies on the app-key rejection path

`src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:617`

On sandbox create `401/403`, this logs the raw upstream gateway response body:

```csharp
_logger.LogError(
    "Sandbox create failed for workspace {WorkspaceId}: sandbox_auth_failed "
        + "({StatusCode}) {Body}",
    workspaceId,
    (int)response.StatusCode,
    body
);
```

This path is specifically handling `X-Sbx-App-Key` rejection. Even though the exception message returned to callers is client-safe, logging the raw gateway auth body violates the PR’s “never-log” invariant if the gateway includes submitted credential material, header values, or other secret-bearing diagnostics in its error response.

Please avoid logging the raw body for auth failures, or pass it through an explicit redaction/scrubbing routine before logging. The current client-safe exception detail can stay as-is.

---

### Must — `Provision` creates S2S conversations without binding them to the creating caller

`samples/LmStreaming.Sample/Controllers/ConversationsController.cs:214`

`Provision` creates the thread metadata and returns a `ThreadId`, but it never reads or persists the caller credential/app id from `X-Sbx-App-Id`:

```csharp
await store.UpdateMetadataAsync(
    threadId,
    existing =>
    {
        var propertiesBuilder = existing?.Properties?.ToBuilder()
            ?? ImmutableDictionary.CreateBuilder<string, object>();

        propertiesBuilder[MultiTurnAgentPool.ProviderPropertyKey] = request.ProviderId;
        propertiesBuilder[MultiTurnAgentPool.WorkspacePropertyKey] = request.WorkspaceId;
        propertiesBuilder[MultiTurnAgentPool.ModePropertyKey] = request.ModeId;
        ...
    },
    ct);
```

The documented contract says a conversation is bound to the caller identity that first created it. For the headless REST flow, `Provision` is the endpoint that creates/reserves the conversation. Today the binding only happens later in `SendMessage` via `MultiTurnAgentPool.GetOrCreateAgent(...)`, so a provisioned S2S conversation is unowned until the first message. That leaves a race/window where another caller — including the plain UI/default identity — can send first and bind the thread under the wrong app id, causing the original S2S caller to get a later `409`.

Please bind ownership during `Provision`: read the caller credential, persist at least the owning app id in thread metadata, and have `SendMessage` / mode/provider switches validate against that persisted owner before creating or reusing the agent. Add regression coverage for S2S `Provision` followed by UI/other-app `SendMessage` returning `409`, and same-app `SendMessage` succeeding.

## Verdict

Request changes.

gautam-achieveai:
**[Gautam's review bot]** ✅ **APPROVE** — Iteration 5 incremental review

The constant consolidation in \75f7c1b5\ cleanly addresses the LOW from my previous review. \TryBuildCallerCredential\ now uses \SandboxCredential.AppIdHeader\ / \.AppKeyHeader\, matching \IsServiceToServiceRequest\, and the duplicate private constants are removed. No behavior change.

**Re: the 2 new 'Must' findings from the parallel review (issue-4895192716):**

1. **Raw gateway body in server-side logs** — I disagree this is blocking. The body is already excluded from the client-facing exception message (lines 624–634). Server-side logging of upstream error responses is standard diagnostic practice. The 'never-log' invariant in this PR protects the *app key*, not arbitrary upstream error payloads. For a sample app this is appropriate.

2. **Provision doesn't eagerly bind caller ownership** — Valid observation, but not blocking. The bind-on-first-use model is documented, consistent across the codebase, and the race window is negligible for the S2S sample flow (provision → immediate send). Adding early binding would over-complicate a sample app's provisioning endpoint.

**Summary:** All 3 original blockers + 3 medium findings from my first review are thoroughly addressed. Test coverage is comprehensive (494 + 322 tests). The code is ready to merge.

gautam-achieveai:
## Response to the two follow-up "Must" findings

Judging these on the merits (independent of the parallel APPROVE):

**1. Raw gateway body logged on the auth-rejection path — FIXED (`45893f22`).**
Agreed this is worth closing. On sandbox-create `401/403` the server log no longer dumps the raw upstream body — that response is the one most likely to echo submitted credential material, so logging it undercut the PR's never-log invariant. It now logs `app id + status code` only; the client-facing `detail` was already sanitized. Non-auth `5xx` logging is unchanged (not an auth boundary, useful diagnostics).

**2. `Provision` doesn't eagerly bind caller ownership — DEFERRED → #162.**
Valid observation, but not blocking for this slice:
- The thread id is a server-minted unguessable `thread-{guid:N}`, returned only to the provisioning caller who sends immediately — the unowned window is sub-second and not practically exploitable, and there's no accidental (non-adversarial) path (the UI never sends to a guid it wasn't handed).
- Bind-on-first-use is the documented, codebase-consistent model; eager persisted-ownership binding is a design expansion (a durable ownership layer the in-memory pool freezing doesn't currently consult) beyond a sample's provisioning endpoint.
- Captured as a proper hardening follow-up (with the durable-across-restart upside and the suggested regression tests) in **#162** rather than dropped.

Updated verification: solution builds clean · **494** LmStreaming + **322** daemon tests green · format gate ✓. CI re-running on `45893f22`.

gautam-achieveai:
Reviewed PR #157 across 12 files — 2 Must, 1 Should, 0 Consider.

## Findings

### Must — Bind caller identity during `Provision`, not only on first `SendMessage`
- **File:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:211-244`
- **Related:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:372-385`, `docs/deployment/AUTH_ENFORCE.md:95-110`

`Provision` reserves a new thread and persists provider/workspace/mode metadata, but it never reads or records the caller’s `X-Sbx-App-Id` / `X-Sbx-App-Key`. The ownership check only begins later in `SendMessage`, where `TryBuildCallerCredential(...)` is passed into `agentPool.GetOrCreateAgent(...)`.

That conflicts with the documented contract that “a conversation is bound to the caller identity (`X-Sbx-App-Id`) that first created it.” With the current implementation, the S2S caller that provisions the thread is not actually the bound owner; whichever caller first sends a message for that thread becomes the owner. If another service learns or receives the provisioned `threadId`, it can capture the conversation before the original provisioner sends the first message.

**Fix:** Bind ownership as part of `Provision` when the thread is created/reserved. For example, persist or reserve the creator app id at provision time, then enforce that owner in `SendMessage`, `SwitchMode`, and `SwitchProvider` before creating/recreating the pooled agent. If `Provision` is intentionally not an ownership boundary, update the runbook/API contract so callers do not rely on provision-time ownership.

---

### Must — Do not log raw gateway auth failure bodies
- **File:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:617-623`
- **Related:** `docs/deployment/AUTH_ENFORCE.md:140-146`

On sandbox create `401/403`, the code logs the raw gateway response body:

```csharp
_logger.LogError(
    "Sandbox create failed for workspace {WorkspaceId}: sandbox_auth_failed ({StatusCode}) {Body}",
    workspaceId,
    (int)response.StatusCode,
    body);
```

This is the credential-rejection path, and the gateway/proxy body is outside this component’s control. If it echoes rejected headers, request details, or diagnostics containing `X-Sbx-App-Key`, `X-S2S-Auth`, bearer tokens, etc., those secrets are written to server logs. This also contradicts the new runbook’s “Never-log invariants,” which state these secret values are never written to logs.

**Fix:** For auth failures, avoid logging the raw body entirely. Log status, workspace, app id, and a safe error code/correlation id. If body content is required for diagnostics, pass it through an explicit redactor before logging.

---

### Should — Apply the same bounded/sanitized logging policy to non-auth gateway create failures
- **File:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:637-642`
- **Related:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1141-1151`

The non-auth sandbox-create failure path also logs the full raw gateway body. This file already has `TruncateBody(...)` for gateway error bodies, but create failures bypass it. A reverse-proxy HTML error page or gateway diagnostic can flood logs and may include reflected request details, workspace paths, or sensitive auth-provider configuration.

**Fix:** Use a bounded, sanitized body snippet here too, or log only status plus a correlation/error code. At minimum, apply `TruncateBody(body)` and redact known secret/header patterns before logging.

## Verdict

Request changes.

gautam-achieveai:
[Revobot]

# PR #157 Re-review (round 02) — Revobot

Re-reviewed PR #157 (per-app sandbox gateway credentials + inbound S2S auth, issue #153) at head `7888ceb8`, focused on the delta since `45893f22` across ~9 touched core files — 0 Must, 3 Should (all carried over), 1 Consider.

The core auth surface (`SandboxCredential`, `SandboxCredentialConflictException`, `ChatWebSocketManager`, `MarketplaceCatalogClient`) is byte-for-byte unchanged since round 01. The delta is a merge of origin/main that brings wire-compatible, tangential changes into PR-touched files. No new correctness, security, or compatibility issues.

## Merge-in changes reviewed (all sound)
- **Wire contract preserved:** `SwitchMode`/`SwitchProvider` moved from anonymous objects to typed `SwitchModeResponse`/`SwitchProviderResponse` (samples/LmStreaming.Sample/Models/ConversationDtos.cs:73,98) with `[JsonPropertyName]` keeping `modeId`/`modeName`/`providerId` camelCase; the new `warning` field is additive and nullable — forward/backward compatible for the SPA and headless callers.
- **No double-stamping:** `GatewayHttpClient()` now wraps every gateway-bound client in `GatewayAuthHandler` (samples/LmStreaming.Sample/Program.cs), which is idempotent — `TryAddWithoutValidation` is a no-op when the header is already present (src/LmAgentInfra/Sandbox/GatewayAuthHandler.cs:37-45) — so it composes cleanly with `MarketplaceCatalogClient`'s per-request `StampHeaders`.
- **`CreateSessionAsync`** dropping local workspace-dir creation is the remote-gateway refactor (#165); it does not touch the credential / `SendGatewayAsync` seam.
- Credential-conflict catch blocks, the cross-actor guard, and the marker-gated S2S filter are all unchanged.

## Should (all carried over from round 01 — still open)
1. **`CreateSessionAsync` 401/403 → `sandbox_auth_failed` mapping untested** — src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:~600-640. This is the one branch deliberately keeping the raw gateway auth body OUT of the exception message and log; a regression could silently start echoing credential-bearing upstream output. No test asserts `sandbox_auth_failed`. Add a create-401/403 test asserting the marker + appId and NOT a sentinel body value.
2. **`ChatWebSocketManager.SendCredentialConflictErrorAsync` untested** — samples/LmStreaming.Sample/WebSocket/ChatWebSocketManager.cs:465-498. The only cross-actor sink with no coverage of its emitted frame/close; the frame must carry app ids only.
3. **Controller `SwitchMode`/`SwitchProvider` 409 credential-conflict mapping untested** — samples/LmStreaming.Sample/Controllers/ConversationsController.cs:585,702. Pool-level `RecreateAgentWith*Async` conflict is covered in `MultiTurnAgentPoolCallerCredentialTests`; the two controller catch blocks that map it to a 409 are not.

## Consider
4. **`DestroyWorkspaceSessionAsync` sequential per-partition teardown** — src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs. Low cardinality per run; defer (unchanged from round 01).

The three open Should items are the same test-coverage gaps flagged in round 01 — carried forward, not escalated, since the merge did not touch that behavior. The auth work itself remains well-structured and secret-hygiene invariants stay enforced structurally.

## Verdict
Approve with comments.

PR #165, opened 2026-07-07
Make sandbox WorkspaceBasePath optional; client never FS-accesses the workspace (#164)

### LINKED ISSUE #164: Sandbox client requires WorkspaceBasePath and pre-creates workspace dirs — breaks remote gateways
## Problem
The sandbox client assumes the workspace lives on a filesystem it can access locally:

- `SandboxGatewayOptions.ResolveWorkspace(override)` **throws** `InvalidOperationException` ("Cannot resolve a workspace override because no workspace base path is configured") when `WorkspaceBasePath`/`WorkspacePath` is unset.
- `SandboxSessionRegistry` calls `Directory.CreateDirectory(...)` for the per-workspace directory on **every** session-create, including when adopting an already-running gateway.

Both break when the sandbox gateway runs on a **remote machine**: the client can neither resolve an absolute local path nor create a directory on the gateway's filesystem. Against the enforcing gateway (ADR 0029), which roots workspaces under a per-app hash and creates them if missing, the client's `mkdir` only produced an **orphan** directory that was never mounted.

### Repro
Adopt a gateway with `SandboxGateway:AutoSpawn=false` and no `WorkspaceBasePath`, then start a Workspace-Agent conversation → `InvalidOperationException` → unhandled 500 → WebSocket crash, before any gateway call.

## Fix
- `ResolveWorkspace(override)` returns just the **leaf** (the workspace identifier) with a null base/full-path when no base is configured — no throw. A rooted override is still rejected; the escape check still applies when a base IS set.
- `SandboxSessionRegistry` no longer creates the workspace directory; it forwards the leaf and lets the gateway own creation + mounting.
- `WorkspaceBasePath` becomes **optional**, used only by the spawn path to provision a LOCAL gateway.

## Impact
- Workspace-Agent conversations work against a remote (or per-app-rooting) gateway with **no** `WorkspaceBasePath` configured.
- The client stops creating orphan workspace directories.

## Tests
- `ResolveWorkspace(override)` resolves the leaf with no base configured (no throw); rooted override still throws.
- Registry does not create the workspace directory on disk while still forwarding the leaf.

gautam-achieveai:
## Summary
Makes the sandbox client treat the workspace as a **remote-owned resource**: it forwards the workspace *identifier* (leaf) to the gateway and never touches the workspace filesystem itself. `WorkspaceBasePath` becomes **optional**.

Fixes #164.

## Why
The client assumed the workspace was on a shared local filesystem:
- `ResolveWorkspace(override)` threw when no `WorkspaceBasePath` was configured.
- `SandboxSessionRegistry` pre-created the workspace dir via `Directory.CreateDirectory` on every session-create.

Both break when the gateway is on a **remote machine**, and the enforcing gateway (ADR 0029) already owns creation (per-app rooting + create-if-missing) — so the client's `mkdir` only produced an orphan directory.

## Changes
- `SandboxGatewayOptions.ResolveWorkspace(override)` — returns the leaf with a null base/full-path when no base is configured (no throw). A rooted override, or one containing a `..` traversal segment, is rejected **regardless of base** (FS-free string check); the escape check still applies when a base IS set.
- `SandboxSessionRegistry` — removed the per-session `Directory.CreateDirectory`; forwards the leaf, gateway owns creation/mounting.
- Docs updated to mark `WorkspaceBasePath` optional.

## Client/gateway boundary now
| Path | Client workspace FS access |
|---|---|
| Adopt a gateway (local or **remote**) | **None** — sends the leaf; gateway owns everything |
| Spawn a **local** gateway (`AutoSpawn=true`) | `EnsureWorkspaceDirectory` still pre-creates locally (client owns the gateway it launches) |

## Tests
- `ResolveWorkspace(override)` resolves the leaf with no base (no throw); rooted + `..`-traversal overrides throw regardless of base.
- Registry: **no-base** session create does not throw and forwards the leaf (#164 repro through `GetOrCreateSessionAsync`); and the client does **not** create the workspace dir on disk (`Directory.Exists(base/leaf) == false`).

## Notes
- **Rebased onto current `main`**; the earlier Scriban bump was dropped as redundant — `main` already remediated the NU1903 advisory (7.2.0 → 7.2.2, `658f7c92`), so this PR inherits that. #166 closed accordingly.
- Deliberately separate from #157 (auth) — different concern, based on `main`.
- The spawn-path `EnsureWorkspaceDirectory` is the one remaining client-side workspace `mkdir` (only runs when the client spawns a local gateway) — could be delegated to the gateway too; noted for discussion.

### DIFF
--- a/src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
+++ b/src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
@@ -77,12 +77,13 @@ public sealed class SandboxGatewayOptions
     /// Resolves a per-workspace directory leaf into base/leaf/full-path. When
     /// <paramref name="relPathOverride"/> is null/whitespace this is identical to
     /// <see cref="ResolveWorkspace()"/>. Otherwise <paramref name="relPathOverride"/> is treated as
-    /// the workspace leaf. A rooted override is always rejected (via
-    /// <see cref="InvalidOperationException"/>) — a workspace identifier is never an absolute path.
+    /// the workspace leaf. A rooted override, or one containing a <c>..</c> traversal segment, is
+    /// always rejected (via <see cref="InvalidOperationException"/>) regardless of whether a base is
+    /// configured — a workspace identifier is never an absolute path and never traverses upward.
     /// <para>
     /// If a local base is configured (<see cref="WorkspaceBasePath"/>/<see cref="WorkspacePath"/>),
-    /// the leaf is resolved under it and rejected when it escapes the base (e.g. <c>"../evil"</c>), so
-    /// a chosen workspace can never mount a directory outside the configured base. If NO base is
+    /// the leaf is resolved under it and additionally rejected if it escapes the base, so a chosen
+    /// workspace can never mount a directory outside the configured base. If NO base is
     /// configured, resolution returns just the leaf with a null base and null full path: the gateway
     /// may be remote, or root the workspace under its own per-app base (ADR 0029), so the client
     /// forwards the leaf and lets the gateway own directory creation instead of resolving/creating a
@@ -102,6 +103,18 @@ public sealed class SandboxGatewayOptions
         {
             throw new InvalidOperationException(
                 $"Workspace directory '{relPathOverride}' must be relative to the workspace base."
+            );
+        }
+
+        // Defense-in-depth: a workspace identifier never legitimately contains a parent-directory
+        // traversal segment. Reject '..' regardless of whether a base is configured — on the no-base
+        // path the leaf is forwarded straight to the (possibly remote/permissive) gateway, so this is
+        // the client's only containment guard there; with a base, it fails fast before the escape
+        // check below would catch the same thing.
+        if (Array.Exists(relPathOverride.Split('/', '\\'), segment => segment == ".."))
+        {
+            throw new InvalidOperationException(
+                $"Workspace directory '{relPathOverride}' must not contain '..' path segments."
             );
         }
 
--- a/src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
+++ b/src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
@@ -29,6 +29,13 @@ public sealed class SandboxGatewayOptions
     /// <summary>
     /// Host base directory the gateway resolves the workspace under.
     /// Becomes <c>WORKSPACE_BASE_PATH</c> when the gateway is spawned.
+    /// <para>
+    /// OPTIONAL. Needed only for the same-host <b>spawn</b> path (and the same-host adopt
+    /// optimization where the client pre-creates the workspace directory). When it is unset, the
+    /// client forwards the workspace leaf to the gateway and lets the gateway own directory
+    /// creation — so a workspace-backed session works against a REMOTE or per-app-rooting gateway
+    /// (ADR 0029) with no local base configured. See <see cref="ResolveWorkspace(string)"/>.
+    /// </para>
     /// </summary>
     public string? WorkspaceBasePath { get; set; }
 
@@ -69,11 +76,18 @@ public sealed class SandboxGatewayOptions
     /// <summary>
     /// Resolves a per-workspace directory leaf into base/leaf/full-path. When
     /// <paramref name="relPathOverride"/> is null/whitespace this is identical to
-    /// <see cref="ResolveWorkspace()"/>. Otherwise the configured base from
-    /// <see cref="ResolveWorkspace()"/> is reused and <paramref name="relPathOverride"/> is treated
-    /// as the workspace leaf. The override is rejected (via <see cref="InvalidOperationException"/>)
-    /// when it is rooted or escapes the base directory (e.g. <c>"../evil"</c>), so a chosen
-    /// workspace can never mount a directory outside the configured base.
+    /// <see cref="ResolveWorkspace()"/>. Otherwise <paramref name="relPathOverride"/> is treated as
+    /// the workspace leaf. A rooted override is always rejected (via
+    /// <see cref="InvalidOperationException"/>) — a workspace identifier is never an absolute path.
+    /// <para>
+    /// If a local base is configured (<see cref="WorkspaceBasePath"/>/<see cref="WorkspacePath"/>),
+    /// the leaf is resolved under it and rejected when it escapes the base (e.g. <c>"../evil"</c>), so
+    /// a chosen workspace can never mount a directory outside the configured base. If NO base is
+    /// configured, resolution returns just the leaf with a null base and null full path: the gateway
+    /// may be remote, or root the workspace under its own per-app base (ADR 0029), so the client
+    /// forwards the leaf and lets the gateway own directory creation instead of resolving/creating a
+    /// local path. A base is thus OPTIONAL — required only for the same-host spawn optimization.
+    /// </para>
     /// </summary>
     public (string? BasePath, string? Leaf, string? FullPath) ResolveWorkspace(string? relPathOverride)
     {
@@ -82,19 +96,25 @@ public sealed class SandboxGatewayOptions
             return ResolveWorkspace();
         }
 
-        var (basePath, _, _) = ResolveWorkspace();
-        if (string.IsNullOrWhiteSpace(basePath))
-        {
-            throw new InvalidOperationException(
-                "Cannot resolve a workspace override because no workspace base path is configured."
-            );
-        }
-
+        // A workspace override is always a RELATIVE leaf (a workspace identifier), never an absolute
+        // path — reject a rooted value up front, independent of whether a local base is configured.
         if (Path.IsPathRooted(relPathOverride))
         {
             throw new InvalidOperationException(
                 $"Workspace directory '{relPathOverride}' must be relative to the workspace base."
             );
+        }
+
+        var (basePath, _, _) = ResolveWorkspace();
+        if (string.IsNullOrWhiteSpace(basePath))
+        {
+            // No local base is configured — e.g. the client adopts a gateway that may be REMOTE, or
+            // that roots the workspace under its own per-app base (ADR 0029). The client neither
+            // resolves an absolute path nor pre-creates the directory on a filesystem it may not even
+            // share: it forwards the leaf (the workspace identifier) and lets the gateway own
+            // creation. WorkspaceBasePath / WorkspacePath is therefore OPTIONAL, needed only for the
+            // same-host spawn path where the client pre-creates the workspace dir as an optimization.
+            return (null, relPathOverride, null);
         }
 
         var baseFull = Path.TrimEndingDirectorySeparator(Path.GetFullPath(basePath));
--- a/src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
+++ b/src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
@@ -384,23 +384,14 @@ public sealed partial class SandboxSessionRegistry : IAsyncDisposable
             throw new SandboxSessionUnavailableException(workspaceId, statusCode: null, ex.Message, ex);
         }
 
-        var (_, workspaceLeaf, workspaceFullPath) = _options.ResolveWorkspace(workspaceRef.DirectoryRelPath);
+        // Resolve only the workspace LEAF — the identifier sent to the gateway as the `workspace`
+        // field. The client deliberately does NOT touch the workspace filesystem here: the sandbox
+        // gateway may be on a remote machine, and it owns workspace directory creation
+        // (create-if-missing) and mounting. A local WorkspaceBasePath/WorkspacePath is used only by
+        // the spawn path to provision a LOCAL gateway (SandboxGatewayLifetime.EnsureWorkspaceDirectory)
+        // — it is neither required nor resolved to a local path when adopting a (possibly remote) gateway.
+        var (_, workspaceLeaf, _) = _options.ResolveWorkspace(workspaceRef.DirectoryRelPath);
         var workspaceRelPath = workspaceLeaf ?? string.Empty;
-
-        // Ensure the workspace directory exists before the gateway mounts it — the gateway rejects a
-        // leaf that doesn't exist under WORKSPACE_BASE_PATH. Idempotent with the spawn-time creation;
-        // this also covers the adopt-an-existing-gateway path (no spawn ran).
-        if (!string.IsNullOrWhiteSpace(workspaceFullPath))
-        {
-            try
-            {
-                _ = Directory.CreateDirectory(workspaceFullPath);
-            }
-            catch (Exception ex)
-            {
-                _logger.LogWarning(ex, "Could not create workspace directory '{WorkspacePath}'", workspaceFullPath);
-            }
-        }
 
         var requestUri = $"{_gateway.GatewayBaseUrl}/api/v1/sandboxes";
         var (authProviders, network) = BuildAuthProviders();
--- a/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
+++ b/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
@@ -173,6 +173,19 @@ public class SandboxGatewayOptionsTests
     }
 
     [Fact]
+    public void ResolveWorkspace_WithTraversalOverride_Throws_EvenWithNoBaseConfigured()
+    {
+        // Defense-in-depth (PR #165 review): a '..' traversal segment is never a valid workspace
+        // identifier, so it is rejected even with no base configured — the no-base path forwards the
+        // leaf straight to the gateway, so this is the client's only containment guard there.
+        var options = new SandboxGatewayOptions();
+
+        var act = () => options.ResolveWorkspace(Path.Combine("..", "evil"));
+
+        act.Should().Throw<InvalidOperationException>();
+    }
+
+    [Fact]
     public void ResolveWorkspace_WithEscapingOverride_Throws()
     {
         var basePath = Path.Combine(Path.GetTempPath(), "ws-base");
--- a/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
+++ b/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
@@ -143,6 +143,36 @@ public class SandboxGatewayOptionsTests
     }
 
     [Fact]
+    public void ResolveWorkspace_WithOverride_ReturnsLeafWithoutBaseOrFullPath_WhenNoBaseConfigured()
+    {
+        // A remote gateway (or one that roots the workspace under its own per-app base, ADR 0029) has
+        // no local filesystem the client can resolve against or pre-create in. WorkspaceBasePath is
+        // therefore OPTIONAL: with no base configured, resolving an override must NOT throw — it yields
+        // just the leaf (the workspace identifier the gateway needs), with a null base and null full
+        // path so the registry skips local directory pre-creation and lets the gateway own creation.
+        var options = new SandboxGatewayOptions(); // no WorkspaceBasePath, no WorkspacePath
+
+        var (resolvedBase, leaf, full) = options.ResolveWorkspace("projA");
+
+        resolvedBase.Should().BeNull();
+        leaf.Should().Be("projA");
+        full.Should().BeNull();
+    }
+
+    [Fact]
+    public void ResolveWorkspace_WithRootedOverride_Throws_EvenWithNoBaseConfigured()
+    {
+        // The "override must be relative" invariant is independent of whether a base is configured —
+        // a rooted value is never a valid workspace leaf.
+        var options = new SandboxGatewayOptions();
+
+        var rooted = Path.Combine(Path.GetTempPath(), "elsewhere");
+        var act = () => options.ResolveWorkspace(rooted);
+
+        act.Should().Throw<InvalidOperationException>();
+    }
+
+    [Fact]
     public void ResolveWorkspace_WithEscapingOverride_Throws()
     {
         var basePath = Path.Combine(Path.GetTempPath(), "ws-base");
--- a/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
+++ b/tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
@@ -169,7 +169,9 @@ public class SandboxGatewayOptionsTests
         var rooted = Path.Combine(Path.GetTempPath(), "elsewhere");
         var act = () => options.ResolveWorkspace(rooted);
 
-        act.Should().Throw<InvalidOperationException>();
+        // Pin WHICH guard fires: the rooted check must run ahead of the no-base return (pre-PR the
+        // no-base guard threw first with a different message), so this stays sensitive to the reorder.
+        act.Should().Throw<InvalidOperationException>().WithMessage("*must be relative to the workspace base*");
     }
 
     [Fact]
--- a/tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryWorkspaceTests.cs
+++ b/tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryWorkspaceTests.cs
@@ -58,6 +58,23 @@ public class SandboxSessionRegistryWorkspaceTests
         second.WorkspaceRelPath.Should().Be("projA");
     }
 
+    [Fact]
+    public async Task GetOrCreateSession_DoesNotCreateWorkspaceDirectoryOnDisk_GatewayOwnsIt()
+    {
+        // The client must NEVER touch the workspace filesystem — the gateway may be on a remote
+        // machine and owns workspace directory creation (create-if-missing) and mounting. Even with a
+        // local base configured, creating a session forwards the leaf WITHOUT the client mkdir-ing it.
+        using var baseDir = new TempWorkspaceBase();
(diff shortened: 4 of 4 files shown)


gautam-achieveai (commented):
[Gautam's review bot] **APPROVE** — Clean, well-scoped fix for #164.

**Summary**: The change correctly makes `WorkspaceBasePath` optional by returning `(null, leaf, null)` when no base is configured, and removes the client-side `Directory.CreateDirectory` call that was creating orphan directories when the gateway is remote. Tests cover the key invariants (no-base returns leaf without throwing, rooted paths still rejected, no FS writes from registry).

**One observation** (non-blocking): When no base is configured, `../evil` passes through as a leaf without rejection (the escape check only fires when a base exists). This is defensible since the gateway owns containment — but a defense-in-depth check rejecting `..` segments in the leaf regardless of base presence would harden the client against a misconfigured or permissive gateway. Consider for a future pass if the threat model warrants it.

---

**Inline note** on `SandboxGatewayOptions.cs:117` — When no base is configured, a value like `\"../evil\"` passes through here as-is (only the rooted check fires). The gateway is documented as owning containment, so this is correct per the design — but if the threat model ever requires client-side hardening, consider rejecting any `relPathOverride` containing `..` segments here as well (a workspace *identifier* should never need path traversal sequences). Non-blocking.

gautam-achieveai:
**[Gautam's review bot]** Incremental review of iteration 2 (commit df425de6).

**Changes:** Scriban 7.2.0 -> 7.2.5 to remediate NU1902/NU1903 advisories (GHSA-6q7j-xr26-3h2c, GHSA-q6rr-fm2g-g5x8, GHSA-7jvp-hj45-2f2m). The temporary `WarningsNotAsErrors` demotion for NU1902 in `Directory.Build.props` is correctly removed — NuGet audit is back to zero-tolerance.

**Verification:** Solution builds clean (0 warnings, 0 errors) with `TreatWarningsAsErrors=true`.

**Verdict:** :white_check_mark: **APPROVE** — both the sandbox workspace change (iter 1) and the dependency remediation (iter 2) are correct and well-documented.

gautam-achieveai:
Thanks for the review. Applied the non-blocking defense-in-depth suggestion in `a349d3d2`: `ResolveWorkspace(override)` now rejects any `..` traversal segment **regardless of whether a base is configured**. On the no-base path the leaf is forwarded straight to the (possibly remote/permissive) gateway, so this is the client's only containment guard there; with a base it fails fast before the existing escape check. A workspace identifier should never traverse upward — added a no-base traversal-rejection test. Build clean with the NuGet audit on; options tests 14/14.

tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryWorkspaceTests.cs
```diff
@@ -58,6 +58,23 @@ public async Task GetOrCreateSession_FirstCreationWins_SecondCallReturnsCachedSe
         second.WorkspaceRelPath.Should().Be("projA");
     }
 
+    [Fact]
```
gautam-achieveai, line 61: [gautam's bot] [BLOCKER] **HIGH** (Test Coverage / Regression Durability)

The literal thesis of this PR — "a workspace-backed session works with **no `WorkspaceBasePath` configured**" — has **no end-to-end regression test** through `GetOrCreateSessionAsync`. This is exactly the #164 repro (adopt a gateway with no base -> old code threw `InvalidOperationException` -> 500 -> WebSocket crash).

`ResolveWorkspace_WithOverride_ReturnsLeaf...WhenNoBaseConfigured` covers `ResolveWorkspace` **in isolation**, and `GetOrCreateSession_DoesNotCreateWorkspaceDirectoryOnDisk_GatewayOwnsIt` exercises the registry but **with a real base** (`CreateRegistry` always sets one). So the no-base path *through the registry* has zero coverage. A future refactor that reintroduces a base dependency in `CreateSessionAsync` would pass every registry test and silently reopen #164 on a green build.

**Suggestion** — add a registry-level regression test (fails red pre-PR, green post-PR):
```csharp
[Fact]
public async Task GetOrCreateSession_NoBaseConfigured_DoesNotThrow_AndForwardsLeaf()
{
    // #164 repro: adopting a gateway with NO WorkspaceBasePath must NOT throw
    // (the throw became a 500 -> WebSocket crash). Client forwards the leaf; gateway owns creation.
    await using var registry = CreateRegistryNoBase(out var captured);

    var act = async () => await registry.GetOrCreateSessionAsync(new WorkspaceRef("ws-nobase", "projRemote"));

    var session = await act.Should().NotThrowAsync();
    session.Subject.WorkspaceRelPath.Should().Be("projRemote");
    captured.LastWorkspace.Should().Be("projRemote"); // leaf forwarded to the gateway
}
```
Refactor `CreateRegistry` to accept a nullable base (leaves the existing four tests unchanged):
```csharp
private static SandboxSessionRegistry CreateRegistryNoBase(out CapturedRequest captured)
    => CreateRegistry(workspaceBasePath: null, out captured);

private static SandboxSessionRegistry CreateRegistry(string? workspaceBasePath, out CapturedRequest captured)
{
    // ...
    WorkspaceBasePath = workspaceBasePath, // null => the #164 no-base path
    // ...
}
```

src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
```diff
@@ -82,19 +96,25 @@ public sealed class SandboxGatewayOptions
...
-            throw new InvalidOperationException(
-                $"Workspace directory '{relPathOverride}' must be relative to the workspace base."
-            );
+            // No local base is configured — e.g. the client adopts a gateway that may be REMOTE, or
+            // that roots the workspace under its own per-app base (ADR 0029). The client neither
+            // resolves an absolute path nor pre-creates the directory on a filesystem it may not even
+            // share: it forwards the leaf (the workspace identifier) and lets the gateway own
+            // creation. WorkspaceBasePath / WorkspacePath is therefore OPTIONAL, needed only for the
+            // same-host spawn path where the client pre-creates the workspace dir as an optimization.
+            return (null, relPathOverride, null);
```
gautam-achieveai, line 130: [gautam's bot] **MEDIUM** (Security / Path Traversal — defense-in-depth)

In the no-base path, a **relative** override containing `..` (e.g. `"../evil"`) is returned as the leaf **without** the escape-containment check — that check only runs when a base is configured (below). The value then flows verbatim into `CreateSandboxRequest.workspace` and is POSTed to the gateway. So client-side containment is silently dropped **only on this path**, while the base-configured path still enforces it — an asymmetric regression across a deploy/trust boundary.

Whether this is exploitable depends on the gateway (see the `[QUESTION]` on this line). Even if the gateway re-validates, dropping the sibling path's check is a hardening regression worth closing.

**Suggestion** — restore a cheap, FS-free containment check in the no-base branch (does **not** violate "client never FS-accesses the workspace" — pure string check, no `Directory`/`Path.GetFullPath` against a real base):
```csharp
if (string.IsNullOrWhiteSpace(basePath))
{
    // No local base -> forward the leaf; gateway owns creation. Still reject a leaf that is
    // not a self-contained relative segment path: '..' traversal is never a valid workspace id.
    if (relPathOverride.Split('/', '\\').Any(s => s == ".."))
        throw new InvalidOperationException(
            $"Workspace directory '{relPathOverride}' must be a relative path without '..' segments.");
    return (null, relPathOverride, null);
}
```
(If F2 is fixed, write the F4 characterization test against the **post-fix** behavior — `"../evil"` with no base *throws* — not the current pass-through.)

tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
```diff
@@ -142,6 +142,36 @@ public void ResolveWorkspace_WithNullOrEmptyOverride_MatchesParameterless()
...
+    public void ResolveWorkspace_WithRootedOverride_Throws_EvenWithNoBaseConfigured()
+    {
+        // The "override must be relative" invariant is independent of whether a base is configured —
+        // a rooted value is never a valid workspace leaf.
+        var options = new SandboxGatewayOptions();
+
+        var rooted = Path.Combine(Path.GetTempPath(), "elsewhere");
+        var act = () => options.ResolveWorkspace(rooted);
+
+        act.Should().Throw<InvalidOperationException>();
```
gautam-achieveai, line 187: [gautam's bot] **LOW** (Test Quality — signal integrity)

This test asserts only `Throw<InvalidOperationException>()`. But **pre-PR this call already threw** `InvalidOperationException` — via the *no-base* guard, which ran first and never reached the rooted guard (different message). So the test is **green on both sides of the change it purports to protect** and contributes zero signal about the actual substance here: moving the rooted guard *ahead of* the base check.

**Suggestion** — pin *which* guard fires so the test is sensitive to the reorder (would have failed pre-PR, where the message was the no-base one):
```csharp
act.Should().Throw<InvalidOperationException>()
   .WithMessage("*must be relative to the workspace base*");
```

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -384,24 +384,15 @@ private async Task<SandboxSession> CreateSessionAsync(WorkspaceRef workspaceRef,
...
         }
 
-        var (_, workspaceLeaf, workspaceFullPath) = _options.ResolveWorkspace(workspaceRef.DirectoryRelPath);
+        // Resolve only the workspace LEAF — the identifier sent to the gateway as the `workspace`
+        // field. The client deliberately does NOT touch the workspace filesystem here: the sandbox
+        // gateway may be on a remote machine, and it owns workspace directory creation
+        // (create-if-missing) and mounting. A local WorkspaceBasePath/WorkspacePath is used only by
+        // the spawn path to provision a LOCAL gateway (SandboxGatewayLifetime.EnsureWorkspaceDirectory)
+        // — it is neither required nor resolved to a local path when adopting a (possibly remote) gateway.
+        var (_, workspaceLeaf, _) = _options.ResolveWorkspace(workspaceRef.DirectoryRelPath);
```
gautam-achieveai, line 393: [gautam's bot] [QUESTION] **Clarification Needed** (Rollout / Deploy Compatibility)

This line's hunk removes the client-side `Directory.CreateDirectory` on the adopt path, so the client now depends on the **gateway** to create the workspace directory if missing (ADR 0029). There is no version gate / capability handshake in the client (only a `GET /health` liveness probe).

**Question:** Is ADR 0029 "create-if-missing + per-app rooting" already deployed on **every** gateway this client can reach, and is the deploy ordered **gateway-first** (gateway never older than the client)?

**Why this matters:** A new client talking to an **older** gateway that required the workspace dir to pre-exist would fail session creation (new-client -> old-server; old-client -> new-server is safe). If the answer is "yes, fleet-wide + gateway-first," this is a non-issue. If not, the mkdir removal is premature and needs a capability/version gate (or keep the idempotent same-host mkdir when a local base is set).

**Possible answers:** (a) Yes — all gateways are >= the ADR-0029 build and we deploy gateway-first. (b) No — some gateways still require pre-existing dirs; we'll gate/keep the client mkdir until the fleet is upgraded. (c) Client + gateway ship from one release train with no independent lag.

src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
gautam-achieveai, line 130: [gautam's bot] [QUESTION] **Clarification Needed** (Trust Boundary — server-side containment)

With no base configured, this line returns the raw leaf (including a `..`-containing value) and the client no longer runs the escape check (see the MEDIUM finding here). The value is sent to the gateway as `CreateSandboxRequest.workspace`.

**Question:** Does the gateway **independently** canonicalize and re-root the `workspace` leaf under its per-app base and reject/neutralize `..` traversal, or does it **trust** `workspace` as an already-contained value (as every prior client guaranteed)?

**Why this matters:** It determines whether skipping the client-side check is a true path-traversal risk (gateway trusts the client -> escalate to HIGH, server-side containment required) or a defense-in-depth-only gap (gateway re-validates -> MEDIUM). Either way the client should not be the sole enforcer for a value crossing an independent deploy boundary.

**Possible answers:** (a) Gateway hash-roots + rejects traversal server-side; client value is untrusted (defense-in-depth only). (b) Gateway trusts `workspace` as contained -> restore the client check **and** add server-side containment.

gautam-achieveai:
[gautam's bot]
# PR Review: Make sandbox WorkspaceBasePath optional; client never FS-accesses the workspace

**Verdict: 🔴 REQUEST CHANGES** — initial review.

## What this PR does (and does well)
Makes the sandbox client treat the workspace as **remote-owned**: `ResolveWorkspace(override)` now returns `(null, leaf, null)` instead of throwing when no base is configured, and `SandboxSessionRegistry` stops pre-creating the workspace directory (the gateway owns creation). This correctly fixes the #164 crash for remote / per-app-rooting gateways.

Genuinely good aspects:
- **Scope is tight and on-target** — every source/test hunk maps 1:1 to #164. The guard reorder (rooted-check moved above the base check) is **load-bearing**, not gratuitous: it preserves the "rooted always rejected" invariant now that the no-base branch returns early.
- **Clean deletion** — the removed `try/catch` went away *with* the operation it guarded (the mkdir); surrounding handlers (domain-wrap + cancellation-aware) are correctly left intact. No debug artifacts, no leftover code.
- **Bonus remediation in the latest push** — bumps `Scriban 7.2.0 -> 7.2.5` (patches GHSA-6q7j-xr26-3h2c / GHSA-q6rr-fm2g-g5x8) and removes the now-unneeded `NU1902` build suppression. Net-positive; resolves the advisory the description flagged as "tracked separately." (Minor nit: the props comment says `NU1902` while the PR body earlier said `NU1903` — align the reference.)

## Why REQUEST CHANGES
One blocker and one security-adjacent gap, plus two open questions the client diff cannot resolve.

| # | Sev | Category | File | Issue |
|---|-----|----------|------|-------|
| F1 | **BLOCKER** (HIGH) | Test / regression durability | `SandboxSessionRegistryWorkspaceTests.cs` | The no-base path — the PR's whole point and the #164 repro — has **no end-to-end registry test**. `CreateRegistry` always sets a base, so a reintroduced base dependency would reopen #164 on a green build. |
| F2 | MEDIUM | Security (path traversal, defense-in-depth) | `SandboxGatewayOptions.cs:117` | No-base path forwards a `..`-containing leaf **without** the escape check the base path still runs. Escalates to **HIGH** if the gateway trusts the value (see Q2). |
| F3 | MEDIUM | Rollout / deploy | `SandboxSessionRegistry.cs:393` | mkdir removal makes the client depend on gateway create-if-missing with no version/capability gate. Tracked as **Q1**. |
| F4 | LOW | Test (characterization) | `SandboxGatewayOptionsTests.cs` | `"../evil"` with no base is untested — pin the intended contract (author against the F2 post-fix behavior). |
| F5 | LOW | Test signal integrity | `SandboxGatewayOptionsTests.cs:172` | Rooted-override test is green pre- and post-PR; add `.WithMessage(...)` so it actually pins the guard reorder. |
| F6 | INFO/nit | Conventions | `SandboxGatewayOptions.cs` | `InvalidOperationException` vs `ArgumentException` for arg validation — leave as-is; the module uses it uniformly. |

F1, F2, and F5 are posted inline. F4/F6 are captured here (non-blocking, low signal). See the two inline `[QUESTION]` threads (Q1, Q2) — non-blocking, but the answers determine whether F2/F3 escalate.

## To clear the verdict
1. **F1** — add the no-base registry regression test (required).
2. Answer **Q1** and **Q2**. If Q2 comes back "gateway trusts the value," also apply **F2** (restore the FS-free `..` check).
3. Optional polish: **F5** (message assertion), **F4** (characterization test).

Severities were confirmed by a grading pass; F1 was escalated to a merge-blocker because a bug fix without a regression test that fails on the old code has not durably fixed the bug.

gautam-achieveai:
[gautam's bot]
## Re-Review Summary: PR #165 — correction (review raced with your pushes)

Heads-up: my inline comments above were generated against commit `df425de6` (iteration 2), but you were pushing while I reviewed. The PR is now at **`a349d3d2`** (iteration 3, `harden(sandbox): reject '..' traversal…`). Re-anchoring here and reconciling — a couple of my inline notes are now **stale/resolved**. Apologies for the noise.

### ✅ Already fixed in `a349d3d2` — please disregard these inline comments
- **F2 (MEDIUM, path-traversal defense-in-depth)** and its **[QUESTION]** on `SandboxGatewayOptions.cs` — **resolved.** `a349d3d2` adds exactly the guard I suggested: `Array.Exists(relPathOverride.Split('/', '\'), s => s == "..")` throwing `InvalidOperationException`, placed **before** the base check so it applies on the no-base path. This is the client's containment guard for the remote/permissive-gateway case. 👍
- **F4 (LOW, characterization test)** — **resolved.** `ResolveWorkspace_WithTraversalOverride_Throws_EvenWithNoBaseConfigured` now pins that behavior.
- **Q2 (gateway trust boundary)** — largely **mooted** as a client concern now that the client rejects `..` itself. Still worth confirming the gateway independently re-roots/validates `workspace` as pure defense-in-depth, but it is **no longer gating**.

### 🔴 Still open at `a349d3d2`
- **F1 — BLOCKER (HIGH): the no-base path still has no end-to-end registry test.** This is the one that matters. The new traversal test and the iter-1 tests exercise `ResolveWorkspace` **in isolation**; `CreateRegistry(string workspaceBasePath, …)` still *always* sets a base (line 78/89), so **no test drives `GetOrCreateSessionAsync` with NO `WorkspaceBasePath`** — the exact #164 repro (throw → 500 → WebSocket crash). See the inline BLOCKER for the drop-in test + the `CreateRegistryNoBase` refactor. This is a small add and it's the regression guard for the bug this PR exists to fix.
- **F5 — LOW (test signal):** `ResolveWorkspace_WithRootedOverride_Throws_EvenWithNoBaseConfigured` still asserts only the exception *type*; add `.WithMessage("*must be relative to the workspace base*")` so it pins the guard reorder (green pre- and post-PR otherwise). Inline above.
- **Q1 — rollout/deploy:** the mkdir removal makes the client depend on gateway create-if-missing (ADR 0029) with no version/capability gate. Non-blocking, but please confirm ADR-0029 gateways lead the client in the deploy order (inline `[QUESTION]` on `SandboxSessionRegistry.cs:393`).
- **F6 — nit:** `InvalidOperationException` vs `ArgumentException` — leave as-is; the module is uniform.

### Verdict: 🔴 REQUEST CHANGES → clears to ✅ on **F1 alone**
Everything else is optional polish or already done. The security item you proactively fixed was the right call. Add the no-base **registry** regression test (F1) and I'm happy to approve.

(Note: another review bot posted an APPROVE on iter 2 earlier. This review adds the F1 test-coverage blocker that pass didn't flag — a bug fix without a regression test that fails on the *old* code hasn't durably fixed the bug. Reconcile before merging.)

tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryWorkspaceTests.cs
gautam-achieveai: Fixed in d9f837d7. Added `GetOrCreateSession_NoBaseConfigured_DoesNotThrow_AndForwardsLeaf`, which drives the #164 repro **through** `GetOrCreateSessionAsync` with no `WorkspaceBasePath` (via a new `CreateRegistryNoBase`), asserting it does not throw and the leaf is forwarded to the gateway. `CreateRegistry` now takes a nullable base; the existing four tests are unchanged. Red on `main` (throws), green with the fix.

src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
gautam-achieveai: Closed in de8b8b70 (pushed just before this review pass). `ResolveWorkspace(override)` now rejects any `..` traversal segment **regardless of base** — a pure string check (`relPathOverride.Split(...).Any(s => s == "..")`, no `Directory`/`Path.GetFullPath`), so it does not violate "client never FS-accesses the workspace". It runs right after the rooted check, so both the no-base and with-base paths enforce it — the asymmetry is closed. Characterization test `ResolveWorkspace_WithTraversalOverride_Throws_EvenWithNoBaseConfigured` asserts the post-fix behavior (`../evil` with no base throws).

tests/LmStreaming.Sample.Tests/Services/SandboxGatewayOptionsTests.cs
gautam-achieveai: Fixed in d9f837d7. The no-base rooted test now asserts `.WithMessage("*must be relative to the workspace base*")`, pinning that the rooted guard fires (pre-PR the no-base guard threw first, with a different message) — so the test is sensitive to the guard reorder.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: Answer: (a), with a caveat. This is a sample app whose sandbox gateway is dev tooling, not an independently-versioned production fleet — there is no client/gateway lag to gate against. Empirically the current enforcing (ADR-0029) gateway DOES create-if-missing + per-app-root: verified live — adopting the enforcing :3003 gateway with **no** base created the session, and the gateway created the workspace dir under its per-app hash (`<app>-<hash>/<leaf>`). Fallback for an older gateway that required a pre-existing dir: set `WorkspaceBasePath` (the spawn/same-host pre-creation path is unchanged). A capability handshake would be over-engineering for a sample; happy to add a version gate if this client ever targets a mixed fleet.

src/LmAgentInfra/Sandbox/SandboxGatewayOptions.cs
gautam-achieveai: Answer: the client no longer drops the check — de8b8b70 rejects `..` before the value is sent, so the client-side asymmetry is closed regardless of the gateway. Whether the gateway ALSO canonicalizes server-side is the gateway repo (SandboxedOstoolsMcpServer) concern: ADR 0029 per-app rooting re-roots `workspace` under the gateway's own per-app base, which should neutralize traversal, but I cannot assert its exact `..` handling from here. With the client now rejecting `..`, it is not the sole enforcer; if the gateway is found to trust `workspace` verbatim, that is a separate gateway-side hardening. Treating as (a) for this PR.

gautam-achieveai:
[Gautam's review bot] **APPROVE** — Iteration 3 (incremental review of de8b8b70..d9f837d7)

All 5 findings from the previous review pass are addressed:

| Finding | Resolution |
|---------|-----------|
| **BLOCKER** — No e2e test for no-base path | ✅ `GetOrCreateSession_NoBaseConfigured_DoesNotThrow_AndForwardsLeaf` drives #164 repro through the registry |
| **MEDIUM** — `../evil` forwarded unchecked in no-base path | ✅ `de8b8b70` rejects `..` segments regardless of base — defense-in-depth closed |
| **LOW** — Rooted test doesn't pin which guard fires | ✅ `.WithMessage()` pinned |
| **Q** — Deploy compatibility | ✅ Answered (sample app, no client/gateway version lag) |
| **Q** — Trust boundary asymmetry | ✅ Closed (client now rejects `..` before forwarding) |

The `..` check implementation (`Array.Exists(split on '/' and '\\', == "..")`) is correct and handles both separator styles. Guard ordering (rooted → traversal → base-escape) is sound. 19 tests pass.

No new findings. Ship it. 🚢

gautam-achieveai:
[gautam's bot]
## Re-Review Summary: PR #165 — iteration 3 (`d9f837d7`)

Re-reviewed the delta since my last pass (`a349d3d2` → `d9f837d7`; branch rebased, the Scriban/build-props changes dropped out of scope — fine). **Every item I raised is now resolved, and both open questions are answered.** Verdict flips **REQUEST CHANGES → ✅ APPROVE WITH COMMENTS**.

### Resolved
- **F1 (was BLOCKER)** ✅ — `GetOrCreateSession_NoBaseConfigured_DoesNotThrow_AndForwardsLeaf` drives the #164 repro **through** `GetOrCreateSessionAsync` with no `WorkspaceBasePath` (new `CreateRegistryNoBase`; `CreateRegistry` now takes a nullable base, existing tests unchanged). Verified the test is non-vacuous: with `WorkspaceBasePath=null`, `ResolveWorkspace("projRemote")` hits the no-base branch and returns `(null, "projRemote", null)` — red on `main` (threw), green now. This is the regression guard the fix needed.
- **F5 (LOW)** ✅ — rooted-override test now asserts `.WithMessage("*must be relative to the workspace base*")`, so it's sensitive to the guard reorder.
- **F2 / F4** ✅ — the `..`-traversal guard (`de8b8b70`) and its characterization test survive the rebase and are intact in `ResolveWorkspace`.

### Questions answered (both non-blocking)
- **Q1 (rollout/deploy)** → **(a)**, well-justified: this is a sample-app dev-tooling gateway with no independently-versioned fleet to gate against, and you verified live that the ADR-0029 gateway does create-if-missing + per-app rooting. Fallback (set `WorkspaceBasePath`) documented. Agree a capability handshake would be over-engineering here.
- **Q2 (server-side containment)** → **(a)**: the client now rejects `..` before send, so it's no longer the sole enforcer; server-side canonicalization is a separate gateway-repo concern. Reasonable. (Left as a note for the gateway side, not this PR.)

### Verified intact at head
Production `ResolveWorkspace` guard order is correct (rooted → `..` → no-base early-return → base-escape check), and the registry mkdir removal is unchanged.

### Non-blocking residual
- **F6 (nit):** `InvalidOperationException` vs `ArgumentException` for arg validation — leave as-is; the module is uniform.
- The Q2 server-side hardening lives in the gateway repo (SandboxedOstoolsMcpServer) — worth a follow-up there, but out of scope for this client PR.

Nice, tight turnaround — the added tests are exactly the right shape (end-to-end at the registry level, not just `ResolveWorkspace` in isolation). No blockers remain. LGTM. 👍

PR #168, opened 2026-07-08
feat(sandbox): authenticate clients to the gateway with per-app bearer (ADR 0029)

gautam-achieveai:
## Context

The sandbox gateway added **per-app bearer authentication** (gateway work-item #109 / ADR 0029). Under `AUTH_ENFORCE` (default **on**), every app-facing request must carry `X-Sbx-App-Id` + `X-Sbx-App-Key`; the gateway binds `app.id` from the credential and scopes each sandbox session to that app (foreign/absent session → `404`, bad/missing creds → `401`).

LmDotnetTools' gateway clients only sent `X-Session-ID` — they work today solely because the gateway they hit runs `AUTH_ENFORCE=off`. This change makes them authenticate under enforcement, with a **distinct app id per client**.

## What changed

**Shared primitives (`src/LmAgentInfra/Sandbox`)**
- `SandboxGatewayOptions.AppKey` — base64 per-app secret (SECRET-doc'd), companion to the existing `AppId`.
- `GatewayAuthHeaders` — single owner of the header names + the "attach only when a key is configured" rule; `ForMcp(sessionId, appId, appKey)` builds the MCP transport header map, `Apply(...)` for the REST path.
- `GatewayAuthHandler : DelegatingHandler` — attaches the two headers to every REST call (mirrors the existing `OperationPolicyHandler`).

**Code-Review Daemon** (app id `code-review-daemon`, env `CRD_SANDBOX_APP_ID` / `CRD_SANDBOX_APP_KEY`)
- REST clients (session registry + gateway lifetime) wrapped with `GatewayAuthHandler`; the two MCP transports (`SandboxOrchestrator` diff-only path, `LiveReviewAgentLoopFactory` tool-assisted path via new `ReviewToolContext.AppId/AppKey`) build headers with `ForMcp`.

**LmStreaming.Sample** (app id `lmstreaming-sample`, `SandboxGateway__AppKey`)
- The three gateway REST clients wrapped via a small local factory; both MCP header maps use `ForMcp`; `AppKey` knob added to `appsettings.Development.json` and documented in `SandboxWorkspaceGuide.md`.

## Backward compatible

Both headers attach **only when `AppKey` is non-empty**. With no key configured the clients send exactly what they did before, so existing `AUTH_ENFORCE=off` deployments are unaffected. Enabling enforcement is an operator step (register the app id + `openssl rand -base64 32` secret in the gateway's `APP_SECRETS*`, set the matching key here).

## Tests

- `GatewayAuthHeadersTests` / `GatewayAuthHandlerTests` — header logic + the DelegatingHandler (attach-when-configured, omit-when-unset, don't-overwrite).
- `GatewayAppAuthWiringTests` — `AppKey` binds from the `SandboxGateway` config section; a handler-wrapped `MarketplaceCatalogClient` emits both headers.
- Full solution builds **0 warnings / 0 errors**; daemon suite **442 green**, LmStreaming suite **502 green**.

## Scope

Client-side only. The gateway/operator side (`APP_SECRETS*`, `AUTH_ENFORCE`, proxy `PROXY_SERVICE_CRED`, deploy order) is configured out-of-band.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
+++ b/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
@@ -223,7 +223,11 @@ internal sealed class DaemonReviewStageExecutor : IReviewStageExecutor
                 EnableReviewerWrites: writeScope.Enabled,
                 WritableToolAllowList: writeScope.WritableAllow,
                 NotesDir: writeScope.NotesDir,
-                ScratchDir: writeScope.ScratchDir);
+                ScratchDir: writeScope.ScratchDir,
+                // Per-app bearer identity for the agent's MCP calls to the gateway (ADR 0029), read from the
+                // same env as GatewayBaseUrl so the tool-assisted session authenticates as the daemon.
+                AppId: Environment.GetEnvironmentVariable("CRD_SANDBOX_APP_ID") ?? "code-review-daemon",
+                AppKey: Environment.GetEnvironmentVariable("CRD_SANDBOX_APP_KEY"));
         }
         catch (Exception ex) when (ex is not OperationCanceledException)
         {
--- a/samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs
+++ b/samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs
@@ -4,6 +4,7 @@ using AchieveAi.LmDotnetTools.AnthropicProvider.Models;
 using AchieveAi.LmDotnetTools.GithubCopilotProvider.Agents;
 using AchieveAi.LmDotnetTools.GithubCopilotProvider.Auth;
 using AchieveAi.LmDotnetTools.LmAgentInfra;
+using AchieveAi.LmDotnetTools.LmAgentInfra.Sandbox;
 using AchieveAi.LmDotnetTools.LmCore.Agents;
 using AchieveAi.LmDotnetTools.LmCore.Core;
 using AchieveAi.LmDotnetTools.LmCore.Middleware;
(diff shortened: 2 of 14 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review complete — clean, well-tested implementation of per-app bearer auth (ADR 0029).

## Summary
Adds GatewayAuthHeaders (static header builder) + GatewayAuthHandler (DelegatingHandler) as shared primitives in LmAgentInfra, wires them through both CodeReviewDaemon and LmStreaming.Sample for REST and MCP paths, with good backward compatibility (no-op when AppKey is unset). Test coverage is solid — 3 new test files covering the primitives, handler behavior, and end-to-end wiring through MarketplaceCatalogClient.

## Findings

### MEDIUM — DRY: duplicated env-var reads for daemon app identity

DaemonReviewStageExecutor.cs (line ~226) reads CRD_SANDBOX_APP_ID / CRD_SANDBOX_APP_KEY from the environment independently from Program.cs (line ~117), both with the same 'code-review-daemon' default. If the variable name or default changes in one place, the other drifts silently. Consider injecting the resolved sandboxAppId/sandboxAppKey into the executor (e.g. via CodeReviewDaemonOptions or a dedicated record) rather than re-reading Environment.GetEnvironmentVariable in a nested class.

### LOW — No base64 validation on AppKey at bind-time

If a user sets a non-base64 value (typo, wrong encoding), it is sent as-is and rejected by the gateway only at runtime. A startup Convert.TryFromBase64String check or a warning log at configuration binding would surface misconfigurations earlier — but this is non-blocking since the gateway will reject it cleanly.

---

Verdict: **APPROVE** — well-structured, backward-compatible, follows existing patterns.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -223,7 +223,11 @@ public DaemonReviewStageExecutor(
                 EnableReviewerWrites: writeScope.Enabled,
                 WritableToolAllowList: writeScope.WritableAllow,
                 NotesDir: writeScope.NotesDir,
-                ScratchDir: writeScope.ScratchDir);
+                ScratchDir: writeScope.ScratchDir,
+                // Per-app bearer identity for the agent's MCP calls to the gateway (ADR 0029), read from the
+                // same env as GatewayBaseUrl so the tool-assisted session authenticates as the daemon.
+                AppId: Environment.GetEnvironmentVariable("CRD_SANDBOX_APP_ID") ?? "code-review-daemon",
```
gautam-achieveai, line 229: **[Gautam's review bot]** DRY: This reads CRD_SANDBOX_APP_ID/CRD_SANDBOX_APP_KEY a second time (the first is in Program.cs line ~117, same default 'code-review-daemon'). If the env-var name or the default ever changes in one place, this silently diverges.

Consider injecting the resolved values (e.g. threading them through CodeReviewDaemonOptions or adding them to the constructor dependencies) rather than re-reading from the environment in a nested executor.

PR #169, opened 2026-07-08
fix(daemon): don't post an empty review — it poisons the head's dedup slot

gautam-achieveai:
## Symptom

A real review never appears on a PR that previously received an empty one. Found live: a fresh **opus** daemon produced a real 4000-char review of #157 but the log showed `already posted as 4895617153 (found via backstop scan); not re-posting` — where `4895617153` is an **empty** `_No review content was produced._` comment a prior gpt-5.5 run had left.

## Root cause (not the old comment — the mechanism that created it)

`DaemonReviewStageExecutor.PostAsync` ran an empty-content review through the full idempotency path:

```csharp
var body = string.IsNullOrWhiteSpace(reviewText) ? "_No review content was produced._" : reviewText;
// ... posts `body` with a head_sha-scoped IdempotencyKey + marker
```

That places a **marked comment** on the provider for the review's `head_sha`. `ReviewPoster`'s backstop scan (`FindPostedCommentAsync`) later finds that marker and **adopts** it — so a subsequent run that *does* produce content (a re-run, or a different reviewer model) is treated as "already posted" and its real review is silently dropped. An empty review permanently claims the head's post slot.

## Fix

Skip the post **and** the retention when the review produced no content:
- No placeholder comment, no idempotency marker → a later real review of the same `head_sha` posts cleanly.
- The run still reaches its terminal stage (its `review_run` row prevents re-review), and the pooled slot + sandbox session are still freed — so nothing leaks or loops.

## Tests

The existing test `Posted_substitutes_a_placeholder_body_when_the_review_is_empty` **enshrined the buggy behavior** (asserted the placeholder was posted). Flipped it to `Posted_does_not_post_a_comment_when_the_review_is_empty` (asserts `PostedBodies` is empty) — RED against the old code, GREEN after the fix. Full daemon suite **442 green**, 0 warnings.

## Note on already-poisoned PRs

This prevents *future* poisoning. PRs that already carry an old empty comment (#157, #88) still need that stale comment deleted for a real review to post — handled operationally, separately from this code fix.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
+++ b/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
@@ -1057,60 +1057,79 @@ internal sealed class DaemonReviewStageExecutor : IReviewStageExecutor
 
         var poster = new ReviewPoster(publisher, _store, _loggerFactory.CreateLogger<ReviewPoster>());
 
-        // ReviewPoster requires a non-blank body even on the collect-only path (it still builds the
-        // idempotency key and records the outbox row), so fall back when the agent produced nothing.
+        // A review that produced NO content must post NOTHING. A "_No review content was produced._"
+        // placeholder would go through the full idempotency path and leave a marked comment on the provider;
+        // the backstop scan (ReviewPoster.FindPostedCommentAsync) later ADOPTS that placeholder and
+        // permanently suppresses a real review of the same head_sha — e.g. a re-run on a model that does
+        // produce content. So skip both the post and the retention when the review is empty. The run still
+        // reaches its terminal stage below (and its review_run row prevents re-review), and the slot/session
+        // are still freed, so nothing is leaked or looped.
         var reviewText = ReadReviewText(run.Id);
-        var body = string.IsNullOrWhiteSpace(reviewText) ? "_No review content was produced._" : reviewText;
-        // The POSTED comment is prefixed with "[BotName]" so a reader knows the content was authored by
-        // the bot on behalf of whichever OAuth app/person's credential actually posted it — the retained
-        // artifact (CommitPooledNotesAsync / PublishToReviewBotAsync below) keeps the raw, unprefixed body.
-        var postedBody = $"[{_options.BotName}]\n\n{body}";
+        var hasContent = !string.IsNullOrWhiteSpace(reviewText);
+        if (hasContent)
+        {
+            // The POSTED comment is prefixed with "[BotName]" so a reader knows the content was authored by
+            // the bot on behalf of whichever OAuth app/person's credential actually posted it — the retained
+            // artifact (CommitPooledNotesAsync / PublishToReviewBotAsync below) keeps the raw, unprefixed body.
+            var postedBody = $"[{_options.BotName}]\n\n{reviewText}";
 
-        var key = new IdempotencyKeyComponents(
-            Provider: provider,
-            OrgOrOwner: repo.OrgOrOwner,
-            Project: repo.Project,
-            // RepoStableId must be non-blank and ':'-free; fall back to the colon-free normalized key.
-            RepoStableId: string.IsNullOrWhiteSpace(repo.RepoStableId) ? repo.NormalizedKey : repo.RepoStableId,
-            PrId: run.PrId,
-            Operation: ReviewPoster.PostReviewCommentOperation,
-            ArtifactKind: ReviewArtifactKind,
-            ArtifactSubject: "summary",
-            // Scope the key to the reviewed COMMIT. head_sha is stable across re-polls and — unlike the PR
-            // updated_at — is not mutated by posting the comment, so a re-poll of the same commit resolves
-            // to the same key and the backstop scan recognizes the already-posted comment (no duplicate).
-            HeadSha: run.HeadSha,
-            VariantId: run.VariantId);
+            var key = new IdempotencyKeyComponents(
+                Provider: provider,
+                OrgOrOwner: repo.OrgOrOwner,
+                Project: repo.Project,
+                // RepoStableId must be non-blank and ':'-free; fall back to the colon-free normalized key.
+                RepoStableId: string.IsNullOrWhiteSpace(repo.RepoStableId) ? repo.NormalizedKey : repo.RepoStableId,
+                PrId: run.PrId,
+                Operation: ReviewPoster.PostReviewCommentOperation,
+                ArtifactKind: ReviewArtifactKind,
+                ArtifactSubject: "summary",
+                // Scope the key to the reviewed COMMIT. head_sha is stable across re-polls and — unlike the PR
+                // updated_at — is not mutated by posting the comment, so a re-poll of the same commit resolves
+                // to the same key and the backstop scan recognizes the already-posted comment (no duplicate).
+                HeadSha: run.HeadSha,
+                VariantId: run.VariantId);
 
-        var request = new PostReviewRequest(
-            run.Id,
-            key,
-            new ReviewCommentTarget(repo, run.PrId),
-            postedBody,
-            LivePostingAuthorized: _options.EnableCommentPosting);
+            var request = new PostReviewRequest(
+                run.Id,
+                key,
+                new ReviewCommentTarget(repo, run.PrId),
+                postedBody,
+                LivePostingAuthorized: _options.EnableCommentPosting);
 
-        _ = await poster.PostReviewAsync(request, cancellationToken).ConfigureAwait(false);
+            _ = await poster.PostReviewAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        else
+        {
+            _logger.LogWarning(
+                "Run {RunId}: review produced no content; not posting a comment or claiming the head's dedup "
+                    + "slot (a placeholder would block a later real review of the same commit).",
+                run.Id);
+        }
 
-        // Retention (design §4.4, the commit gate). A run that leased a pooled slot commits its notes onto
-        // the slot's store checkout scoped to ONLY the PR notes dir, then returns the slot; every other run
-        // uses the host ReviewBot retention checkout exactly as before. The slot is returned even if the
-        // commit throws, and the atomic TryRemove guards against a double-return.
+        // Retention (design §4.4, the commit gate) — only when there is content to retain. A run that leased a
+        // pooled slot commits its notes onto the slot's store checkout scoped to ONLY the PR notes dir, then
+        // returns the slot; every other run uses the host ReviewBot retention checkout. The slot is ALWAYS
+        // returned (finally) and the session ALWAYS torn down (below), so an empty review still frees its
+        // resources; the atomic TryRemove guards against a double-return.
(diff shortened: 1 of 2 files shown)


gautam-achieveai:
[Revobot]

# Review of achieveai/LmDotnetTools#169

Reviewed PR #169 across 2 files — 0 Must, 0 Should, 1 Consider.

## Summary
This PR fixes a real suppression bug in the code-review daemon's `PostAsync`. Previously, a review that produced no prose still posted a `[BotName] _No review content was produced._` placeholder through the full idempotency/backstop path. That placeholder claimed the `head_sha` dedup slot, and the backstop scan (`ReviewPoster.FindPostedCommentAsync`) would later *adopt* it — permanently suppressing a genuine review of the same commit (e.g. a re-run on a model that does produce content). The fix skips the post, the idempotency-key write, and retention (`CommitPooledNotesAsync` / `PublishToReviewBotAsync`) when the review text is blank, logging a warning instead.

## Correctness / safety
The change is sound and I verified its core safety claim against the surrounding code:

- Run identity in `ReviewStore.CreateOrGetReviewRun` is `INSERT OR IGNORE` keyed on `(repo_id, pr_id, head_sha, base_sha, review_kind, variant_id, mode)`. So the `review_run` row — not the outbox row — is what prevents re-review of the same commit. The PR's assertion that "its review_run row prevents re-review" is accurate, meaning skipping the outbox write for an empty review leaks and loops nothing.
- Slot return remains inside a `finally` and session teardown stays unconditional below the `hasContent` branch, so an empty review still frees its slot and tears down its session.
- No reconcile path depends on a `Collected` outbox row existing for an empty review, so dropping that write has no downstream effect.
- The comments in the diff are accurate and match the code; no prompt-injection content was present in the reviewed files.

## Consider

- **Consider — the empty-review path's resource-freeing is untested.** `DaemonReviewStageExecutorTests.cs:628` (`Posted_does_not_post_a_comment_when_the_review_is_empty`) only asserts `PostedBodies.Should().BeEmpty()`. The comment's strongest safety claims — slot returned, session torn down, run reaches terminal stage on the empty path — are not asserted by any test. A future refactor that accidentally moved slot-return or teardown inside the `hasContent` guard would still pass this test. Consider adding an assertion that the run reached terminal stage (and, in the pooled fixture, that the slot was returned). Low severity: current code keeps those outside the guard, so this is regression defense rather than a present bug.

## Verdict
Approve.

gautam-achieveai (commented):
[Gautam's review bot]

**APPROVE** — clean, minimal, well-reasoned fix.

Verified the core safety claims independently:
- `review_run` row (not the outbox) is the re-review gate — skipping the outbox write for empty reviews is safe
- Slot return stays in `finally` — unconditional
- Session teardown stays below the content guard — unconditional
- The `else if (hasContent)` correctly skips `PublishToReviewBotAsync` on the non-pooled path too

Agree with Revobot's existing Consider item about adding regression assertions for slot-return/teardown on the empty path — low priority but worthwhile defense.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/DaemonReviewStageExecutorTests.cs
```diff
@@ -625,17 +625,19 @@ public async Task Posted_throws_when_no_publisher_matches_the_provider()
...
+        // placeholder and permanently suppress a later REAL review of the same commit (e.g. a re-run on a
+        // model that actually produces content). So an empty review must post NOTHING.
         fixture.Factory.TextByProfileId[DaemonAgentFactory.ReviewProfileId] = string.Empty;
         var run = fixture.SeedRun(watermark: "2026-06-29T12:34:56Z");
 
         await RunAllStagesAsync(fixture, run);
 
-        fixture.GitHubPublisher.PostedBodies.Should().ContainSingle()
-            .Which.Should().Be("[Revobot]\n\n_No review content was produced._");
+        fixture.GitHubPublisher.PostedBodies.Should().BeEmpty("an empty review must not claim the head's dedup slot");
```
gautam-achieveai, line 640: [Claude's review bot] **MEDIUM** (Test coverage — behavioral gap)

This assertion guards the *visible symptom* (`PostedBodies` empty), but the invariant the PR exists to protect is stronger: **an empty review must not claim the head's dedup slot**. `ReviewPoster.PostReviewAsync` writes the idempotency/outbox row via `EnqueueOutbox` *before* any publish, and on the default collect-only path (`LivePostingAuthorized: false`) it records that row **without ever populating `PostedBodies`**. So a future regression that re-introduced a placeholder post under collect-only semantics would claim the slot yet keep this test green — the test would miss exactly the class of bug this PR fixes.

**Suggestion:** also assert that no post-review-comment outbox row was recorded, using the `GetOutboxForRun` surface already used elsewhere in this file (e.g. lines 510/570/592):

```csharp
await RunAllStagesAsync(fixture, run);

fixture.GitHubPublisher.PostedBodies.Should().BeEmpty("an empty review must not claim the head's dedup slot");

// Prove the dedup slot itself was not claimed — the actual invariant, not just the visible comment.
fixture.Store.GetOutboxForRun(run.Id)
    .Should().NotContain(e => e.Operation == ReviewPoster.PostReviewCommentOperation,
        "an empty review must not record an idempotency marker for the head_sha");
```

Non-blocking — the shipped production behavior is correct and already has a (weaker) passing regression test; this just hardens the net around the PR's own core claim.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -1057,60 +1057,79 @@ private async Task PostAsync(ReviewRun run, CancellationToken cancellationToken)
...
+                new ReviewCommentTarget(repo, run.PrId),
+                postedBody,
+                LivePostingAuthorized: _options.EnableCommentPosting);
 
-        _ = await poster.PostReviewAsync(request, cancellationToken).ConfigureAwait(false);
+            _ = await poster.PostReviewAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        else
+        {
+            _logger.LogWarning(
```
gautam-achieveai, line 1103: [Claude's review bot] **LOW** (Observability / structured logging)

The empty-review warning captures only `{RunId}`. When triaging *why* a real commit got no review (distinguishing a genuinely trivial diff from a model that silently returned nothing), `HeadSha`, `PrId`, and `VariantId` are the natural correlation keys — and they're already in scope on `run` (used just above at lines 1082/1089/1090).

**Suggestion:** enrich the template (keeps the existing structured, non-interpolated idiom):

```csharp
_logger.LogWarning(
    "Run {RunId} (PR {PrId}, head {HeadSha}, variant {VariantId}): review produced no content; "
        + "not posting a comment or claiming the head's dedup slot (a placeholder would block a later "
        + "real review of the same commit).",
    run.Id, run.PrId, run.HeadSha, run.VariantId);
```

Pure enrichment nit — not a correctness issue.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/DaemonReviewStageExecutorTests.cs
gautam-achieveai, line 640: [Claude's review bot] **LOW** (Test coverage — retention path)

The other half of the fix — retention is skipped for an empty review (`PublishToReviewBotAsync` / `CommitPooledNotesAsync` not called) — is currently uncovered. The existing `Posted_publishes_..._reviewbot` test (~line 471) exercises only the non-empty path.

**Suggestion (optional, lower priority than the outbox assertion above):** add an empty-review variant that sets `ReviewBotRepoUrl` and asserts the retention path is skipped — no `checkout -B review/...`, no `PushReviewBot` outbox row, no `review.md` write. This locks in that an empty review neither posts nor retains.

gautam-achieveai:
[Claude's review bot]

# PR Review: fix(daemon): don't post an empty review — it poisons the head's dedup slot

**Verdict: ✅ APPROVE WITH COMMENTS** &nbsp;·&nbsp; 0 blockers &nbsp;·&nbsp; lightweight review (2 files, +63/-42)

_(Supersedes the earlier [Revobot] summary on this PR — replying here rather than duplicating the thread.)_

## Summary

This is a correct, minimal, and well-documented fix for a real **data-poisoning** defect. Previously, a review that produced no content still posted a `_No review content was produced._` placeholder through the full idempotency path; the backstop scan (`ReviewPoster.FindPostedCommentAsync`) later **adopted** that marked comment for the `head_sha` and permanently suppressed a *later real* review of the same commit (e.g. a re-run on a model that does produce content). The fix gates both the post and the retention behind `if (hasContent)`, while still freeing the pooled slot and tearing down the session.

I verified the change with three targeted specialists (exception/control-flow, test-coverage, over-engineering) and a grading pass. The production behavior was independently re-verified as **correct** — no blocker, no production defect.

## What's good

- **Control flow is correct across all four `{slot leased} × {hasContent}` combinations.** The slot `ReturnAsync` stays in an **unconditional `finally`**, so an empty review that leased a slot still returns it; the lease is atomically `TryRemove`d *before* the `try`, so there's no double-return; `else if (hasContent)` correctly gates `PublishToReviewBotAsync` and no-ops the not-leased/empty case; the terminal `DestroyAsync` teardown is reached on every path (no early return/throw). No swallowed exceptions; all awaits use `ConfigureAwait(false)`.
- **Empty-content path is reachable and does not throw** — the artifact row persists with `ReviewText = ""`, so `ReadReviewText` returns `""` rather than throwing.
- **Scope is appropriately minimal.** No speculative abstractions, no drive-by refactors, no unrequested features. The line churn is inflated purely by a 5-space re-indent of the existing block moving inside `if (hasContent)`. The new `LogWarning` reuses the class's existing structured-logging idiom.
- **The flipped test is valid** — it fails against the old code and passes against the new (correct bug-fix litmus).

## Findings (all non-blocking)

| # | Severity | Category | Location | Summary |
|---|----------|----------|----------|---------|
| F2 | **MEDIUM** | Test coverage | `DaemonReviewStageExecutorTests.cs:640` | Test asserts `PostedBodies` is empty (the visible symptom) but not that **no dedup/outbox row was claimed** — the actual invariant. The collect-only path records the outbox row via `EnqueueOutbox` *without* populating `PostedBodies`, so a placeholder regression would keep this test green. ~2-line fix via `GetOutboxForRun` (already used 3× in this file). **Recommended before/at merge.** |
| F1 | LOW | Observability | `DaemonReviewStageExecutor.cs:1103` | Empty-review `LogWarning` captures only `{RunId}`; enrich with `HeadSha`/`PrId`/`VariantId` (all in scope) for triage. Optional. |
| F3 | LOW | Test coverage | `DaemonReviewStageExecutorTests.cs` | No empty-review variant covering the skipped retention/ReviewBot-push path. Optional, lower priority than F2. |

Two INFO-level notes were **not** posted as they need no action: the 7-line lead-in comment (1060–1066) is proportionate — it documents a genuinely subtle, non-local dedup interaction and should be kept as-is; and the test's restatement of that rationale is a trivial, acceptable duplication.

## Recommendation

Approve with comments. The one item worth acting on is **F2** — a cheap, high-leverage strengthening of the test that guards the exact invariant the PR exists to protect. F1 and F3 are optional nice-to-haves. No context questions.

PR #158, opened 2026-07-06
feat(triggers): Wait/trigger follow-ups — notify mode, new sources, durable restore (#140-#145)

gautam-achieveai:
## Wait/trigger follow-ups (#140–#145)

Implements the six follow-ups deferred from the #107 Wait/trigger primitive (PR #139), as one coherent work item. Design spec and implementation plan are included under `docs/superpowers/`.

### What's included

- **#140 — Notify mode (core).** `Wait` gains `mode: "block" | "notify"` (default `block`, fully backward compatible) plus `maxFires`. Notify-mode arms don't park a deferred tool call; each fire is injected as a `<trigger>`-tagged user turn through the loop's existing internal queue gate (mirroring `ResumeSentinel`) — never interrupting an in-flight generation. A notify wait stays armed until `maxFires` or its `timeout` TTL, with a mandatory terminal envelope on cancel/expiry/restore-loss. Block mode's one-shot latch is untouched.
- **#141 — `file_tail` source (sample-app).** Fires on matching appended lines; path canonicalized against allowed-roots with per-platform case handling, symlink/junction ancestor-walk escape rejection at arm time, `NonBacktracking` regex + match timeout, per-line/per-batch caps, and content redaction so file data can't break out of the envelope. `SupportsRestore: false`.
- **#142 — `process` source (sample-app, Sandbox-gated).** Observes a process exit (via an `IProcessExitObserver` seam) and applies an exit-code/stdout predicate — no execution policy duplicated. Ships with a placeholder observer (see follow-ups).
- **#143 — `schedule` kind (sample-app).** Cron (via Cronos) + fixed-rate intervals; restore skips missed occurrences (no catch-up storm) so a restored schedule is restart-safe. `SupportsRestore: true`.
- **#144 — `subagent` source + core seam.** Fires when a specific background sub-agent completes, suppressing the double-relay by flipping `NotifyParentOnCompletion` and restoring it on cancel. Core adds `SubAgentManager.ObserveCompletionAsync` (non-destructive observation) + `SetNotifyParentOnCompletion`.
- **#145 — Durable notify-watcher restore (core).** A dedicated `notify_waits` SQLite table + `INotifyWaitStore`/`SqliteNotifyWaitStore`; `TriggerRuntime` persists notify arms, deletes on terminal, and `RestoreNotifyWaitsAsync` re-arms restorable waits (or delivers `trigger_lost_on_restart`) on thread rehydrate.

[...]

PR #170, opened 2026-07-08
refactor(daemon): {repo}-{pr} review-store naming + reconcile orphaned notes branches

gautam-achieveai:
## What

Two related changes to the Code-Review Daemon's cross-repo review store (`AchieveAiReviews`):

### 1. Simplify per-PR naming to `{repo}-{pr}` (commit 11f3d163)

The store named a PR's branch/folder/commit with a verbose, redundant scheme. The repo is already identified by its `repos/<Repo>` submodule in the store, so the `provider`/`owner` prefix and the head-sha suffix added noise:

| | before | after |
|---|---|---|
| branch | `review/github/achieveai-lmdotnettools/156` | `review/lmdotnettools-156` |
| folder | `PRs/github/achieveai-…/156/…-<sha>/review.md` | `PRs/lmdotnettools-156/review.md` |
| commit | `Review github achieveai/LmDotnetTools#156 @ 3c57b499` | `Review LmDotnetTools#156` |

Dropping the head-sha makes the folder a **stable, accumulating per-PR working directory keyed by PR number** for the PR's lifetime — room for progress-tracking artifacts, not just a per-commit `review.md`. `provider` is removed from `ReviewBranchManager`'s ctor and the naming helpers; the now-unused `ShortSha` helper is dropped.

### 2. Reconcile orphaned `review/*` branches in the PR-lifecycle sweep (commit d8e7fd66)

The sweep's watch-set came **only** from the daemon's SQLite DB (`ListReviewedPrsAsync`). A reviewed PR whose row was absent from the running daemon's DB (a fresh DB after a restart, or DB churn) never had its notes branch resolved — a merged PR's notes never reached the store default branch, and an abandoned PR's branch was never deleted. This is why `AchieveAiReviews` `main` stopped advancing even as PRs merged.

The sweep now **also** lists the store's actual `review/*` branches (`git ls-remote`, fresh each cycle) and reconciles them: any `review/{repo}-{pr}` branch not already covered by the DB set, whose slug matches a configured poll target, is added as a pollable PR — resolved back to that target's repo identity + provider. Unmatched or legacy-format branches are logged and skipped; a `ls-remote` failure degrades to the DB-only set.

## Why

- Naming: user-requested — the verbose path was redundant with the submodule layout.
- Reconciler: makes `main` advance robustly regardless of DB state; closes the durability gap where DB churn orphaned merged PRs (#156, #158) so their notes never reached `main`.

## Notes / follow-ups

- **Legacy branches are not auto-reconciled.** The parser handles only the new `review/{repo}-{pr}` form; the existing bring-up branches (`review/github/achieveai-…/88/156/157/158/169`) keep their old names and are skipped. If their notes are wanted on `main`, they need a one-time manual carry — happy to do that separately.
- The "folder holds more than `review.md`" capability is *enabled* by the stable per-PR dir but not itself built here.

## Tests

455 daemon tests green (12 new): `OrphanBranchReconcilerTests`, `ReviewBranchNameTests`, plus updated naming expectations across the existing suite. Solution builds 0-warning under `TreatWarningsAsErrors`.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
+++ b/samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
@@ -404,8 +404,8 @@ internal sealed class DaemonReviewStageExecutor : IReviewStageExecutor
                 return false;
             }
 
-            var branch = BuildNotesBranchName(hostGit, hostFileSystem, provider, repo, run);
-            var notesRelPath = BuildNotesRelPath(provider, repo, run.PrId);
+            var branch = BuildNotesBranchName(hostGit, hostFileSystem, repo, run);
+            var notesRelPath = BuildNotesRelPath(repo, run.PrId);
             var policy = DaemonOperationPolicy.BuildForRun(
                 repo, _options.ReviewBotRepoUrl, allowWriteOperations: false,
                 allowedSubmodules: BuildStoreSubmoduleAllowList(run, repo));
@@ -498,18 +498,18 @@ internal sealed class DaemonReviewStageExecutor : IReviewStageExecutor
         return entry?.Path;
     }
 
-    /// <summary>The PR's persistent notes branch name (<c>review/{provider}/{owner-repo}/{pr}</c>) — resolved
+    /// <summary>The PR's persistent notes branch name (<c>review/{repo}-{pr}</c>) — resolved
     /// through <see cref="ReviewBranchManager.BuildReviewBranchName(ReviewBotPublishRequest)"/> so the preparer,
     /// the commit-notes step, and the sweeper all name the branch identically.</summary>
     private string BuildNotesBranchName(
-        GitRunner hostGit, ISandboxFileSystem hostFileSystem, string provider, RepoIdentity repo, ReviewRun run) =>
-        new ReviewBranchManager(hostGit, hostFileSystem, provider, _loggerFactory.CreateLogger<ReviewBranchManager>())
+        GitRunner hostGit, ISandboxFileSystem hostFileSystem, RepoIdentity repo, ReviewRun run) =>
+        new ReviewBranchManager(hostGit, hostFileSystem, _loggerFactory.CreateLogger<ReviewBranchManager>())
             .BuildReviewBranchName(BuildNotesRequest(repo, run, []));
 
-    /// <summary>The PR's persistent notes directory under the store (<c>PRs/{provider}/{owner-repo}/{pr}</c>,
-    /// design §4.3 D3 — one accumulating dir per PR). Each re-review adds a per-commit subdir under it.</summary>
-    private static string BuildNotesRelPath(string provider, RepoIdentity repo, string prId) =>
-        $"PRs/{provider}/{ReviewBotRepoManagerSlug(repo)}/{prId}";
+    /// <summary>The PR's persistent notes directory under the store (<c>PRs/{repo}-{pr}</c>,
+    /// design §4.3 D3 — one accumulating dir per PR, keyed by PR number for the PR's lifetime).</summary>
+    private static string BuildNotesRelPath(RepoIdentity repo, string prId) =>
+        $"PRs/{ReviewBotRepoManagerSlug(repo)}-{prId}";
 
     private static ReviewBotPublishRequest BuildNotesRequest(
         RepoIdentity repo, ReviewRun run, IReadOnlyList<ReviewArtifactFile> files) =>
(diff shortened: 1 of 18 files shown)


gautam-achieveai:
[Revobot]

# PR #170 Review — Revobot

Reviewed PR #170 (ReviewBot notes branch/path scheme simplification + orphan-branch reconciliation) across 8 changed production/test files — 0 Must, 2 Should, 2 Consider.

The change is coherent and well-tested: `review/{provider}/{owner-repo}/{pr}` → `review/{repo}-{pr}`, with a matching reverse parser (`TryParseReviewBranch`) and a new `OrphanBranchReconciler` that folds review branches present on origin but missing from the DB back into the lifecycle sweep. The parser correctly handles repo slugs that themselves contain `-` (last-dash split) and rejects the legacy nested form; `ReadPrNotesFromBranchAsync` derives the notes path from the branch slug so it stays in lock-step with the new scheme. No correctness, security, or compatibility bug blocks the change.

The one thing worth a conscious decision is that repository identity is now collapsed to `RepoName` alone.

---

## Should

### 1. Branch/notes identity is now `RepoName` only — same-named repos across owners/providers collide
`ReviewBranchManager.RepoSlug` (Workspace/Git/ReviewBranchManager.cs:373) is now `Slug(repo.RepoName)`, dropping `OrgOrOwner` and `Project` that the previous `SlugifyRepo` included. Consequences if the operator's `EnabledRepos` ever contains two repos with the same leaf name (e.g. `acme/widgets` and `contoso/widgets`, or a GitHub `owner/widgets` and an ADO `org/proj/widgets`):

- Both PRs share one notes branch `review/widgets-{pr}` and one dir `PRs/widgets-{pr}/` — reviews and merge/delete lifecycle for one repo's PR act on the other's notes. PR numbers also collide across repos, compounding it.
- In `OrphanBranchReconciler`, `targetsBySlug[RepoSlug(target.Repo)] = target` (OrphanBranchReconciler.cs:39) silently last-wins on a duplicate slug, so an orphaned branch is resolved to the wrong repo identity and the sweep polls the wrong PR's lifecycle.

This is a real cross-repo footgun even though the daemon runs against one store. If single-leaf-name uniqueness is an intended invariant of `EnabledRepos`, please state it (a guard/log in `PrPollTargetBuilder` when two entries slug to the same leaf, or in the reconciler when `targetsBySlug` would overwrite, would make it safe). If it is *not* guaranteed, the slug should keep enough of the owner/project to disambiguate. Worth an explicit call in the design note either way.

### 2. Stale comment in `ReadPrNotesFromBranchAsync` still describes the old nested scheme
Program.cs:416 — the comment reads "The notes dir mirrors the branch slug (review/<p>/<slug>/<pr> -> PRs/<p>/<slug>/<pr>)", which is the pre-PR nested form. The code just below correctly does `"PRs/" + branch["review/".Length..]` for the flat scheme. Update the comment so it doesn't mislead the next reader.

---

## Consider

### 3. `ls-remote` passes `-C repoRoot` and `workingDirectory: repoRoot` redundantly
Program.cs:352 sets the working directory to `repoRoot` *and* prepends `-C repoRoot`. Harmless (both point at the same dir) and matches the existing `ReadPrNotesFromBranchAsync` idiom — purely a tidiness note.

### 4. `OrphanBranchReconciler` skip-vs-warn asymmetry between legacy names and unknown slugs
A legacy nested branch (`review/github/acme-widgets/42`) is silently dropped by `TryParseReviewBranch` returning false (OrphanBranchReconciler.cs:49), while a well-formed name whose slug matches no configured repo is logged at Warning (line 56). If any legacy-scheme branches still exist on origin from before this change, they will never be merged/deleted and produce no diagnostic. A single debug/info log on the parse-failure path would make that observable. (Depends on whether legacy branches can exist in your store.)

---

## Correct as written
- `TryParseReviewBranch` rejects empty slug, missing/non-digit PR, and embedded `/`; round-trips slugs with internal dashes — `ReviewBranchNameTests` covers these.
- `OrphanBranchReconciler` preserves the DB set verbatim, dedups via a case-sensitive `Branch` `HashSet`, and never mutates its input.
- `review.md` overwriting per-PR (vs. per-commit subdirs) is fine: git history retains prior versions, and numbered `PR_Findings_NN.md` still accumulate in the same dir.
- No wire/DB schema touched; the scheme is entirely git branch/path strings the daemon owns on both sides — no serializer/deserializer drift.
- No prompt-injection or EUII concerns in the diff.

---

Verdict: **Approve with comments** — no blocker; resolve the same-name collision question (Should #1) and fix the stale comment (Should #2) before or shortly after merge.

PR #172, opened 2026-07-08
feat(daemon): reconcile legacy-named review branches so they auto-clean on close

gautam-achieveai:
## Why

The `{repo}-{pr}` rename (#170) taught the orphan-branch reconciler only the **new** branch-name form. Review branches created *before* the rename — `review/{provider}/{owner-repo}/{pr}` — are silently skipped, so when their PR closes the daemon leaves the branch behind instead of merging-to-main + deleting it. That's the exact orphaning the reconciler exists to prevent (and it just bit #156/#158, which had to be carried + deleted by hand).

The three in-flight legacy branches (#88/157/169) would hit the same fate when they close.

## What

- `ReviewBranchManager.TryParseLegacyReviewBranch` + `LegacyRepoSlug` — reverse of the pre-rename `owner[-project]-repo` naming.
- `OrphanBranchReconciler` now indexes configured poll targets by **both** the new `{repo}` slug and the legacy `{owner-repo}` slug, and tries both parsers per branch — so the sweep self-heals orphans from either naming generation and acts on the branch's *actual* name.

Result: any completed PR's review branch — old-named or new — is merged-to-main and deleted automatically, no hand-cleanup.

## Tests

463 daemon tests green (8 new): legacy parse round-trip/reject, `LegacyRepoSlug`, and reconciler resolves-legacy / skips-unmatched-legacy. Builds 0-warning under `TreatWarningsAsErrors`.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Orchestration/OrphanBranchReconciler.cs
+++ b/samples/CodeReviewDaemon.Sample/Orchestration/OrphanBranchReconciler.cs
@@ -9,11 +9,13 @@ namespace CodeReviewDaemon.Sample.Orchestration;
 /// still gets its notes branch merged-or-deleted when the PR closes. Without this, such a branch is orphaned:
 /// the merged PR's notes never reach the store default branch and the abandoned PR's branch is never deleted.
 /// <para>
-/// The DB-derived set is authoritative and always kept. This only <b>adds</b> orphaned branches whose
-/// <c>review/{repo}-{pr}</c> slug (see <see cref="ReviewBranchManager.TryParseReviewBranch"/>) matches a
-/// configured poll target — the daemon needs that target's <see cref="PrPollTarget.Repo"/> identity and
-/// <see cref="PrPollTarget.Provider"/> to look the PR's lifecycle up. A branch that matches no configured repo
-/// (or is in the legacy nested form) is logged and skipped: its identity cannot be recovered from the name.
+/// The DB-derived set is authoritative and always kept. This only <b>adds</b> orphaned branches whose slug
+/// matches a configured poll target — either the new <c>review/{repo}-{pr}</c> form (see
+/// <see cref="ReviewBranchManager.TryParseReviewBranch"/>) or the legacy <c>review/{provider}/{owner-repo}/{pr}</c>
+/// form (see <see cref="ReviewBranchManager.TryParseLegacyReviewBranch"/>), so orphans from either naming
+/// generation get cleaned up. The daemon needs that target's <see cref="PrPollTarget.Repo"/> identity and
+/// <see cref="PrPollTarget.Provider"/> to look the PR's lifecycle up. A branch that matches no configured repo,
+/// or is unparseable, is logged and skipped: its identity cannot be recovered from the name.
 /// </para>
 /// </summary>
 internal static class OrphanBranchReconciler
@@ -36,7 +38,10 @@ internal static class OrphanBranchReconciler
         var targetsBySlug = new Dictionary<string, PrPollTarget>(StringComparer.OrdinalIgnoreCase);
         foreach (var target in configuredTargets)
         {
+            // Index by both the new {repo} slug and the legacy {owner-repo} slug so orphans from either naming
+            // generation resolve back to the same configured repo.
             targetsBySlug[ReviewBranchManager.RepoSlug(target.Repo)] = target;
+            targetsBySlug[ReviewBranchManager.LegacyRepoSlug(target.Repo)] = target;
         }
 
         foreach (var branch in remoteReviewBranches)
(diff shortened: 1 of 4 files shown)


gautam-achieveai:
[Revobot]

Let me examine the key files to verify the legacy parsing logic.

Let me check the existing `RepoSlug`, `TryParseReviewBranch`, and `RepoIdentity` to verify consistency.

Let me read the relevant portions.

The diff is self-contained enough to review carefully. Let me verify a few concerns about the legacy slug matching.

Key concern: `LegacyRepoSlug` builds `owner[-project]-repo` via `Slug` on each part joined by `-`. `TryParseLegacyReviewBranch` extracts `parts[2]` verbatim (the `{owner-repo}` segment) without re-slugging. The reconciler looks up `targetsBySlug` which was keyed with `LegacyRepoSlug(...)` under `OrdinalIgnoreCase`. So the legacy branch's raw segment must equal the slugged join. For the branch to have been produced originally by the same slugging code, this should match — the test `acme-widgets`/`acme-platform-widgets` confirms round-trip when the raw segment is already slug-form. This is sound as long as legacy branches were generated identically.

Let me confirm no collision risk with the new-scheme dictionary keys.

```
Reviewed PR #172 across 3 files (ReviewBranchManager.cs, OrphanBranchReconciler.cs, + 2 test files) — 0 Must, 1 Should, 2 Consider.
```

## Summary

The PR extends the orphan-branch reconciler to also recover legacy `review/{provider}/{owner-repo}/{pr}` branches, adding `LegacyRepoSlug` and `TryParseLegacyReviewBranch` to `ReviewBranchManager`, indexing configured targets by both slug forms, and falling back to legacy parsing in the reconcile loop. The change is well-scoped, symmetric (slug ↔ parse), and covered by focused tests. No correctness, security, or contract-break issues found.

## Findings

**Should — Slug-key collision between new and legacy forms silently drops a target.**
`samples/CodeReviewDaemon.Sample/Orchestration/OrphanBranchReconciler.cs` ~L41-44. The dictionary is now populated with two keys per target (`RepoSlug` and `LegacyRepoSlug`) under `OrdinalIgnoreCase`. For a single-segment repo with no org/project distinction, or across two *different* configured targets, these keys can collide — e.g. a repo whose `RepoSlug` equals another repo's `LegacyRepoSlug`, or a target where `RepoSlug == LegacyRepoSlug` (a repo with only a `RepoName` and empty owner). Because `targetsBySlug[key] = target` overwrites rather than guards, the later-iterated target wins and the earlier one's orphans resolve to the wrong repo (or a same-target self-overwrite is harmless but masks the collision). Consider detecting/logging when a key would map to a *different* target, or preferring the new-form key on conflict. Add a test with two targets whose slug forms overlap.

**Consider — `parts[2]` is used verbatim, so resolution depends on legacy branches already being in slug-normalized form.**
`ReviewBranchManager.cs` `TryParseLegacyReviewBranch` returns `parts[2]` without re-slugging, while `LegacyRepoSlug` slugs each component. If any legacy branch was ever written with un-slugged casing/characters in the `{owner-repo}` segment that `Slug` would normalize, the lookup (though `OrdinalIgnoreCase`) could still miss on non-alphanumeric differences. This is fine if the legacy writer used identical slugging — worth a one-line comment asserting that invariant, or re-slugging `parts[2]` before returning for robustness.

**Consider — `prPart.All(char.IsAsciiDigit)` plus `int.TryParse` rejects valid large PR numbers silently.**
`ReviewBranchManager.cs` `TryParseLegacyReviewBranch`. A PR segment of all digits that overflows `int` returns `false` and the branch is skipped as "unparseable" rather than logged distinctly. Realistically PR numbers never overflow `int`, so this is only a theoretical edge; noting for parity with how `TryParseReviewBranch` handles it (worth confirming both paths behave the same).

No prompt-injection or suspicious content in the diff.

## Verdict
Approve with comments — sound change; address the slug-key collision concern before merge.

gautam-achieveai (commented):
**[Gautam's review bot]**

## Summary

Clean, well-scoped change. Adds legacy branch reconciliation (`review/{provider}/{owner-repo}/{pr}`) alongside the new `review/{repo}-{pr}` form. The symmetry between `LegacyRepoSlug` (build) and `TryParseLegacyReviewBranch` (parse) is sound, and the test coverage is good.

## Findings

### Should — Guard against slug-key collision in targetsBySlug

`OrphanBranchReconciler.cs` L43-44: Two keys per target (`RepoSlug` + `LegacyRepoSlug`) are inserted into one dictionary. When `RepoSlug(A) == LegacyRepoSlug(B)` for two *different* configured targets, the second target silently overwrites the first. Minimal fix: check `TryAdd` or log a warning when the key already maps to a different target. E.g.:

`csharp
if (targetsBySlug.TryGetValue(key, out var existing) && existing != target)
    logger.LogWarning("Slug collision: '{Key}' maps to both {A} and {B}", key, existing.Repo, target.Repo);
targetsBySlug[key] = target;
`

### Consider — Provider segment extracted but discarded

`TryParseLegacyReviewBranch` validates `parts[1]` (provider) is non-empty but never returns it. The reconciler uses `target.Provider` from the dictionary lookup. Fine for single-provider configs, but if the same slug exists under multiple providers the collision issue means the wrong provider could win.

### Nit — Unnecessary double-validation of PR number

Both `prPart.All(char.IsAsciiDigit)` and `int.TryParse` validate the PR segment. `int.TryParse` alone suffices — minor, not blocking.

## Verdict

**APPROVE** — Solves the problem correctly and is well-tested. The slug-key collision concern (also noted by Revobot) is worth addressing but isn't a ship-stopper for the three in-flight legacy branches since the current config has no colliding slugs.

PR #173, opened 2026-07-09
perf(daemon): stop re-resolving already-closed PRs on every sweep

gautam-achieveai:
## Why

The PR-lifecycle sweep resolves every reviewed PR from the review DB each poll. The DB only grows, so a PR that's already been merged-and-swept (or abandoned-and-deleted) still costs a GitHub lifecycle lookup + a git no-op **every cycle** — visible live as the recurring `merge of notes branch …lmdotnettools-170/-172… into main: True` log lines (idempotent no-ops on branches already merged+deleted), and O(all-PRs-ever-closed) work per sweep.

## What

- `PrLifecycleSweeper` caches branches it has resolved to a terminal state this daemon lifetime and skips them **before** the lifecycle lookup.
- Cached only on *confirmed* terminal resolution: `ResolveMergedAsync` now returns whether the merge actually succeeded, so a failed merge push is **retried** next sweep rather than cached.
- Not persisted — a restart re-confirms each branch once then caches, so it can never wrongly skip a PR whose resolution didn't complete.

## Tests

465 daemon tests green (2 new): resolve-once-then-skip, and retry-when-merge-push-fails. 0 warnings.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs
+++ b/samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs
@@ -95,6 +95,15 @@ internal sealed class PrLifecycleSweeper
     /// </summary>
     private readonly Func<ReviewedPr, CancellationToken, Task>? _extractKnowledgeAsync;
 
+    /// <summary>
+    /// Notes branches this daemon lifetime has already resolved to a terminal lifecycle (merged-and-swept, or
+    /// abandoned-and-deleted). The reviewed-PR list only grows, so without this the sweep would re-run a GitHub
+    /// lifecycle lookup plus a git no-op for every PR it has ever closed, on every poll. Not persisted — a
+    /// restart re-confirms each branch once and then caches it, so this can never wrongly skip a PR whose
+    /// resolution did not actually complete.
+    /// </summary>
+    private readonly HashSet<string> _terminallyResolved = new(StringComparer.Ordinal);
+
     public PrLifecycleSweeper(
         Func<CancellationToken, Task<IReadOnlyList<ReviewedPr>>> listReviewedPrsAsync,
         Func<ReviewedPr, CancellationToken, Task<PrLifecycle>> getPrLifecycleAsync,
@@ -145,6 +154,13 @@ internal sealed class PrLifecycleSweeper
 
     private async Task ResolveAsync(ReviewedPr pr, CancellationToken cancellationToken)
     {
+        // A branch already resolved to a terminal state this lifetime never needs re-resolving; skipping it
+        // avoids a per-poll GitHub lifecycle lookup + git no-op for every PR the daemon has ever closed.
+        if (_terminallyResolved.Contains(pr.Branch))
+        {
+            return;
+        }
+
         var lifecycle = await _getPrLifecycleAsync(pr, cancellationToken).ConfigureAwait(false);
         switch (lifecycle)
         {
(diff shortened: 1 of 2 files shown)


gautam-achieveai:
[Revobot]

This is a small, well-scoped optimization to `PrLifecycleSweeper`. The change caches terminally-resolved branches in-memory so subsequent sweeps skip redundant GitHub lifecycle lookups and git no-ops.

**Reviewed PR #173 across 2 files — 0 Must, 0 Should, 1 Consider.**

## Correctness

The core logic is sound and the boundaries are handled correctly:

- `ResolveMergedAsync` now returns `bool`, and the cache add is gated on it — a failed merge push (`merged == false`) is *not* cached, so it retries. The `mergeNotesBranchOnClose == false` path and the "already gone" path correctly return `true`. Verified against the diff's control flow.
- The `Abandoned` branch caches unconditionally after logging, which is fine — deletion is terminal.
- The `Open` and `default` branches do not cache, which is correct — those states are non-terminal.
- Not persisting the set is deliberate and safe: a restart re-confirms each branch once (one extra lookup) rather than risking a wrongly-skipped unresolved PR. The reasoning in the XML doc matches the code.

## Test coverage

Both new tests are meaningful and target the two behaviors that matter:
- `Sweep_does_not_re_resolve_a_merged_PR_on_a_later_sweep` asserts the lookup count *and* the push count — good, it pins both the cache hit and the absence of a re-merge.
- `Sweep_retries_a_merged_PR_whose_merge_push_failed` correctly forces a push failure and asserts the retry. This is the important negative case that proves the cache is not over-eager.

## Consider

- **Consider** (`PrLifecycleSweeper.cs:104`): The cache keys on `pr.Branch` alone (`StringComparer.Ordinal`). If two distinct PRs across different providers/repos could ever share a branch name, the first to resolve would suppress re-resolution of the other. Given the reviewed-PR list is per-daemon and branch names here are review-specific (e.g. `review/widgets-42`), this is very likely fine in practice — but keying on the same identity tuple used elsewhere (provider + branch, or `PrId`) would be more robust and self-documenting. Worth a moment's thought rather than a required change.

No security, contract, or compatibility concerns — the signature change on `ResolveMergedAsync` is on a `private` method, so no external surface is affected. No prompt-injection content in the diff.

**Verdict: Approve.**

gautam-achieveai:
**[Gautam's review bot]** ✅ **APPROVE**

Clean, well-scoped performance optimization. The in-memory terminal-state cache correctly avoids redundant GitHub lifecycle lookups + git no-ops for already-closed PRs on every poll cycle.

**What I verified:**
- Thread safety: `SweepAsync` runs on a single polling timer — no concurrent `HashSet` access. ✓
- Merged path: caches only when `ResolveMergedAsync` returns `true` (push succeeded); failed push → retry next sweep. ✓
- Abandoned path: `DeleteBranchAsync` is idempotent (`allowFailure: true`), so caching unconditionally is safe. ✓
- Not persisted: restart re-confirms each branch once — can never wrongly skip an unresolved PR. ✓
- Tests: 2 new tests (resolve-once + retry-on-failed-push) cover the critical correctness invariants. All 10 sweeper tests pass.

**Minor observations (non-blocking):**
1. No explicit test for abandoned-PR caching (only merged path tested for skip-on-second-sweep). Low risk given `DeleteBranchAsync` never throws.
2. `_terminallyResolved` grows unbounded over daemon lifetime — negligible in practice (string keys, likely <10K entries over months), but if the daemon ever runs for months with thousands of PRs, a bounded cache or periodic trim could help.

Issue #161, opened 2026-07-06
Trigger follow-up hardening: redaction, rollout flags, backpressure, ISP (deferred from PR #158 review)


gautam-achieveai:
Deferred enhancements from the PR #158 review (achieveai/LmDotnetTools#158, Wait/trigger follow-ups #140–#145). The review bot raised these as Medium/Low items; they are valid but out of scope for #158 (which fixed the verified blockers + low-risk correctness items). Tracking here so they aren't dropped.

### Security / privacy (highest value)
- **Configurable PII/secret redaction for forwarded trigger content.** `FileTailTriggerSource.Redact` only caps length + escapes angle brackets (envelope-injection defense); it does not redact emails, API keys, bearer tokens, connection strings, etc. from matched log lines that flow into model/history. Add a real redaction/metadata-only mode and rename `Redact` → `SanitizeEnvelopeContent` to avoid implying privacy redaction. (`FileTailTriggerSource.cs:259`)
- **`notify_waits.args` persists trigger args verbatim** — for `file_tail` this can include absolute paths/patterns revealing usernames/project layout. Avoid persisting args for non-restorable sources, or persist minimal/opaque source-specific state (or document sensitivity + redact/encrypt at rest). (`SqliteNotifyWaitStore.cs:21`)

### Rollout / safety
- **Gate sample trigger enablement behind a flag** (e.g. `LM_ENABLE_TRIGGERS`). `Program.cs` currently enables trigger tools + sample sources for every conversation by default; notify waits can enqueue future `<trigger>` turns (extra runs/cost) until timeout/maxFires/cancel. Consider split flags for wait tools, notify mode, and each source. (`Program.cs:939`)
- **Hide or clearly mark the inert `process` kind.** It registers with `NoopProcessExitObserver` (never fires) when sandbox is on — a model/user can arm a `process` wait that parks until timeout. Skip registering until a real Bash-tool exit observer is wired, or gate behind an explicit experimental/disabled flag. (`SampleTriggerRegistrations.cs:63`)

### Reliability
- **Notify-delivery backpressure/overflow policy.** Notify fires write to the bounded input channel with `CancellationToken.None`; a full channel can block source tasks indefinitely inside `sink.FireAsync`, stalling polling/schedule progress. Use the runtime shutdown token / a bounded enqueue timeout and define an explicit overflow policy (drop/coalesce with warning, reject-until-drained, or durable queue). (`TriggerRuntime.cs:584`)

### Design / maintainability
[...]

Issue #145, opened 2026-07-05
Wait/trigger: durable notify-watcher restore across restart


gautam-achieveai:
Follow-up to #107 (PR #139) and notify mode (#140). v1 keeps notify waits in-memory/non-durable (per the #107 non-goal). This tracks persisting live-watcher arming records and safely re-arming them across a process restart, including the reconcile decisions (`active`/`trigger_lost_on_restart`) already defined for block waits.

gautam-achieveai:
Implemented in #158 (merged to main at bce61aa0): a dedicated `notify_waits` SQLite table + `INotifyWaitStore`/`SqliteNotifyWaitStore`; `TriggerRuntime` persists notify arms, deletes on terminal, and `RestoreNotifyWaitsAsync` re-arms restorable waits (or delivers `trigger_lost_on_restart`) on thread rehydrate. This is opt-in and NOT yet wired into the sample host — a host must set `TriggerOptions.NotifyWaitStore` + `ThreadId` and run `SqliteSchemaInitializer` for it to take effect. Leaving this open until the sample host wiring lands (would make a `schedule` notify wait survive restart end-to-end).

gautam-achieveai:
[gautam-achieveai's bot] Starting analysis.

gautam-achieveai:
[gautam-achieveai's bot] **Implementation Plan for #145: Wait/trigger — durable notify-watcher restore across restart**

## Understanding

**Type:** Task / enhancement
**Summary:** Persist live notify-watcher arming records and safely re-arm them across a process restart, reusing the `active` / `trigger_lost_on_restart` reconcile decisions already defined for block waits. **The core primitive already exists on `main`** (PR #158 added `INotifyWaitStore`/`SqliteNotifyWaitStore`, the `notify_waits` table, `TriggerRuntime.RestoreNotifyWaitsAsync`, and `TriggerOptions.NotifyWaitStore`/`ThreadId`; `MultiTurnAgentLoop.OnThreadRecoveredAsync` already calls `RestoreNotifyWaitsAsync` on every thread recovery). **This issue is purely a sample-host wiring gap** — `samples/LmStreaming.Sample` never constructs a store nor sets `NotifyWaitStore`/`ThreadId`, so notify waits there are still non-durable.

**Acceptance Criteria:**
1. The sample host persists notify-watcher arming records to durable storage (SQLite), so they survive a process restart.
[...]

Issue #143, opened 2026-07-05
Wait/trigger: cron + interval timer capabilities


gautam-achieveai:
Follow-up to #107 (PR #139). Extend the built-in `timer` source (currently one-shot delay|deadline) with:
- `interval` (fixed-rate from arm, with a floor) — multi-fire, notify-mode oriented.
- `cron` (timezone-documented; full DST rigor a separate concern).

Cron needs a dependency (e.g. **Cronos**, MIT, no native deps) — none is referenced today. Multi-fire intervals depend on notify mode (#140).

gautam-achieveai:
Implemented in #158 (merged to main at bce61aa0): `schedule` trigger kind in LmStreaming.Sample — cron (via Cronos) + fixed-rate intervals. Restore skips missed occurrences (no catch-up storm), so a restored schedule is restart-safe. `SupportsRestore: true`.

Issue #140, opened 2026-07-05
Wait/trigger: notify mode (keep working + injected <trigger> messages)


gautam-achieveai:
Follow-up to #107 (PR #139). Add `mode:"notify"` to the `Wait` tool: instead of parking, return a `trigger_id` immediately and inject a self-describing `<trigger kind id label>…</trigger>` message on each fire while the agent keeps working (until `CancelWait`, `maxFires`, or the deadline).

Design notes carried from the #107 plan:
- Inject through the loop's **existing internal gate** (`EnqueueRawAsync` + a new `QueuedInput` sentinel, mirroring `ResumeSentinel`) — no public deferral API. Define a single safe injection point with FIFO/sequence ordering.
- Add ordering tests for fires during: model generation, normal tool execution, an unresolved block deferral, and history restore.
- Notify waits are **in-memory / non-durable** in v1 (per the issue's non-goal); durable notify-watcher restore is a separate follow-up.
- Add `mode`/`maxFires` params to the tool surface; `SupportsNotify` capability on sources.

gautam-achieveai:
Implemented in #158 (merged to main at bce61aa0): `Wait` now supports `mode: "block" | "notify"` (default `block`, backward compatible) plus `maxFires`. Notify-mode fires are injected as `<trigger>`-tagged user turns through the loop's existing queue gate without interrupting in-flight generation; block mode is unchanged.

PR #174, opened 2026-07-09
NotifyMessage: typed out-of-band notifications tied to originating tool calls (#171)

### LINKED ISSUE #171: NotifyMessage: typed out-of-band notifications tied to originating tool calls
## Summary
Introduce a first-class `NotifyMessage` message type for **out-of-band notifications** pushed into a running `MultiTurnAgent` conversation from asynchronous sources. Today every such notification is injected as a plain `Role.User` `TextMessage`, so the LLM treats it as human input and the UI renders it as a user bubble. `NotifyMessage` maps to a **user-role** message for the LLM (so the model still reads it), but its content is a **self-describing text envelope** that names the originating tool call, and it renders as a **visually distinct notification pill** in the UI (not a user message).

Async sources in scope: (1) async sub-agent completion, (2) monitors that fire on a change, (3) async long-running bash commands, (4) timers/cron firing or resuming the conversation.

## Value
- **LLM correctness:** the model understands an injected event is an *asynchronous response to a specific prior tool call*, not a fresh human instruction — avoiding confused re-replies.
- **Users:** notifications are clearly distinguishable from their own messages in the chat UI.
- **Platform:** one typed primitive that every async producer (sub-agents, monitors, async bash, timers/cron — #140/#143/#106) emits, instead of each hand-rolling a `Role.User` `TextMessage`.

## Design (decided)
- **Correlation via text envelope only** — no provider `tool_result` pairing. `NotifyMessage` maps to a user-role message whose content is a self-describing envelope naming the source tool call, e.g.
  `<notification kind="subagent-completion" in-response-to="Spawn:agent-7" label="build-fixer">…payload…</notification>`
  Structured fields (`SourceToolCallId`, `SourceToolName`, `NotifyKind`, `Label`, `Detail`) are also carried on the type for the UI and future use. Sources with no originating tool call (timer/cron) omit `in-response-to` and still produce a valid envelope + pill.
- **Single PR:** the foundation and the migration of all current injectors land together.

### Core type & serialization
- [ ] `NotifyMessage : IMessage, ICanGetText` exists in `src/LmCore/Messages`, `Role.User`, exposing `Text` (envelope), `SourceToolCallId?`, `SourceToolName?`, `NotifyKind`, `Label?`, `Detail?`, plus standard `IMessage` identity fields.
- [ ] Discriminator `$type:"notify"` registered in both `GetDiscriminatorFromType` and `GetTypeFromDiscriminator` (`IMessageJsonConverter`); round-trips through the production serializer with all structured fields intact.
- [ ] Structural-inference fallback: a `NotifyMessage` whose `$type` is stripped deserializes back to `NotifyMessage` (keyed on a notify-only field), **not** to `TextMessage`.

### Provider mapping — no silent drops on any backend
- [ ] Anthropic (`AnthropicRequest`), OpenAI chat (`ChatCompletionRequest`), OpenAI Responses (`MessageMapper`), Copilot (`CopilotEventTranslator.BuildPrompt`), and Codex (`CodexEventTranslator.BuildPrompt`) each emit the `NotifyMessage` as a **user-role message containing the envelope text**. A unit test per backend asserts the envelope is present (no `is TextMessage` / `default:` silent drop).
- [ ] Anthropic consecutive-same-role merge: a `NotifyMessage` injected adjacent to a pending user message still produces a valid Anthropic request and keeps the envelope legible. (If injection can ever place a notify between an assistant `tool_use` and its `tool_result`, that request must also stay valid — verify reachability first.)

### Injection, ordering & persistence
- [ ] Notifications inject through the loop's internal gate (`EnqueueRawAsync` + a `QueuedInput`), FIFO/sequence-ordered, appended to history, persisted, and replayed on restore.
- [ ] Injecting a `NotifyMessage` while an unresolved block-wait/deferral exists does **not** trip the `ExecuteTurnAsync` deferred-guard (no `RunFailed`); the notify is delivered after the deferral resolves. Regression: a sub-agent completes while the parent is parked on a `Wait`.
- [ ] The emit path **publishes** the `NotifyMessage` to subscribers (not only `AddToHistory`) so the pill renders live and on resume, with the published copy carrying the same identity as the persisted copy.

### Client / UI (Vue 3 sample)
- [ ] `MessageType` token + TS interface + type guard added; `handleMessage` routes `NotifyMessage` (not "unknown → skipped"); `role` tagging is consistent across live-create and reload.
- [ ] `displayItems` renders `NotifyMessage` as a distinct notification pill **before** the `role === 'user'` catch-all; `MessageList.vue` + component render it with a stable `data-testid`.
- [ ] Merge-key disambiguation: two `NotifyMessage`s in one run both render, in order, live and after reload (no `other-<runId>-default-0` collision).

### Producer migration (all current injectors, this PR)
- [ ] `SubAgentManager.SendToParentAsync` emits a `NotifyMessage` (`kind=subagent-completion`, source = spawning tool call id/name) instead of a `Role.User` `TextMessage`.
- [ ] `ContextDiscoveryInjector.InjectAsync` emits a `NotifyMessage` (`kind=context-discovery`); a reusable emit helper is provided that trigger notify-mode (#140) and monitors/async-bash/timers (#143) can call.
- [ ] Backward-compat: reloading a pre-migration conversation that persisted the old `context_discovery` `TextMessage`+metadata still renders a single context pill (no duplicate, no user bubble).

### Non-functional
- [ ] A single emit-time debug log records each notification (`kind`, `sourceToolCallId`, `threadId`, `runId`) for troubleshooting. No new config surface, telemetry taxonomy, or rollout flags in this PR — payload size-cap, redaction, and rollout/kill-switch flags are owned by #161.

## Out of scope / follow-ups
- Payload size-cap + PII/secret redaction, backpressure, rollout/kill-switch flags, and ISP hardening (#161). The reusable emit helper introduced here stays minimal — those knobs get added where #161 lands, not bolted onto the helper now.
- Trigger notify-mode delivery (#140), cron/interval sources (#143), durable notify-watcher restore (#145), background fan-out re-enable (#106). This issue provides the typed message + emit helper those consume.
- Strict provider `tool_result` correlation — rejected: fire-and-forget sources have no open tool call to bind to.

## Related
#140, #143, #145, #161, #106 (in the #107 / PR #139 Wait/trigger lineage).

gautam-achieveai:
## Summary

Introduces a first-class **`NotifyMessage`** type so out-of-band notifications pushed into a running `MultiTurnAgent` (async sub-agent completion, monitors, async bash, timers/cron) stop injecting as plain `Role.User` `TextMessage`s. A notification is delivered like any input — **injected into the next turn if a run is in progress, or it starts a turn when the agent is idle** — maps to a **user-role message** for the LLM whose content is a self-describing envelope naming the originating tool call, and renders as a **distinct notification pill** (not a user bubble) in the UI.

Fixes #171.

## Key decisions

- **Text-envelope correlation only** (no provider `tool_result` pairing). `NotifyMessage.Text` is a **computed get-only projection** of the structured fields (`notify_kind`, `source_tool_call_id`, `source_tool_name`, `label`, `detail`) via a single envelope builder, so the LLM envelope and the UI fields can never desync. Envelope: `<notification kind="…" [in-response-to="Name:id"] [label="…"]>detail</notification>`; the payload's closing marker is sanitized to prevent envelope breakout.
- **Delivery reuses the existing gate** (`SendAsync`) — no new emit API. The loop gains notify-awareness: it publishes the notification body live (so the pill renders immediately and lands in the replay buffer), and — the one edge — a notification arriving while the agent is **parked on an unresolved block-`Wait`** is folded into history and delivered by the resume path instead of tripping the deferred-guard (`RunFailed`). A regular user input while deferred keeps its existing fail-fast behavior.
- **Serialization registered before any producer emits**: `notify` in `GetDiscriminatorFromType`, `GetTypeFromDiscriminator`, and an `InferTypeFromProperties` guard **before** the `text` check (keyed on `notify_kind`) — load-bearing because the conversation store persists without `$type`, so a rehydrated notify must infer as `NotifyMessage`, not `TextMessage` (otherwise history recovery would silently downgrade it).
- **No silent drops on any backend**: every provider/CLI request builder pattern-matches concrete `TextMessage`, so an explicit `NotifyMessage` arm was added to Anthropic, OpenAI-Responses, and the Copilot/Codex CLI prompt builders; OpenAI chat needs no change (it maps via `ICanGetText`).
- **Replay-buffer reset keyed on a new runId** (was: every `RunAssignmentMessage`), so a same-run injection no longer truncates a reconnecting client's replay.
- **Rejected**: metadata-on-`TextMessage` (contradicts the explicit distinct-type requirement) and `NotifyMessage : TextMessage` inheritance (`ShadowPropertiesJsonConverter<TextMessage>` is closed-generic and would erase the subtype).

## What changed

- **LmCore**: `NotifyMessage` + `NotifyKinds` + envelope builder; `notify` registered in `IMessageJsonConverter` (3 sites).
- **LmMultiTurn**: gate delivery + live pill publish + parked-Wait safety + replay-reset-on-runId; `SubAgentManager.SendToParentAsync` → `NotifyMessage`; Copilot/Codex `BuildPrompt` arms.
- **Providers**: Anthropic `AddBasicMessageContents`, OpenAI-Responses `MessageMapper` arms.
- **LmAgentInfra**: `ContextDiscoveryInjector` → `NotifyMessage` (retires the legacy `context_discovery` metadata emit; keeps the const for back-compat rendering of old rows).
- **LmWorkflow**: the dormant injected-completion case matches `NotifyMessage` and parses `Detail`, so the #106 fan-out re-enable isn't silently broken.
- **Client (LmStreaming.Sample)**: `notify` token/interface/guard; `NotificationPill.vue`; `displayItems` notification branch **before** the `role==='user'` catch-all (also unifies legacy `context_discovery` rows into one pill); merge-key disambiguation via the per-message `generationId`.

## Testing

- **C#**: full solution builds clean (0 warnings / 0 errors). Green per project: `LmCore.Tests` 784, `LmMultiTurn.Tests` 431 (incl. new notify loop/build-prompt tests: idle-wake, mid-run inject, parked-Wait no-RunFailed + delivered-on-resume, serialization round-trip incl. `$type`-stripped inference), `LmWorkflow.Tests` 195, provider mapping tests on all 5 backends + Anthropic tool_result-adjacency, updated `SubAgentManagerTests` + `ContextDiscoveryInjectorTests`.
- **Client**: 311 Vitest passing (incl. 7 new) — notify routes to a pill not a user bubble, two-in-one-run distinct/ordered live and after reload, legacy `context_discovery` unified pill; `vue-tsc` type-check + production build clean.
- **Bug-focused review**: an adversarial pass over the loop concurrency, serialization, envelope, provider mapping, client, and producers found no correctness issues (the Anthropic `[tool_use][tool_result][notify]` ordering holds because the deferred placeholder is updated in place, not appended).

## Deferred

- **Browser (Playwright) E2E** for the pill (live / reload / two-in-one-run) is not yet run in this PR; the equivalent scenarios are covered at the Vitest unit level and the `data-testid="notification-pill"` / `data-notify-kind` hooks are in place for it.
- Payload size-cap, PII/secret redaction, rollout/kill-switch flags → #161. Trigger notify-mode delivery → #140; cron/interval → #143; durable restore → #145; background fan-out re-enable → #106.

### DIFF
--- a/samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue
+++ b/samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue
@@ -1,0 +1,180 @@
+<script setup lang="ts">
+import { computed, ref } from 'vue';
+import type { NotificationDisplayData, NotifyMessage } from '@/types';
+
+/**
+ * Presentational pill for an out-of-band notification (async sub-agent completion, sandbox
+ * context-discovery, monitors, timers). Distinct from a user bubble. Accepts either the normalized
+ * {@link NotificationDisplayData} that `displayItems` produces, or a raw {@link NotifyMessage}
+ * (normalized here) so callers/tests can pass whichever they hold.
+ */
+const props = defineProps<{
+  notification: NotificationDisplayData | NotifyMessage;
+}>();
+
+/** Normalize the prop to a single shape regardless of whether a raw NotifyMessage was passed. */
+const data = computed<NotificationDisplayData>(() => {
+  const n = props.notification;
+  // Only a raw NotifyMessage carries `$type`; NotificationDisplayData never does.
+  if ('$type' in n) {
+    return {
+      notifyKind: n.notify_kind,
+      label: n.label,
+      sourceToolName: n.source_tool_name,
+      sourceToolCallId: n.source_tool_call_id,
+      detail: n.detail,
+      text: n.text,
+    };
+  }
+  return n;
+});
+
+/** Icon per well-known kind; a bell is the generic fallback for future kinds. */
+const icon = computed<string>(() => {
+  switch (data.value.notifyKind) {
+    case 'context-discovery':
+      return '\u{1F4C4}'; // 📄
+    case 'subagent-completion':
+      return '\u{1F916}'; // 🤖
+    default:
+      return '\u{1F514}'; // 🔔
+  }
+});
+
+/** Human-friendly heading per well-known kind; unknown kinds show the raw kind string. */
+const kindLabel = computed<string>(() => {
+  switch (data.value.notifyKind) {
+    case 'context-discovery':
+      return 'Context loaded';
+    case 'subagent-completion':
+      return 'Sub-agent completed';
+    default:
+      return data.value.notifyKind;
+  }
+});
+
+/** The primary label shown on the header: the file path for context, else the notification label. */
+const primaryLabel = computed<string | null>(() => {
+  if (data.value.notifyKind === 'context-discovery') {
+    return data.value.contextPath ?? null;
+  }
+  return data.value.label ?? null;
+});
+
+/** Expandable body: the pre-rendered detail if present, otherwise the full envelope text. */
+const bodyText = computed<string | null>(() => data.value.detail ?? data.value.text ?? null);
+const hasBody = computed<boolean>(() => !!bodyText.value && bodyText.value.trim().length > 0);
+
+const expanded = ref(false);
+function toggle(): void {
+  if (hasBody.value) {
+    expanded.value = !expanded.value;
+  }
+}
+</script>
+
+<template>
+  <div
+    class="notification-pill"
+    data-testid="notification-pill"
+    :data-notify-kind="data.notifyKind"
+  >
+    <div
+      class="notification-header"
+      :class="{ clickable: hasBody }"
+      @click="toggle"
+    >
+      <span class="notification-icon" aria-hidden="true">{{ icon }}</span>
+      <span class="notification-kind">{{ kindLabel }}</span>
+      <span
+        v-if="data.sourceToolName"
+        class="notification-source"
+        data-testid="notification-source"
+      >&larr; {{ data.sourceToolName }}</span>
+      <span
+        v-if="primaryLabel"
+        class="notification-label"
+        data-testid="notification-label"
+      >{{ primaryLabel }}</span>
+      <span
+        v-if="data.contextTruncated"
+        class="notification-truncated"
+        data-testid="notification-truncated"
+      >(truncated)</span>
+      <span v-if="hasBody" class="notification-expand" aria-hidden="true">{{ expanded ? '▼' : '▶' }}</span>
+    </div>
+    <pre v-if="expanded && hasBody" class="notification-body" data-testid="notification-body">{{ bodyText }}</pre>
+  </div>
+</template>
+
+<style scoped>
+.notification-pill {
+  display: inline-flex;
+  flex-direction: column;
+  gap: 6px;
+  max-width: 100%;
+  padding: 6px 10px;
+  background: #eef2ff;
+  border: 1px solid #c7d2fe;
+  border-radius: 12px;
+  color: #3730a3;
+  font-size: 13px;
+}
+
+.notification-header {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  flex-wrap: wrap;
+  user-select: none;
+}
+
+.notification-header.clickable {
+  cursor: pointer;
+}
+
+.notification-icon {
+  font-size: 15px;
+  flex-shrink: 0;
+}
+
+.notification-kind {
+  font-weight: 600;
+}
+
+.notification-source {
+  color: #4f46e5;
+  font-family: monospace;
+  font-size: 12px;
+}
+
+.notification-label {
+  font-family: monospace;
+  color: #4338ca;
+}
+
+.notification-truncated {
+  color: #9a3412;
+  font-size: 12px;
+}
+
+.notification-expand {
+  color: #6366f1;
+  font-size: 10px;
+  margin-left: auto;
+}
+
+.notification-body {
+  margin: 0;
+  padding: 8px;
+  background: #ffffff;
+  border: 1px solid #e0e7ff;
+  border-radius: 6px;
+  font-size: 12px;
+  line-height: 1.4;
+  white-space: pre-wrap;
+  word-wrap: break-word;
+  overflow-x: auto;
+  color: #1e1b4b;
+}
+</style>
--- a/samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue
+++ b/samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue
@@ -1,33 +1,18 @@
 <script setup lang="ts">
-import { computed, ref } from 'vue';
-import type { NotificationDisplayData, NotifyMessage } from '@/types';
+import { computed, ref, toRef } from 'vue';
+import type { NotificationDisplayData } from '@/types';
 
 /**
  * Presentational pill for an out-of-band notification (async sub-agent completion, sandbox
- * context-discovery, monitors, timers). Distinct from a user bubble. Accepts either the normalized
- * {@link NotificationDisplayData} that `displayItems` produces, or a raw {@link NotifyMessage}
- * (normalized here) so callers/tests can pass whichever they hold.
+ * context-discovery, monitors, timers). Distinct from a user bubble. Takes the normalized
+ * {@link NotificationDisplayData} that `useChat`'s `displayItems` produces — the single normalization
+ * site for both new NotifyMessages and legacy context_discovery rows.
  */
 const props = defineProps<{
-  notification: NotificationDisplayData | NotifyMessage;
+  notification: NotificationDisplayData;
 }>();
 
-/** Normalize the prop to a single shape regardless of whether a raw NotifyMessage was passed. */
-const data = computed<NotificationDisplayData>(() => {
-  const n = props.notification;
-  // Only a raw NotifyMessage carries `$type`; NotificationDisplayData never does.
-  if ('$type' in n) {
-    return {
-      notifyKind: n.notify_kind,
-      label: n.label,
-      sourceToolName: n.source_tool_name,
-      sourceToolCallId: n.source_tool_call_id,
-      detail: n.detail,
-      text: n.text,
-    };
-  }
-  return n;
-});
+const data = toRef(props, 'notification');
 
 /** Icon per well-known kind; a bell is the generic fallback for future kinds. */
 const icon = computed<string>(() => {
--- a/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
@@ -5,10 +5,12 @@ import type {
   UsageMessage,
   ToolCallResultMessage,
   DisplayItem,
+  NotificationDisplayData,
   MessageStatus,
   ReasoningMessage,
   ToolsCallMessage,
   ToolCallMessage,
+  NotifyMessage,
   AuthEvent,
   AuthRequiredEvent,
 } from '@/types';
@@ -29,6 +31,7 @@ import {
   isServerToolUseMessage,
   isServerToolResultMessage,
   isTextWithCitationsMessage,
+  isNotifyMessage,
   normalizeReasoningVisibility,
 } from '@/types';
 import { sendChatMessage } from '@/api/chatClient';
@@ -97,6 +100,15 @@ function getMergeKey(msg: Message, turnSeq = 0): string {
   const messageOrderIdx = msg.messageOrderIdx ?? 0;
   const mergeKind = getMergeKind(msg);
 
+  // A notification carries a distinct generationId ('notify:<guid>' stamped by the backend on the
+  // ONE message object), so two notifications in one run never collide on the shared
+  // runId/messageOrderIdx (both 0), and the live-published copy merges with its reload-parsed twin
+  // (same generationId). Keyed explicitly so a null/missing messageOrderIdx can't fold two distinct
+  // notifications onto one pill.
+  if (isNotifyMessage(msg)) {
+    return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`;
+  }
+
   // Individual tool calls — streaming OR finalized — are keyed by tool_call_id. Several concurrent
   // tool calls in one turn share runId/generationId/messageOrderIdx and differ only by tool_call_id
   // (e.g. GPT-5.5 via the OpenAI Responses API); without it they collapse into a single pill.
@@ -131,12 +143,29 @@ function getMergeKey(msg: Message, turnSeq = 0): string {
   return `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`;
 }
 
-function getMergeKind(msg: Message): 'text' | 'reasoning' | 'tools' | 'tool' | 'other' {
+function getMergeKind(msg: Message): 'text' | 'reasoning' | 'tools' | 'tool' | 'notify' | 'other' {
+  if (isNotifyMessage(msg)) return 'notify';
   if (isTextMessage(msg) || isTextUpdateMessage(msg)) return 'text';
   if (isReasoningMessage(msg) || isReasoningUpdateMessage(msg)) return 'reasoning';
   if (isToolsCallMessage(msg) || isToolsCallUpdateMessage(msg)) return 'tools';
   if (isToolCallMessage(msg) || isToolCallUpdateMessage(msg)) return 'tool';
   return 'other';
+}
+
+/**
+ * Normalize a NotifyMessage into the shape the notification pill renders. Producing the pill data
+ * here (rather than passing the raw message) lets the legacy `context_discovery` TextMessage path
+ * feed the SAME pill via a hand-built {@link NotificationDisplayData}.
+ */
+function notifyToDisplayData(msg: NotifyMessage): NotificationDisplayData {
+  return {
+    notifyKind: msg.notify_kind,
+    label: msg.label,
+    sourceToolName: msg.source_tool_name,
+    sourceToolCallId: msg.source_tool_call_id,
+    detail: msg.detail,
+    text: msg.text,
+  };
 }
 
 /**
@@ -329,7 +358,34 @@ export function useChat(options: UseChatOptions = {}) {
     }
 
     for (const msg of sortedMessages) {
-      if (msg.role === 'user') {
+      const content = msg.content;
+      // Out-of-band notifications render as a distinct pill, NEVER a user bubble — so this branch
+      // must precede the `role === 'user'` catch-all (a NotifyMessage maps to Role.User, and a
+      // reload-parsed one is tagged role 'user'). The legacy pre-migration path — a context_discovery
+      // marker flattened onto a Role.User TextMessage — is folded into the SAME branch so already
+      // persisted rows render one unified context pill (no duplicate, no user bubble) too.
+      if (isNotifyMessage(content)) {
+        flushPill();
+        items.push({
+          type: 'notification',
+          id: msg.id,
+          notification: notifyToDisplayData(content),
+          runId: msg.runId,
+        });
+      } else if (isTextMessage(content) && content.context_discovery != null) {
+        flushPill();
+        items.push({
+          type: 'notification',
+          id: msg.id,
+          notification: {
+            notifyKind: 'context-discovery',
+            contextPath: content.context_discovery.path,
+            contextTruncated: content.context_discovery.truncated,
+            text: content.text,
+          },
+          runId: msg.runId,
+        });
+      } else if (msg.role === 'user') {
         flushPill();
         items.push({
           type: 'user-message',
@@ -772,8 +828,10 @@ export function useChat(options: UseChatOptions = {}) {
     const isUpdate = isTextUpdateMessage(msg) || isReasoningUpdateMessage(msg) ||
                      isToolsCallUpdateMessage(msg) || isToolCallUpdateMessage(msg);
 
-    // Determine if this is a complete (non-update) content message
-    const isCompleteMessage = isTextMessage(msg) || isReasoningMessage(msg) || isToolsCallMessage(msg) || isToolCallMessage(msg);
+    // Determine if this is a complete (non-update) content message. A NotifyMessage is a terminal,
+    // non-streamed message (out-of-band notification) — routed here so it is NOT dropped as an
+    // "unknown message type"; the displayItems notification branch renders it as a pill.
+    const isCompleteMessage = isTextMessage(msg) || isReasoningMessage(msg) || isToolsCallMessage(msg) || isToolCallMessage(msg) || isNotifyMessage(msg);
 
     if (!isUpdate && !isCompleteMessage) {
       // Unknown message type - skip
--- a/src/LmCore/Messages/NotifyMessage.cs
+++ b/src/LmCore/Messages/NotifyMessage.cs
@@ -74,12 +74,16 @@ public record NotifyMessage : IMessage, ICanGetText
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? Detail { get; init; }
 
+    private string? _cachedText;
+
     /// <summary>
-    ///     The self-describing envelope the LLM reads. Computed from the structured fields — never set
-    ///     directly, so it cannot drift from them.
+    ///     The self-describing envelope the LLM reads. Computed once from the structured fields (all
+    ///     <c>init</c>-only, so the value is immutable) and cached — a <see cref="NotifyMessage" /> lives in
+    ///     history and is re-mapped by the active provider on every subsequent turn, so this avoids
+    ///     rebuilding the envelope each access. Never set directly, so it cannot drift from the fields.
     /// </summary>
     [JsonPropertyName("text")]
-    public string Text => BuildEnvelope(NotifyKind, SourceToolName, SourceToolCallId, Label, Detail);
+    public string Text => _cachedText ??= BuildEnvelope(NotifyKind, SourceToolName, SourceToolCallId, Label, Detail);
 
     /// <inheritdoc />
     public string? GetText()
@@ -214,6 +218,12 @@ public class NotifyMessageJsonConverter : ShadowPropertiesJsonConverter<NotifyMe
 {
     protected override NotifyMessage CreateInstance()
     {
+        // NotifyKind is required (see Create), but the converter fills it from the wire's notify_kind.
+        // A hand-crafted/corrupt $type:"notify" payload with the field missing would deserialize to an
+        // empty kind rather than throwing. We deliberately do NOT throw here: this converter also runs on
+        // the unguarded history-rehydration path, so a hard failure would brick recovery of an entire
+        // conversation over one bad row. The gap is unreachable via our own serializer (Create always sets
+        // the field, and structural inference requires notify_kind to route here at all).
         return new NotifyMessage { NotifyKind = string.Empty };
     }
 }
--- a/src/LmCore/Messages/NotifyMessage.cs
+++ b/src/LmCore/Messages/NotifyMessage.cs
@@ -1,0 +1,219 @@
+using System.Collections.Immutable;
+using System.Text;
+using System.Text.Json.Serialization;
+using AchieveAi.LmDotnetTools.LmCore.Utils;
+
+namespace AchieveAi.LmDotnetTools.LmCore.Messages;
+
+/// <summary>
+///     Well-known <see cref="NotifyMessage.NotifyKind" /> values. This is an <b>open</b> set — additional
+///     kinds (monitor/bash/timer/cron) are introduced by their producing features without changing LmCore.
+/// </summary>
+public static class NotifyKinds
+{
+    /// <summary>A background sub-agent finished and reported its result to the parent.</summary>
+    public const string SubAgentCompletion = "subagent-completion";
+
+    /// <summary>A context file discovered by the sandbox was injected into the conversation.</summary>
+    public const string ContextDiscovery = "context-discovery";
+}
+
+/// <summary>
+///     An out-of-band notification pushed into a running conversation from an asynchronous source
+///     (async sub-agent completion, monitors, async bash, timers/cron). It is delivered like any other
+///     input — injected into the next turn when a run is in progress, or starting a turn when the agent
+///     is idle — maps to a <see cref="Role.User" /> message for the LLM whose content is a self-describing
+///     envelope naming the originating tool call, and renders as a distinct notification pill (not a user
+///     bubble) in the UI.
+/// </summary>
+/// <remarks>
+///     <para>
+///         <see cref="Text" /> is a computed projection of the structured fields via the envelope builder,
+///         so the envelope shown to the LLM and the structured fields read by the UI can never desync. The
+///         <see cref="ShadowPropertiesJsonConverter{T}" /> writes the computed value and skips it on read
+///         (get-only), recomputing from the fields — a lossless round-trip.
+///     </para>
+///     <para>
+///         <see cref="Detail" /> is dropped verbatim into the envelope body and is treated as opaque —
+///         consumers must not parse it. The only structured fields the UI/observers should key on are
+///         <see cref="NotifyKind" />, <see cref="SourceToolName" />, <see cref="SourceToolCallId" /> and
+///         <see cref="Label" />.
+///     </para>
+/// </remarks>
+[JsonConverter(typeof(NotifyMessageJsonConverter))]
+public record NotifyMessage : IMessage, ICanGetText
+{
+    /// <summary>
+    ///     Discriminating kind of notification (see <see cref="NotifyKinds" />). Required — a notification
+    ///     without a kind is not a valid <see cref="NotifyMessage" />; the serializer keys structural
+    ///     inference on this field.
+    /// </summary>
+    [JsonPropertyName("notify_kind")]
+    public required string NotifyKind { get; init; }
+
+    /// <summary>
+    ///     Id of the tool call this notification responds to, if any. Null for sources with no originating
+    ///     tool call (timer/cron/context-discovery); the envelope then omits <c>in-response-to</c>.
+    /// </summary>
+    [JsonPropertyName("source_tool_call_id")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? SourceToolCallId { get; init; }
+
+    /// <summary>Name of the tool call this notification responds to, if any.</summary>
+    [JsonPropertyName("source_tool_name")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? SourceToolName { get; init; }
+
+    /// <summary>Short human/UI label (e.g. sub-agent template name, discovered file path).</summary>
+    [JsonPropertyName("label")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Label { get; init; }
+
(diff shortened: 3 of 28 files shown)


gautam-achieveai:
[Revobot]

# Revobot review — PR achieveai/LmDotnetTools#174

Reviewed PR #174 across ~25 files — 0 Must, 0 Should, 2 Consider.

This PR adds `NotifyMessage`, an out-of-band notification that maps to a `Role.User` turn for the LLM (a self-describing `<notification>` envelope) yet renders as a distinct pill in the Vue client. It cleanly replaces two ad-hoc TextMessage-based paths (sub-agent completion and sandbox context-discovery). The change is broad but disciplined, and the test coverage is genuinely thorough — serialization/structural-inference, envelope-format invariants, all five provider mappers, loop-level parked-on-deferral delivery, replay-buffer preservation, and client pill/merge-key behaviour each have targeted tests.

## Correctness / contract review

No correctness, security, or compatibility breaks found. Specifically verified:

- **Serialization contract is symmetric.** `notify` is registered in both directions of `IMessageJsonConverter` (`GetDiscriminatorFromType` and `GetTypeFromDiscriminator`), and a `notify_kind` guard is added to `InferTypeFromProperties` *before* the generic `text` → `TextMessage` fallback (`IMessageJsonConverter.cs:305`). Since `IMessage` discrimination is converter-based (no `[JsonDerivedType]` attributes), map registration is the correct mechanism. The store persists without `$type`, so the structural-inference path is load-bearing and is covered by `Deserialize_NotifyMessage_WithoutTypeDiscriminator_...` plus a `TextMessage` no-regression test.

- **Backward compatibility with pre-migration rows is preserved.** The C# `MetadataKey` constant and the client `TextMessage.context_discovery` field are both retained, and `useChat` displayItems folds a legacy `context_discovery` TextMessage into the same notification pill, verified by the legacy-context reload test.

- **Envelope injection is defended.** `BuildEnvelope` XML-escapes attribute values and `SanitizeBody` neutralizes an embedded `</notification>` closing marker; the breakout attempt is pinned by `Envelope_SanitizesClosingMarker_InPayload_PreventingBreakout`. `Text` is a computed get-only projection, so a stale/hostile persisted `text` is ignored on read (`Text_IsComputedFromFields_...`). Good handling of the untrusted-`Detail` boundary.

- **Replay-buffer change is sound.** `PublishToAllAsync` now only clears the in-flight buffer on a RunAssignment for a *new* run (`incomingRunId != _replayRunId`), preserving it for a same-run injection. First-run behaviour is still correct because `_replayRunActive` starts false.

- **Provider coverage is complete.** Anthropic, OpenAI chat (free via `ICanGetText`), OpenAI Responses, Codex and Copilot each handle `NotifyMessage`, each with a regression test asserting the envelope reaches the request — closing the "silent drop on some backend" gap the Responses `default:` arm would otherwise leave.

## Consider (non-blocking)

1. **Stale comment in `WorkflowRuntime.cs` (~lines 467-471).** The retained comment states "MultiTurnAgentLoop does not publish injected sub-agent completions to its subscribers, so this case does not fire in V1." That premise no longer holds — `MultiTurnAgentLoop.PublishIfNotifyAsync` now publishes injected notifications. The `case NotifyMessage { NotifyKind: SubAgentCompletion, Detail: {} detail }` logic itself is correct (it parses `Detail`, which carries the raw `<sub-agent>` block); only the explanatory comment is now misleading. Worth tightening so a future reader doesn't conclude the branch is dead. (Consider)

2. **Notify merge-key block is structurally identical to the generic fallback.** In `getMergeKey` (`useChat.ts:108`) the notify branch returns `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`, the same shape the `other`-kind fallback would produce (they differ only because `mergeKind === 'notify'`). The explicit branch is harmless and its comment documents intent well, but it is not strictly necessary; fine to keep for clarity or collapse for brevity — reviewer's discretion. (Consider)

## Verdict
**Approve.** The design is coherent, backward-compatible, injection-aware, and well-tested across the serialize/deserialize boundary and every provider; the two notes are documentation/tidiness only.

gautam-achieveai (commented):
[Gautam's review bot] Deep review of #171 — NotifyMessage: typed out-of-band notifications.

## Verdict: APPROVE

Excellent feature PR. Clean architecture, atomic commits, thorough test coverage (serialization round-trips, structural inference regression, loop behavior under deferral, all 5 provider backends, Vue pill rendering + legacy compat). Build is clean with 0 warnings, all new tests pass.

### Highlights
- Self-describing envelope pattern — computed Text from structured fields makes desync impossible
- Replay buffer fix — same-run injection (WasInjected) no longer truncates a reconnecting client's replay buffer
- Parked-on-deferral safety — notifications fold into history when the agent is blocked on a Wait, delivered on resume
- Backward compat — legacy context_discovery TextMessages still render as a unified pill

### Minor observations (non-blocking)

1. **NotifyMessageJsonConverter.CreateInstance()** (NotifyMessage.cs L219) returns NotifyKind = string.Empty — the deserialization path doesn't enforce the ThrowIfNullOrWhiteSpace invariant that Create() does. A malformed JSON (missing notify_kind) would produce a valid-looking object with an empty kind. Consider post-deserialization validation or a sentinel value.

2. **allNotifications LINQ on every loop iteration** (MultiTurnAgentLoop.cs L253-254): realInputs.All(i => ... i.Input.Messages.All(...)) allocates enumerators each pass. Negligible for normal batch sizes, but a foreach early-out would be zero-alloc.

3. **Receipt telemetry gap**: AppendParkedInputsAsync doesn't record input-receipt IDs (the normal StartRunAsync path does). Parked notifications won't appear in input-receipt telemetry — fine for V1, worth a follow-up if telemetry completeness matters.

src/LmMultiTurn/MultiTurnAgentBase.cs
```diff
@@ -988,18 +1003,25 @@ protected ValueTask PublishToAllAsync(IMessage message, CancellationToken ct)
...
-            if (message is RunAssignmentMessage)
+            // Maintain the in-flight run's replay buffer. A RunAssignmentMessage for a NEW run opens a
+            // fresh buffer; RunCompletedMessage closes it (after which a joining subscriber must NOT
+            // replay it — the client already has completed messages via persisted REST history, so
+            // replaying would duplicate). A same-run injection assignment (WasInjected, same RunId — e.g.
+            // an out-of-band NotifyMessage folded into the active run) must NOT clear the buffer, or a
+            // client reconnecting after the injection would lose the run's earlier deltas. Snapshotting
+            // subscribers under the SAME lock SubscribeAsync uses to register + snapshot the buffer
+            // guarantees this message reaches each subscriber exactly once (replay XOR live).
+            if (message is RunAssignmentMessage ram)
```
gautam-achieveai, line 1014: [Claude's review bot] **MEDIUM** (Test coverage — load-bearing branch)

This same-run preservation branch is the mechanism that makes mid-run reconnect correct after a notification: a `WasInjected` `RunAssignmentMessage` with the same `RunId` must **not** clear the replay buffer, while a new-`runId` assignment still must.

I verified this branch is **untested**: `tests/LmMultiTurn.Tests/MultiTurnAgentReplayTests.cs` is not modified by this PR, and its `RunAssignment` test helper only ever builds `new RunAssignment(runId, genId)` — it never sets `WasInjected`. So a one-line regression to the `||` condition here would pass the entire existing suite. Client-visible impact: a client reconnecting mid-run after a notification loses the run's earlier streamed deltas (they still reappear on a full reload from persisted history).

**Suggestion:** add a regression test that (1) publishes a new-run `RunAssignmentMessage` + a few deltas, then a `WasInjected` same-`RunId` assignment, subscribes a *reconnecting* consumer, and asserts it replays the earlier deltas; and (2) asserts a *different*-`runId` assignment still clears the buffer. Cheap now, and it pins both sides of the new condition.

src/LmCore/Messages/NotifyMessage.cs
```diff
@@ -0,0 +1,219 @@
...
+/// <summary>
+///     JSON converter for <see cref="NotifyMessage" /> using the shadow-properties pattern. Mirrors
+///     <c>TextMessageJsonConverter</c>; the computed get-only <see cref="NotifyMessage.Text" /> is written
+///     on serialize and skipped on read (recomputed from the structured fields).
+/// </summary>
+public class NotifyMessageJsonConverter : ShadowPropertiesJsonConverter<NotifyMessage>
+{
+    protected override NotifyMessage CreateInstance()
+    {
+        return new NotifyMessage { NotifyKind = string.Empty };
```
gautam-achieveai, line 227: [Claude's review bot] **LOW** (Robustness)

`CreateInstance()` returns `NotifyKind = string.Empty`, and `ShadowPropertiesJsonConverter` then overwrites it from the JSON `notify_kind` **if present**. A malformed/corrupt payload carrying `$type:"notify"` but no `notify_kind` would deserialize into a `NotifyMessage` with `NotifyKind == ""` — silently bypassing the `ArgumentException.ThrowIfNullOrWhiteSpace(notifyKind)` invariant that the `Create()` factory enforces.

This is narrow (the structural-inference path requires `notify_kind` to route to `NotifyMessage` at all, so it's only reachable via an explicit `$type:"notify"` with the field missing), and it's currently untested.

**Suggestion:** add a post-deserialization guard (throw, or fall back to a documented sentinel) so the converter path upholds the same "kind is required" invariant as `Create()` — or add a short comment noting the deliberate gap. Low priority.

src/LmCore/Messages/NotifyMessage.cs
```diff
@@ -0,0 +1,219 @@
...
+    [JsonPropertyName("detail")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Detail { get; init; }
+
+    /// <summary>
+    ///     The self-describing envelope the LLM reads. Computed from the structured fields — never set
+    ///     directly, so it cannot drift from them.
+    /// </summary>
+    [JsonPropertyName("text")]
+    public string Text => BuildEnvelope(NotifyKind, SourceToolName, SourceToolCallId, Label, Detail);
```
gautam-achieveai: [Claude's review bot] **LOW** (Performance — GC/allocation)

`Text` (and `GetText()`, which forwards to it) rebuilds the envelope on **every** access — a fresh `StringBuilder`, up to five `string.Replace` calls in `EscapeAttribute`, and one `SanitizeBody` `string.Replace` that scans/reallocates the entire `Detail` body. Because a `NotifyMessage` stays in `ConversationHistory`, the active provider re-maps it (reading `Text`) on every subsequent turn — so the envelope is rebuilt ~2–3× at delivery plus once per future turn.

This is GC pressure only, never correctness, and it's genuinely **low frequency** (notifications are async out-of-band events, only one provider is active per request, and the re-map count scales with turn count — tens — not tokens). So this is a nice-to-have, not a blocker.

**Suggestion (optional):** memoize once — the fields are `init`-only, so the value is immutable:
```csharp
private string? _cachedText;

[JsonPropertyName("text")]
public string Text => _cachedText ??= BuildEnvelope(NotifyKind, SourceToolName, SourceToolCallId, Label, Detail);
```
(A nullable backing field is enough; `Lazy<string>` would add its own allocation. Confirm it isn't itself serialized by the shadow converter.)

samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue
```diff
@@ -0,0 +1,180 @@
...
+import type { NotificationDisplayData, NotifyMessage } from '@/types';
+
+/**
+ * Presentational pill for an out-of-band notification (async sub-agent completion, sandbox
+ * context-discovery, monitors, timers). Distinct from a user bubble. Accepts either the normalized
+ * {@link NotificationDisplayData} that `displayItems` produces, or a raw {@link NotifyMessage}
+ * (normalized here) so callers/tests can pass whichever they hold.
+ */
+const props = defineProps<{
+  notification: NotificationDisplayData | NotifyMessage;
```
gautam-achieveai: [Claude's review bot] **LOW** (Dead flexibility / over-generalization)

This prop accepts a union `NotificationDisplayData | NotifyMessage`, and the `'$type' in n` normalization branch just below (`data` computed) exists to handle the raw-`NotifyMessage` case. I checked the callers: the only production use of `NotificationPill` is `MessageList.vue`, which always binds `:notification="item.notification"` — a `NotificationDisplayData` produced by `displayItems`. Nothing in production passes a raw `NotifyMessage`; the normalization branch is reached only by its own unit test, and it duplicates `notifyToDisplayData` in `useChat.ts`.

**Suggestion:** narrow the prop to `NotificationDisplayData` only and drop the in-component normalization branch (and the raw-message test that exists solely to cover it). Keeps a single normalization site (`useChat.ts`) and simplifies the component. Non-blocking.

samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts
```diff
@@ -97,6 +100,15 @@ function getMergeKey(msg: Message, turnSeq = 0): string {
   const messageOrderIdx = msg.messageOrderIdx ?? 0;
   const mergeKind = getMergeKind(msg);
 
+  // A notification carries a distinct generationId ('notify:<guid>' stamped by the backend on the
+  // ONE message object), so two notifications in one run never collide on the shared
+  // runId/messageOrderIdx (both 0), and the live-published copy merges with its reload-parsed twin
+  // (same generationId). Keyed explicitly so a null/missing messageOrderIdx can't fold two distinct
+  // notifications onto one pill.
+  if (isNotifyMessage(msg)) {
```
gautam-achieveai, line 108: [Claude's review bot] **TRIVIAL** (Cleanliness — your discretion)

This explicit `notify` branch returns `${mergeKind}-${runId}-${generationId}-${messageOrderIdx}`, which is the same shape the generic fallback at the bottom of `getMergeKey` produces (they differ only because `mergeKind === 'notify'` — which `getMergeKind` already yields for a notify). So the branch is functionally redundant: removing it, a notify would fall through to the fallback and compute an identical key.

It's harmless and the comment documents the intent (the `notify:<guid>` `generationId` is what actually disambiguates two notifications in one run). Fine to keep for readability or collapse for brevity — entirely your call. Flagging only for awareness.

src/LmWorkflow/Runtime/WorkflowRuntime.cs
```diff
@@ -464,14 +464,15 @@ public void ObserveMessage(IMessage message)
 
                 break;
 
-            // Injected <sub-agent> completion (DORMANT / forward-built in V1): MultiTurnAgentLoop does not
+            // Injected sub-agent completion (DORMANT / forward-built in V1): MultiTurnAgentLoop does not
             // publish injected sub-agent completions to its subscribers, so this case does not fire in V1.
             // Even if it did, the background receipt path now fails fast (ObserveSpawnResult) and never records
             // an agent_id correlation, so ObserveInjectedResult would find no mapping and surface the result as
             // harmless "unmatched". Re-enabled in the follow-up that makes injected completions observable.
```
gautam-achieveai: [Claude's review bot] **LOW** (Documentation accuracy)

The retained comment says *"MultiTurnAgentLoop does not publish injected sub-agent completions to its subscribers, so this case does not fire in V1."* That premise no longer holds in this same PR: `MultiTurnAgentLoop.PublishIfNotifyAsync` now publishes injected `NotifyMessage`s (including sub-agent completions), so this `case` **is** reachable if a host wires the loop's message stream into `ObserveMessage`.

It's not a correctness bug — the `case NotifyMessage { NotifyKind: SubAgentCompletion, Detail }` logic is right, and even when it fires the background-receipt path records no `agent_id` correlation, so `ObserveInjectedResult` surfaces a harmless "unmatched". But the first sentence now misleads a future reader into thinking the branch is dead.

**Suggestion:** tighten the comment to say injected notifications are now published (so the branch is reachable) but remain safe because the receipt path never records a correlation in V1.

gautam-achieveai:
[Claude's review bot]

# PR Review: NotifyMessage — typed out-of-band notifications tied to originating tool calls (#171)

## Summary
- **Files reviewed:** 25 (+1589 / −71), 6 atomic commits
- **Findings:** 0 Critical, 0 High, 2 Medium, 4 Low, 1 Trivial
- **Context questions:** 0
- **Test coverage:** strong overall (28 new/updated test methods) — one meaningful gap (see M1)
- **Domain areas touched:** LmCore (new type + serializer), LmMultiTurn (run loop + replay buffer), 4 provider request builders, LmWorkflow, LmAgentInfra, Vue sample client
- **Backward/forward compat:** SAFE — every schema change is additive or has a working migration (verified against a rolling-deploy model)

## Verdict
**APPROVE WITH COMMENTS**

This is a well-designed, disciplined foundational feature. The implementation is coherent, backward-compatible, injection-aware, and well-tested across the serialize/deserialize boundary and every provider. No correctness, security, or compatibility **break** was found by a deep manual pass or by four specialized sub-reviews (schema-compatibility, test-coverage, performance, over-engineering). The two Medium items are "should fix / track" — neither is a correctness bug or data-loss in shipped SDK surface — so they do not block, but they are worth addressing.

## Strengths (genuinely good work)
- **Desync-proof envelope.** `NotifyMessage.Text` is a computed get-only projection of the structured fields (`NotifyMessage.cs:82`), so the LLM envelope and the UI fields can never drift. The `ShadowPropertiesJsonConverter` writes it and recomputes on read — a lossless round-trip.
- **Serialization is symmetric and the inference guard is correctly ordered.** `notify` is registered in both discriminator directions, and the `notify_kind` guard in `InferTypeFromProperties` (`IMessageJsonConverter.cs:305`) runs **before** the generic `text` check. This is load-bearing because the store persists without `$type`; a `$type`-stripped notify correctly rehydrates as `NotifyMessage`, not `TextMessage` — and there's a `TextMessage` no-regression test for the other side.
- **Envelope is defended against breakout.** Attribute values are XML-escaped and the payload's `</notification>` closing marker is sanitized; a hostile-payload breakout attempt is pinned by a test.
- **Replay-buffer fix is correct.** Keying the reset on a *new* `runId` (instead of every `RunAssignmentMessage`) so a same-run injection preserves a reconnecting client's earlier deltas is the right fix, and the `RunCompleted` lifecycle still clears correctly (verified).
- **Parked-on-`Wait` safety is sound and rigorously tested.** A notification arriving while parked on an unresolved deferral is folded into history and delivered on resume instead of tripping `RunFailed`; the test asserts *both* no-`RunFailed` and delivery-in-the-turn-2-payload.
- **No silent drops on any backend.** Anthropic, OpenAI Responses, Codex and Copilot each got an explicit `NotifyMessage` arm (OpenAI chat maps via `ICanGetText`), each with a regression test.
- **Well-scoped.** The +1589 breadth is inherent to introducing a first-class message type; the author declined a speculative emit-helper abstraction and deferred `#161`'s knobs. No drive-by refactors.

## Medium Issues
| # | File | Line | Blocker? | Issue | Fix |
|---|------|------|----------|-------|-----|
| M1 | src/LmMultiTurn/MultiTurnAgentBase.cs | 1014 | No | The load-bearing replay-buffer branch (preserve on same-run `WasInjected` injection vs. clear on new `runId`) is **untested**. Verified: `MultiTurnAgentReplayTests.cs` is not modified by this PR and its `RunAssignment` helper never sets `WasInjected`, so a one-line regression to the `\|\|` would pass the entire suite. Client-visible impact: a client reconnecting mid-run after a notification loses the run's earlier streamed deltas (they still appear on a full reload). | Add a regression test: publish a new-run assignment + deltas, then a `WasInjected` same-run assignment, subscribe a *reconnecting* consumer and assert it replays the earlier deltas; also assert a different-`runId` assignment *does* clear. |
| M2 | samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts | 836 | No | Rollout sequencing: a new server sending `$type:"notify"` to an **old cached SPA** during the deploy window is dropped as "unknown message type" — live pills silently vanish until a hard-refresh. Non-destructive (envelope persists, renders on next reload) and the affected client is a sample app. Explicitly deferred to #161. | Track under #161 (already deferred). Optionally gate producer emission behind the #161 flag so the server keeps emitting `TextMessage` until clients are known-updated. |

## Low / Trivial Issues
| # | File | Line | Blocker? | Issue | Fix |
|---|------|------|----------|-------|-----|
| L1 | src/LmWorkflow/Runtime/WorkflowRuntime.cs | 471 | No | Stale comment: "MultiTurnAgentLoop does not publish injected sub-agent completions… so this case does not fire in V1." This PR's `PublishIfNotifyAsync` now **does** publish injected notifications, so the branch **is** reachable if a host wires the loop stream into `ObserveMessage`. Not a bug (the fail-fast receipt path keeps it "harmless unmatched") — but the comment misleads. | Update the comment: injected notifications are now published (branch reachable) but still safe because the background-receipt path records no `agent_id`, so `ObserveInjectedResult` surfaces "unmatched". |
| L2 | src/LmCore/Messages/NotifyMessage.cs | 217 | No | `NotifyMessageJsonConverter.CreateInstance()` returns `NotifyKind = string.Empty`. A corrupt `$type:"notify"` payload missing `notify_kind` yields `NotifyKind=""`, bypassing the `ThrowIfNullOrWhiteSpace` invariant that `Create()` enforces. Only reachable via a hand-crafted/corrupt payload (inference requires `notify_kind`). Untested. | Add a post-deserialization validation (throw or sentinel), or a documented note. |
| L3 | src/LmCore/Messages/NotifyMessage.cs | 82 | No | `Text` rebuilds its envelope (StringBuilder + up to 5 `string.Replace` + an O(len(detail)) `SanitizeBody` scan) on **every** access. The message persists in history and is re-mapped by the active provider each subsequent turn. GC pressure only, never correctness; calibrated LOW (notifications are low-frequency async events, one active provider, re-map scales with turns not tokens). | Optional lazy memoization: `private string? _cachedText; Text => _cachedText ??= BuildEnvelope(...)` — fields are init-only so the value is immutable. |
| L4 | samples/LmStreaming.Sample/ClientApp/src/components/NotificationPill.vue | 12 | No | The prop accepts a union `NotificationDisplayData \| NotifyMessage` with a normalization branch reached **only** by its own test — the sole production caller (`MessageList.vue`) always passes `NotificationDisplayData`, and the branch duplicates `notifyToDisplayData` in `useChat.ts`. | Narrow the prop to `NotificationDisplayData`; drop the in-component normalization branch and its test. |
| T1 | samples/LmStreaming.Sample/ClientApp/src/composables/useChat.ts | 108 | No | The notify branch in `getMergeKey` returns a key of identical shape to the generic fallback at L143 (`mergeKind` is `'notify'` in both). Functionally redundant; the comment documents intent well. | Keep for clarity or collapse for brevity — your discretion. |

## Testing Assessment
Coverage is strong: serialization round-trip incl. `$type`-stripped inference + a `TextMessage` no-regression; envelope escaping/sanitization breakout; all five provider mappers incl. Anthropic `tool_result` adjacency; loop idle-wake, mid-run inject, and the parked-`Wait` no-`RunFailed`+delivered-on-resume path; client pill routing (not a user bubble), two-in-one-run distinct/ordered live and after reload, and the legacy `context_discovery` unified pill. **Gaps:** the same-run replay-buffer preservation branch (M1 — the one worth adding) and the `CreateInstance()` empty-`NotifyKind` path (L2).

## Security Review
No issues. The one untrusted boundary — the opaque `Detail` payload dropped into the envelope — is handled: attributes are XML-escaped and the closing marker is sanitized to prevent envelope breakout. (Payload size-cap and PII/secret redaction are explicitly deferred to #161.)

## Recommendations
1. **Add the M1 regression test** — it's cheap now (freshest context) and guards the feature's core mid-run-reconnect guarantee against a silently-reintroducible one-line regression.
2. Update the L1 stale comment in `WorkflowRuntime.cs` in this PR, since this PR is what makes the branch reachable.
3. L2–L4, T1 are optional polish — address at your discretion.

---
*Independent deep review (worktree checkout). Cross-checked against two prior bot reviews; findings M1 (replay test gap) and M2 (client rollout drop) are additive to what they surfaced. No correctness/security/compat BLOCKER found.*

gautam-achieveai:
[Revobot]

Notes and findings persisted. Here is the finished review.

---

# Revobot re-review (round 02) — PR achieveai/LmDotnetTools#174

Re-reviewed the delta since 2a46540 across 3 files (1 production, 2 test) — 0 Must, 0 Should, 0 new Consider.

## Delta
One commit landed since the last review: render the notification pill in the active-conversation section and add a browser E2E test. Round 01 already approved the full PR; this follow-up is small and correct.

## Assessment

- **`MessageList.vue` (+6) — genuine bug fix.** The template renders two group lists, `splitGroups.history` and `splitGroups.current` (the active conversation). At the previously-reviewed commit the `item.type === 'notification'` branch existed only in the history section, so a *live* sub-agent completion — which lands in the active group — fell through every `v-else-if` arm and rendered nothing on screen. This commit adds the identical branch to the active section (`MessageList.vue:277`), so out-of-band notifications actually appear as they arrive. Minimal and correct.

- **`NotificationPillTests.cs` (+110) / `UiHelpers.cs` (+9) — good E2E coverage.** Drives the real background path (`Agent` with `run_in_background: true` → parent goes idle → researcher completes → `NotifyMessage` published live over the WebSocket) and pins the two decisive properties: a `notification-pill[data-notify-kind="subagent-completion"]` appears (`NotificationPillTests.cs:94`), and the user-group count stays at exactly 1 (`:105`), proving the `Role.User` notify never renders as a second user bubble. Runs for both `test` and `test-anthropic`. The run‑2-vs-idle-wake race is handled by DOM polling with a generous timeout and documented, rather than asserted on stream-idle. All referenced helpers (`UserMessageGroups`, `ToolCallPills`, `NotificationPills`, `WaitForCountAtLeastAsync`) exist, so the test compiles.

No correctness, security, or contract concerns in the delta.

## Prior notes (unchanged by this commit, non-blocking)
- Stale comment in `WorkflowRuntime.cs` (~L467) still claims the loop "does not publish injected sub-agent completions" — no longer true, though the branch logic itself is correct. (Consider)
- The explicit notify `getMergeKey` branch (`useChat.ts`) is structurally identical to the generic fallback; harmless, kept for documented clarity. (Consider)

## Verdict
Approve — the delta fixes a real live-render gap and lands solid E2E coverage; nothing new to block on.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental review** of commit `a740f78` - render notification pill in active-conversation section + browser E2E.

## Verdict: APPROVE

The incremental change is clean and correct:

1. **MessageList.vue** - Adds NotificationPill to the active-conversation section (parity with the scrollback section from a prior commit). Without this, live notifications wouldn't render until the user scrolled back.

2. **NotificationPillTests.cs** - Thorough E2E test exercising the full background sub-agent -> NotifyMessage -> pill pipeline. Validates: notification-pill[data-notify-kind=subagent-completion] appears, Agent tool-call pill is present, and the notify (mapped to Role.User) does NOT render as a second user bubble. Covers both test and test-anthropic provider modes.

3. **UiHelpers.cs** - Simple locator helper, consistent with existing ToolCallPills()/UserMessageGroups() patterns.

Build: **0 warnings, 0 errors**. No new concerns beyond the 3 minor observations from the initial review (all non-blocking, unchanged by this iteration).

gautam-achieveai:
[gautam-achieveai's bot] Addressed the review feedback in `5610951e`. Disposition of each point:

| # | Finding | Severity | Disposition |
|---|---------|----------|-------------|
| 1 | Replay-buffer same-run preservation branch is untested (`MultiTurnAgentBase.cs`) | **MEDIUM** | **Fixed** — added two regression tests to `MultiTurnAgentReplayTests` pinning both sides of the runId-keyed reset: a `WasInjected` **same-runId** `RunAssignment` does NOT clear the buffer (a reconnecting subscriber still replays the run's earlier deltas), and a **new-runId** `RunAssignment` DOES clear it. A one-line regression to the `||` condition now fails. |
| 2 | `NotifyMessage.Text` rebuilds the envelope on every access | LOW | **Fixed** — memoized via a nullable `_cachedText` backing field (fields are init-only ⇒ immutable). Confirmed the field isn't serialized (the shadow converter only writes `[JsonPropertyName]` properties). |
| 3 | `CreateInstance` empty-kind bypasses the `Create()` invariant | LOW | **Documented** — added a comment on the deliberate gap. Chose not to throw: this converter also runs on the *unguarded* history-rehydration path, so a hard failure would brick recovery of a whole conversation over one bad row. The gap is unreachable via our own serializer (`Create` always sets the field; structural inference requires `notify_kind` to route here at all). |
| 4 | `NotificationPill.vue` union prop / dead raw-`NotifyMessage` flexibility | LOW | **Fixed** — narrowed the prop to `NotificationDisplayData` (production's only shape), dropped the in-component normalization branch and its unit test, so `useChat.displayItems` is the single normalization site. |
| 5 | `getMergeKey` notify branch is redundant with the fallback | TRIVIAL | **Kept** (your discretion) — retained for the in-place comment documenting the non-obvious `notify:<guid>` `generationId` disambiguation invariant, at the point a reader looks for notify merge-key logic. |
| 6 | `WorkflowRuntime` comment says injected completions aren't published | LOW | **Fixed** — corrected: injected notifications *are* now published (#171), so the case is reachable but stays safe in V1 (the background receipt path records no `agent_id`, so it surfaces a harmless "unmatched"). |
| 7 | `allNotifications` LINQ allocates per drain (`MultiTurnAgentLoop.cs`) | obs | **Fixed** — replaced with a zero-allocation `foreach` helper (`AllMessagesAreNotifications`), and reordered so the cheap `resumeSentinels`/`_deferred` checks short-circuit first. |
| 8 | `AppendParkedInputsAsync` doesn't record input-receipt IDs | obs | **Deferred** (V1) — as noted, fine for now; worth a follow-up if input-receipt telemetry completeness matters for parked notifications. |

Verification: full solution builds **0/0**; `LmCore.Tests` NotifyMessage 10, `LmMultiTurn.Tests` replay/notify/deferred 28 (incl. the 2 new replay tests), `LmWorkflow` builds, client `vue-tsc` type-check clean + notify Vitest green.

gautam-achieveai:
[Revobot]

I'll review this PR, focusing on changes since the last review. Let me examine the key changed files.

Since this is a re-review (round 03), let me focus on what changed since a740f78 and re-check prior findings. The diff shows the full PR scope. Let me examine the core new files and the areas most prone to bugs.

Let me look at the critical files.


name: Read
input:
  file_path: /workspace/store/repos/LmDotnetTools/src/LmCore/Messages/NotifyMessage.cs

gautam-achieveai (commented):
[Gautam's review bot] **Incremental review** of commit `5610951` - addresses all prior review feedback.

## Verdict: APPROVE

All 6 observations from iterations 1-7 have been addressed:
- Dead `NotifyMessage` union removed from `NotificationPill.vue` (now `toRef` on single-type prop)
- `_cachedText` lazy cache on `Text` eliminates per-access StringBuilder allocation
- `AllMessagesAreNotifications()` extracted as zero-alloc foreach (replaces LINQ `.All()`)
- `CreateInstance()` comment explains empty-kind tolerance for history rehydration
- `WorkflowRuntime` comment updated to reflect live publish semantics
- Two new replay buffer tests cover same-run injection preservation and new-run reset

Build: **0 warnings, 0 errors**. All 10 replay tests pass.

### One new observation (LOW - non-blocking)

**Record equality and _cachedText** (`NotifyMessage.cs` L77): The private `_cachedText` field participates in the compiler-synthesized Equals/GetHashCode for this record type. Two structurally identical `NotifyMessage` instances will compare unequal if one has accessed `.Text` (populating the cache) and the other hasn't. Practically harmless since these live in history lists by reference, never as dictionary keys - but worth a mental note if NotifyMessage equality ever becomes load-bearing.

PR #175, opened 2026-07-09
feat(daemon): persist review conversations + optional JSONL logging

gautam-achieveai:
## Why

The review agents' `MultiTurnAgentLoop` was created with **no conversation store**, so every run's messages — `Skill` loads, sub-agent `Task` dispatches, tool results — were streamed and discarded. Nothing was persisted, so "which skills/agents did that review use?" was unanswerable from any store or log.

## What (both opt-in; unset ⇒ unchanged)

- **`CodeReviewDaemon:ConversationStorePath`** — wires a `FileConversationStore` into the primary loop **and** its sub-agents (`SubAgentOptions.DefaultConversationStoreFactory`), so each thread's full history persists to `<path>/<threadId>/messages.json` (JSON, DuckDB-queryable). `persistRunLedger` on when a store is present.
- **`CodeReviewDaemon:LogFilePath`** — adds a Serilog `CompactJsonFormatter` JSONL file sink (daily-rolled) **alongside** the console logger, so the daemon's own logs are DuckDB-queryable. Isolated in `Hosting/DaemonLogging` to avoid a `using Serilog;` / `ILogger` clash in `Program.cs`.

## Verify

467 daemon tests green; 0-warning. Live: with `ConversationStorePath` set, each review writes `messages.json` (tool calls incl. Skill/Task) queryable via `duckdb -c "SELECT ... FROM read_json('.../messages.json')"`.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs
+++ b/samples/CodeReviewDaemon.Sample/Agents/LiveReviewAgentLoopFactory.cs
@@ -8,6 +8,7 @@ using AchieveAi.LmDotnetTools.LmCore.Agents;
 using AchieveAi.LmDotnetTools.LmCore.Core;
 using AchieveAi.LmDotnetTools.LmCore.Middleware;
 using AchieveAi.LmDotnetTools.LmMultiTurn;
+using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence;
 using AchieveAi.LmDotnetTools.McpMiddleware.Extensions;
 using AchieveAi.LmDotnetTools.OpenAiResponsesProvider.Agents;
 using AchieveAi.LmDotnetTools.OpenAiResponsesProvider.Models;
@@ -50,16 +51,21 @@ internal sealed class LiveReviewAgentLoopFactory : IReviewAgentLoopFactory, IDis
 {
     private readonly ILoggerFactory _loggerFactory;
     private readonly CodeReviewDaemonOptions _options;
+    private readonly IConversationStore? _conversationStore;
     private readonly ICopilotTokenProvider _tokenProvider = new CliCredentialCopilotTokenProvider();
     private readonly CopilotSessionContext _session = new();
     private readonly object _agentGate = new();
     private IStreamingAgent? _anthropicAgent;
     private IStreamingAgent? _openAiAgent;
 
-    public LiveReviewAgentLoopFactory(ILoggerFactory loggerFactory, CodeReviewDaemonOptions options)
+    public LiveReviewAgentLoopFactory(
+        ILoggerFactory loggerFactory,
+        CodeReviewDaemonOptions options,
+        IConversationStore? conversationStore = null)
     {
         _loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
         _options = options ?? throw new ArgumentNullException(nameof(options));
+        _conversationStore = conversationStore;
     }
 
     /// <summary>
@@ -150,6 +156,15 @@ internal sealed class LiveReviewAgentLoopFactory : IReviewAgentLoopFactory, IDis
                 string.Join(",", keptContracts.Select(c => c.Name)));
         }
 
+        // Persist the conversation (primary + sub-agents) when a store is configured, so every review's tool
+        // calls — Skill loads and sub-agent Task dispatches — are auditable after the run. Sub-agents share the
+        // same store (keyed per thread) via DefaultConversationStoreFactory unless one is already supplied.
+        var subAgentOptions = toolContext?.SubAgentOptions;
+        if (_conversationStore is not null && subAgentOptions is { DefaultConversationStoreFactory: null })
+        {
+            subAgentOptions = subAgentOptions with { DefaultConversationStoreFactory = _ => _conversationStore };
+        }
+
         var loop = new MultiTurnAgentLoop(
             providerAgent,
             registry,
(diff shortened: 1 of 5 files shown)


gautam-achieveai:
[Revobot]

Notes persisted. Here is the finished review.

---

Reviewed PR #175 across 4 changed files (+5 supporting files read) — 0 Must, 0 Should, 2 Consider.

This is a small, cleanly-scoped, opt-in change to the `CodeReviewDaemon.Sample` project that adds two operator features: optional conversation persistence (`ConversationStorePath` → `FileConversationStore`) and optional structured JSONL logging (`LogFilePath` → Serilog). Both default to off and preserve existing behavior exactly. No correctness bugs, security issues, or contract/compatibility breaks were found.

**Correctness / compatibility — verified sound:**
- The riskiest line, `persistRunLedger: _conversationStore is not null` (`LiveReviewAgentLoopFactory.cs:183`), is safe: `MultiTurnAgentBase`'s constructor throws `ArgumentException` if `persistRunLedger` is true but the store doesn't implement `IRunLedgerStore` — and `FileConversationStore` implements both `IConversationStore` and `IRunLedgerStore` (`FileConversationStore.cs:9`). The flag is only ever true when the store is a `FileConversationStore`, so the guard never trips.
- Sharing one `FileConversationStore` across the primary loop and concurrent sub-agents (via `DefaultConversationStoreFactory = _ => _conversationStore`) is safe: writes are keyed per `threadId` into isolated directories and serialized by a `SemaphoreSlim`.
- The new optional constructor parameter (`IConversationStore? conversationStore = null`) is backward compatible; existing callers and the routing tests (which use static methods) are unaffected. The DI re-registration in `Program.cs:222` preserves the prior `ILoggerFactory`/options resolution.
- Serilog package versions (`Serilog 4.2.0`, `Extensions.Logging 9.0.0`, `Sinks.File 6.0.0`, `Formatting.Compact 3.0.0`) match `LmTestUtils` exactly, as the csproj comment claims.
- Comments are accurate and not misleading; no prompt-injection content was encountered in the diff or files read.

**Maintainability / coverage:**

- **Consider** — No retention bound on either persistence path (`Hosting/DaemonLogging.cs:26-28`). The JSONL sink uses `MinimumLevel.Debug()` with `rollingInterval: Day` but no `retainedFileCountLimit` or `fileSizeLimitBytes`, so on a long-running daemon the log dir grows without bound; `ConversationStorePath` similarly keeps every review's full history forever. Both are opt-in operator paths, hence Consider rather than a blocker — but a bound (or a documented cleanup expectation) would prevent a slow disk-fill.

- **Consider** — No tests were added. Nothing exercises the `ConversationStorePath`/`LogFilePath` defaults, the `persistRunLedger: store != null` wiring, or the sub-agent `DefaultConversationStoreFactory` hookup. The persistRunLedger↔IRunLedgerStore coupling is a silent runtime invariant; a cheap unit test (in the existing `CodeReviewDaemonOptionsTests`) asserting the defaults and that a set path produces a persisting loop would lock it in.

Verdict: **Approve with comments** — sound and backward-compatible; the two Consider items (retention bounds, a little test coverage) are worth a follow-up but need not block.

gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #175.

## Summary

Well-structured opt-in features (conversation persistence + JSONL logging) that don't break existing behavior. Both default to off, and the daemon's serial execution model makes the shared `FileConversationStore` singleton safe. The `SubAgentOptions` patching logic correctly guards against overwriting an existing factory.

## Findings

### MEDIUM — Startup crash on bad `ConversationStorePath` (no graceful degradation)

**File:** `Program.cs:214-216`

`FileConversationStore`'s constructor calls `Directory.CreateDirectory` which can throw `IOException`, `UnauthorizedAccessException`, `PathTooLongException` etc. on misconfigured paths. Since conversation persistence is a **non-critical, opt-in** audit feature, a bad path should degrade to "no persistence" with a clear log message rather than crashing the daemon at startup.

**Suggestion:** Wrap in try-catch in `Program.cs` and fall back to `null`:
`csharp
IConversationStore? conversationStore = null;
if (!string.IsNullOrWhiteSpace(daemonOptions.ConversationStorePath))
{
    try { conversationStore = new FileConversationStore(daemonOptions.ConversationStorePath); }
    catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or ArgumentException)
    {
        logger.LogError(ex, "Conversation store at '{Path}' unavailable; persistence disabled",
            daemonOptions.ConversationStorePath);
    }
}
`

### MEDIUM — `MinimumLevel.Debug()` on JSONL file sink is too noisy for an audit log

**File:** `DaemonLogging.cs:26`

The stated purpose is "DuckDB-queryable" audit/diagnostics. With `Debug`, when an operator lowers log levels for troubleshooting (common in production), the JSONL file captures all 43+ `LogDebug` sites in `LmMultiTurn` x sub-agents x turns — potentially thousands of entries per review, making the audit data hard to query for its intended purpose.

Additionally, no `fileSizeLimitBytes` or `retainedFileCountLimit` is set — unbounded daily growth on disk.

**Suggestion:** Use `MinimumLevel.Information()` (or make it configurable), and add retention:
`csharp
.WriteTo.File(new CompactJsonFormatter(), path,
    rollingInterval: RollingInterval.Day,
    fileSizeLimitBytes: 100_000_000,
    retainedFileCountLimit: 14)
`

### MEDIUM — Serilog silently drops logs on invalid `LogFilePath`

**File:** `Program.cs:40-43`

Serilog's file sink swallows file-creation failures by default (its SelfLog goes nowhere). If an operator configures a path to a non-existent or inaccessible directory, they get **no error** and **no JSONL output** — they believe logging is active but nothing is written. Pre-validate or wrap with a try-catch similar to the store path.

### LOW — No new tests for the new features

The commit claims "467 tests green" but no new tests exercise: (a) the sub-agent options patching logic (only patches when `DefaultConversationStoreFactory` is null), (b) the new `ConversationStorePath`/`LogFilePath` config binding, (c) `DaemonLogging.AddJsonlFileSink` guard clauses. Consider at minimum adding the new options to the existing `Binds_every_flag_from_the_CodeReviewDaemon_section` test.

---

**Verdict:** Approve with suggestions. The PR is well-designed and properly opt-in. The startup resilience items are the most actionable — a single try-catch each would prevent the daemon from crashing on config typos.

PR #178, opened 2026-07-09
feat(#145): durable notify-watcher restore across restart in LmStreaming.Sample

### LINKED ISSUE #145: Wait/trigger: durable notify-watcher restore across restart

gautam-achieveai:
## Summary

Fixes #145. Notify-mode waits in the `LmStreaming.Sample` host were process-lifetime only — nothing persisted their arming records, so a restart silently dropped them. The core durability primitive already exists on `main` (from PR #158): `INotifyWaitStore`/`SqliteNotifyWaitStore`, the `notify_waits` table, `TriggerRuntime.RestoreNotifyWaitsAsync`, `TriggerOptions.NotifyWaitStore`/`ThreadId`, and `MultiTurnAgentLoop.OnThreadRecoveredAsync` already calls restore on every thread recovery. **This was purely a sample-host wiring gap** — the sample never constructed a store nor set `NotifyWaitStore`/`ThreadId`.

## Changes

- **Register a standalone SQLite-backed `INotifyWaitStore`** in the sample host — its own db file, configurable via the `LmStreaming:NotifyWaitDbPath` key (default beside the existing `conversations/` folder). The conversation store stays `FileConversationStore`; only notify-wait persistence uses SQLite.
- **Attach the store + per-thread `ThreadId`** onto the `TriggerOptions` the pool's agent factory already builds, via a record `with` expression at the call site. `SampleTriggerRegistrations.Build`'s signature is **unchanged**. Gated to test-mode only, matching the existing `triggerOptions: isTestMode ? … : null` pass-through; broader real-provider rollout is tracked in #161.
- **New `NotifyWaitDurableRestoreTests`** — a `WebApplicationFactory<Program>` integration test that seeds notify rows + thread metadata on disk, boots the real host over that state, and asserts against **real on-disk SQLite** that:
  - a `schedule` wait (restorable) re-arms and survives the restart, and
  - an unregistered/non-restorable `process` wait delivers `trigger_lost_on_restart` and is cleared.

  This is the one scenario no existing in-memory test covered.

## Design note — deviation from the approved plan (DI registration)

The approved plan proposed registering `ISqliteConnectionFactory` as its own DI singleton so the container disposes it. During implementation this was found to be **wrong for this host**: `ISqliteConnectionFactory` is `IAsyncDisposable`-only, and a container-tracked `IAsyncDisposable`-only singleton makes **synchronous** `ServiceProvider.Dispose()` throw `InvalidOperationException`. The sample host is torn down synchronously in tests (`BrowserWebAppFactory.Dispose()` → `IHost.Dispose()`), so tracking the factory would have regressed every E2E test that opens a conversation. 

Final approach: register only the non-disposable `INotifyWaitStore`, constructing `SqliteConnectionFactory` **inline**. The factory is reclaimed by GC at process end — a benign leak (it holds only `SemaphoreSlim`s, no OS handle). Rationale is documented inline in `Program.cs`.

## Known limitations (accepted, documented in code)

- Test-mode only today (real-provider rollout tracked in #161).
- No pruning of `notify_waits` rows for threads never reopened after a restart (acceptable for a sample app).
- The reused `SqliteSchemaInitializer` creates a few unused sibling tables in the notify db (intentional — reuses shared infra).

## Verification

- Build (sample + test project): clean, 0 warnings / 0 errors; husky `format-verify` passes.
- New tests: **2/2 pass** against real on-disk SQLite through the full `Program` host.
- `LmMultiTurn.Tests` trigger/notify/persistence: **43/43 pass**.
- `LmStreaming.Sample.Tests`: 556 passed, 1 skipped, 5 failed — **all 5 failures are pre-existing on clean `origin/main`** (unrelated ContextDiscovery/GatewayAuth tests tied to the #174 NotifyMessage merge), verified by stashing this change and re-running: the same 5 fail. This change is innocent.

## Files

| File | Action |
|------|--------|
| `samples/LmStreaming.Sample/Program.cs` | Modify — Sqlite using; `INotifyWaitStore` DI registration; resolve in pool closure; `with`-attach on `TriggerOptions` |
| `tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs` | Add — `WebApplicationFactory<Program>` restart test (2 facts) |

### DIFF
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -395,9 +395,16 @@ try
     // IAsyncDisposable-only singleton makes ServiceProvider.Dispose() (synchronous) throw
     // "only implements IAsyncDisposable". The sample host is torn down synchronously in tests
     // (BrowserWebAppFactory calls IHost.Dispose()), so tracking the factory would regress every E2E
-    // test that creates an agent. The inline factory is instead reclaimed by GC at process end — a
-    // benign leak (it only holds SemaphoreSlims, which allocate no OS handle here), which is the right
-    // trade for a sample host that must stay sync-dispose-safe.
+    // test that creates an agent. The factory OBJECT itself only holds SemaphoreSlims (no OS handle)
+    // and is reclaimed by GC at process end; the real cost of never disposing it is that the
+    // PROCESS-WIDE Microsoft.Data.Sqlite connection pool — not owned by this object — stays populated
+    // with live WAL connections (this factory sets Pooling=true + Cache=Shared and runs
+    // PRAGMA journal_mode=WAL), each an open OS file handle plus its -wal/-shm sidecars, until the
+    // process exits or SqliteConnection.ClearAllPools() is called. Acceptable here ONLY because this
+    // path is test-mode-gated and the tests clear the pool in their finally. FOLLOW-UP (#161): before
+    // flipping the real-provider gate, register real async disposal (e.g.
+    // IHostApplicationLifetime.ApplicationStopped -> SqliteConnection.ClearAllPools()) so the pooled
+    // WAL handles are released deterministically instead of lingering to process exit.
     // Configurable path (default a sibling of the conversations/ folder) mirrors
     // CodeReviewDaemon.Sample's CodeReviewDaemon:DatabasePath so a WebApplicationFactory test can
     // UseSetting an isolated file.
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -20,6 +20,7 @@ using AchieveAi.LmDotnetTools.LmCore.Middleware;
 using AchieveAi.LmDotnetTools.LmCore.Utils;
 using AchieveAi.LmDotnetTools.LmMultiTurn;
 using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence;
+using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence.Sqlite;
 using AchieveAi.LmDotnetTools.LmMultiTurn.SubAgents;
 using AchieveAi.LmDotnetTools.LmStreaming.AspNetCore.Extensions;
 using AchieveAi.LmDotnetTools.LmStreaming.Sample.Triggers;
@@ -388,6 +389,33 @@ try
     _ = builder.Services.AddSingleton<IConversationStore>(conversationStore);
     _ = builder.Services.AddSingleton<IRunLedgerStore>(conversationStore);
 
+    // #145: durable notify-wait persistence. Register ONLY the (non-disposable) INotifyWaitStore and
+    // construct its SqliteConnectionFactory INLINE — deliberately NOT as a separately-tracked
+    // singleton. ISqliteConnectionFactory is IAsyncDisposable-only; a container-tracked
+    // IAsyncDisposable-only singleton makes ServiceProvider.Dispose() (synchronous) throw
+    // "only implements IAsyncDisposable". The sample host is torn down synchronously in tests
+    // (BrowserWebAppFactory calls IHost.Dispose()), so tracking the factory would regress every E2E
+    // test that creates an agent. The inline factory is instead reclaimed by GC at process end — a
+    // benign leak (it only holds SemaphoreSlims, which allocate no OS handle here), which is the right
+    // trade for a sample host that must stay sync-dispose-safe.
+    // Configurable path (default a sibling of the conversations/ folder) mirrors
+    // CodeReviewDaemon.Sample's CodeReviewDaemon:DatabasePath so a WebApplicationFactory test can
+    // UseSetting an isolated file.
+    // NOTE (intentional): SqliteNotifyWaitStore self-initializes via the shared SqliteSchemaInitializer,
+    // which also creates currently-unused sibling tables (messages/thread_metadata/run_ledger/
+    // accepted_inputs) in this db file — accepted to reuse shared infra rather than fork a
+    // notify-only initializer.
+    // KNOWN LIMITATION (#145, accepted): notify_waits rows for threads that are never reopened after a
+    // restart are never pruned (restore/cleanup only runs when that thread's loop is reconstructed).
+    // Acceptable for a sample app; no pruning logic added.
+    var notifyWaitDbPath = builder.Configuration["LmStreaming:NotifyWaitDbPath"];
+    if (string.IsNullOrWhiteSpace(notifyWaitDbPath))
+    {
+        notifyWaitDbPath = Path.Combine(AppContext.BaseDirectory, "notify-waits.db");
+    }
+    _ = builder.Services.AddSingleton<INotifyWaitStore>(_ =>
+        new SqliteNotifyWaitStore(new SqliteConnectionFactory(notifyWaitDbPath)));
+
     // The REST status resolver (ConversationsController dependency) reads the conversation store plus
     // its run ledger. Resolve the ledger from the registered IConversationStore so a test host that
     // swaps the store (see BrowserWebAppFactory) keeps both halves pointing at the same instance —
@@ -472,6 +500,10 @@ try
         var functionRegistry = sp.GetRequiredService<FunctionRegistry>();
         var agentFactory = sp.GetRequiredService<Func<string, IStreamingAgent>>();
         var conversationStore = sp.GetRequiredService<IConversationStore>();
+        // #145: the durable notify-wait store is a process singleton (like conversationStore); captured
+        // here so the inner per-thread agent factory can attach it to TriggerOptions (see the Build(...)
+        // call site below). ThreadId is per-thread via context.ThreadId inside that factory.
+        var notifyWaitStore = sp.GetRequiredService<INotifyWaitStore>();
         var providerRegistry = sp.GetRequiredService<ProviderRegistry>();
         var codexLifetime = sp.GetRequiredService<CodexMcpServerLifetime>();
         var mockHostLifetime = sp.GetRequiredService<MockProviderHostLifetime>();
@@ -1047,9 +1079,20 @@ try
                     // inside its own ctor, so a subagent-kind source can't be handed the manager
                     // directly — it resolves it lazily once the loop (and thus the manager) exists.
                     MultiTurnAgentLoop agent = null!;
+                    // #145: attach the durable notify-wait store + this thread's id so notify-mode waits
+                    // survive a process restart (TriggerRuntime restores/reconciles them on thread
+                    // recovery). Set via a record `with` on Build's result so SampleTriggerRegistrations.Build
+                    // keeps its trigger-source-only responsibility. Test-mode/mock-provider ONLY today — same
+                    // gate as the triggerOptions pass-through at the loop ctor below; real-provider rollout is
+                    // tracked in #161. (When !isTestMode the whole triggerOptions object is discarded there,
+                    // so leaving the store null keeps it from being attached to a thrown-away options value.)
                     var triggerOptions = SampleTriggerRegistrations.Build(
                         sandboxEnabled: sandboxSession is not null,
-                        subAgentManagerAccessor: () => agent?.SubAgentManager);
+                        subAgentManagerAccessor: () => agent?.SubAgentManager) with
+                    {
+                        NotifyWaitStore = isTestMode ? notifyWaitStore : null,
+                        ThreadId = isTestMode ? threadId : null,
+                    };
 
                     agent = new MultiTurnAgentLoop(
                         providerAgent,
--- a/tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
+++ b/tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
@@ -1,0 +1,292 @@
+using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence.Sqlite;
+using AchieveAi.LmDotnetTools.LmStreaming.Sample.Triggers;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace LmStreaming.Sample.Tests.Triggers;
+
+/// <summary>
+/// Issue #145 integration test: proves the sample host's durable notify-wait wiring
+/// (<c>Program.cs</c> DI registration of <see cref="INotifyWaitStore"/> + the <c>TriggerOptions</c>
+/// <c>with</c>-attach in the pool's agent factory) actually persists notify-watcher arming records
+/// and reconciles them across a simulated restart against REAL on-disk SQLite — the one gap no
+/// existing in-memory test covers.
+/// </summary>
+/// <remarks>
+/// <para>
+/// A "restart" is simulated as: (1) seed the durable notify-wait store + thread metadata on disk
+/// directly (write side), then (2) boot the real <c>Program</c> host over that same on-disk
+/// state and open the thread. Booting the host resolves the DI-registered store and drives the pool
+/// agent factory, whose <c>RunAsync</c> auto-recovery calls
+/// <c>TriggerRuntime.RestoreNotifyWaitsAsync</c> — the wiring under test. Seeding directly (rather
+/// than driving the mock model to arm a wait) keeps the test focused on the new host wiring; the
+/// model-driven arm path is already covered at the loop level.
+/// </para>
+/// <para>
+/// Thread metadata MUST be seeded: <c>MultiTurnAgentBase.RecoverAsync</c> only invokes
+/// <c>OnThreadRecoveredAsync</c> (→ notify restore) when the thread has metadata; a metadata-less
+/// thread short-circuits and restore never runs.
+/// </para>
+/// <para>
+/// Tests are serialized (see <c>AssemblyInfo.cs</c> <c>DisableTestParallelization</c>) because the
+/// host reads the process-global <c>LM_PROVIDER_MODE</c>.
+/// </para>
+/// </remarks>
+public sealed class NotifyWaitDurableRestoreTests
+{
+    private const string ScheduleWaitId = "call_schedule_1";
+    private const string ProcessWaitId = "call_process_1";
+    private const string ProcessBarrierWaitId = "call_process_barrier";
+
+    /// <summary>
+    /// Boots the real <c>Program</c> over an isolated notify-wait db + conversation store.
+    /// Uses the default in-memory <c>TestServer</c> (we only need <see cref="WebApplicationFactory{TEntryPoint}.Services"/>,
+    /// not HTTP), so — unlike <c>BrowserWebAppFactory</c> — no Kestrel swap / cast workaround is needed.
+    /// </summary>
+    private sealed class NotifyRestoreWebAppFactory : WebApplicationFactory<Program>
+    {
+        private readonly string _notifyDbPath;
+        private readonly string _conversationsPath;
+
+        public NotifyRestoreWebAppFactory(string notifyDbPath, string conversationsPath)
+        {
+            _notifyDbPath = notifyDbPath;
+            _conversationsPath = conversationsPath;
+
+            // Program.cs reads LM_PROVIDER_MODE at the top level, before any host-builder callback.
+            // 'test' selects the scripted agent factory — the agent never needs a script here because
+            // RunAsync runs recovery (and thus notify restore) before it processes any input.
+            Environment.SetEnvironmentVariable("LM_PROVIDER_MODE", "test");
+        }
+
+        protected override void ConfigureWebHost(IWebHostBuilder builder)
+        {
+            // Production avoids the Vite dev-server auto-spawn (matches BrowserWebAppFactory / DaemonWebAppFactory).
+            builder.UseEnvironment("Production");
+
+            // Point the #145 notify-wait store at this test's isolated db (mirrors
+            // CodeReviewDaemon.Sample's CodeReviewDaemon:DatabasePath test-override via UseSetting).
+            builder.UseSetting("LmStreaming:NotifyWaitDbPath", _notifyDbPath);
+
+            builder.ConfigureTestServices(services =>
+            {
+                // Isolate the conversation store to this test's temp dir so the seeded metadata (which
+                // gates recovery) is what the host recovers. Same override seam BrowserWebAppFactory uses.
+                services.RemoveAll<IConversationStore>();
+                services.AddSingleton<IConversationStore>(new FileConversationStore(_conversationsPath));
+            });
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            base.Dispose(disposing);
+            if (disposing)
+            {
+                Environment.SetEnvironmentVariable("LM_PROVIDER_MODE", null);
+            }
+        }
+    }
+
+    [Fact]
+    public async Task ProcessNotifyWait_UnregisteredKindAfterRestart_DeliversTriggerLostAndClearsRow()
+    {
+        var (root, conversationsPath, notifyDbPath) = NewTempPaths();
+        var threadId = "notify-restore-process-" + Guid.NewGuid().ToString("N");
+        try
+        {
+            var now = NowMs();
+            // A gateway-less host builds SampleTriggerRegistrations with sandboxEnabled:false, so the
+            // 'process' kind is never registered. Restore therefore hits the unregistered-kind branch,
+            // delivers trigger_lost_on_restart, and clears the row once the terminal envelope is accepted.
+            await SeedAsync(conversationsPath, notifyDbPath, threadId,
+            [
+                new NotifyWaitRecord(
+                    WaitId: ProcessWaitId,
+                    ThreadId: threadId,
+                    Kind: ProcessTriggerSource.KindName,
+                    Args: "{\"handle\":\"proc-abc\"}",
+                    Label: null,
+                    MaxFires: null,
+                    FiresSoFar: 0,
+                    TimeoutAtUnixMs: now + 3_600_000,
+                    ArmedAtUnixMs: now,
+                    Status: "active"),
+            ]);
+
+            await using var host = new NotifyRestoreWebAppFactory(notifyDbPath, conversationsPath);
+            var notifyStore = host.Services.GetRequiredService<INotifyWaitStore>();
+            var pool = host.Services.GetRequiredService<MultiTurnAgentPool>();
+            var mode = SystemChatModes.GetById(SystemChatModes.DefaultModeId)!;
+
+            // Opening the thread runs the real agent-factory closure (the #145 wiring) and its
+            // RunAsync auto-recovery → RestoreNotifyWaitsAsync.
+            _ = pool.GetOrCreateAgent(threadId, mode, requestedProviderId: "test", requestResponseDumpFileName: null);
+
+            var active = await PollActiveUntilAsync(notifyStore, threadId, rows => rows.Count == 0);
+
+            active.Should().BeEmpty(
+                "the process kind is not registered in a gateway-less host, so restore delivers " +
+                "trigger_lost_on_restart and clears the durable row");
+        }
+        finally
+        {
+            SqliteConnection.ClearAllPools();
+            TryDeleteDir(root);
+        }
+    }
+
+    [Fact]
+    public async Task ScheduleNotifyWait_RestorableKind_SurvivesRestartAndReArms()
+    {
+        var (root, conversationsPath, notifyDbPath) = NewTempPaths();
+        var threadId = "notify-restore-schedule-" + Guid.NewGuid().ToString("N");
+        try
+        {
+            var now = NowMs();
+            await SeedAsync(conversationsPath, notifyDbPath, threadId,
+            [
+                // Restorable: the schedule kind is always registered and re-arms deterministically from
+                // (interval, armedAt). A long interval guarantees it never fires during the test, so the
+                // row stays active after a successful re-arm.
+                new NotifyWaitRecord(
+                    WaitId: ScheduleWaitId,
+                    ThreadId: threadId,
+                    Kind: ScheduleTriggerSource.KindName,
+                    Args: "{\"intervalSeconds\":86400}",
+                    Label: null,
+                    MaxFires: null,
+                    FiresSoFar: 0,
+                    TimeoutAtUnixMs: now + 3_600_000,
+                    ArmedAtUnixMs: now,
+                    Status: "active"),
+
+                // Non-restorable barrier: its deletion is the positive signal that restore ran this pass.
+                // Both rows are processed in one synchronous RestoreNotifyWaitsAsync loop, so once the
+                // barrier is gone the schedule row has also been processed — making the schedule-survival
+                // assertion race-free (rather than indistinguishable from "restore hasn't run yet").
+                new NotifyWaitRecord(
+                    WaitId: ProcessBarrierWaitId,
+                    ThreadId: threadId,
+                    Kind: ProcessTriggerSource.KindName,
+                    Args: "{\"handle\":\"proc-xyz\"}",
+                    Label: null,
+                    MaxFires: null,
+                    FiresSoFar: 0,
+                    TimeoutAtUnixMs: now + 3_600_000,
+                    ArmedAtUnixMs: now,
+                    Status: "active"),
+            ]);
+
+            await using var host = new NotifyRestoreWebAppFactory(notifyDbPath, conversationsPath);
+            var notifyStore = host.Services.GetRequiredService<INotifyWaitStore>();
+            var pool = host.Services.GetRequiredService<MultiTurnAgentPool>();
+            var mode = SystemChatModes.GetById(SystemChatModes.DefaultModeId)!;
+
+            _ = pool.GetOrCreateAgent(threadId, mode, requestedProviderId: "test", requestResponseDumpFileName: null);
+
+            // Converge to exactly {schedule}: the process barrier is gone (restore ran) AND the schedule
+            // row is still active (it re-armed rather than being lost).
+            var active = await PollActiveUntilAsync(
+                notifyStore,
+                threadId,
+                rows => rows.Count == 1 && rows[0].WaitId == ScheduleWaitId);
+
+            active.Select(r => r.WaitId).Should().BeEquivalentTo(
+                [ScheduleWaitId],
+                "the restorable schedule kind re-arms and survives the restart while the non-restorable " +
+                "process barrier is cleared");
+        }
+        finally
+        {
+            SqliteConnection.ClearAllPools();
+            TryDeleteDir(root);
+        }
+    }
+
+    // --- helpers -----------------------------------------------------------------------------
+
+    private static long NowMs() => DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+    private static (string Root, string ConversationsPath, string NotifyDbPath) NewTempPaths()
+    {
+        var root = Path.Combine(Path.GetTempPath(), "lm-notify-restore-test", Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(root);
+        return (root, Path.Combine(root, "conversations"), Path.Combine(root, "notify-waits.db"));
+    }
+
+    /// <summary>
+    /// Seeds the on-disk conversation metadata (required for recovery to fire) and the durable notify
+    /// rows directly, then releases the SQLite connection pool's file handle so the host opens the db
+    /// cleanly (needed on Windows, where the shared-cache pool otherwise pins the file).
+    /// </summary>
+    private static async Task SeedAsync(
+        string conversationsPath,
+        string notifyDbPath,
+        string threadId,
+        IReadOnlyList<NotifyWaitRecord> rows)
+    {
+        var conversationStore = new FileConversationStore(conversationsPath);
+        await conversationStore.SaveMetadataAsync(
+            threadId,
+            new ThreadMetadata { ThreadId = threadId, LastUpdated = NowMs(), LatestRunId = "run_seed" });
+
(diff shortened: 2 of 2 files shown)


gautam-achieveai:
[Revobot]

Reviewed PR #178 across 2 changed files (grounded against 8 files in the checkout) — 0 Must, 0 Should, 2 Consider.

This PR wires the durable notify-wait store (#145) into `LmStreaming.Sample`: it registers `INotifyWaitStore` in DI, captures it in the agent factory, and attaches it (plus the per-thread id) to `TriggerOptions` via a record `with`, so notify-mode waits survive a process restart. It adds one real-on-disk-SQLite integration test proving the restart/reconcile path. The change is test-mode-only today (#161 tracks real-provider rollout).

**Correctness / contracts — clean.** I verified the load-bearing claims rather than trusting the (extensive) comments:
- The inline `SqliteConnectionFactory` is genuinely untracked (it's passed to `new SqliteNotifyWaitStore(...)`, never registered), and `SqliteNotifyWaitStore` implements only `INotifyWaitStore` — it is **not** `IDisposable`/`IAsyncDisposable`. So `ServiceProvider.Dispose()` (synchronous) will not throw on it, and the E2E-regression rationale in the comment holds. `SqliteConnectionFactory` is indeed `IAsyncDisposable`-only, matching the stated hazard. (`Program.cs:392-417`, `SqliteNotifyWaitStore.cs:9`, `SqliteConnectionFactory.cs:110`)
- The `isTestMode` gate is internally consistent: the store is attached only when `isTestMode` (`Program.cs:1093-1094`) and the whole `triggerOptions` is discarded when `!isTestMode` (`Program.cs:1133`), so there's no dangling attach to a thrown-away value.
- SQL is fully parameterized (no injection); the schema is reused via the shared initializer (no wire/DB-compat impact — new db file only); the test's premises check out (process kind is unregistered in a gateway-less host per `SampleTriggerRegistrations.cs:64`; `RestoreNotifyWaitsAsync` exists; parallelization is disabled per `AssemblyInfo.cs`). No EUII, secrets, or prompt-injection in the diff.

**Maintainability / tests — two minor notes (non-blocking):**
- *Consider* — `NotifyWaitDurableRestoreTests.cs` (factory ctor / `Dispose`): the factory sets the process-global `LM_PROVIDER_MODE=test` in its constructor and clears it to `null` on dispose. This is safe today **only** because `DisableTestParallelization=true`, which the doc-comment correctly calls out. Consider save/restore of the prior value instead of hard-clearing to `null`, so a caller that already had it set isn't wiped — a small robustness hedge if parallelization is ever re-enabled.
- *Consider* — `Program.cs:392-417`: the ~20-line DI rationale block dwarfs the 2-line registration. Every claim is accurate, and for a sample app the documentation is arguably a feature; flagging for awareness only.

The test itself is well-constructed: the non-restorable "barrier" row makes the schedule-survival assertion race-free rather than indistinguishable from "restore hasn't run yet," and cleanup uses `SqliteConnection.ClearAllPools()` for Windows file-handle safety.

Verdict: **Approve with comments.**

gautam-achieveai:
[Gautam's review bot]

**Deep review** — full diff analysis, build validation (0 warnings/0 errors), type/interface verification against source.

**Verdict: Approve.**

Reviewed 2 changed files (+336/-1) grounded against TriggerOptions, SqliteNotifyWaitStore, SqliteConnectionFactory, INotifyWaitStore, MultiTurnAgentPool, and the test project's global usings / AssemblyInfo.cs.

**Correctness — clean.** Independent verification confirms:
- The SqliteConnectionFactory is genuinely untracked in DI (only passes through the INotifyWaitStore factory lambda) — the sync-dispose hazard rationale holds.
- The isTestMode gate is self-consistent: lines 1093-1094 set NotifyWaitStore/ThreadId only in test mode, and when !isTestMode the entire 	riggerOptions is discarded at the loop ctor (line ~1133), so no orphan state.
- TriggerOptions is a sealed record with init-only properties — the with expression is the correct mutation pattern and preserves immutability.
- Test seeding → recovery path is valid: FileConversationStore.SaveMetadataAsync creates the metadata gate that MultiTurnAgentBase.RecoverAsync checks before invoking OnThreadRecoveredAsync → RestoreNotifyWaitsAsync.

**Test design — solid.** The non-restorable "barrier" row technique eliminates the race between "restore hasn't run yet" and "schedule survived" — once the barrier is gone, both rows have been processed in the same synchronous loop. SqliteConnection.ClearAllPools() handles Windows file-handle pinning correctly.

**Minor observations (non-blocking):**
1. The test's ConfigureTestServices replaces IConversationStore but not IRunLedgerStore — the old ledger instance still points at AppContext.BaseDirectory/conversations. This is harmless for the current test scope (recovery only reads IConversationStore metadata), but if future tests in this fixture assert on run ledger state, they'll hit the wrong store. Consider aligning both overrides if the fixture grows.
2. I concur with Revobot's env-variable save/restore note — storing + restoring the prior LM_PROVIDER_MODE value (instead of clearing to 
ull) would make the fixture robust against future test-ordering changes.

No additional issues found. The implementation cleanly closes #145 for the sample host, with appropriate scope-gating (#161) and documented limitations.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -388,6 +389,33 @@ HttpClient GatewayHttpClient(TimeSpan? timeout = null)
...
     _ = builder.Services.AddSingleton<IRunLedgerStore>(conversationStore);
 
+    // #145: durable notify-wait persistence. Register ONLY the (non-disposable) INotifyWaitStore and
+    // construct its SqliteConnectionFactory INLINE — deliberately NOT as a separately-tracked
+    // singleton. ISqliteConnectionFactory is IAsyncDisposable-only; a container-tracked
+    // IAsyncDisposable-only singleton makes ServiceProvider.Dispose() (synchronous) throw
+    // "only implements IAsyncDisposable". The sample host is torn down synchronously in tests
+    // (BrowserWebAppFactory calls IHost.Dispose()), so tracking the factory would regress every E2E
+    // test that creates an agent. The inline factory is instead reclaimed by GC at process end — a
+    // benign leak (it only holds SemaphoreSlims, which allocate no OS handle here), which is the right
```
gautam-achieveai: [Claude's bot] **MEDIUM** (Code Health / Documentation Accuracy)

This justification is **factually inaccurate about what leaks**, and the inaccuracy sits right on the `#161` seam. The factory *object* does only hold `SemaphoreSlim`s — but the connection string sets `Pooling=true` + `Cache=Shared` and the first connection runs `PRAGMA journal_mode=WAL`. Once `GetConnectionAsync` runs even once, live connections are retained by **Microsoft.Data.Sqlite's process-wide pool** (not owned by the factory) — each an open OS file handle plus `-wal`/`-shm` — until `SqliteConnection.ClearAllPools()` or process exit. The new test's own `ClearAllPools()`-in-`finally`, whose remark says it's *"needed on Windows, where the shared-cache pool otherwise pins the file"*, is direct proof those handles exist.

There is **no shipped runtime defect today** (production is gated off via `NotifyWaitStore = isTestMode ? … : null`), so this is not blocking — but the comment will actively mislead whoever does the `#161` real-provider rollout, who inherits a singleton that is never disposed *and* never has its pool cleared.

**Suggestion:** Don't claim "no OS handle." Reword to: *never disposing the factory leaks nothing through the factory itself, but it leaves the process-wide connection pool populated with live WAL connections that hold OS file handles until process exit or `SqliteConnection.ClearAllPools()`. Acceptable here only because this path is test-mode-gated and the tests clear the pool.* Before `#161` flips the gate, register real async disposal (e.g. `IHostApplicationLifetime.ApplicationStopped` → `SqliteConnection.ClearAllPools()`).

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
```diff
@@ -0,0 +1,292 @@
...
+            await SeedAsync(conversationsPath, notifyDbPath, threadId,
+            [
+                // Restorable: the schedule kind is always registered and re-arms deterministically from
+                // (interval, armedAt). A long interval guarantees it never fires during the test, so the
+                // row stays active after a successful re-arm.
+                new NotifyWaitRecord(
+                    WaitId: ScheduleWaitId,
+                    ThreadId: threadId,
+                    Kind: ScheduleTriggerSource.KindName,
+                    Args: "{\"intervalSeconds\":86400}",
```
gautam-achieveai, line 159: [Claude's bot] **MEDIUM** (Test Coverage)

The restorable `schedule` fixture uses a 24h interval (`intervalSeconds:86400`), so after restore it re-arms but **never fires within the test window**. That means the single most valuable durable property — *a restored wait actually fires again and persists an updated `fires_so_far`* (re-arm mainline `TriggerRuntime.RestoreNotifyWaitsAsync` → `ArmCoreAsync`; persistence in `OnSourceFiredAsync`) — is not verified end-to-end. The litmus otherwise passes (both tests fail without the wiring), so this is a coverage gap on already-proven wiring, not a correctness defect.

**Suggestion:** Add a third `[Fact]` with a short interval (e.g. `intervalSeconds:1`) that asserts the re-armed wait fires post-restore and the persisted `fires_so_far` advances (or the row terminates at `maxFires`). ~40 LOC reusing the existing helpers.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
```diff
@@ -0,0 +1,292 @@
...
+                    Label: null,
+                    MaxFires: null,
```
gautam-achieveai, line 161: [Claude's bot] **MEDIUM** (Test Coverage)

Both seeded rows use `MaxFires: null`, so the **`FiresSoFar >= MaxFires` → delete-without-re-arm** reconcile branch (`TriggerRuntime.RestoreNotifyWaitsAsync`, approx. lines 502–506) is never exercised. A crash between persist-and-delete could reach this branch in production.

**Suggestion:** Seed a `schedule` row with `MaxFires:1, FiresSoFar:1` and assert the row is cleared with no envelope delivered. ~25 LOC.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
```diff
@@ -0,0 +1,292 @@
...
+                    FiresSoFar: 0,
+                    TimeoutAtUnixMs: now + 3_600_000,
```
gautam-achieveai, line 163: [Claude's bot] **MEDIUM** (Test Coverage)

Both rows use a far-future `TimeoutAtUnixMs (now + 3_600_000)`, so the **`deadline <= now` → `timed_out` terminal + cleanup** branch (`TriggerRuntime.RestoreNotifyWaitsAsync`, approx. lines 512–528) is never exercised — yet "a restored wait whose ceiling elapsed while offline" is a core **AC2** reconcile decision named in the approved plan for #145.

**Suggestion:** Seed a restorable row with `TimeoutAtUnixMs = now - 1000` and assert a `timed_out` envelope is delivered and the row is cleaned up. ~20 LOC.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -388,6 +389,33 @@ HttpClient GatewayHttpClient(TimeSpan? timeout = null)
...
+    // notify-only initializer.
+    // KNOWN LIMITATION (#145, accepted): notify_waits rows for threads that are never reopened after a
+    // restart are never pruned (restore/cleanup only runs when that thread's loop is reconstructed).
+    // Acceptable for a sample app; no pruning logic added.
+    var notifyWaitDbPath = builder.Configuration["LmStreaming:NotifyWaitDbPath"];
+    if (string.IsNullOrWhiteSpace(notifyWaitDbPath))
+    {
+        notifyWaitDbPath = Path.Combine(AppContext.BaseDirectory, "notify-waits.db");
+    }
+    _ = builder.Services.AddSingleton<INotifyWaitStore>(_ =>
```
gautam-achieveai, line 423: [Claude's bot] **LOW** (Exception Handling / Robustness)

The **restore-time** read (`RestoreNotifyWaitsAsync` → `store.LoadActiveAsync`) is not wrapped, so a corrupt or inaccessible `NotifyWaitDbPath` would throw out of `OnThreadRecoveredAsync` during recovery. Note the arm/save/fires paths **are** already best-effort try/caught, so this is the one restore-time IO edge left undefended. Low blast radius for a test-mode-gated sample host, but the PR newly makes the db path *configurable*, which widens the chance of a bad value.

**Suggestion:** Wrap the restore-time load (or the `OnThreadRecoveredAsync` restore step) in a logged best-effort try/catch mirroring the arm-path pattern, and add a negative test that points the path at an un-creatable location to lock in the intended behavior (fail-loud vs. degrade).

samples/LmStreaming.Sample/Program.cs
```diff
@@ -388,6 +389,33 @@ HttpClient GatewayHttpClient(TimeSpan? timeout = null)
...
+    // benign leak (it only holds SemaphoreSlims, which allocate no OS handle here), which is the right
+    // trade for a sample host that must stay sync-dispose-safe.
+    // Configurable path (default a sibling of the conversations/ folder) mirrors
+    // CodeReviewDaemon.Sample's CodeReviewDaemon:DatabasePath so a WebApplicationFactory test can
+    // UseSetting an isolated file.
+    // NOTE (intentional): SqliteNotifyWaitStore self-initializes via the shared SqliteSchemaInitializer,
+    // which also creates currently-unused sibling tables (messages/thread_metadata/run_ledger/
+    // accepted_inputs) in this db file — accepted to reuse shared infra rather than fork a
+    // notify-only initializer.
+    // KNOWN LIMITATION (#145, accepted): notify_waits rows for threads that are never reopened after a
```
gautam-achieveai, line 415: [Claude's bot] **LOW** (Maintainability — tracked follow-up)

Acknowledging the documented limitation: `notify_waits` rows for threads that are never reopened after a restart are never pruned (cleanup only runs when that thread's loop is reconstructed), and there is no `timeout_at < now` sweep. Correctly accepted for a sample host.

**Suggestion:** Track for `#161` — before real-provider rollout, add a startup sweep of expired rows (`DELETE … WHERE timeout_at < now`) or a bounded retention policy so this doesn't become unbounded growth in a real multi-tenant host.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -1047,9 +1079,20 @@ HttpClient GatewayHttpClient(TimeSpan? timeout = null)
...
+                    // survive a process restart (TriggerRuntime restores/reconciles them on thread
+                    // recovery). Set via a record `with` on Build's result so SampleTriggerRegistrations.Build
+                    // keeps its trigger-source-only responsibility. Test-mode/mock-provider ONLY today — same
+                    // gate as the triggerOptions pass-through at the loop ctor below; real-provider rollout is
+                    // tracked in #161. (When !isTestMode the whole triggerOptions object is discarded there,
+                    // so leaving the store null keeps it from being attached to a thrown-away options value.)
                     var triggerOptions = SampleTriggerRegistrations.Build(
                         sandboxEnabled: sandboxSession is not null,
-                        subAgentManagerAccessor: () => agent?.SubAgentManager);
+                        subAgentManagerAccessor: () => agent?.SubAgentManager) with
```
gautam-achieveai, line 1098: [Claude's bot] **LOW** (Design — informational)

Appending `with { NotifyWaitStore = …, ThreadId = … }` to `Build(...)`'s result (rather than threading the two values through `Build`) is a reasonable SRP trade — it keeps `SampleTriggerRegistrations.Build` trigger-source-only and `TriggerOptions` is an immutable sealed record so the `with` allocates a fresh, thread-safe instance. The mild smell is that `TriggerOptions` is now effectively two-phase-constructed (Build yields a partial value the caller must remember to complete), which is easy to miss at a future second call site.

**Suggestion:** No change required. Optionally document on `TriggerOptions` that `NotifyWaitStore`/`ThreadId` are host-attached after `Build`, so a future call site doesn't forget the `with`.

gautam-achieveai:
[Claude's bot] **LOW** (Performance — pre-existing, out of scope for this PR)

Not in this diff, but flagging as context because **this PR is the first to place `SqliteConnectionFactory` on the per-thread recovery path**: its `PooledConnection` wrapper (`src/LmMultiTurn/Persistence/Sqlite/SqliteConnectionFactory.cs`, approx. lines 143–147 — `connection.Close(); connection.Dispose(); Open();`) performs **two physical opens plus a synchronous `Open()`** on an otherwise-async path per `GetConnectionAsync`. During recovery, each `LoadActiveAsync`/`DeleteAsync`/`ArmCoreAsync` pays that cost. With pooling it's mostly pool churn, not fresh file opens, so impact is modest.

**Suggestion:** Do **not** fix in this PR (it's unrelated pre-existing code). File a tracking issue against the factory to return the already-open connection wrapped in a disposer that releases the semaphore, avoiding the close/re-`Open()` dance and the sync open.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
```diff
@@ -0,0 +1,292 @@
...
+            var pool = host.Services.GetRequiredService<MultiTurnAgentPool>();
+            var mode = SystemChatModes.GetById(SystemChatModes.DefaultModeId)!;
+
+            // Opening the thread runs the real agent-factory closure (the #145 wiring) and its
+            // RunAsync auto-recovery → RestoreNotifyWaitsAsync.
+            _ = pool.GetOrCreateAgent(threadId, mode, requestedProviderId: "test", requestResponseDumpFileName: null);
+
+            var active = await PollActiveUntilAsync(notifyStore, threadId, rows => rows.Count == 0);
+
+            active.Should().BeEmpty(
```
gautam-achieveai, line 131: [Claude's bot] [QUESTION] **Clarification Needed**

**Code:** `active.Should().BeEmpty(...)` — asserting the non-restorable `process` row (and the barrier row in the other test) is **deleted**.

**Question:** Deletion of a non-restorable row only happens if `TryDeliverRestoreNotifyAsync` returns `true`; when the bounded input channel is **full**, `RestoreNotifyWaitsAsync` intentionally **retains** the row for redelivery on the next recovery. In a freshly-booted test host the channel should be empty (restore runs before the run loop reads input), so delivery should always succeed here — but is that guaranteed, or could a slow/full channel at restore time flip these delete-based assertions to flaky?

**Why this matters:** If the retention path is reachable during the test, asserting on *row-absence* would be intermittently wrong; asserting on *delivery-accepted* would be the more robust signal.

**Possible answers:** (a) The channel is guaranteed empty at restore (restore precedes any input read) → assertions are safe as written; (b) there is a backpressure path that could retain the row → consider asserting delivery rather than only row-absence.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -388,6 +389,33 @@ HttpClient GatewayHttpClient(TimeSpan? timeout = null)
...
 
+    // #145: durable notify-wait persistence. Register ONLY the (non-disposable) INotifyWaitStore and
+    // construct its SqliteConnectionFactory INLINE — deliberately NOT as a separately-tracked
+    // singleton. ISqliteConnectionFactory is IAsyncDisposable-only; a container-tracked
+    // IAsyncDisposable-only singleton makes ServiceProvider.Dispose() (synchronous) throw
+    // "only implements IAsyncDisposable". The sample host is torn down synchronously in tests
+    // (BrowserWebAppFactory calls IHost.Dispose()), so tracking the factory would regress every E2E
+    // test that creates an agent. The inline factory is instead reclaimed by GC at process end — a
+    // benign leak (it only holds SemaphoreSlims, which allocate no OS handle here), which is the right
+    // trade for a sample host that must stay sync-dispose-safe.
```
gautam-achieveai: [Claude's bot] [QUESTION] **Clarification Needed**

**Code:** the inline, intentionally-undisposed `new SqliteConnectionFactory(notifyWaitDbPath)` in the `INotifyWaitStore` singleton registration.

**Question:** When `#161` flips the real-provider gate on, this same singleton factory will be resolved by the synchronously-disposed E2E host (`BrowserWebAppFactory` → `IHost.Dispose()`). Nothing on that path calls `SqliteConnection.ClearAllPools()` and the factory is intentionally never disposed — will that leave WAL file handles pinned (Windows) after each E2E test?

**Why this matters:** This is the concrete downstream consequence of the leak-characterization issue flagged on the registration comment above. Deciding it now (rather than rediscovering it as E2E flakiness during #161) is cheap.

**Possible answers:** (a) `BrowserWebAppFactory` already clears pools / disposes on teardown → fine; (b) it does not, and #161 should add pool clearing or a disposable wrapper as part of enabling the real-provider path.

gautam-achieveai:
[Claude's bot]
# PR Review: feat(#145): wire durable notify-watcher restore into LmStreaming.Sample

**Verdict: ✅ APPROVE WITH COMMENTS** — 0 blockers · 4 MEDIUM · 4 LOW · 2 questions

This PR closes the sample-host wiring gap for #145: it registers a standalone SQLite-backed `INotifyWaitStore` in `LmStreaming.Sample`, captures it in the pool's agent factory, and attaches it + the per-thread `ThreadId` to `TriggerOptions` via a record `with` (test-mode gated), so notify-mode waits survive a process restart. It adds one real-on-disk-SQLite `WebApplicationFactory<Program>` integration test covering the restore/reconcile path. The core durability primitive already exists from #158 — this is purely the wiring + a test.

I reviewed independently (not rubber-stamping the two prior bot approvals) across code alignment, plan alignment, architecture, performance/resource lifecycle, test coverage, exception handling, and security — reading the actual `SqliteConnectionFactory` / `SqliteNotifyWaitStore` / `TriggerRuntime` sources at head, not just the diff.

## What's solid ✅
- **Plan alignment is faithful.** All approved steps (A1–A5) are implemented; the PR touches exactly 2 files with no drive-by refactors or unrequested scope.
- **The documented DI deviation is correct.** `ISqliteConnectionFactory` is `IAsyncDisposable`-only, so registering it as a tracked singleton *would* make synchronous `ServiceProvider.Dispose()` throw. Registering only the non-disposable `INotifyWaitStore` (verified: implements only `INotifyWaitStore`) genuinely avoids — not relocates — the throw.
- **The test is well-built and load-bearing.** Litmus passes: remove the wiring and both facts fail. Not over-mocked (real host, real on-disk SQLite, real `TriggerRuntime` restore). The non-restorable "barrier row" is a genuinely race-free signal that restore completed.
- **Concurrency & immutability are sound.** No cross-thread `ThreadId` leak (`threadId` is a fresh lambda local; `TriggerOptions` is an immutable sealed record). Singleton store is stateless-per-op with thread-scoped queries and PK `(thread_id, wait_id)`.
- **Security clean.** Parameterized SQL only; the configurable db path is operator-trust config; no hardcoded secrets. No temp/debug code.
- I also **verified the "beside `conversations/`" comment is accurate** (both sit directly under `AppContext.BaseDirectory`) — a suspected inaccuracy that did *not* pan out.

## Findings

### MEDIUM
| # | File:Line | Issue |
|---|-----------|-------|
| G1 | `Program.cs:399` | The **"benign leak / no OS handle" comment is factually wrong about what leaks** — pooled WAL connections (process-wide pool) hold OS file handles until `ClearAllPools()`/process exit; the test's own `ClearAllPools()`-on-Windows remark proves it. No shipped defect (prod is gated off), but it actively misleads the #161 rollout. **Fix the comment.** |
| G2 | `…RestoreTests.cs:159` | Restorable `schedule` fixture uses `intervalSeconds:86400`, so the re-armed wait **never fires** — the headline durable property (fires again + persists `fires_so_far`) is untested. |
| G3 | `…RestoreTests.cs:161` | `MaxFires:null` on both rows → the `FiresSoFar >= MaxFires` delete branch is uncovered. |
| G4 | `…RestoreTests.cs:163` | Far-future timeout on both rows → the `deadline <= now` → `timed_out` + cleanup branch (a named **AC2** decision) is uncovered. |

### LOW
| # | File:Line | Issue |
|---|-----------|-------|
| G5 | `Program.cs:416` | Restore-time `LoadActiveAsync` is unguarded against a bad/inaccessible db path (arm/save paths already are). |
| G6 | `Program.cs:408` | Documented unbounded `notify_waits` growth — track a sweep/retention policy for #161. |
| G7 | (general) | Pre-existing `PooledConnection` double-open + sync `Open()`, newly on the recovery hot path — file a tracking issue, don't fix here. |
| G8 | `Program.cs:1091` | `with`-at-call-site leaves `TriggerOptions` mildly two-phase-constructed — informational. |

### Questions (non-blocking)
| # | File:Line | Question |
|---|-----------|----------|
| Q1 | `…RestoreTests.cs:131` | Could a full/slow input channel at restore retain the non-restorable row and make the delete-based assertions flaky? |
| Q2 | `Program.cs:400` | When #161 flips the gate, will the sync-disposed E2E host leave WAL handles pinned since the factory is never disposed / pool never cleared? |

## Verdict rationale
Nothing blocks. **G1** is the only above-LOW item with real weight, but it's a documentation-accuracy problem with **zero shipped runtime defect** (production is verifiably gated off via `isTestMode`), so it doesn't meet the Critical/High bar, and one MEDIUM isn't a systemic pattern. The G2–G4 cluster are coverage gaps on *already-proven* wiring, and G6–G7 are pre-existing/accepted-and-tracked. Net: the PR is a clean, well-scoped, well-tested improvement to the codebase.

**Recommended before merge:** fix **G1**'s comment (it misdirects #161) and add **G4**'s elapsed-TTL fixture (a named AC2 path). G2/G3 and the two questions are worthwhile fast-follows.

This aligns with the two prior bot approvals (`[Revobot]`, `[Gautam's review bot]`) while adding the G1 correction they both missed.

---
<sub>🤖 Independent review by Claude. Findings posted as inline comments above. This review did not approve or merge the PR — that's left to a human.</sub>

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: Reworded the comment (Program.cs ~395-406) to drop the inaccurate benign-leak / SemaphoreSlim-only characterization. It now states that never disposing the factory leaves the process-wide Microsoft.Data.Sqlite connection pool (Pooling=true + Cache=Shared, PRAGMA journal_mode=WAL) populated with live WAL connections — each an open OS file handle plus its -wal/-shm sidecars — until process exit or SqliteConnection.ClearAllPools(); acceptable only because this path is test-mode-gated and the tests clear the pool in their finally, with a forward note that #161 should register deterministic disposal (IHostApplicationLifetime.ApplicationStopped -> ClearAllPools()) before the real-provider gate is flipped. Pushed in a1a43811.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: Added ScheduleNotifyWait_ReArmsAndFiresAfterRestart_PersistsAdvancedFireCount — seeds a capped (MaxFires:2) schedule row with intervalSeconds:1, boots the host, and polls until the restored wait is active with fires_so_far >= 1. That proves the row re-armed, fired on the wall-clock timer, and persisted the advanced count (a capped wait persists every fire; only uncapped waits debounce). Pushed in a1a43811.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: Added ScheduleNotifyWait_ExhaustedFireBudgetAfterRestart_ClearsRowWithoutReArming — seeds a schedule row with MaxFires:1 / FiresSoFar:1 and a far-future TTL plus a non-restorable barrier, then asserts the thread converges to empty: the exhausted row is deleted without re-arming (no double-fire), while the barrier's deletion proves restore ran this pass. Pushed in a1a43811.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: Added ScheduleNotifyWait_TtlElapsedWhileOffline_TimesOutAndClearsRow — seeds a restorable schedule row whose deadline already elapsed (TimeoutAtUnixMs in the past) plus a non-restorable barrier, then asserts convergence to empty: restore resolves the elapsed row as timed_out and clears it (contrasting the survive-test's live-TTL row that re-arms), with the barrier deletion proving restore ran. Pushed in a1a43811.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Won't Fix: Deferred to #161 for follow-up. The restore-time load hardening (fail-loud vs degrade on a bad/corrupt NotifyWaitDbPath) is a core TriggerRuntime.RestoreNotifyWaitsAsync behavioral decision affecting every notify-restore consumer, out of scope for this test-mode-gated sample-wiring PR.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Won't Fix: Deferred to #161 for follow-up. Accepted limitation for a sample host (documented in code); a startup expired-row sweep / retention policy belongs with the real-provider rollout.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] By Design: The with-attach keeps SampleTriggerRegistrations.Build trigger-source-only, and TriggerOptions is an immutable sealed record so the with allocates a fresh thread-safe instance. The host-attached NotifyWaitStore/ThreadId step is documented at the call site (Program.cs ~1082-1088). No change required.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
gautam-achieveai: [gautam-achieveai's bot] (a) The channel is guaranteed empty at restore. RestoreNotifyWaitsAsync runs synchronously inside RunAsync auto-recovery BEFORE the run loop reads any input, and the bounded input channel comfortably fits the <=2 restore envelopes this test delivers, so TryDeliverRestoreNotifyAsync always succeeds and the non-restorable row is deleted. The retention path is only reachable under a full channel, which cannot occur before any input has been read — so the row-absence assertion is safe and non-flaky.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] (b) BrowserWebAppFactory.Dispose does not call SqliteConnection.ClearAllPools() nor dispose the factory. Test-mode ('test'/'test-anthropic') already resolves INotifyWaitStore today, but because the notify db lives at a fixed shared path (AppContext.BaseDirectory/notify-waits.db) rather than the per-test temp dir that Dispose deletes, the pinned WAL handle does not currently break teardown. When #161 enables the real-provider path it should add pool clearing (IHostApplicationLifetime.ApplicationStopped -> SqliteConnection.ClearAllPools(), or a disposable factory wrapper). Tracked in #161.

gautam-achieveai:
[Claude's bot]
## Re-Review Summary: PR #178

**Verdict: INCOMPLETE**

GitHub authentication succeeded and PR metadata, changed-file scope, linked issue, existing review history, and CI status were retrieved. The PR changes `samples/LmStreaming.Sample/Program.cs` and adds `tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs`; its two reported GitHub checks are passing.

The required isolated worktree/scratchpad inspection and source-level specialist review could not be completed because the sandbox gateway became unhealthy: its container no longer has the required `sandbox` user account, preventing subsequent filesystem and shell operations. The required `test-coverage-review` specialist was also unavailable in the installed skill catalog.

No verified findings or context questions are being posted. No approval or merge action was taken.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
```diff
@@ -0,0 +1,475 @@
...
+                //
+                // Signal chosen: fires_so_far >= 1 while the row is still active. It is observable during
+                // the ~1s window between the first fire (fires_so_far -> 1, row stays active) and the
+                // second/final fire (which tears the row down), comfortably inside the ~10s poll budget.
+                // The alternative "row terminates at maxFires" would also work but proves strictly less.
+                new NotifyWaitRecord(
+                    WaitId: ScheduleWaitId,
+                    ThreadId: threadId,
+                    Kind: ScheduleTriggerSource.KindName,
+                    Args: "{\"intervalSeconds\":1}",
```
gautam-achieveai, line 362: [Claude's bot] **HIGH** (Test Reliability — CI flake)

The newly added `ScheduleNotifyWait_ReArmsAndFiresAfterRestart_PersistsAdvancedFireCount` is the one test in this suite that depends on a **real wall-clock timer**. With `intervalSeconds:1` + `MaxFires:2`, it polls for `FiresSoFar >= 1` inside a bounded budget — but the observable "row active with `fires_so_far == 1`" state is a **transient ~1s window** between the 1st fire (row stays active) and the 2nd/final fire (which tears the row down). On a loaded/slow CI agent, host boot + recovery latency + GC pauses can compress or overshoot that window, making this intermittently fail even though the wiring is correct. Flaky tests tax **every** future PR that runs this shared, parallelization-disabled suite, so the blast radius exceeds this PR.

The fix is cheap and fully deterministic:

**Preferred** — assert the *monotonic terminal* outcome instead of the transient mid-state. With `MaxFires:2`, poll until the row is **cleared** (budget exhausted). Row-deletion is latched — it can never be "missed" by overshooting:
```csharp
// seed MaxFires: 2, intervalSeconds: 1
var active = await PollActiveUntilAsync(notifyStore, threadId, rows => rows.Count == 0);
active.Should().BeEmpty("both scheduled fires elapse post-restart and the exhausted row is cleared without re-arming");
```

**Alternatively** keep proving a live re-arm but make `FiresSoFar >= 1` a *latched* state: raise `MaxFires` (e.g. `10`) so the row stays active long after the 1st fire, so later polls still observe `>= 1`.

Either removes the race while keeping the "fired on the wall-clock timer after restore" proof. (The sibling `SurvivesRestartAndReArms` test is fine — its `intervalSeconds:86400` + synchronous barrier make it race-free; only this new fact carries the timing risk.)

gautam-achieveai:
[Claude's bot]
## Re-Review Summary: PR #178

**Verdict: APPROVE WITH COMMENTS** (supersedes the prior INCOMPLETE re-review — GitHub auth is now working and the full review completed).

Re-reviewed the current head (`a1a43811`) against the earlier `[Claude's bot]` review round. The author has addressed the prior findings — the `Program.cs` DI comment was reworded to correctly describe the WAL-pool handle retention (dropping the inaccurate "benign SemaphoreSlim-only leak" wording), and the requested edge-case tests (exhausted-budget, TTL-elapsed, re-arms-and-fires) were added. Those prior items are resolved; I did **not** re-raise them, and I explicitly withdraw any "test scope creep" concern — those three facts were added at the prior review's request and are legitimate distinct restore branches.

The production change remains a clean, minimal, test-mode-gated wiring fix (register a non-disposable `INotifyWaitStore`, attach it + `ThreadId` to `TriggerOptions` via a record `with`). The DI decision to construct the `IAsyncDisposable`-only `SqliteConnectionFactory` inline rather than as a container-tracked singleton is sound for this synchronously-disposed sample host, and the #161 follow-up (call `SqliteConnection.ClearAllPools()` on `ApplicationStopped`) is correctly noted — verified in source that factory `DisposeAsync()` alone drains only its semaphores and does **not** clear the pool, so `ClearAllPools()` is the load-bearing call for #161.

### New finding this pass (1)

| Severity | File | Line | Issue |
|---|---|---|---|
| **HIGH** (non-blocking) | `tests/.../NotifyWaitDurableRestoreTests.cs` | 358 | The newly added `ScheduleNotifyWait_ReArmsAndFiresAfterRestart_PersistsAdvancedFireCount` asserts a **transient** mid-state (`fires_so_far == 1`, active) that exists for only ~1s between the 1st and 2nd (terminal) fire of a `MaxFires:2` / `intervalSeconds:1` timer. On a loaded CI agent, boot/GC latency can overshoot that window → intermittent failure. Fix is cheap + deterministic: assert the monotonic terminal outcome (row **cleared** after budget exhausted) or raise `MaxFires` so `>= 1` is latched. Detail in the inline comment. |

This is escalated to HIGH (not a merge blocker) because a wall-clock-race test taxes every future PR in this shared, parallelization-disabled suite — but the deterministic fix is trivial. If the flake cannot be made deterministic or quarantined, this crosses into REQUEST CHANGES.

### Non-blocking nits (optional, may defer/decline)
- **LOW** — `NotifyWaitStore` / `ThreadId` are a coupled pair gated by the same `isTestMode` flag but set independently; nothing structurally enforces they move together in a future edit. Cosmetic — a single tuple assignment would make the invariant explicit.

**Recommendation:** address the HIGH test-flake item (or quarantine it), then this is good to merge. No blockers; the functional fix and its coverage are sound.

tests/LmStreaming.Sample.Tests/Triggers/NotifyWaitDurableRestoreTests.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: Raised `MaxFires` from 2 to 100 in `ScheduleNotifyWait_ReArmsAndFiresAfterRestart_PersistsAdvancedFireCount`. The row now stays active for ~100s (far beyond the ~10s poll budget), so `fires_so_far >= 1 while active` is a latched, monotonic state rather than a transient ~1s window — a slow/loaded CI agent can no longer overshoot it into a torn-down row. Kept the live re-arm+fire proof (deliberately not the poll-until-cleared variant, which would duplicate the exhausted/timed-out facts). Verified deterministic across repeated local runs. Fixed in e167331a.

Issue #187, opened 2026-07-12
Add typed .NET SDK for sandbox gateway management


gautam-achieveai:
## Summary

Add a lightweight `net8.0;net9.0` package wrapping `SandboxedOstoolsMcpServer`. Applications can:

- Discover marketplaces, plugins, skills, agents, and session-discovered items.
- Create, get, list, and delete sandboxes.
- Execute non-interactive argv-shaped commands through the gateway Bash tool in POSIX-capable sandboxes.
- Read, write, and list UTF-8 files.
- Operate entirely through remote gateway APIs without host filesystem access.

## Value

Review bots and other applications can create isolated workspaces, clone repositories, run setup commands, manage files, perform work, and clean up through one supported SDK.

## Acceptance Criteria

- [ ] Client accepts `serverAddress`, `appId`, and base64 `clientSecret`.
- [ ] Every app-facing REST/MCP request sends `X-Sbx-App-Id` and `X-Sbx-App-Key`; `/health` does not.
- [ ] Secrets are excluded from logs, URLs, exceptions, diagnostics, and `ToString()`.
- [ ] Empty `401` responses map to an authorization error; foreign and missing sessions both remain `404`.
- [ ] Expose typed create/get/list/delete operations with explicit deletion—disposing the client never deletes sandboxes.
- [ ] Create supports logical workspace identifiers and marketplace aliases without resolving or creating host paths.
- [ ] Expose marketplace preview plus `ListDiscoveredAsync(sessionId)` over the existing session-discovery REST endpoint; V1 does not add proactive discovery or crawling.
- [ ] Move existing lifecycle/catalog protocol logic from `LmAgentInfra` and samples into the package rather than duplicating it.
- [ ] Command API accepts non-empty argv and an optional workspace-relative working directory.
- [ ] V1 command execution explicitly requires a gateway Bash/POSIX-capable sandbox. Arguments are safely POSIX-quoted; the API does not claim native cross-platform argv semantics.
- [ ] Results expose `ExitCode` and `CombinedOutput`.
- [ ] Commands automatically bypass the gateway’s 20 KB/500-line `exec` truncation by capturing output to a collision-resistant, command-scoped workspace path.
- [ ] A command-scoped completion manifest records exit code, output length, and integrity metadata. Missing command responses recover from that manifest without SDK-level reruns of side-effecting commands.
[...]

PR #186, opened 2026-07-10
Sub-agent characteristics factory + modelintelligence/effort front matter

gautam-achieveai:
## Summary
- Add typed `modelintelligence` and `effort` front matter with shared validation diagnostics.
- Resolve model characteristics once per spawn while preserving the existing agent factory fallback.
- Route Copilot sub-agents by configured tier and shape reasoning effort for the selected transport.
- Preserve characteristics through mid-session discovery and provider recreation; keep the daemon parse-only.

## Key Decisions
- The characteristics factory is additive, so existing `AgentFactory` callers remain compatible.
- Explicit model and spawn overrides win; unresolved tiers inherit the parent provider.
- Provider-specific effort clamping and wire shapes remain in `GithubCopilotProvider`.
- Session refresh publishes immutable template snapshots to prevent mixed-provider routing. `Templates` is a point-in-time snapshot; callers must reread it to observe later registrations or rebinds.
- New record properties preserve legacy constructors, deconstruction, equality, and hash behavior.

## Testing
- Solution build: 0 warnings, 0 errors.
- `LmMultiTurn.Tests`, `GithubCopilotProvider.Tests`, and `CodeReviewDaemon.Sample.Tests` pass on clean reruns.
- New parser, mapper, resolver, shaper, composition, session, and daemon parity regressions pass.
- `LmStreaming.Sample.Tests` retains pre-existing `origin/main` failures in context notification ordering and gateway auth-header expectations.

## Related Issues
Fixes #177

PR #201, opened 2026-07-13
feat(#198): route sandbox-discovered directory context to the opening sub-agent

### LINKED ISSUE #198: When SubAgent opens a file in a directory, context is loaded in primary agent.
## Summary
When a **sub-agent** opens a file in a sub-directory, the directory-level context file (`CLAUDE.md` / `AGENTS.md`) discovered there is injected into the **primary/parent** conversation instead of the sub-agent that opened it. Route that discovered directory context to the **sub-agent that triggered it**, while preserving today's behavior for boot/root context and for any discovery that arrives without a triggering-agent identifier.

This is the **app-side** (LmDotnetTools) half of the fix and is independently shippable and verifiable via synthetic webhooks. It is a follow-up to #78 (which built the context injector).

## Background
Directory-context discovery is gateway-webhook driven and keyed **only by `session_id`**. The sandbox gateway watches the workspace filesystem and POSTs a batch envelope to `ContextDiscoveryController.NotifyAsync`; `ContextDiscoveryInjector.InjectAsync` resolves targets via `SandboxSessionRegistry.GetThreads(sessionId)` and fans the file out to **every registered top-level conversation thread**, delivering it as a user turn. Sub-agents (`SubAgentState.AgentId`, thread `subagent-{agentId}`) are never registered as targets and their sandbox file ops reuse the parent's MCP client (frozen `X-Session-ID` = primary session), so a sub-agent's file-open is indistinguishable from the primary's. Result: directory context opened by a sub-agent lands on the primary agent.

## Value
- **Persona:** an operator running a sub-agent that navigates into a sub-directory carrying its own rules.
- The sub-agent doing the work receives that directory's rules (correct attribution), and the primary conversation is no longer polluted with context for directories it never entered — a cleaner primary context window and fewer irrelevant instructions in the parent.

## Scope

**In scope (app-side; verifiable now via synthetic webhooks):**
1. **Contract** — add an optional `agent_id` to the inbound discovery contract (`ContextDiscoveryItem` / `ContextDiscoveryPayload`) and carry it through the controller projection. Backward compatible: absent ⇒ null, unknown-field tolerant.
2. **Sub-agent resolution** — resolve `agent_id` to a live sub-agent by **reusing the existing chain** (`GetThreads(sessionId)` → `MultiTurnAgentPool.TryGet(threadId)` → `MultiTurnAgentLoop.SubAgentManager`), not a new parallel registry duplicating `_sessionThreads`. Resolution is scoped within the envelope's `session_id` (so cross-conversation isolation is structural). If an explicit `agent_id → sub-agent` map is needed, colocate it with the existing per-conversation `_subAgentBindings` and reuse its teardown hooks (`UnregisterThreadFromAllSessions` / session destroy / dispose); it must reflect per-agent lifecycle (registered at spawn, removed on run completion **and** disposal), not only manager disposal.
3. **Injector routing** — extend `ContextDiscoveryInjector.InjectAsync` with a routing branch: when a `context_file` item carries an `agent_id` that resolves to a **currently-Running** sub-agent, deliver **only** to that sub-agent via a new lifecycle-safe `SubAgentManager` entry point (e.g. `TryDeliverToRunningAsync`) that gates on `Status == Running`, reuses the manager's running-branch send, and **never** calls `RestartRunAsync` / bypasses via `state.Agent.SendAsync`.
4. **Graceful fallback** — absent `agent_id` ⇒ the existing session fan-out, byte-identical to today (guard-only change); present-but-unresolved/stale ⇒ fall back to session fan-out, log at debug, never error.
5. **Dedup** — extend `SandboxSessionRegistry.TryMarkDiscoverySeen` so the key is `(target, kind, path)` where `target = agent_id ?? "__session__"`. The sentinel keeps the fallback path's per-session dedup behavior identical to today.
6. **Observability** — the injector logs the resolved target (`primary` vs `subagent-{id}`) and the routed / fallback / dropped outcome per `context_file` discovery (log line; no new diagnostics schema).

**Out of scope (dependencies / follow-ups):**
- Gateway stamping the triggering-agent id onto the discovery webhook + the per-request attribution it needs (cf. #187).
- App-side emission of a per-sub-agent discriminator on sub-agent sandbox file ops (needs a per-call MCP header seam that does not exist today — sub-agents reuse the parent's frozen MCP client).
- Nested sub-agents (sub-agents spawning sub-agents) — resolution covers direct sub-agents of a conversation only.
- A user-visible "Context loaded" pill for sub-agent-scoped context (see Decisions).
- Retry/redelivery of a discovery that arrives before/after the sub-agent is live (see Decisions — drop, don't retry).

## Acceptance Criteria
- [ ] A discovery webhook **without** `agent_id` deserializes and behaves byte-identically to today (regression test over the existing fan-out path).
- [ ] A synthetic `context_file` webhook carrying an `agent_id` that resolves to a Running sub-agent injects the formatted context as a user turn into that **sub-agent's** conversation store, and the **primary's** store does **not** receive it.
- [ ] The same webhook when the target sub-agent is **not** currently Running (completed / disposed / never-registered) ⇒ the context is **dropped**, the sub-agent is **not** restarted, and no spurious completion is relayed to the parent; the drop is logged.
- [ ] A present-but-unresolved/stale `agent_id` ⇒ falls back to the session fan-out (delivered to the primary thread(s)), logs at debug, and never throws.
- [ ] Routed delivery goes through a `SubAgentManager` entry point that no-ops when the target is not Running (no `RestartRunAsync`, no `state.Agent.SendAsync` bypass).
- [ ] Dedup is per `(target, kind, path)`: the primary and a sub-agent (or two sub-agents) each entering the **same** directory each receive that directory's context; the fallback path (`target = "__session__"`) dedups identically to today.
- [ ] The injector logs the resolved target and the routed / fallback / dropped outcome for each `context_file` discovery.

## Decisions
- **Route to sub-agent only:** a discovery resolving to a sub-agent that is no longer live is **dropped**, not redirected to the primary — keeps the primary clean; observable via the injector log.
- **Drop, don't retry:** delivery timing stays mark-seen-before-deliver (uniform with today); a discovery that arrives before the sub-agent is registered or after it finishes is dropped rather than retried. Retry-until-live can be revisited in the live-path PR against real gateway redelivery timing.
- Sub-agent-scoped context is delivered into the sub-agent's own (non-user-facing) conversation, so it will **not** surface as a "Context loaded" pill in the primary view. Acceptable (it's the sub-agent's context); flag if a user-visible indicator is desired.
- Until the gateway echoes `agent_id`, the routed path is dormant-but-correct; the fallback path (all real traffic today) is unchanged.

## Dependencies / sequencing
- The lifecycle-safe delivery must be safe against concurrent sub-agent disposal. The `SubAgentState` disposal lease from #177/#186 is not on the base branch as of writing — either add minimal synchronization in the new entry point or sequence this after that lease lands. See #163 for the related concurrency hazards.

## Related
#78 (built the injector) · #76 / #77 (sub-agent discovery/activation) · #187 (gateway SDK) · #163 / #177 (SubAgentManager lifecycle & terminal-disposal races the routing must respect).

gautam-achieveai:
## Summary

When a **sub-agent** opens a file in a sub-directory, the directory-level context file (`CLAUDE.md`/`AGENTS.md`) discovered there was injected into the **primary/parent** conversation instead of the sub-agent that opened it — because discovery is keyed only by `session_id` and the injector fanned out to registered top-level threads only. This routes such a discovery to the **sub-agent that opened the file**, while preserving today's behavior for boot/root context and for any discovery without a triggering-agent id.

This is the **app-side** half. It is gated behind a **default-off** flag and is dormant in production until the sandbox gateway starts stamping `agent_id` (tracked separately as #187); it is fully proven now via synthetic-webhook and instruction-based tests.

Fixes #198.

## How it works

- Optional `agent_id` rides the discovery webhook (`ContextDiscoveryItem` → per-item projection → `ContextDiscoveryPayload`), backward compatible.
- A new role interface **`ISubAgentContextSink`** (LmMultiTurn, implemented by `MultiTurnAgentLoop`) lets `ContextDiscoveryInjector` reach a running sub-agent via `is`-check — no concrete cast, not on `IMultiTurnAgent`.
- Delivery goes through a new side-effect-free **`SubAgentState.TryBeginInjectLease()`** + `SubAgentManager.TryDeliverToRunningAsync` returning a tri-state `{Delivered, TargetNotDeliverable, NotOwned}`; it never restarts a finished sub-agent and never mutates the sub-agent's parent-relay state.
- The injector aggregates across the session's sinks and routes/drops/falls-back accordingly; dedup key becomes `(target, kind, path)`; a per-outcome diagnostics counter (routed/dropped/fallback) is the operator signal (routed context renders no "Context loaded" pill).

## Key Decisions

1. **Deliver via a purpose-built inject-only lease, not `BeginContinuation`.** `BeginContinuation` mutates `_notifyParentOnCompletion` and claims the restart lock as its first act; a context delivery through it would silently drop/double the sub-agent's parent relay and race the Wait/trigger state machine. `TryBeginInjectLease()` admits only `Running && !terminating`, touches neither.
2. **Drop-clean-when-done (owner decision).** A context delivery to a finishing/finished sub-agent is dropped — it must never cause a spurious extra run, a second parent relay, or a run against a disposed provider.
3. **Sub-agent only; present-but-unresolved drops (does not fall back to primary).** Because completed sub-agents linger in the manager, "unresolved" means the pre-registration race or a nested sub-agent — falling back to the primary there would re-introduce the exact #198 pollution. Only a `null`/blank `agent_id` fans out to the primary.
4. **Retry-friendly dedup.** An all-`NotOwned` result is dropped **without** marking-seen, so the gateway's redelivery heals the pre-registration race; `Delivered`/`TargetNotDeliverable` mark under `agent_id`; fallback marks under the `__session__` sentinel. The flag gates routing **and** the dedup-target predicate together, so flag-off is byte-identical to today.
5. **Blank-`agent_id` normalization** (`!IsNullOrWhiteSpace`) so a blank id can't create a rogue `""` dedup bucket.

## Testing

- **LmMultiTurn.Tests: 473/473.** Lease grant/refuse (+ notify/restart untouched), `Delivered` with the context text on the sub-agent's next turn (parent untouched), completed ⇒ `TargetNotDeliverable` with no restart and no second parent relay, unknown ⇒ `NotOwned`, teardown-race ⇒ not resurrected.
- **LmStreaming.Sample.Tests: 656 passed** (+1 skipped). Contract round-trip + mixed-batch per-item projection, registry dedup (two sub-agents, back-compat, per-sub-agent eviction), full injector routing/aggregation suite (flag-off-with-`agent_id`-present byte-identical, present-unresolved-drop, blank-id, not-marked-on-`NotOwned` retry), real-loop cast, and a deterministic routed-path HTTP integration test that drives HTTP → injector → real `MultiTurnAgentLoop` → `SubAgentManager` → a live `ctx-probe` sub-agent and asserts routed + primary-untouched.
- A `PromptExamples.md` **webhook-routing fixture** (backgrounded sub-agent parked on a long `Wait`) documents the equivalent instruction-chain flow for manual/browser testing.
- `dotnet format whitespace` clean.

Two failing tests were confirmed **pre-existing and unrelated** (not touched by this change): the `CodeReviewDaemon.Sample.Tests` build break (stale `DiscoveredSubAgentTemplateBuilder.Build(...)` call after the #186 merge) and `GatewayAppAuthWiringTests.Catalog_client_sends_no_bearer_headers_when_key_is_unset`.

## Known limitation

The completion-boundary guard closes the common case (once terminal disposal begins, delivery is refused). One tiny residual window remains — between the loop publishing the final `RunCompletedMessage` and the monitor calling `BeginTerminalDisposalAsync` (status still `Running`, `_terminating` not yet set) — where a delivery could still wake one spurious run. Fully closing it needs tagging context injections inside `MultiTurnAgentLoop.RunLoopAsync` (an invasive loop change) and is deferred; the routed path is dormant (flag-off) until #187, so this is not reachable in production today.

## Dependencies / out of scope

- **#187** — the gateway must stamp the triggering `agent_id` on the discovery webhook (and leave it null on the session-level root file). Until then the routed path is dormant.
- Out of scope: nested sub-agent routing (documented drop), a per-sub-agent MCP file-op header, a user-visible pill for sub-agent context.

### DIFF
--- a/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
+++ b/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
@@ -198,13 +198,21 @@ public sealed class ContextDiscoveryInjector
     /// <summary>
     /// Routes the discovery to the sub-agent identified by <c>agent_id</c>. Marks the dedup ledger
     /// under the agent id FIRST (so a redelivery of an already-resolved discovery drops); then asks
-    /// each session thread's <see cref="ISubAgentContextSink"/> in turn. Aggregation: the first
-    /// <see cref="SubAgentContextDeliveryResult.Delivered"/> wins (routed); otherwise if any sink
-    /// owns the id but can't take it (<see cref="SubAgentContextDeliveryResult.TargetNotDeliverable"/>)
-    /// the discovery is dropped WITHOUT fan-out and the mark kept (terminal); if no sink owns it
-    /// (all <see cref="SubAgentContextDeliveryResult.NotOwned"/>) the discovery is dropped and the
-    /// mark REMOVED so a gateway redelivery can retry once the sub-agent registers (pre-registration
-    /// race). A present-but-unresolved <c>agent_id</c> never falls back to the primary.
+    /// each session thread's <see cref="ISubAgentContextSink"/> in turn. Aggregation across the
+    /// non-delivered outcomes: the first <see cref="SubAgentContextDeliveryResult.Delivered"/> wins
+    /// (routed); otherwise, in priority order —
+    /// <list type="number">
+    ///   <item>if any sink owns the id but can't take it
+    ///   (<see cref="SubAgentContextDeliveryResult.TargetNotDeliverable"/>) the discovery is dropped
+    ///   WITHOUT fan-out and the mark KEPT (terminal — a retry can't succeed);</item>
+    ///   <item>else if any sink THREW, the mark is likewise KEPT: a throwing sink may already have
+    ///   accepted (delivered) the send, so un-marking could DOUBLE-INJECT the same context on a
+    ///   gateway retry — safer to drop-and-hold;</item>
+    ///   <item>else (every sink cleanly returned <see cref="SubAgentContextDeliveryResult.NotOwned"/>)
+    ///   the discovery is dropped and the mark REMOVED so a gateway redelivery can retry once the
+    ///   sub-agent registers (pre-registration race).</item>
+    /// </list>
+    /// A present-but-unresolved <c>agent_id</c> never falls back to the primary.
     /// </summary>
     private async Task<int> RouteToSubAgentAsync(ContextDiscoveryPayload body, CancellationToken ct)
     {
@@ -225,6 +233,7 @@ public sealed class ContextDiscoveryInjector
         IReadOnlyList<IMessage> messages = [message];
 
         var anyNotDeliverable = false;
+        var anyAmbiguous = false;
         foreach (var threadId in _registry.GetThreads(body.SessionId!))
         {
             if (!_pool.TryGet(threadId, out var agent) || agent is not ISubAgentContextSink sink)
@@ -243,13 +252,18 @@ public sealed class ContextDiscoveryInjector
             }
             catch (Exception ex)
             {
-                // A misbehaving sink must not strand routing — treat it as not-owned and keep scanning.
+                // A misbehaving sink must not strand routing — keep scanning the other threads so a
+                // healthy sink can still Deliver. But flag the failure as AMBIGUOUS: because the throw
+                // could have happened AFTER the sink accepted (delivered) the send, we can no longer be
+                // sure this discovery wasn't delivered. The final aggregation uses this to KEEP the mark
+                // rather than un-marking, so a gateway retry can't double-inject.
                 _logger.LogWarning(
                     ex,
                     "ContextDiscovery context_file: sink delivery threw for agent {AgentId} on thread {ThreadId}; "
-                    + "treating as not-owned.",
+                    + "treating as ambiguous (delivery may have occurred).",
                     agentId,
                     threadId);
+                anyAmbiguous = true;
                 continue;
             }
 
@@ -286,10 +300,28 @@ public sealed class ContextDiscoveryInjector
             return 0;
         }
 
-        // No sink owns the id: the sub-agent hasn't registered yet (pre-registration race), or it is a
-        // nested / cross-session agent. Drop and un-mark ONLY this discovery's (target, kind, path) so the
-        // gateway's redelivery can retry once the sub-agent is live — precise so a concurrent delivery's
-        // mark for another path of the same agent survives — but still never fall back to the primary.
+        if (anyAmbiguous)
+        {
+            // A sink THREW after being asked to deliver, and no other sink Delivered. We cannot tell
+            // whether the throwing sink already accepted (delivered) the send, so KEEP the mark rather
+            // than un-marking: un-marking would let a gateway redelivery retry and could DOUBLE-INJECT
+            // the same context if that sink had in fact delivered. Drop-and-hold (no retry) is the safe
+            // choice — still never fall back to the primary.
+            _logger.LogWarning(
+                "ContextDiscovery context_file: ambiguous sink failure for {AgentId} on {Path} in session "
+                + "{SessionId}; keeping the dedup mark (no gateway retry) to avoid a possible double-inject.",
+                agentId,
+                body.Path,
+                body.SessionId);
+            _diagnostics.RecordRoutingOutcome(ContextRoutingOutcome.Dropped);
+            return 0;
+        }
+
+        // No sink owns the id AND none threw: the sub-agent hasn't registered yet (pre-registration
+        // race), or it is a nested / cross-session agent. Drop and un-mark ONLY this discovery's
+        // (target, kind, path) so the gateway's redelivery can retry once the sub-agent is live —
+        // precise so a concurrent delivery's mark for another path of the same agent survives — but
+        // still never fall back to the primary.
         _registry.UnmarkDiscoverySeen(body.SessionId!, agentId, body.Kind!, body.Path!);
         _logger.LogDebug(
             "ContextDiscovery context_file: no live sub-agent owns {AgentId} for {Path} in session {SessionId}; "
--- a/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
+++ b/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
@@ -69,6 +69,19 @@ public sealed class ContextDiscoveryInjector
     }
 
     /// <summary>
+    /// Back-compat constructor matching the pre-#198 signature: routing defaults OFF (behaviour
+    /// identical to pre-#198) for callers that do not supply options/diagnostics.
+    /// </summary>
+    public ContextDiscoveryInjector(
+        SandboxSessionRegistry registry,
+        MultiTurnAgentPool pool,
+        ContextDiscoveryFormatter formatter,
+        ILogger<ContextDiscoveryInjector> logger)
+        : this(registry, pool, formatter, new ContextDiscoveryOptions(), new ContextDiscoveryDiagnostics(), logger)
+    {
+    }
+
+    /// <summary>
     /// Best-effort: delivers the formatted file body to its target — the opening sub-agent (routed)
     /// or every live thread of the session (fallback). Returns the number of conversations the
     /// message was actually delivered to (0 when nothing is live, the discovery is a duplicate, the
@@ -274,9 +287,10 @@ public sealed class ContextDiscoveryInjector
         }
 
         // No sink owns the id: the sub-agent hasn't registered yet (pre-registration race), or it is a
-        // nested / cross-session agent. Drop and UN-mark so the gateway's redelivery can retry once the
-        // sub-agent is live — but still never fall back to the primary.
-        _registry.EvictDiscoverySeenForTarget(body.SessionId!, agentId);
+        // nested / cross-session agent. Drop and un-mark ONLY this discovery's (target, kind, path) so the
+        // gateway's redelivery can retry once the sub-agent is live — precise so a concurrent delivery's
+        // mark for another path of the same agent survives — but still never fall back to the primary.
+        _registry.UnmarkDiscoverySeen(body.SessionId!, agentId, body.Kind!, body.Path!);
         _logger.LogDebug(
             "ContextDiscovery context_file: no live sub-agent owns {AgentId} for {Path} in session {SessionId}; "
             + "dropping and allowing gateway retry.",
--- a/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
+++ b/src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
@@ -1,15 +1,17 @@
 using AchieveAi.LmDotnetTools.LmCore.Messages;
 using AchieveAi.LmDotnetTools.LmAgentInfra.Agents;
 using AchieveAi.LmDotnetTools.LmAgentInfra.Sandbox;
+using AchieveAi.LmDotnetTools.LmMultiTurn;
 
 namespace AchieveAi.LmDotnetTools.LmAgentInfra.Context;
 
 /// <summary>
-/// Routes a <c>context_file</c> discovery from the gateway webhook into every live agent thread
-/// bound to the same sandbox session: looks up the thread set, dedups gateway retries, and
-/// enqueues a <see cref="NotifyMessage"/> (kind <c>context-discovery</c>) carrying the formatted
-/// file body — delivered to the model via <c>ICanGetText</c> and rendered as a distinct pill by the
-/// chat UI.
+/// Routes a <c>context_file</c> discovery from the gateway webhook to its target conversation(s):
+/// either the specific sub-agent that opened the file (when routing is enabled and the gateway
+/// stamped an <c>agent_id</c>) or — the fallback, and all traffic today — every live agent thread
+/// bound to the same sandbox session. It dedups gateway retries and enqueues a
+/// <see cref="NotifyMessage"/> (kind <c>context-discovery</c>) carrying the formatted file body —
+/// delivered to the model via <c>ICanGetText</c> and rendered as a distinct pill by the chat UI.
 /// </summary>
 /// <remarks>
 /// <para>
@@ -19,9 +21,18 @@ namespace AchieveAi.LmDotnetTools.LmAgentInfra.Context;
 /// switch in flight cannot translate into a webhook retry storm.
 /// </para>
 /// <para>
-/// Dedup is per <c>(sessionId, kind, path)</c>: redelivery of the same discovery is dropped for
-/// the session as a whole, so multi-threaded sessions only inject the file once (on first
-/// delivery, fanned out across every registered thread).
+/// Dedup is per <c>(target, kind, path)</c> where <c>target</c> is the resolved sub-agent
+/// <c>agent_id</c> for a routed delivery, or <see cref="SandboxSessionRegistry.SessionDiscoveryTarget"/>
+/// for the session fan-out: the primary and a sub-agent (or two sub-agents) entering the same
+/// directory each receive that directory's context once.
+/// </para>
+/// <para>
+/// Routing (gated by <see cref="ContextDiscoveryOptions.RouteToOpeningSubAgent"/>) is deliberately
+/// sub-agent-only: a discovery whose <c>agent_id</c> resolves to a sub-agent that is no longer
+/// running is DROPPED, and one that resolves to no owner at all (the pre-registration race, or a
+/// nested/cross-session sub-agent) is also dropped — NEVER redirected to the primary, which would
+/// re-introduce the exact context pollution this feature fixes. Only a null/blank <c>agent_id</c>
+/// (or the flag being off) fans out to the primary.
 /// </para>
 /// </remarks>
 public sealed class ContextDiscoveryInjector
@@ -37,24 +48,31 @@ public sealed class ContextDiscoveryInjector
     private readonly SandboxSessionRegistry _registry;
     private readonly MultiTurnAgentPool _pool;
     private readonly ContextDiscoveryFormatter _formatter;
+    private readonly ContextDiscoveryOptions _options;
+    private readonly ContextDiscoveryDiagnostics _diagnostics;
     private readonly ILogger<ContextDiscoveryInjector> _logger;
 
     public ContextDiscoveryInjector(
         SandboxSessionRegistry registry,
         MultiTurnAgentPool pool,
         ContextDiscoveryFormatter formatter,
+        ContextDiscoveryOptions options,
+        ContextDiscoveryDiagnostics diagnostics,
         ILogger<ContextDiscoveryInjector> logger)
     {
         _registry = registry ?? throw new ArgumentNullException(nameof(registry));
         _pool = pool ?? throw new ArgumentNullException(nameof(pool));
         _formatter = formatter ?? throw new ArgumentNullException(nameof(formatter));
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _diagnostics = diagnostics ?? throw new ArgumentNullException(nameof(diagnostics));
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
     }
 
     /// <summary>
-    /// Best-effort: enqueues the formatted file body into every agent thread that the session is
-    /// currently routing. Returns the number of threads the message was actually sent to (0 when
-    /// no thread is live, the discovery is a duplicate, or the session is unknown).
+    /// Best-effort: delivers the formatted file body to its target — the opening sub-agent (routed)
+    /// or every live thread of the session (fallback). Returns the number of conversations the
+    /// message was actually delivered to (0 when nothing is live, the discovery is a duplicate, the
+    /// routed target isn't running, or the session is unknown).
     /// </summary>
     public async Task<int> InjectAsync(ContextDiscoveryPayload body, CancellationToken ct)
     {
@@ -78,7 +96,28 @@ public sealed class ContextDiscoveryInjector
             return 0;
         }
 
-        if (!_registry.TryMarkDiscoverySeen(body.SessionId, body.Kind!, body.Path!))
+        // Route only when the flag is on AND the gateway stamped a NON-BLANK agent_id. Using
+        // IsNullOrWhiteSpace (not just != null) is load-bearing: a blank id would otherwise create a
+        // rogue "" dedup bucket distinct from __session__ and double-inject into the primary.
+        var routeEnabled = _options.RouteToOpeningSubAgent && !string.IsNullOrWhiteSpace(body.AgentId);
+
+        return routeEnabled
+            ? await RouteToSubAgentAsync(body, ct).ConfigureAwait(false)
+            : await FanOutToSessionAsync(body, ct).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Today's behavior (and every real delivery until the gateway stamps <c>agent_id</c>): dedup
+    /// under <see cref="SandboxSessionRegistry.SessionDiscoveryTarget"/>, then fan the file out to
+    /// every live thread the session is routing.
+    /// </summary>
+    private async Task<int> FanOutToSessionAsync(ContextDiscoveryPayload body, CancellationToken ct)
+    {
+        if (!_registry.TryMarkDiscoverySeen(
+                body.SessionId!,
+                SandboxSessionRegistry.SessionDiscoveryTarget,
+                body.Kind!,
+                body.Path!))
         {
             _logger.LogDebug(
                 "ContextDiscovery context_file: duplicate delivery for {Path} in session {SessionId}; dropping.",
@@ -87,7 +126,9 @@ public sealed class ContextDiscoveryInjector
             return 0;
         }
 
-        var threads = _registry.GetThreads(body.SessionId);
+        _diagnostics.RecordRoutingOutcome(ContextRoutingOutcome.Fallback);
+
+        var threads = _registry.GetThreads(body.SessionId!);
         if (threads.Count == 0)
         {
             // No live agent thread for this session yet. Discovery arrived between session
@@ -101,8 +142,7 @@ public sealed class ContextDiscoveryInjector
             return 0;
         }
 
-        var truncated = body.Truncated ?? false;
-        var text = _formatter.BuildInjectedMessage(body.Path!, body.Content!, truncated);
+        var message = BuildMessage(body);
 
         var injected = 0;
         foreach (var threadId in threads)
@@ -116,13 +156,6 @@ public sealed class ContextDiscoveryInjector
 
             try
             {
-                // Emit a typed notification (kind=context-discovery). The formatted file body stays in
-                // Detail so it still reaches the LLM (via ICanGetText), and the UI renders it as a pill.
-                var message = NotifyMessage.Create(
-                    NotifyKinds.ContextDiscovery,
-                    detail: text,
-                    label: body.Path);
-
                 _ = await agent.SendAsync([message], inputId: null, parentRunId: null, ct).ConfigureAwait(false);
                 injected++;
             }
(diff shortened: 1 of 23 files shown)


gautam-achieveai:
[Revobot]

Reviewed PR 201 across 22 files — 5 Must, 4 Should, 2 Consider.

### Must — Preserve the public `TryMarkDiscoverySeen` overload

`src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1188`

The existing public `TryMarkDiscoverySeen(string sessionId, string kind, string path)` was replaced by a four-parameter signature. This breaks source consumers and can cause `MissingMethodException` for compiled consumers of the packaged library.

Retain the three-parameter overload and delegate to the new implementation using `SessionDiscoveryTarget`.

### Must — Preserve the public `ContextDiscoveryInjector` constructor

`src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:55-61`

The existing public constructor was replaced with one requiring `ContextDiscoveryOptions` and `ContextDiscoveryDiagnostics`. Direct callers, explicit DI factories, and compiled consumers can break.

Keep a compatibility constructor or introduce the new dependencies without removing the existing signature.

### Must — Prevent concurrent session teardown from deleting active deduplication state

`src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1143-1145`

`UnregisterThreadFromAllSessions` checks `set.IsEmpty` and then removes `_discoverySeen`, but `RegisterThread` can concurrently add a thread after the emptiness check. An active session can therefore lose its deduplication ledger and inject the same context repeatedly.

Synchronize the membership and ledger transition, or conditionally remove the ledger only if the same session set is still current and empty.

### Must — Drain context-delivery leases for borrowed providers

`src/LmMultiTurn/SubAgents/SubAgentManager.cs:455-472`  
`src/LmMultiTurn/SubAgents/SubAgentState.cs:282-290`

`TryDeliverToRunningAsync` takes an inject lease for every sub-agent, but terminal disposal drains and cancels leases only when `OwnedProviderAgent` is non-null. For a borrowed provider, the loop can become terminal while a context `SendAsync` remains in flight, allowing a send to race terminal handling or target a completed loop.

Drain/cancel active inject leases independently of whether the provider is owned; keep provider disposal conditional on ownership.

### Must — Bound or evict routed deduplication entries

`src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1199-1202`

Every distinct `(agentId, kind, path)` creates a retained ledger entry for the lifetime of the session. The new target-eviction method is used for unresolved routing, but no verified production sub-agent teardown path removes entries for completed routed agents. Long-lived sessions can therefore accumulate unbounded per-agent/path state.

Add reliable teardown eviction, bounded retention, or TTL cleanup, and cover the lifecycle path with a test.

### Should — Avoid duplicating the test-agent implementation

`tests/LmStreaming.Sample.Tests/TestDoubles/RecordingSinkAgent.cs:14-143`

This duplicates most of `RecordingMultiTurnAgent.cs`. The two implementations can drift whenever `IMultiTurnAgent` changes.

Share the common recording/lifecycle implementation through a base class or composition, leaving `RecordingSinkAgent` responsible only for sink-specific behavior.

### Should — Extract shared registry test infrastructure

`tests/LmStreaming.Sample.Tests/Services/ContextDiscoveryInjectorRoutingTests.cs:365-391`

The registry factory and delegating stub handler duplicate helpers in neighboring context-discovery tests. Extract shared test helpers so gateway setup and constructor changes remain synchronized.

### Should — Reduce sensitive content in Information-level logs

`samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs:116-123`  
`src/LmMultiTurn/SubAgents/SubAgentManager.cs:195-197, 378-380, 708-710`

The changes log workspace-supplied descriptions, paths, session IDs, sub-agent tasks, and continuation prompts. These values can contain user content, usernames, proprietary text, or credentials. Truncation limits volume but does not redact sensitive data.

Prefer logging operation type, lengths, and pseudonymized identifiers. If content logging is required for diagnostics, protect it behind an explicit local-only setting with redaction.

### Should — Restrict the diagnostics endpoint

`samples/LmStreaming.Sample/Controllers/ContextDiscoveryDiagnosticsController.cs:21-50`

The endpoint exposes active and historical session IDs, discovered filesystem paths, and the configured webhook URL. It is mapped with the general controllers, and no operator-specific authorization is shown here.

Restrict it to an authenticated operator policy or development-only hosting, and minimize or pseudonymize session/path fields.

### Consider — Avoid full paths and session IDs in routing logs

`src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:122-125, 175-180, 202-207, 245-249, 266-271, 280-285`

Routing logs repeatedly persist complete discovered paths and session IDs. Consider logging a safe filename/category and a pseudonymized session identifier instead.

### Consider — Sanitize error details returned by `Peek`

`src/LmMultiTurn/SubAgents/SubAgentManager.cs:796-807`

`Peek` exposes exception-derived error text through `SendToParentError` and related fields. Provider or tool exceptions can contain request details or user-supplied values.

Consider returning a stable error code and sanitized summary while retaining full exception details only in protected logs.

**Verdict: Request changes**

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1153,23 +1174,64 @@ public IReadOnlyList<string> GetThreads(string sessionId)
...
+    /// storms re-firing the same discovery event.
     /// </summary>
-    public bool TryMarkDiscoverySeen(string sessionId, string kind, string path)
+    /// <remarks>
+    /// <paramref name="target"/> discriminates WHO the discovery was delivered to: a sub-agent's
+    /// <c>agent_id</c> for a routed delivery, or <see cref="SessionDiscoveryTarget"/> for the session
+    /// fan-out / root-file seed. Keying on it lets the primary and a sub-agent (or two sub-agents)
+    /// entering the same directory each receive that directory's context exactly once.
+    /// </remarks>
+    public bool TryMarkDiscoverySeen(string sessionId, string target, string kind, string path)
```
gautam-achieveai, line 1182: [Revobot] [BLOCKER] **HIGH** (API Compatibility)

This replaces the public three-parameter `TryMarkDiscoverySeen(sessionId, kind, path)` API instead of extending it. Existing consumers of the packaged library can fail to compile after upgrading (and previously compiled consumers can fail at runtime).

**Suggestion:** Restore the original overload and forward it to the target-aware overload with `SessionDiscoveryTarget`; retain this four-parameter API for new callers.

src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
```diff
@@ -37,24 +48,31 @@ public sealed class ContextDiscoveryInjector
     private readonly SandboxSessionRegistry _registry;
     private readonly MultiTurnAgentPool _pool;
     private readonly ContextDiscoveryFormatter _formatter;
+    private readonly ContextDiscoveryOptions _options;
+    private readonly ContextDiscoveryDiagnostics _diagnostics;
     private readonly ILogger<ContextDiscoveryInjector> _logger;
 
     public ContextDiscoveryInjector(
```
gautam-achieveai, line 55: [Revobot] [BLOCKER] **HIGH** (API Compatibility)

Adding options and diagnostics removed the former public constructor that accepted registry, pool, formatter, and logger. External consumers constructing this packaged public type directly will no longer compile.

**Suggestion:** Restore the previous signature as a delegating compatibility constructor with default-off options and suitable diagnostics, while retaining the expanded constructor as the primary DI path.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1120,9 +1128,22 @@ public void UnregisterThreadFromAllSessions(string threadId)
...
+            {
+                continue;
+            }
+
+            // Once a session has lost its last live thread, no primary conversation OR sub-agent
+            // remains that a dedup mark could protect, so drop the whole session's dedup ledger to
+            // bound its per-target growth (blind-spot #5). A later conversation on the same session
+            // legitimately re-receives that directory's context. Guarded by IsEmpty so a session
+            // that still routes another live conversation keeps its (shared + sub-agent) marks.
+            if (set.IsEmpty)
```
gautam-achieveai: [Revobot] [BLOCKER] **HIGH** (Concurrency / Correctness)

The `set.IsEmpty` check and discovery-ledger removal are independent. A concurrent `RegisterThread` can add a live thread after the empty check but before removal, deleting dedup state for an active session and allowing duplicate context delivery.

**Suggestion:** Coordinate membership transition and ledger eviction atomically per session, revalidating that the same membership set is current and empty. Add a deterministic concurrent last-unregister/register regression test.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1153,23 +1174,64 @@ public IReadOnlyList<string> GetThreads(string sessionId)
...
-        if (string.IsNullOrWhiteSpace(sessionId) || string.IsNullOrWhiteSpace(kind) || string.IsNullOrWhiteSpace(path))
+        if (string.IsNullOrWhiteSpace(sessionId)
+            || string.IsNullOrWhiteSpace(target)
+            || string.IsNullOrWhiteSpace(kind)
+            || string.IsNullOrWhiteSpace(path))
         {
             return false;
         }
 
         var set = _discoverySeen.GetOrAdd(
```
gautam-achieveai, line 1193: [Revobot] **MEDIUM** (Resource Management)

Target-specific discovery marks for completed sub-agents remain until the whole session loses its last thread. Although target eviction exists, no production sub-agent teardown path calls it, so long-lived sessions accumulate target/path keys without a bound.

**Suggestion:** Connect permanent sub-agent removal to exact target-ledger cleanup, or apply a documented TTL/size bound. Test that a completed target is reclaimed while other live targets remain intact.

src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
```diff
@@ -148,4 +181,121 @@ public async Task<int> InjectAsync(ContextDiscoveryPayload body, CancellationTok
...
+                body.Path,
+                body.SessionId);
+            _diagnostics.RecordRoutingOutcome(ContextRoutingOutcome.Dropped);
+            return 0;
+        }
+
+        // No sink owns the id: the sub-agent hasn't registered yet (pre-registration race), or it is a
+        // nested / cross-session agent. Drop and UN-mark so the gateway's redelivery can retry once the
+        // sub-agent is live — but still never fall back to the primary.
+        _registry.EvictDiscoverySeenForTarget(body.SessionId!, agentId);
```
gautam-achieveai: [Revobot] **MEDIUM** (Concurrency / Deduplication)

This unresolved tuple rolls back via target-wide eviction. That can erase marks created by successful concurrent deliveries for other paths owned by the same target, allowing later retries to inject those files again.

**Suggestion:** Roll back only the exact `(session, target, kind, path)` mark. Reserve whole-target eviction for quiesced terminal cleanup, and add a concurrent distinct-path interleaving test.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
```diff
@@ -436,24 +420,128 @@ public async Task<string> SendMessageAsync(
...
+    private static async Task<bool> InjectIntoRunningLoopAsync(
+        SubAgentState state,
+        IReadOnlyList<IMessage> messages,
+        CancellationToken ct)
+    {
+        using var linkedCts = state.LinkLifecycleToken(ct);
+        var payload = messages as List<IMessage> ?? [.. messages];
+        try
+        {
+            _ = await state.Agent.SendAsync(payload, ct: linkedCts.Token);
```
gautam-achieveai, line 497: [Revobot] [BLOCKER] **HIGH** (Message Delivery Correctness)

The delivery lease ends when `Agent.SendAsync` returns after enqueueing, not when the context is consumed. Completion can observe no pending work and dispose the run after enqueue admission while this method still returns `Delivered`; the retained dedup mark then suppresses retries for context the target never consumed.

**Suggestion:** Make queue admission atomic with terminal completion, or retain/acknowledge the lease through durable consumption. Add a deterministic enqueue-versus-final-completion test proving that `Delivered` cannot be followed by silent loss.

gautam-achieveai:
[Revobot] ## Re-Review Summary: PR #201

## Final Quality Gate: REQUEST CHANGES

Retained **8** findings: **6 HIGH blockers** and **2 MEDIUM** findings. Six findings were posted inline; two could not be anchored to GitHub's changed-line map:

### [BLOCKER] HIGH — Lifecycle / Concurrency
`src/LmMultiTurn/SubAgents/SubAgentState.cs:282`

`BeginTerminalDisposalAsync` cancels and drains context-injection leases only when `OwnedProviderAgent` is non-null. With a borrowed provider, terminal status can be published while an admitted `SendAsync` remains active, so delivery can overlap completion or be reported accepted after the run is terminal.

**Suggestion:** Always cancel the lifecycle token and drain active injection leases before terminal publication; keep only provider disposal conditional on ownership. Add a borrowed-provider completion-versus-injection test.

### [BLOCKER] HIGH — Lifecycle / Disposal
`src/LmMultiTurn/SubAgents/SubAgentManager.cs:866`

`DisposeAsync` cancels and disposes sub-agent resources without first rejecting new context-delivery admissions or draining leases already acquired by `TryDeliverToRunningAsync`. Teardown can therefore race `SendAsync` against agent/provider disposal.

**Suggestion:** Add a manager/state disposal transition that rejects new leases, cancels lifecycle tokens, and drains admitted sends before disposing resources. Cover teardown racing a blocked `SendAsync` deterministically.

The blockers comprise two public API compatibility breaks and four lifecycle/concurrency defects affecting deduplication, completion, delivery acknowledgment, and manager disposal. Passing CI does not exercise downstream package compatibility or deterministic race interleavings.

### [QUESTION] Specification clarification

Should #198's acceptance text be updated to match approved plan v2's intentional behavior for a present but unresolved `agent_id`—drop without primary fan-out and permit retry? The approved plan is treated as authoritative here, so this is non-blocking, but leaving the older fallback wording creates contradictory guidance.

GitHub would not accept a formal request-changes review because the authenticated account owns this PR; this comment records the review verdict instead.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in e4060636 — restored the 3-arg `TryMarkDiscoverySeen(sessionId, kind, path)` as a delegating overload that forwards to the 4-arg with `SessionDiscoveryTarget`. Both signatures exist now; existing/external callers compile unchanged.

src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in e4060636 — restored the pre-#198 ctor `(registry, pool, formatter, logger)` as a delegating compatibility constructor (routing defaults OFF, own diagnostics). The 6-arg ctor stays the DI-selected primary path (all its extra deps are registered singletons, so no ambiguity).

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in e4060636 — removed the `set.IsEmpty` → ledger-remove block entirely (the check-then-act race you describe). The discovery ledger is now cleared only on session destroy/dispose, as before #198, so there is no membership-vs-eviction interleaving.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Correct — the per-target eviction was never wired to a sub-agent-teardown path (its only caller was the injector rollback), so the "per-sub-agent eviction" framing was inaccurate. Fixed in e4060636: removed the dead `EvictDiscoverySeenForTarget`. Per-target marks are now bounded by session lifetime (cleared on session destroy); per-sub-agent reclamation is deferred to the gateway-integration work (#187), where the routed path actually activates.

src/LmAgentInfra/Context/ContextDiscoveryInjector.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in e4060636 — the rollback now un-marks only the exact `(session, target, kind, path)` via a new `UnmarkDiscoverySeen`, so a concurrent sibling delivery's mark for the same target survives. Added `AllNotOwned_UnmarksOnlyItsPath_LeavingConcurrentDeliveredPathIntact` covering the interleaving.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
gautam-achieveai: [gautam-achieveai's bot] Acknowledged — this is the completion-boundary residual already called out under "Known limitation" in the PR description. Fully closing it (atomic queue-admission vs terminal completion) needs an invasive `MultiTurnAgentLoop.RunLoopAsync` change; per the issue owner's explicit decision the accepted behaviour is drop-clean-when-done, and the routed path is dormant (flag default-off) until the gateway stamps agent_id (#187), so it is not reachable in production today. Deferred to the #187 integration rather than taken on in this PR.

gautam-achieveai:
[Revobot]

Reviewed PR 201 across 20 files — 4 Must, 4 Should, 0 Consider.

- **Must — Borrowed-provider lifecycle race:** `src/LmMultiTurn/SubAgents/SubAgentState.cs:232-240,280-291` admits inject leases for borrowed providers, but terminal disposal drains/cancels leases only when `OwnedProviderAgent` is non-null. A borrowed-provider `SendAsync` can remain in flight while terminal completion proceeds. Drain active leases regardless of provider ownership; keep ownership conditional only for provider disposal. Add a deterministic teardown-race test.
- **Must — Unbounded routed dedup retention:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1131-1137,1193-1196` retains every `(target, kind, path)` until the complete sandbox session is destroyed. `UnmarkDiscoverySeen` only handles unresolved deliveries; successful routes have no production cleanup when sub-agents end. Add generation-safe cleanup or an explicit TTL/size bound without restoring the previous race.
- **Must — Unauthenticated diagnostics and sensitive operational data:** `samples/LmStreaming.Sample/Controllers/ContextDiscoveryDiagnosticsController.cs:21-50` exposes active/historical session IDs, filesystem paths, and the webhook URL, while the route is mapped with general controllers at `samples/LmStreaming.Sample/Program.cs:1428-1429` without evident operator authorization or development-only restriction. Related logs also emit workspace-controlled descriptions, paths, session IDs, agent IDs, and user-supplied prompts (`samples/LmStreaming.Sample/Controllers/ContextDiscoveryController.cs:116-123`, `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:135-193,215-299`, `src/LmMultiTurn/SubAgents/SubAgentManager.cs:378-380,708-710`). Restrict/minimize the endpoint and log metadata or pseudonymized identifiers.
- **Must — Public DTO compatibility:** `samples/LmStreaming.Sample/Controllers/ContextDiscoveryDiagnosticsController.cs:55-59` changes the public positional constructor from three parameters to four. Existing consumers constructing `ContextDiscoveryDiagnosticsResponse(DiscoveryEnabled, WebhookUrl, Sessions)` will break. Preserve a three-argument compatibility constructor or retain the original constructor shape.

- **Should — Ambiguous sink exceptions can duplicate context:** `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:235-253,289-300` treats every non-cancellation sink exception as `NotOwned`, then removes the dedup mark. If delivery succeeded before the exception, a retry can inject the same context twice. Retain the mark for ambiguous outcomes or define a sink contract guaranteeing exceptions occur before acceptance; add a regression test.
- **Should — Missing compatibility-overload tests:** The restored overloads at `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:75-82` and `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1204-1205` are not exercised. Add tests using the old constructor/signature to protect the compatibility guarantee.
- **Should — Swallowed cancellation callback failures:** `src/LmMultiTurn/SubAgents/SubAgentState.cs:299-305` discards `AggregateException` from lifecycle cancellation callbacks. Continue teardown, but report the exception through the existing logging/diagnostics path.
- **Should — Test lifecycle failure modes:** The new tests cover normal routing and precise rollback, but not borrowed-provider lease draining or throw-after-acceptance behavior. Add deterministic tests alongside the fixes.

The constructor and registry overload compatibility fixes, last-thread ledger race fix, and exact-key rollback fix from the previous review are now resolved. Focused tests could not run because `dotnet` is unavailable in the review environment.

**Request changes**

gautam-achieveai:
[gautam-achieveai's bot] Triage of the 4 Must / 4 Should (verified against the code before acting).

### Fixed
- **Should — Ambiguous sink exceptions can duplicate context** — valid. Fixed in `fbd320cc`: a sink whose `TryDeliverContextAsync` throws is now treated as *ambiguous* — the routed dedup mark is **kept** (never un-marked), so a gateway retry can't double-inject context a delivery may have accepted before throwing. Clean `NotOwned` still un-marks (retry heals the pre-registration race); `TargetNotDeliverable` still keeps the mark. Added a throwing-sink regression test.
- **Should — Missing compatibility-overload tests** — added in `fbd320cc`: a test constructing `ContextDiscoveryInjector` via the restored 4-arg compat ctor (routing off ⇒ fans out), and a registry test exercising the restored 3-arg `TryMarkDiscoverySeen` (dedups under `__session__`, sharing the bucket with the 4-arg form).

### Deferred / declined — with reasons
- **Must — Borrowed-provider lifecycle race.** `SubAgentState.BeginTerminalDisposalAsync` (`:272-306`) is **pre-existing #186 code this PR does not touch** — the diff only *adds* `TryBeginInjectLease`. Its owned-provider-only drain is deliberate (documented in-code): a borrowed provider is **not disposed** at completion, so an in-flight send races no disposal — this is the completion-boundary residual already called out under *Known limitation*, not a new crash. Making the shared #186 drain unconditional touches the deadlock-free lease invariants #186 established and is out of scope for #198; tracked for the completion-boundary/#187 follow-up.
- **Must — Unbounded routed dedup retention.** The routed path is **dormant** (flag default-off) until the gateway stamps `agent_id` (#187). Per-target marks are bounded by session lifetime — the same characteristic as the pre-existing `__session__` marks (never unbounded in practice; sessions are destroyed). A TTL/size bound is premature for a path carrying no production traffic; proper generation-safe per-sub-agent reclamation is deferred to the #187 integration, where it can be designed against the real activation lifecycle without reintroducing the race just removed.
- **Must — Unauthenticated diagnostics / sensitive logs.** The diagnostics endpoint and its fields (session ids, paths, webhook URL) are **pre-existing**; this PR's diff to that controller is 6 lines adding an aggregate routing counter (routed/dropped/fallback counts — no new sensitive data). The routing logs emit a file path + Guid agent/session ids at **debug** level, consistent with the existing injector logging, in the LmStreaming **sample** app. Endpoint hardening for the sample is a separate, pre-existing concern.
- **Must — Public DTO compatibility.** `ContextDiscoveryDiagnosticsResponse` lives in `samples/LmStreaming.Sample`, which is **not** a published package (no `PackageId`/`IsPackable`) — there are no external consumers to break. (Contrast the LmAgentInfra compat shims added last round — that assembly *is* packaged.)
- **Should — Swallowed cancellation-callback failures.** `SubAgentState.cs:299-306` is pre-existing #186 teardown code (untouched here) with a documented rationale — the cancellation has fully propagated; a throwing linked-token callback must not abort the terminal transition. `SubAgentState` is a lightweight internal state type with no logging dependency by design; adding diagnostics plumbing there is out of scope for #198.

Net this round: 1 correctness fix + compat tests. The remaining items are pre-existing #186/sample code, a dormant-path bound already documented as deferred, or a false-positive compat flag on un-packaged sample code — declined here rather than adding complexity to a dormant path.

gautam-achieveai:
[Revobot]

Reviewed PR 201 across 22 files — 5 Must, 2 Should, 0 Consider.

## Must

### 1. Borrowed-provider lifecycle race — **Must**
- **File:** `src/LmMultiTurn/SubAgents/SubAgentState.cs:272-315`
- **Related:** `src/LmMultiTurn/SubAgents/SubAgentManager.cs:455-472`
- `TryBeginInjectLease()` admits routed context delivery for borrowed-provider sub-agents, but terminal disposal drains and cancels active leases only when `OwnedProviderAgent` is non-null. A borrowed-provider run can therefore become terminal while its routed `SendAsync` is still in flight.
- Drain/cancel active inject leases for every terminal transition; keep provider disposal conditional on ownership. Add a deterministic borrowed-provider teardown test.

### 2. Manager disposal does not honor the new lease — **Must**
- **File:** `src/LmMultiTurn/SubAgents/SubAgentManager.cs:859-909`
- `DisposeAsync()` cancels, stops, and disposes agents without first setting them as terminating or waiting for `_activeSendLeases`. A discovery delivery can acquire the lease at lines 455–472 and race `StopAsync` or `DisposeAsync`, defeating the lease’s safety contract.
- Add a manager-shutdown admission barrier and drain leases before stopping or disposing each agent, with a corresponding race test.

### 3. Unauthenticated diagnostics expose sensitive operational data — **Must**
- **Files:** `samples/LmStreaming.Sample/Controllers/ContextDiscoveryDiagnosticsController.cs:15-50`; `samples/LmStreaming.Sample/Program.cs:1428-1429`
- The diagnostics endpoint returns active and historical session IDs, workspace paths, and the configured webhook URL. It is mapped through general `MapControllers()` wiring, with no authorization or development-only restriction evident in the reviewed code.
- Protect the endpoint with operator authentication or restrict it to development, and redact or pseudonymize session IDs and paths rather than exposing deployment topology to unauthenticated callers.

### 4. Cancellation can permanently poison routed dedup — **Must**
- **File:** `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:221-252`
- The routed dedup mark is claimed before cancellation is checked. If the token is already cancelled, the first sink call throws and the mark remains even though no delivery was attempted or accepted. A later gateway delivery is then silently discarded as a duplicate.
- Check cancellation before claiming the mark. Retain the mark only after a delivery attempt whose outcome may be ambiguous.

### 5. Ambiguous sink failure can double-deliver within one request — **Must**
- **File:** `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:253-278`
- After a sink throws, routing continues to other sinks even though the surrounding code acknowledges that the throwing sink may already have accepted the message. If a later sink returns `Delivered`, the same context can be injected into both conversations during one request.
- Stop routing after an ambiguous exception while retaining the dedup mark, or introduce an idempotent delivery identifier handled by every sink. The added exception test covers retry suppression with one sink, not this multi-sink case.

## Should

### 6. Unbounded target-scoped dedup retention — **Should**
- **File:** `src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs:1131-1137,1193-1196`
- Successful `(target, kind, path)` entries remain until the entire sandbox session is destroyed. Expanding the key to include every sub-agent multiplies retained state by historical agents and paths, while the implementation explicitly defers target reclamation to a future gateway dependency.
- Add lifecycle-safe target cleanup or a bounded/expiring ledger, with coverage for successful-route cleanup or retention bounds.

### 7. Missing behavioral coverage for lifecycle and public diagnostics edges — **Should**
- **Files:** `tests/LmMultiTurn.Tests/SubAgentManagerTests.cs:590-818`; `tests/LmStreaming.Sample.Tests/Controllers/ContextDiscoveryDiagnosticsControllerTests.cs:12-66`
- The new tests do not cover borrowed-provider terminal teardown, manager disposal racing an in-flight context send, or the controller-level `Routing` response values.
- Add deterministic race tests for the first two cases and assert routed, dropped, and fallback counters through the diagnostics controller.

Resolved since the prior review: compatibility constructors/overloads, precise rollback, and the clear-on-last-thread ledger race. Focused tests could not be run because `dotnet` is unavailable in the review environment.

**Verdict: Request changes**

gautam-achieveai:
[gautam-achieveai's bot] Round-3 triage (verified against the code before acting).

### Fixed (`27eb3d7b`)
- **Must 4 — Cancellation can poison routed dedup** — valid. `InjectAsync` now checks cancellation **before** claiming any mark, and the routed cancelled-rethrow path un-marks (no delivery ⇒ no stale mark that would silently drop a later gateway redelivery). Tests added (pre-cancelled + mid-delivery-cancel), RED-verified.
- **Must 5 — Ambiguous multi-sink double-deliver** — hardened. Routing now **stops** after an ambiguous sink exception (keeps the mark, drops), so a later sink can't also deliver in the same request; a `Delivered` result still short-circuits first. Test added. (In practice at most one sink owns a given `agent_id` — a unique per-spawn GUID — so this pins the contract defensively.)
- **Should 7 (partial)** — added a diagnostics-**controller** test asserting the routed/dropped/fallback counts on the response, plus the injector edge tests above.

### Declined — with evidence
- **Must 1 — Borrowed-provider lease drain.** `SubAgentState.BeginTerminalDisposalAsync` is pre-existing #186 code this PR does not touch (the diff only *adds* `TryBeginInjectLease`). My delivery uses the **identical** lease + lifecycle-token-linked send as the existing `SendMessageAsync` Inject branch, so it adds no new exposure; for a borrowed provider nothing is disposed, so an in-flight send races no disposal — the documented completion-boundary residual, deferred (Known limitation / #187).
- **Must 2 — Manager disposal vs the lease.** `SubAgentManager.DisposeAsync` (`:859-914`) is pre-existing #186 code (untouched). It cancels each run's `Cts` **first** — which cancels my linked inject send (⇒ `TargetNotDeliverable`) — and awaits `RunTask` before disposing. My path uses the same lease + linked-token machinery as `SendMessageAsync`, so a lease acquired during shutdown yields a **cancelled** send, not a send through a disposed agent. Adding a shutdown drain barrier changes shared #186 disposal semantics and isn't needed given that coordination.
- **Must 3 — Unauthenticated diagnostics.** The endpoint + its fields are pre-existing; this PR's diff to that controller is a 6-line aggregate counter (no new sensitive data). Debug logs emit a file path + Guid ids in the LmStreaming **sample** app. Endpoint hardening for the sample is a separate, pre-existing concern.
- **Should 6 — Unbounded target dedup.** The routed path is **dormant** (flag-off) until #187; per-target marks are bounded by session lifetime — the same characteristic as the pre-existing `__session__` marks. Generation-safe per-sub-agent reclamation is deferred to the #187 integration where the path activates; a TTL/size bound is premature for a path with no production traffic.
- **Should 7 (race tests).** The borrowed-provider-teardown and disposal-race tests are for the declined Must-1/Must-2 (pre-existing #186 behavior), so not added.

Net this round: 2 correctness fixes + coverage. Three Musts and one Should are re-raises of pre-existing #186 code (untouched by this PR) or a dormant-path bound already documented as deferred — declined rather than modifying shared lifecycle machinery or adding complexity to a dormant path.

gautam-achieveai:
[Revobot]

Reviewed PR 201 across 3 changed files — 0 Must, 1 Should, 2 Consider.

- **Should** — Add fallback-path coverage for the global pre-cancellation guard in `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:94-97`. The current test at `tests/LmStreaming.Sample.Tests/Services/ContextDiscoveryInjectorRoutingTests.cs:224-243` only exercises cancellation with routing enabled and a populated `agent_id`. Add a case with routing disabled or a blank `agent_id` that verifies cancellation occurs before the session-level dedup mark is created.
- **Consider** — Remove internal review bookkeeping such as `Must-4 / Must-5` and `FIX A/B` from committed comments in `tests/LmStreaming.Sample.Tests/Services/ContextDiscoveryInjectorRoutingTests.cs:222-278`; replace them with behavior-focused explanations or a stable issue reference.
- **Consider** — Consolidate the duplicated `IMultiTurnAgent`/`ISubAgentContextSink` scaffolding in `tests/LmStreaming.Sample.Tests/Services/ContextDiscoveryInjectorRoutingTests.cs:521-653` with the existing `RecordingSinkAgent`, using an optional delivery callback instead of maintaining a second full test double.

The changed production routing behavior has no identified correctness, security, compatibility, performance, logging, architecture, or lifecycle-invariant defects. `git diff --check` passed; tests could not be run because `dotnet` is unavailable in the review environment.

**Approve with comments**

gautam-achieveai:
[gautam-achieveai's bot] Round-4 addressed in `b49d6551` (test-only — production routing unchanged, per your approve-with-comments):
- **Should** — added `PreCancelledToken_FallbackPath_DoesNotMarkSession`, pinning that the pre-cancellation guard fires before the `__session__` (fallback) mark too.
- **Consider** — removed the internal review-bookkeeping labels (Must-4/Must-5/FIX A/B) from test comments in favour of behavior-focused prose (kept the #198/AC references).
- **Consider** — consolidated the duplicate `ProgrammableSink` double into `RecordingSinkAgent` via an optional `DeliverBehavior` hook and deleted the second double + its harness helper.

70 ContextDiscovery tests green; `dotnet format whitespace` clean. Thanks for the review pass.

gautam-achieveai:
[Revobot]

Reviewed PR #201 across 12 files — 1 Must, 0 Should, 0 Consider.

### Must — Do not roll back deduplication after an ambiguous cancellation

- **File:** `src/LmAgentInfra/Context/ContextDiscoveryInjector.cs:255-262`
- After invoking `TryDeliverContextAsync`, an `OperationCanceledException` causes the routed `(session, agent, kind, path)` dedup mark to be removed. However, the sink ultimately calls `SendAsync`, which may accept or enqueue the message before observing cancellation. A gateway retry can therefore pass deduplication and inject the same context twice.
- Treat cancellation after sink invocation as ambiguous and retain the mark, or change the sink contract to provide a definitive “not accepted” result. Only roll back when cancellation is known to occur before delivery begins.

The new two-file delta adds useful coverage for routing, fallback behavior, per-item projection, blank IDs, deduplication, and sink outcomes. No additional issues were found in the delta itself. Targeted tests could not be executed because the environment does not provide `dotnet`.

**Verdict: Request changes**

gautam-achieveai:
[gautam-achieveai's bot] Round-5 Must — valid, and it's a real flaw in my own round-3 cancellation guard (not a re-raise). Fixed in `1ece41d8`.

The round-3 change un-marked the routed dedup mark on ANY `OperationCanceledException` from a sink call. But a cancellation that lands **after** the sink was invoked is ambiguous — the sink's `SendAsync` may have accepted/enqueued the message before observing cancellation — so un-marking there could let a retry double-inject. The routed cancellation branch now **keeps** the mark and rethrows. The already-cancelled case (cancellation **before** delivery) is still handled by `ThrowIfCancellationRequested` at the top of `InjectAsync`, which throws before the mark is ever created, so there's no stale-mark to roll back. Test updated to assert the mark is kept (70 ContextDiscovery tests green).

gautam-achieveai:
[Revobot]

Reviewed PR 201 across 2 changed files — 0 Must, 0 Should, 0 Consider.

The cancellation change correctly retains the routed deduplication mark when cancellation occurs after a sink has been invoked, because delivery may have been accepted before cancellation was observed. The corresponding test update covers this behavior, and no correctness, security, compatibility, performance, maintainability, or coverage issues were identified.

**Approve**

PR #203, opened 2026-07-15
Fix chat client whole-page scrollbar + clipped header Clear button (#202)

### LINKED ISSUE #202: LmStreaming chat client: spurious whole-page scrollbar + clipped header Clear button
## Summary

While testing the #199 rich per-tool rendering in the LmStreaming.Sample chat client, two layout defects surfaced. Neither breaks the tool pills themselves, but both are visible in normal use.

### 1. Spurious whole-page scrollbar
The chat client is a fixed full-height shell — only the conversation list (`.message-list`) should scroll. When tool pills render, a second **whole-page** vertical scrollbar appears next to the conversation scrollbar.

Root cause: the tool pills render `.sr-only` accessibility labels with `position: absolute`, but every ancestor up to the document is `position: static`. With no positioned ancestor, those labels' containing block resolves to the document itself, so `.chat-layout { overflow: hidden }` cannot clip them (`overflow` only clips descendants whose containing block *is* that element). Below the fold of the scrolled conversation they lengthen `documentElement` past the viewport.

### 2. Clipped header "Clear" button
The header action row (Workspace / Provider / Mode selectors + Marketplaces + Clear) is wider than the content column, and `.header-actions` was `flex-shrink: 0` with no wrapping, so the trailing **Clear** button overflows the row and is clipped off the right edge on typical laptop widths (~1280–1440px). It produced no scrollbar (clipped by `.chat-layout`) but the button was unreachable.

## Fix
- `.tool-pill { position: relative }` — makes the pill the containing block for its own absolute `.sr-only` labels so they stay inside the pill and scroll within `.message-list` instead of the page.
- Header wraps (`flex-wrap` on `.chat-header` + `.header-actions`, right-aligned) so controls drop to a second row instead of overflowing.

## Files
- `samples/LmStreaming.Sample/ClientApp/src/components/ToolPill.vue`
- `samples/LmStreaming.Sample/ClientApp/src/components/ChatLayout.vue`
- `samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts` (regression guard)

## Verification
Reproduced and verified fixed in-browser (`test-anthropic` mock, real tool pills) across collapsed / expanded / scrolled and 1024×600 / 768 / 1280 / 1440: the page has no scrollbar (`pageV=0`, `maxTop=0`, `html.scrollHeight == viewport`) while the conversation still scrolls internally, and every header button stays within the viewport. 459/459 vitest green.

gautam-achieveai:
## Summary

Two UI layout fixes for the LmStreaming.Sample chat client, found while testing the #199 rich per-tool rendering: a spurious whole-page scrollbar and a clipped header "Clear" button. Both are small, contained CSS fixes plus a regression test.

## Changes
- **Page scroll:** `.tool-pill { position: relative }` makes the pill the containing block for its absolutely-positioned `.sr-only` accessibility labels, so they scroll inside `.message-list` instead of escaping `.chat-layout`'s clip and lengthening the document into a whole-page scrollbar.
- **Header:** `.chat-header` / `.header-actions` now wrap (right-aligned), so the Workspace/Provider/Mode selectors + Marketplaces + Clear controls drop to a second row instead of clipping the "Clear" button off the right edge on typical laptop widths.
- **Test:** `AppShellLayout.test.ts` locks the invariants — the pill is a positioning context, the shell stays height-capped + clipped, and `.message-list` remains the single internal scroll region.

## Testing
- 459/459 vitest green; `vue-tsc` clean.
- Verified in-browser (`test-anthropic` mock, real tool pills) across collapsed / expanded / scrolled and 1024×600 / 768 / 1280 / 1440: the page has no scrollbar (`pageV=0`, `maxTop=0`, `html.scrollHeight == viewport`) while the conversation still scrolls internally, and every header button stays within the viewport.

## Related Issues
Fixes #202

### DIFF
--- a/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts
@@ -59,3 +59,36 @@ describe('Conversation-scroll containment (page-scroll regression)', () => {
     expect(messageList).toMatch(/\.message-list\s*\{[^}]*overflow-y:\s*auto/);
   });
 });
+
+/**
+ * Header-wrap regression guard (Clear-button clip).
+ *
+ * The header lays the title (`h1`, `flex: 1`) beside a right-aligned control row
+ * (`.header-actions`: workspace + provider + mode selectors, "Marketplaces", "Clear").
+ * On typical laptop widths those controls are collectively wider than the 900px content
+ * column, so with a single non-wrapping flex row the trailing "Clear" button overflowed
+ * the right edge and was clipped off-screen.
+ *
+ * The fix lets the row reflow: `.chat-header` wraps the control row below the title, and
+ * `.header-actions` wraps its own buttons, so "Clear" always stays within the viewport.
+ *
+ * This is the fast structural guard (source-text) sitting under the behavioral
+ * `ChatClientLayoutRegressionTests` Browser.E2E test, which actually renders the header at
+ * 1280px and asserts the Clear button falls inside the viewport. happy-dom computes no
+ * real layout, so the on-screen containment can only be proven in the C# Chromium suite;
+ * this keeps a cheap, always-on check that the wrap declarations are not removed.
+ */
+describe('Header wrapping (Clear-button clip regression)', () => {
+  const chatLayout = fs.readFileSync(
+    path.resolve(__dirname, '../../components/ChatLayout.vue'),
+    'utf-8'
+  ) as string;
+
+  it('lets .chat-header wrap so the control row drops below the title instead of clipping', () => {
+    expect(chatLayout).toMatch(/\.chat-header\s*\{[^}]*flex-wrap:\s*wrap/);
+  });
+
+  it('lets .header-actions wrap its own controls rather than overflowing the row', () => {
+    expect(chatLayout).toMatch(/\.header-actions\s*\{[^}]*flex-wrap:\s*wrap/);
+  });
+});
--- a/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from 'vitest';
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * App-shell / conversation-scroll regression guard.
+ *
+ * The chat client is a FIXED full-height shell: the page/body must never scroll —
+ * only the conversation list (`.message-list`) scrolls, internally.
+ *
+ * A whole-page scrollbar appeared once (alongside the correct conversation
+ * scrollbar) because the tool pills render `.sr-only` accessibility labels with
+ * `position: absolute`, and EVERY ancestor up to the document was `position: static`.
+ * With no positioned ancestor, those labels' containing block resolved to the
+ * initial containing block (the document), so `.chat-layout { overflow: hidden }`
+ * could NOT clip them — they sat at their in-flow depth (below the fold of the
+ * scrolled conversation) and lengthened `documentElement` into a spurious page
+ * scrollbar. `overflow: hidden` only clips descendants whose containing block is
+ * that element, so clipping an ancestor was not enough.
+ *
+ * The fix: `.tool-pill` is `position: relative`, becoming the containing block for
+ * its own absolutely-positioned `.sr-only` labels so they stay inside the pill and
+ * are absorbed by `.message-list`'s internal scroll instead of the page.
+ *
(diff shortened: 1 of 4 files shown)


gautam-achieveai:
[Revobot]

Reviewed PR 203 across 3 files — 0 Must, 2 Should, 0 Consider.

## Should — Header wrapping has no regression coverage

- **File:** `samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts:53-60`
- `ChatLayout.vue:614-660` changes `.chat-header` and `.header-actions` to wrap so the Clear button remains reachable at constrained widths, but the new test only checks tool-pill, shell, and message-list CSS. The tests would pass against the pre-change header, leaving the user-visible clipping regression unguarded.
- Add a component or browser-level regression at an affected viewport that asserts the Clear button remains within the viewport/header bounds.

## Should — Page-scroll regression test checks source text, not behavior

- **File:** `samples/LmStreaming.Sample/ClientApp/src/__tests__/components/AppShellLayout.test.ts:46-60`
- The regexes verify that CSS declarations exist, but do not render the layout or assert that the document does not scroll while `.message-list` does. A later cascade, specificity, flex-sizing, or containing-block regression could restore the page scrollbar while this test remains green.
- Add a Playwright/browser regression with overflowing conversation/tool content that verifies document overflow is capped and `.message-list` is the actual scroll region. The source checks may remain as fast structural guards, but should not be the only behavioral protection.

The production CSS changes are small and coherent: `position: relative` correctly establishes a containing block for the pill’s absolutely positioned accessibility labels, and flex wrapping addresses the constrained-width header issue. No security, compatibility, temporary-code, or performance defect was identified in the changed production lines.

Approve with comments.

gautam-achieveai:
[Gautam's bot]

Thanks for the review — both Should findings are addressed in 70a1e517.

**Should #2 (page-scroll test checks source, not behavior) + Should #1 (header wrap has no regression coverage)** — Added a behavioral Browser.E2E regression, `ChatClientLayoutRegressionTests`, that drives the real chat client under headless Chromium (same precedent as `ProviderDropdownScrollTests`, added because a source-level CSS assertion alone couldn't prove browser behavior). It scripts 12 `calculate` tool-call turns plus a long final text turn to overflow a pinned **1280×800** viewport, then asserts the observable layout:

- **(a) the document does not scroll** — `document.scrollingElement` overflow (`scrollHeight - clientHeight`) ≤ 1px AND a live scroll probe (`scrollTop = 9999` clamps back to 0) — checked BOTH with pills collapsed AND after expanding every pill (where the absolutely-positioned `.sr-only` leak was largest);
- **(b) `.message-list` IS the scroll region** — its `scrollHeight - clientHeight > 0`, so the "page doesn't scroll" assertion is not vacuous;
- **(c) the header "Clear" button stays within the viewport** — its bounding-box right edge (`X + Width`) ≤ 1280.

This is exactly the cascade/specificity/flex/containing-block regression surface the source-text checks can't cover, since vitest runs in happy-dom which computes no real layout.

**Kept the existing source-text checks** as the fast structural guards (as you suggested), and **added a matching fast guard for the header**: `AppShellLayout.test.ts` now also asserts `.chat-header` and `.header-actions` declare `flex-wrap: wrap`.

**RED/GREEN proof:** with the production fixes temporarily reverted, the new E2E test FAILS — expanded pills re-introduce a **~1482px** whole-page overflow (assertion (a)), and without the header wrap "Clear" sits at right ≈ 1346px (assertion (c)); restoring the fixes turns it GREEN. vitest stays green at **461/461** (+2 for the header guard).

gautam-achieveai:
Revobot — Reviewed PR 203 across 4 files — 0 Must, 0 Should, 1 Consider.

The production CSS changes are small and are backed by structural Vitest guards and a non-vacuous Chromium test covering message-list overflow, document non-scrollability, and Clear-button containment. No security, correctness, schema, or performance blockers were found.

Verdict: Approve with comments.

gautam-achieveai:
[Gautam's bot]

Consider (test robustness) addressed in `03e1d249`: the pill-expansion step in `ChatClientLayoutRegressionTests` now targets `[data-testid="tool-call-pill"] .tool-pill__header` explicitly and clicks through Playwright's actionability checks (no raw DOM `click()`), asserting one header per pill so the intended interaction is explicit and fails clearly if a header is missing or not actionable. Behavior is unchanged and the test is still green (1/1).

gautam-achieveai (commented):
Revobot **Consider** (Test robustness)

The expansion script targets every button currently under each tool pill. This works with today’s collapsed markup because only `.tool-pill__header` exists before expansion, but a future collapsed control would also be clicked, and raw DOM click bypasses Playwright actionability checks.

**Suggestion:** Consider targeting `[data-testid="tool-call-pill"] .tool-pill__header` explicitly and asserting the expected count so the test expresses the intended interaction and fails clearly if a header is not actionable.

PR #212, opened 2026-07-19
feat(#210): pre-defined egress keys for sandbox egress auth

### LINKED ISSUE #210: Add support for egress-policy with following 1) pre-defined key (header name + value), 2) UX shows a dialogue with header name (defaults to auth header for access token) and value text box.
## Summary

Add a "pre-defined key" egress-auth policy to the Workspace Agent sandbox: per-entry credentials — a
list of custom headers (injected verbatim) or an OAuth2 token the app mints and auto-refreshes
(refresh-token or client-credentials grant) — that the gateway injects on outbound requests to a
user-specified host, managed through a new dialog, with a deferred re-prompt when a required key is
missing or marked invalid.

Today sandbox egress auth is OAuth-only (interactive sign-in mints a token injected as
`Authorization: Bearer <token>`) and only for three fixed providers (github/ado/m365) whose hosts are
compile-time constants. This lets a user authenticate egress to **any** service with a credential
they already hold.

## Value

Lets a user authenticate sandbox egress to any service using a credential they already hold (a
cookie, a token, an API key) or an OAuth2 client credential — no interactive sign-in, no code change
— while preserving default-deny egress and never exposing the secret to the sandbox.

## Credential kinds

- **Custom Headers + Value** — a **list** of `{ headerName, value }` pairs injected verbatim (cookie,
  `Authorization` token, API key, any custom headers).
- **OAuth2 Refresh Token** — `grant_type=refresh_token`; app injects `Bearer <access>` and auto-refreshes.
- **OAuth2 Client-Credentials** — `grant_type=client_credentials` (client id + secret); mints + auto-refreshes.

Each entry is generic OAuth2 per entry (token endpoint URL + client id + secret + refresh token +
scopes live on the entry — no provider presets).

## Acceptance Criteria

**Credential kinds & injection**
- [ ] An entry = `{ host(s), kind, kind-fields }`, `kind ∈ { custom-headers, refresh-token, client-credentials }`.
- [ ] custom-headers: the webhook injects every `headerName: value` pair verbatim (no Bearer/Basic), no expiry.
- [ ] refresh-token: app runs `grant_type=refresh_token`, injects `Bearer <access>` under `headerName`, auto-refreshes within the expiry skew; minted tokens persist.
- [ ] client-credentials: app runs `grant_type=client_credentials`, injects `Bearer <access>`, auto-refreshes; persists.
- [ ] Each configured entry emits one webhook `SandboxAuthProvider` + one host-scoped `SandboxNetworkRule` (allow, port 443) at session create; no entries ⇒ blocks omitted (existing behavior unchanged).
- [ ] The webhook injects an entry's headers only when `destination_host` matches that entry's host(s) (fail closed on unknown id / host mismatch).

**Validation & security**
- [ ] Host accepts exact host or `*.suffix`; rejects `*`, empty, scheme/port/path, `0.0.0.0`, loopback, `169.254.169.254`, trailing-dot — with a dialog error. Same matcher builds the rule and gates the webhook.
- [ ] A host colliding with a managed OAuth host (github/ado/m365) is rejected; `(host, headerName)` unique within an entry.
- [ ] Every header name matches RFC 7230 token charset and is not hop-by-hop/forbidden (`Host`, `Content-Length`, `Connection`, `Transfer-Encoding`, `Content-Type`); names/values reject CR/LF/control. (`Cookie` allowed.)
- [ ] Static/custom-header rules stay 443-only (secret never egresses in cleartext).
- [ ] Secret material persists under the gitignored `oauth-tokens/` dir via atomic writes; never logged; GET returns masked/omitted values only; DELETE removes it from disk.
- [ ] The CRUD API is same-origin/local-UI only and is not in any sandbox allow-rule.
- [ ] The headless daemon fails closed (deny, no prompt) for a missing/invalid key.

**UX dialog**
- [ ] A new Egress Auth dialog (opened from the chat header) lists entries (secrets masked) and supports create/edit/delete; kind selector drives the fields; custom-headers supports add/remove header rows; `headerName` defaults to `Authorization`.
- [ ] Adding/removing an entry applies to sessions created afterward; existing sessions unchanged (OAuth parity) — the UI states this.

**Re-prompt on missing / marked-invalid key**
- [ ] When a sandbox request matches an entry whose key is missing or invalidated, the webhook holds and the chat client shows a banner whose CTA opens the Egress Auth dialog prefilled to the failing host/entry. On submit the held call resolves allow; on timeout it denies (no infinite loop).
- [ ] "Invalidated" is set internally on a token-mint failure (refresh-token / client-credentials only). Custom-header entries are only ever "missing" — no app-side invalidation, no public invalidate endpoint.
- [ ] The auth WS frame carries a `kind` discriminator (`oauth` | `predefined-key`) so the banner routes correctly.

## Out of scope (separate follow-ups)

- Auto-detecting a post-egress 401/403 (target rejected the injected key) — needs a gateway webhook-contract change in SandboxedOstoolsMcpServer.
- "Paste a Chrome capture → extract headers" import for custom-headers entries (the entry model is a header list precisely so this drops in later).

gautam-achieveai:
## Summary

Adds **pre-defined egress keys** (#210): per-entry credentials the sandbox gateway injects on
outbound requests to a user-specified host, alongside the existing OAuth webhook. Three credential
kinds:

- **Custom Headers + Value** — a *list* of `{ headerName, value }` pairs injected verbatim (a session
  cookie, an `Authorization` token, an API key, any custom header(s)).
- **OAuth2 Refresh Token** — the app runs `grant_type=refresh_token` and injects `Bearer <access>`, auto-refreshing.
- **OAuth2 Client-Credentials** — the app runs `grant_type=client_credentials`, injects `Bearer <access>`, auto-refreshing.

Managed at runtime via a new **Egress Auth** dialog + REST CRUD (`api/auth/egress-keys`), so a user
can authenticate sandbox egress to *any* service with a credential they already hold — no interactive
sign-in, no code change.

## Design

Reuses the existing egress-auth machinery rather than duplicating it:
- One `PredefinedKeyProvider : IOAuthTokenProvider` (kind switch); the webhook's `Resolve` falls back
  to a runtime `PredefinedKeyRegistry`, with a polymorphic host-gate + `AllowCustom` header builder.
- `SandboxSessionRegistry.BuildAuthProviders` emits a `webhook` auth-provider + host-scoped port-443
  `allow` rule per entry (a new **optional** registry param → `null` = fail-closed, zero churn to the
  ~30 existing callers and the daemon).
- A missing / mint-invalidated key re-prompts through the **existing deferred-auth hold + banner**;
  the banner routes `predefined-<id>` providers to the Egress Auth dialog (the provider-id prefix is
  the discriminator — no new WS-frame field). Updating the entry resolves the held call automatically.
- `OAuthTokenEndpointClient` (refresh/client-credentials grants) is new; `GitHubOAuthProvider` is left
  untouched. `FileOAuthTokenStore` shares a small `AtomicJsonFile` helper with the registry.

Scope was trimmed after an over-engineering + class-design review: no one-impl interface, no separate
store type, no config-seeding, no `/invalidate` endpoint.

## Security

- Secrets persist under the gitignored `oauth-tokens/predefined-keys.json` (minted tokens keyed
  `predefined-<id>`); **never logged**, GET returns masked values only, DELETE removes them.
- Default-deny egress preserved: each entry opens **only its own host**, HTTPS/**443 only**. Host
  patterns are SSRF-validated (rejects `*`, loopback, `169.254.169.254`, scheme/port/path, and
  managed-OAuth-host collisions); the same matcher builds the rule and gates the webhook.
- Header names must be RFC 7230 tokens and may not be hop-by-hop/framing headers (`Cookie` allowed).
- The webhook injects a key's headers **only** when the destination host matches that entry (anti-oracle).
- The headless CodeReviewDaemon fails closed (no keys, no prompt).

## Testing

- Backend: **758** unit tests + **8** over-HTTP webhook E2E (incl. create-key → inject-verbatim →
  host-mismatch-deny), 0 failures.
- Frontend: **492** Vitest passing, `vue-tsc` clean.
- Daemon build green (fail-closed via the optional param).

## Out of scope

Detecting a **post-egress** 401/403 (a static key the *target* rejects) needs a gateway change —
filed as **achieveai/SandboxedOstoolsMcpServer#143**. This PR scopes the re-prompt to
missing / mint-invalid only.

Docs: `samples/LmStreaming.Sample/AuthProviderGuide.md` gains a "Pre-defined egress keys" section.

Closes #210

### DIFF
--- a/samples/LmStreaming.Sample/ClientApp/src/components/EgressAuthModal.vue
+++ b/samples/LmStreaming.Sample/ClientApp/src/components/EgressAuthModal.vue
@@ -1,0 +1,751 @@
+<script setup lang="ts">
+import { computed, onMounted, onBeforeUnmount, ref } from 'vue';
+import type { EgressHeaderPair, EgressKeyKind, EgressKeyRequest, EgressKeyView } from '@/types/egressAuth';
+import { useEgressAuth, egressDialogRequest } from '@/composables/useEgressAuth';
+
+const emit = defineEmits<{ close: [] }>();
+
+const { egressKeys, isLoading, loadEgressKeys, saveEgressKey, removeEgressKey } = useEgressAuth();
+
+// --- Editor state -----------------------------------------------------------
+const editorOpen = ref(false);
+const editingId = ref<string | null>(null);
+const host = ref('');
+const kind = ref<EgressKeyKind>('custom-headers');
+const headerName = ref('Authorization');
+const tokenEndpoint = ref('');
+const clientId = ref('');
+const clientSecret = ref('');
+const refreshToken = ref('');
+const scopesText = ref('');
+const headerRows = ref<EgressHeaderPair[]>([{ name: '', value: '' }]);
+
+const hostError = ref('');
+const serverError = ref('');
+const saving = ref(false);
+
+const isEditing = computed(() => editingId.value !== null);
+const isOAuth = computed(() => kind.value !== 'custom-headers');
+const editorTitle = computed(() => (isEditing.value ? 'Edit egress key' : 'New egress key'));
+
+const KIND_OPTIONS: { value: EgressKeyKind; label: string }[] = [
+  { value: 'custom-headers', label: 'Custom Headers' },
+  { value: 'refresh-token', label: 'OAuth2 Refresh Token' },
+  { value: 'client-credentials', label: 'OAuth2 Client-Credentials' },
+];
+
+function kindLabel(k: EgressKeyKind): string {
+  return KIND_OPTIONS.find((o) => o.value === k)?.label ?? k;
+}
+
+function resetForm(): void {
+  editingId.value = null;
+  host.value = '';
+  kind.value = 'custom-headers';
+  headerName.value = 'Authorization';
+  tokenEndpoint.value = '';
+  clientId.value = '';
+  clientSecret.value = '';
+  refreshToken.value = '';
+  scopesText.value = '';
+  headerRows.value = [{ name: '', value: '' }];
+  hostError.value = '';
+  serverError.value = '';
+}
+
+function startCreate(prefillHost?: string): void {
+  resetForm();
+  if (prefillHost) {
+    host.value = prefillHost;
+  }
+  editorOpen.value = true;
+}
+
+function startEdit(view: EgressKeyView): void {
+  resetForm();
+  editingId.value = view.id;
+  host.value = view.host;
+  kind.value = view.kind;
+  headerName.value = view.headerName || 'Authorization';
+  scopesText.value = view.scopes.join(' ');
+  if (view.kind === 'custom-headers') {
+    // Prefill header NAMES only; values are masked, so leave them blank to keep
+    // the stored secret on update.
+    headerRows.value =
+      view.headerNames.length > 0
+        ? view.headerNames.map((n) => ({ name: n, value: '' }))
+        : [{ name: '', value: '' }];
+  }
+  editorOpen.value = true;
+}
+
+function cancelEditor(): void {
+  editorOpen.value = false;
+  resetForm();
+}
+
+function addHeaderRow(): void {
+  headerRows.value.push({ name: '', value: '' });
+}
+
+function removeHeaderRow(index: number): void {
+  if (headerRows.value.length > 1) {
+    headerRows.value.splice(index, 1);
+  }
+}
+
+function parseScopes(text: string): string[] {
+  return text
+    .split(/[\s,]+/)
+    .map((s) => s.trim())
+    .filter((s) => s.length > 0);
+}
+
+/**
+ * Builds the upsert request. Blank optional fields are OMITTED (not sent as null),
+ * so on update the server preserves the stored value — this is what keeps secrets
+ * (clientSecret / refreshToken / blank custom-header values) intact when left blank.
+ */
+function buildRequest(): EgressKeyRequest {
+  const req: EgressKeyRequest = {
+    id: editingId.value,
+    host: host.value.trim(),
+    kind: kind.value,
+  };
+
+  if (kind.value === 'custom-headers') {
+    req.headers = headerRows.value
+      .filter((r) => r.name.trim().length > 0)
+      .map((r) => ({ name: r.name.trim(), value: r.value }));
+  } else {
+    req.headerName = headerName.value.trim() || 'Authorization';
+    req.scopes = parseScopes(scopesText.value);
+
+    const te = tokenEndpoint.value.trim();
+    if (te) req.tokenEndpoint = te;
+
+    const ci = clientId.value.trim();
+    if (ci) req.clientId = ci;
+
+    if (clientSecret.value.length > 0) req.clientSecret = clientSecret.value;
+
+    if (kind.value === 'refresh-token' && refreshToken.value.length > 0) {
+      req.refreshToken = refreshToken.value;
+    }
+  }
+
+  return req;
+}
+
+async function handleSave(): Promise<void> {
+  hostError.value = '';
+  serverError.value = '';
+
+  if (!host.value.trim()) {
+    hostError.value = 'Host is required';
+    return;
+  }
+
+  saving.value = true;
+  try {
+    await saveEgressKey(buildRequest());
+    editorOpen.value = false;
+    resetForm();
+  } catch (e) {
+    serverError.value = e instanceof Error ? e.message : 'Failed to save egress key';
+  } finally {
+    saving.value = false;
+  }
+}
+
+async function handleDelete(id: string): Promise<void> {
+  serverError.value = '';
+  try {
+    await removeEgressKey(id);
+    // If we were editing the deleted key, drop the editor.
+    if (editingId.value === id) {
+      cancelEditor();
+    }
+  } catch (e) {
+    serverError.value = e instanceof Error ? e.message : 'Failed to delete egress key';
+  }
+}
+
+// --- Modal shell ------------------------------------------------------------
+function handleClose(): void {
+  emit('close');
+}
+
+function handleBackdropClick(event: MouseEvent): void {
+  if (event.target === event.currentTarget) {
+    handleClose();
+  }
+}
+
+function handleKeydown(event: KeyboardEvent): void {
+  if (event.key === 'Escape') {
+    handleClose();
+  }
+}
+
+onMounted(() => {
+  document.addEventListener('keydown', handleKeydown);
+  loadEgressKeys();
+  // Honour a programmatic open (openEgressDialog(host)): start in create mode with
+  // the host prefilled.
+  const requestState = egressDialogRequest.value;
+  if (requestState.open) {
+    startCreate(requestState.prefillHost);
+  }
+});
+
+onBeforeUnmount(() => document.removeEventListener('keydown', handleKeydown));
+</script>
+
+<template>
+  <div class="modal-backdrop" data-testid="egress-auth-modal" @click="handleBackdropClick">
+    <div class="modal-container">
+      <div class="modal-header">
+        <h2 class="modal-title">Egress Auth Keys</h2>
+        <button
+          class="close-btn"
+          data-testid="egress-auth-modal-close"
+          title="Close"
+          @click="handleClose"
+        >
+          &times;
+        </button>
+      </div>
+
+      <div class="modal-content">
+        <!-- Existing keys ------------------------------------------------- -->
+        <section class="keys-section">
+          <div class="section-head">
+            <h3 class="section-title">Configured keys</h3>
+            <button
+              type="button"
+              class="btn btn-primary"
+              data-testid="egress-add-button"
+              @click="startCreate()"
+            >
+              + Add key
+            </button>
+          </div>
+
+          <p v-if="isLoading" class="muted">Loading…</p>
+          <p v-else-if="egressKeys.length === 0" class="muted">
+            No egress keys configured yet.
+          </p>
+
+          <ul v-else class="key-list">
+            <li
+              v-for="key in egressKeys"
+              :key="key.id"
+              class="key-item"
+              data-testid="egress-key-item"
+              :data-key-id="key.id"
+            >
+              <div class="key-main">
+                <span class="key-host">{{ key.host }}</span>
+                <span class="kind-badge">{{ kindLabel(key.kind) }}</span>
+              </div>
+              <div class="key-meta">
+                <span v-if="key.headerNames.length" class="meta-line">
+                  Headers: {{ key.headerNames.join(', ') }}
+                </span>
+                <span v-else-if="key.headerName" class="meta-line">
+                  Header: {{ key.headerName }}
+                </span>
+                <span v-if="key.scopes.length" class="meta-line">
+                  Scopes: {{ key.scopes.join(' ') }}
+                </span>
+                <span v-if="key.hasClientSecret" class="indicator">client secret set</span>
+                <span v-if="key.hasRefreshToken" class="indicator">refresh token set</span>
+              </div>
+              <div class="key-actions">
+                <button
+                  type="button"
+                  class="btn btn-secondary btn-sm"
+                  @click="startEdit(key)"
+                >
+                  Edit
+                </button>
+                <button
+                  type="button"
+                  class="btn btn-danger btn-sm"
+                  data-testid="egress-delete-button"
+                  @click="handleDelete(key.id)"
+                >
+                  Delete
+                </button>
+              </div>
+            </li>
+          </ul>
+        </section>
+
+        <!-- Editor -------------------------------------------------------- -->
+        <section v-if="editorOpen" class="editor-section">
+          <h3 class="section-title">{{ editorTitle }}</h3>
+
+          <form class="editor-form" @submit.prevent="handleSave">
+            <div class="form-group">
+              <label class="form-label" for="egress-kind">Kind</label>
+              <select
+                id="egress-kind"
+                v-model="kind"
+                class="form-input"
+                data-testid="egress-kind-select"
+                :disabled="saving"
+              >
+                <option v-for="opt in KIND_OPTIONS" :key="opt.value" :value="opt.value">
+                  {{ opt.label }}
+                </option>
+              </select>
+            </div>
+
+            <div class="form-group">
+              <label class="form-label" for="egress-host">
+                Host <span class="required">*</span>
+              </label>
+              <input
+                id="egress-host"
+                v-model="host"
+                type="text"
+                class="form-input"
+                :class="{ error: hostError }"
+                data-testid="egress-host-input"
+                placeholder="api.example.com"
+                :disabled="saving"
+              />
+              <span v-if="hostError" class="error-message">{{ hostError }}</span>
+            </div>
+
+            <!-- Custom headers -->
+            <template v-if="!isOAuth">
+              <div class="form-group">
+                <label class="form-label">Headers</label>
+                <div
+                  v-for="(row, index) in headerRows"
+                  :key="index"
+                  class="header-row"
+                >
+                  <input
+                    v-model="row.name"
+                    type="text"
+                    class="form-input header-name"
+                    data-testid="egress-header-name-input"
+                    placeholder="Header name (e.g. Authorization)"
+                    :disabled="saving"
+                  />
+                  <input
+                    v-model="row.value"
+                    type="text"
+                    class="form-input header-value"
+                    data-testid="egress-header-value-input"
+                    placeholder="Header value"
+                    :disabled="saving"
+                  />
+                  <button
+                    type="button"
+                    class="btn btn-secondary btn-sm"
+                    :disabled="headerRows.length <= 1 || saving"
+                    @click="removeHeaderRow(index)"
+                  >
+                    &times;
+                  </button>
+                </div>
+                <button
+                  type="button"
+                  class="btn btn-secondary btn-sm add-header-btn"
+                  data-testid="egress-add-header-row"
+                  :disabled="saving"
+                  @click="addHeaderRow"
+                >
+                  + Add header
+                </button>
+                <span v-if="isEditing" class="hint">Leave a value blank to keep the current secret.</span>
+              </div>
+            </template>
+
+            <!-- OAuth kinds -->
+            <template v-else>
+              <div class="form-group">
+                <label class="form-label" for="egress-header-name">Header name</label>
+                <input
+                  id="egress-header-name"
+                  v-model="headerName"
+                  type="text"
+                  class="form-input"
+                  placeholder="Authorization"
+                  :disabled="saving"
+                />
+              </div>
+
+              <div class="form-group">
+                <label class="form-label" for="egress-token-endpoint">Token endpoint</label>
+                <input
+                  id="egress-token-endpoint"
+                  v-model="tokenEndpoint"
+                  type="text"
+                  class="form-input"
+                  placeholder="https://auth.example.com/oauth/token"
+                  :disabled="saving"
+                />
+                <span v-if="isEditing" class="hint">Leave blank to keep current.</span>
+              </div>
+
+              <div class="form-group">
+                <label class="form-label" for="egress-client-id">Client id</label>
+                <input
+                  id="egress-client-id"
+                  v-model="clientId"
+                  type="text"
+                  class="form-input"
+                  placeholder="client id"
+                  :disabled="saving"
+                />
+                <span v-if="isEditing" class="hint">Leave blank to keep current.</span>
+              </div>
+
+              <div class="form-group">
+                <label class="form-label" for="egress-client-secret">Client secret</label>
+                <input
+                  id="egress-client-secret"
+                  v-model="clientSecret"
+                  type="password"
+                  class="form-input"
+                  placeholder="client secret"
+                  autocomplete="off"
+                  :disabled="saving"
+                />
+                <span v-if="isEditing" class="hint">Leave blank to keep current.</span>
+              </div>
+
+              <div v-if="kind === 'refresh-token'" class="form-group">
+                <label class="form-label" for="egress-refresh-token">Refresh token</label>
+                <input
+                  id="egress-refresh-token"
+                  v-model="refreshToken"
+                  type="password"
+                  class="form-input"
+                  placeholder="refresh token"
+                  autocomplete="off"
+                  :disabled="saving"
+                />
+                <span v-if="isEditing" class="hint">Leave blank to keep current.</span>
+              </div>
+
+              <div class="form-group">
+                <label class="form-label" for="egress-scopes">Scopes</label>
+                <input
+                  id="egress-scopes"
+                  v-model="scopesText"
+                  type="text"
+                  class="form-input"
+                  placeholder="space- or comma-separated scopes"
+                  :disabled="saving"
+                />
+              </div>
+            </template>
+
+            <span v-if="serverError" class="error-message" data-testid="egress-error">
+              {{ serverError }}
+            </span>
+
+            <div class="form-actions">
+              <button
+                type="button"
+                class="btn btn-secondary"
+                :disabled="saving"
+                @click="cancelEditor"
+              >
+                Cancel
+              </button>
+              <button
+                type="submit"
+                class="btn btn-primary"
+                data-testid="egress-save-button"
+                :disabled="saving"
+              >
+                {{ saving ? 'Saving…' : 'Save' }}
+              </button>
+            </div>
+          </form>
+        </section>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.modal-backdrop {
+  position: fixed;
+  inset: 0;
+  background: rgba(0, 0, 0, 0.5);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  z-index: 1000;
+  padding: 20px;
+}
+
+.modal-container {
+  background: white;
+  border-radius: 12px;
+  box-shadow: 0 20px 50px rgba(0, 0, 0, 0.2);
+  width: 100%;
+  max-width: 640px;
+  max-height: 90vh;
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+}
+
+.modal-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 16px 20px;
+  border-bottom: 1px solid #eee;
+}
+
+.modal-title {
+  margin: 0;
+  font-size: 18px;
+  font-weight: 600;
+  color: #333;
+}
+
+.close-btn {
+  width: 32px;
+  height: 32px;
+  padding: 0;
+  background: transparent;
+  border: none;
+  border-radius: 4px;
+  font-size: 24px;
+  line-height: 1;
+  color: #666;
+  cursor: pointer;
+  transition: background 0.2s, color 0.2s;
+}
+
+.close-btn:hover {
+  background: #f8f9fa;
+  color: #333;
+}
+
+.modal-content {
+  flex: 1;
+  overflow-y: auto;
+  padding: 20px;
+}
+
+.section-head {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 12px;
+}
+
+.section-title {
+  margin: 0;
+  font-size: 15px;
+  font-weight: 600;
+  color: #333;
+}
+
+.muted {
+  color: #888;
+  font-size: 14px;
+  margin: 8px 0;
+}
+
+.key-list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+}
+
+.key-item {
+  border: 1px solid #e0e0e0;
+  border-radius: 8px;
+  padding: 12px;
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+}
+
+.key-main {
+  display: flex;
+  align-items: center;
+  gap: 10px;
+}
+
+.key-host {
+  font-weight: 600;
+  color: #333;
+  word-break: break-all;
+}
+
+.kind-badge {
+  font-size: 12px;
+  padding: 2px 8px;
+  border-radius: 999px;
+  background: #e7f1ff;
+  color: #0b5ed7;
+  white-space: nowrap;
+}
+
+.key-meta {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 4px 12px;
+  font-size: 12px;
+  color: #666;
+}
+
+.indicator {
+  color: #198754;
+}
+
+.key-actions {
+  display: flex;
+  gap: 8px;
+  margin-top: 4px;
+}
+
+.editor-section {
+  margin-top: 24px;
+  padding-top: 20px;
+  border-top: 1px solid #eee;
+}
+
+.editor-form {
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+  margin-top: 12px;
+}
+
+.form-group {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+}
+
+.form-label {
+  font-size: 14px;
+  font-weight: 500;
+  color: #333;
+}
+
+.required {
+  color: #dc3545;
+}
+
+.form-input {
+  padding: 10px 12px;
+  border: 1px solid #ddd;
+  border-radius: 4px;
+  font-size: 14px;
+  font-family: inherit;
+  transition: border-color 0.2s, box-shadow 0.2s;
+}
+
+.form-input:focus {
+  outline: none;
+  border-color: #0d6efd;
+  box-shadow: 0 0 0 2px rgba(13, 110, 253, 0.25);
+}
+
+.form-input.error {
+  border-color: #dc3545;
+}
+
+.header-row {
+  display: flex;
+  gap: 8px;
+  margin-bottom: 8px;
+}
+
+.header-name {
+  flex: 1;
+}
+
+.header-value {
+  flex: 1.4;
+}
+
+.add-header-btn {
+  align-self: flex-start;
+}
+
+.hint {
+  font-size: 12px;
+  color: #888;
+}
+
+.error-message {
+  font-size: 12px;
+  color: #dc3545;
+}
+
+.form-actions {
+  display: flex;
+  justify-content: flex-end;
+  gap: 12px;
+  padding-top: 12px;
+  border-top: 1px solid #eee;
+}
+
+.btn {
+  padding: 10px 20px;
+  border: none;
+  border-radius: 6px;
+  font-size: 14px;
+  font-weight: 500;
+  cursor: pointer;
+  transition: background 0.2s, opacity 0.2s;
+}
+
+.btn:disabled {
+  opacity: 0.6;
+  cursor: not-allowed;
+}
+
+.btn-sm {
+  padding: 6px 12px;
+  font-size: 13px;
+}
+
+.btn-primary {
+  background: #0d6efd;
+  color: white;
+}
+
+.btn-primary:hover:not(:disabled) {
+  background: #0b5ed7;
+}
+
+.btn-secondary {
+  background: #6c757d;
+  color: white;
+}
+
+.btn-secondary:hover:not(:disabled) {
+  background: #5a6268;
+}
+
+.btn-danger {
+  background: #dc3545;
+  color: white;
+}
+
+.btn-danger:hover:not(:disabled) {
+  background: #c82333;
+}
+</style>
--- a/src/LmAgentInfra/Auth/AtomicJsonFile.cs
+++ b/src/LmAgentInfra/Auth/AtomicJsonFile.cs
@@ -28,8 +28,31 @@ internal static class AtomicJsonFile
         }
 
         var json = JsonSerializer.Serialize(value, options);
-        var tempFilePath = filePath + ".tmp";
-        await File.WriteAllTextAsync(tempFilePath, json, Encoding.UTF8, ct).ConfigureAwait(false);
-        File.Move(tempFilePath, filePath, overwrite: true);
+
+        // Unique same-directory staging file (not a fixed "<file>.tmp") so concurrent writers / processes
+        // never collide, and delete it if the write or move fails so a secret-bearing partial file is never
+        // left behind on a cancelled/failed write.
+        var tempFilePath = filePath + "." + Path.GetRandomFileName() + ".tmp";
+        try
+        {
+            await File.WriteAllTextAsync(tempFilePath, json, Encoding.UTF8, ct).ConfigureAwait(false);
+            File.Move(tempFilePath, filePath, overwrite: true);
+        }
+        catch
+        {
+            try
+            {
+                if (File.Exists(tempFilePath))
+                {
+                    File.Delete(tempFilePath);
+                }
+            }
+            catch
+            {
+                // Best-effort cleanup of the staging file; surface the original failure below.
+            }
+
+            throw;
+        }
     }
 }
--- a/src/LmAgentInfra/Auth/AtomicJsonFile.cs
+++ b/src/LmAgentInfra/Auth/AtomicJsonFile.cs
@@ -1,0 +1,35 @@
+using System.Text;
+using System.Text.Json;
+
+namespace AchieveAi.LmDotnetTools.LmAgentInfra.Auth;
+
+/// <summary>
+/// Shared atomic JSON-file write: serialize + stage to a temp file + <see cref="File.Move(string, string, bool)"/>
+/// over the target, so a concurrent reader never observes a partially-written file. Extracted so both
+/// <see cref="FileOAuthTokenStore"/> and <c>PredefinedKeyRegistry</c> use one implementation of the
+/// tricky part. Callers own their own locking (both hold a per-store gate) and their own reads.
+/// </summary>
+internal static class AtomicJsonFile
+{
+    /// <summary>
+    /// Serializes <paramref name="value"/> to <paramref name="filePath"/> atomically, creating the
+    /// parent directory if needed.
+    /// </summary>
+    public static async Task WriteAsync<T>(
+        string filePath,
+        T value,
+        JsonSerializerOptions options,
+        CancellationToken ct = default)
+    {
+        var dir = Path.GetDirectoryName(filePath);
+        if (!string.IsNullOrEmpty(dir))
+        {
+            _ = Directory.CreateDirectory(dir);
+        }
+
+        var json = JsonSerializer.Serialize(value, options);
+        var tempFilePath = filePath + ".tmp";
+        await File.WriteAllTextAsync(tempFilePath, json, Encoding.UTF8, ct).ConfigureAwait(false);
+        File.Move(tempFilePath, filePath, overwrite: true);
+    }
+}
--- a/src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
+++ b/src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
@@ -1,0 +1,68 @@
+using System.Net.Http.Headers;
+using System.Text.Json;
+
+namespace AchieveAi.LmDotnetTools.LmAgentInfra.Auth;
+
+/// <summary>Parsed OAuth2 token-endpoint response — a success token set or an OAuth <c>error</c>.</summary>
+/// <param name="AccessToken">The <c>access_token</c>, or null on error / absent.</param>
+/// <param name="RefreshToken">A rotated <c>refresh_token</c>, or null when the endpoint did not return one.</param>
+/// <param name="ExpiresIn">The <c>expires_in</c> lifetime in seconds, or 0 when absent.</param>
+/// <param name="Error">The OAuth <c>error</c> code (or a synthetic transport error), or null on success.</param>
+internal sealed record OAuthTokenEndpointResponse(string? AccessToken, string? RefreshToken, int ExpiresIn, string? Error);
+
+/// <summary>
+/// Minimal OAuth2 token-endpoint client: form-POSTs a grant to a configurable token endpoint
+/// (<c>Accept: application/json</c>) and parses the standard <c>access_token</c>/<c>refresh_token</c>/
+/// <c>expires_in</c>/<c>error</c> response. Modeled on <c>GitHubOAuthProvider.PostTokenAsync</c> but
+/// standalone and endpoint-agnostic so <c>PredefinedKeyProvider</c> can drive the
+/// <c>refresh_token</c> and <c>client_credentials</c> grants against an arbitrary provider, and so it
+/// is unit-testable against a mock <see cref="HttpMessageHandler"/>.
+/// </summary>
+/// <remarks>SECURITY: never logs the form (it carries the client secret / refresh token) or the response body.</remarks>
+internal sealed class OAuthTokenEndpointClient(HttpClient http)
+{
+    private readonly HttpClient _http = http ?? throw new ArgumentNullException(nameof(http));
+
+    /// <summary>
+    /// POSTs <paramref name="form"/> to <paramref name="tokenEndpoint"/> and parses the token/error
+    /// response. Never throws for a non-success status or an unparseable body — those surface as an
+    /// <see cref="OAuthTokenEndpointResponse.Error"/> so the caller decides whether to deny/invalidate.
+    /// </summary>
+    public async Task<OAuthTokenEndpointResponse> PostAsync(
+        string tokenEndpoint,
+        IReadOnlyDictionary<string, string> form,
+        CancellationToken ct = default)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(tokenEndpoint);
+        ArgumentNullException.ThrowIfNull(form);
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, tokenEndpoint)
+        {
+            Content = new FormUrlEncodedContent(form),
+        };
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+        using var response = await _http.SendAsync(request, ct).ConfigureAwait(false);
+        var body = await response.Content.ReadAsStringAsync(ct).ConfigureAwait(false);
+        if (string.IsNullOrWhiteSpace(body))
+        {
+            return new OAuthTokenEndpointResponse(null, null, 0, "empty_response");
+        }
+
+        try
+        {
+            using var doc = JsonDocument.Parse(body);
+            var root = doc.RootElement;
+            return new OAuthTokenEndpointResponse(
+                root.TryGetProperty("access_token", out var at) ? at.GetString() : null,
+                root.TryGetProperty("refresh_token", out var rt) ? rt.GetString() : null,
+                root.TryGetProperty("expires_in", out var ei) && ei.TryGetInt32(out var seconds) ? seconds : 0,
+                root.TryGetProperty("error", out var er) ? er.GetString() : null);
+        }
+        catch (JsonException)
+        {
+            // Non-JSON body (HTML error page, etc.). Surface as a token-free error; never log the body.
+            return new OAuthTokenEndpointResponse(null, null, 0, "unparseable_response");
+        }
+    }
+}
--- a/src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
+++ b/src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
@@ -44,25 +44,43 @@ internal sealed class OAuthTokenEndpointClient(HttpClient http)
 
         using var response = await _http.SendAsync(request, ct).ConfigureAwait(false);
         var body = await response.Content.ReadAsStringAsync(ct).ConfigureAwait(false);
+
+        // Best-effort parse of the standard OAuth token/error fields (a structured `error` may accompany
+        // either a success or a 4xx credential rejection).
+        string? accessToken = null;
+        string? refreshToken = null;
+        var expiresIn = 0;
+        string? error = null;
         if (string.IsNullOrWhiteSpace(body))
         {
-            return new OAuthTokenEndpointResponse(null, null, 0, "empty_response");
+            error = "empty_response";
+        }
+        else
+        {
+            try
+            {
+                using var doc = JsonDocument.Parse(body);
+                var root = doc.RootElement;
+                accessToken = root.TryGetProperty("access_token", out var at) ? at.GetString() : null;
+                refreshToken = root.TryGetProperty("refresh_token", out var rt) ? rt.GetString() : null;
+                expiresIn = root.TryGetProperty("expires_in", out var ei) && ei.TryGetInt32(out var seconds) ? seconds : 0;
+                error = root.TryGetProperty("error", out var er) ? er.GetString() : null;
+            }
+            catch (JsonException)
+            {
+                // Non-JSON body (HTML error page, etc.). Surface as a token-free error; never log the body.
+                error = "unparseable_response";
+            }
         }
 
-        try
+        // A non-success status must NEVER yield a token: return the parsed OAuth `error` when present,
+        // else a normalized `http_<status>` so the caller can tell a transient 429/5xx from a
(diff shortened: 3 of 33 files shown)


samples/LmStreaming.Sample/ClientApp/src/components/EgressAuthModal.vue
```diff
@@ -0,0 +1,751 @@
...
+  host.value = view.host;
+  kind.value = view.kind;
+  headerName.value = view.headerName || 'Authorization';
+  scopesText.value = view.scopes.join(' ');
+  if (view.kind === 'custom-headers') {
+    // Prefill header NAMES only; values are masked, so leave them blank to keep
+    // the stored secret on update.
+    headerRows.value =
+      view.headerNames.length > 0
+        ? view.headerNames.map((n) => ({ name: n, value: '' }))
```
gautam-achieveai, line 76: Must — Editing an existing custom-header key sends every masked header as `{ name, value: "" }`. The server treats this non-empty list as a replacement and rejects each blank value in `EgressKeysController` (lines 80-101), so changing only the host or opening/saving an entry fails instead of preserving the secret. Use explicit per-header patch semantics (omit unchanged values or merge blank values with the stored value) and add a backend round-trip test.

src/LmAgentInfra/Controllers/EgressKeysController.cs
```diff
@@ -0,0 +1,232 @@
...
+[Route("api/auth/egress-keys")]
+public sealed class EgressKeysController(PredefinedKeyRegistry registry, ILogger<EgressKeysController> logger)
+    : ControllerBase
+{
+    /// <summary>Lists the configured entries with all secret material masked/omitted.</summary>
+    [HttpGet]
+    public IActionResult List() => Ok(registry.Entries.Select(ToView).ToList());
+
+    /// <summary>Creates a new entry (blank id) or updates an existing one (secrets left blank are preserved).</summary>
+    [HttpPost]
```
gautam-achieveai, line 37: Must — This introduces unauthenticated state-changing credential CRUD. `MapControllers()` exposes this route, but the controller has no authorization or local-only enforcement; the comment that it is "same-origin/local-UI only" is not an access-control boundary. Any reachable caller can enumerate metadata and create, update, or delete credentials (including retargeting stored secrets). Require an enforced management authorization/CSRF boundary or technically bind/restrict the endpoint to the intended local interface.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /// <summary>
+    /// Swaps the entry in place (edit), clears the minted-token cache, and clears the invalidated
+    /// flag so the next <see cref="GetAccessTokenAsync"/> re-mints / re-reads the updated credential.
+    /// </summary>
+    public void UpdateEntry(PredefinedKeyEntry entry)
+    {
+        ArgumentNullException.ThrowIfNull(entry);
+        _entry = entry;
```
gautam-achieveai: Must — `UpdateEntry` clears only `_cachedToken`; it does not invalidate/version the persisted `OAuthTokenRecord`, and it is not synchronized with `_refreshGate`. The next acquisition can reload the old access/rotated refresh token under the unchanged provider ID, while an in-flight mint can save a token made with the previous credential after this update. Serialize updates through the same gate and remove or version the persisted token before publishing the new entry.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <summary>The current entry for <paramref name="id"/>, or null when unknown (for edit-preserves-secret merges).</summary>
+    internal PredefinedKeyEntry? Find(string id) =>
+        _providers.TryGetValue(id, out var provider) ? provider.Entry : null;
+
+    /// <summary>Creates a new entry or updates an existing one in place (same id), then persists.</summary>
+    internal async Task UpsertAsync(PredefinedKeyEntry entry, CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(entry);
+
+        if (_providers.TryGetValue(entry.Id, out var existing))
```
gautam-achieveai: Must — Runtime registry state is changed before `PersistAsync` succeeds. A failed or cancelled write can return an error while the new provider is already active in memory; delete has the inverse split and can reappear after restart. Also, the persistence gate does not serialize the logical mutation with snapshot creation. Build and persist a candidate snapshot under one mutation gate, then publish it only after the write succeeds (and define token-cleanup failure handling for delete).

src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
```diff
@@ -0,0 +1,68 @@
...
+        ArgumentException.ThrowIfNullOrWhiteSpace(tokenEndpoint);
+        ArgumentNullException.ThrowIfNull(form);
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, tokenEndpoint)
+        {
+            Content = new FormUrlEncodedContent(form),
+        };
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+        using var response = await _http.SendAsync(request, ct).ConfigureAwait(false);
```
gautam-achieveai, line 45: Should — The client ignores `response.IsSuccessStatusCode` and accepts an `access_token` from any JSON response, without preserving the HTTP status for the provider's error classification. Require a successful status before accepting a token and carry a normalized status/category for non-success responses so transient 429/5xx responses are not handled like credential updates.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
```diff
@@ -238,9 +240,32 @@ async Task TryNotifyAuthDeniedAsync(AuthWebhookTarget? target, string providerId
...
     }
 
-    /// <summary>Resolves a registered provider by the route id, matching <see cref="IOAuthTokenProvider.ProviderId"/> case-insensitively.</summary>
+    /// <summary>
+    /// Resolves a registered OAuth provider by the route id (case-insensitive), falling back to a
+    /// runtime predefined-key provider (<c>predefined-&lt;id&gt;</c>) from the registry.
+    /// </summary>
     private IOAuthTokenProvider? Resolve(string provider) =>
-        providers.FirstOrDefault(p => string.Equals(p.ProviderId, provider, StringComparison.OrdinalIgnoreCase));
+        providers.FirstOrDefault(p => string.Equals(p.ProviderId, provider, StringComparison.OrdinalIgnoreCase))
```
gautam-achieveai, line 261: Should — The nested best-effort forwarder catches `Exception`, which also catches request `OperationCanceledException`. This defeats the explicit cancellation-preserving catches in the action and can continue policy/notification work after the gateway disconnects. Add `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` before the broad catch in all three notification helpers.

tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryBuildAuthProvidersTests.cs
```diff
@@ -92,7 +92,52 @@ public async Task No_providers_emitted_when_nothing_configured()
...
+            await keys.UpsertAsync(new PredefinedKeyEntry
+            {
+                Id = "e1",
+                Host = "api.internal.example.com",
+                Kind = PredefinedKeyKind.CustomHeaders,
+                Headers = [new PredefinedHeader("X-Key", "v")],
+            });
+
+            await using var registry = CreateRegistry(new AuthOptions(), keys);
+
```
gautam-achieveai, line 111: Should — This test asserts only the predefined provider ID, but the production change also creates the security-sensitive host-scoped network rule and webhook payload. Capture/assert the serialized pair: exact entry host, `ports: [443]`, `authProvider` linkage, webhook endpoint, and custom-header cache TTL. Otherwise a detached or overbroad rule can regress while this test remains green.

gautam-achieveai (commented):
Reviewed PR 212 across 32 files — 4 Must, 3 Should, 0 Consider.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
```diff
@@ -238,9 +240,32 @@ async Task TryNotifyAuthDeniedAsync(AuthWebhookTarget? target, string providerId
...
+        ?? predefinedKeys?.TryResolve(provider);
```
gautam-achieveai, line 262: [Gautam's review bot] **P2 — Existing sessions are coupled to mutable registry state.** BuildAuthProviders snapshots this stable predefined-<id> and the entry host only when the session is created, but every later webhook callback resolves that ID through the current global registry here. After the gateway cache expires, deleting the entry makes an already-running session deny requests; editing a same-host credential changes what that session injects; editing the host makes the old frozen rule fail while the new host is absent from that session's network rules. This contradicts #210 and the guide's promise that existing sessions are unchanged. Version/snapshot provider IDs per session (or retain immutable provider versions for active sessions) so CRUD only affects subsequently created sessions.

gautam-achieveai (commented):
[Gautam's review bot] Deep review completed. Added one non-duplicate finding; the seven existing inline findings were not repeated.

samples/LmStreaming.Sample/ClientApp/src/components/EgressAuthModal.vue
gautam-achieveai: Fixed in 0ab2916f. `EgressKeysController.BuildCustomHeadersEntry` now does per-header preserve: a blank value on an **existing** header name keeps the stored secret (mirrors the OAuth secret-preserve), a blank value on a **new** header errors, and omitting the header list entirely preserves all headers. Added round-trip tests `Update_custom_headers_with_blank_value_preserves_stored_value` and `Post_rejects_blank_value_for_a_new_header`.

src/LmAgentInfra/Controllers/EgressKeysController.cs
gautam-achieveai: By design for this local-dev sample, not an oversight — flagging as won't-fix within #210 (documented). The whole sample is unauthenticated: the OAuth sign-in controllers (`GitHubAuthController`/`AdoAuthController`/`M365AuthController`) and the workspaces/providers/conversations endpoints, and the guide states the backend runs on the user's own machine (Kestrel binds localhost). Two real boundaries apply: (1) the app host is **unreachable from the sandbox** — default-deny egress opens no allow-rule for `127.0.0.1:5000`, so a sandboxed agent cannot call this API; (2) the localhost bind. Adding enforced auth to only this endpoint would be inconsistent with the rest of the sample. I reworded the class comment (0ab2916f) to state this threat model explicitly rather than imply an access-control guarantee. A management-auth boundary for the sample as a whole is a separate hardening item.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
gautam-achieveai: Fixed in 0ab2916f. `UpdateEntry` is now async and runs under `_refreshGate` (serialized with any in-flight mint) and removes the persisted `OAuthTokenRecord` for the provider id — so an edited credential re-mints on the next acquisition instead of reloading the stale access/rotated-refresh token, and a mint from the pre-edit credential can't publish afterward. `PredefinedKeyRegistry.UpsertAsync` awaits it.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Fixed in 0ab2916f. `UpsertAsync`/`RemoveAsync` now build a candidate entry-set, persist it **first**, and publish the in-memory change only after the write succeeds — all under one `_persistGate`. A failed/cancelled write no longer diverges memory from disk, and a deleted entry can't reappear after restart.

src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
gautam-achieveai: Fixed in 0ab2916f. `PostAsync` now rejects any token from a non-success status and returns the parsed OAuth `error` when present, else a normalized `http_<status>` — so a transient 429/5xx is distinguishable from a definitive credential rejection (only `invalid_grant`-family codes invalidate the key). Added `PostAsync_rejects_token_from_non_success_status` and `PostAsync_surfaces_oauth_error_on_4xx`.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
gautam-achieveai: Fixed in 0ab2916f. Added `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` before the broad catch in all three `TryNotify*` forwarder helpers, so a gateway disconnect propagates instead of continuing notification work. (These helpers pre-date #210, but the fix is small and correct.)

tests/LmStreaming.Sample.Tests/Services/SandboxSessionRegistryBuildAuthProvidersTests.cs
gautam-achieveai: Fixed in 0ab2916f. Added a `BuildAuthProvidersForTest()` seam; the test now asserts the full pair for a predefined entry — provider (`type: webhook`, endpoint `…/api/auth/webhook/predefined-e1`, custom-headers `cacheTtlSeconds: 30`) and rule (`action: allow`, `hosts: [entry host]`, `ports: [443]`, `authProvider: predefined-e1`).

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
gautam-achieveai: [Gautam's review bot] **P1 — Preserve the rotated refresh token on credential-preserving edits.** `MintAsync` stores a server-rotated refresh token only in the token record, while a masked edit omits `refreshToken` and preserves the entry's original value. Unconditionally removing this record therefore discards the only current refresh token; after a host/header-only edit, the next mint retries the already-consumed original token and can fail with `invalid_grant`. Only discard the stored refresh token when the user explicitly replaces that credential; otherwise invalidate the access token while retaining the rotated refresh token (or require explicit re-entry for incompatible edits). Add a test that mints a rotated refresh token, performs a non-secret edit, and verifies the next mint uses the rotated token.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `4f0fc44a...0ab2916f` completed. Five prior fixes were verified and resolved; the local-dev management-endpoint concern was resolved as an accepted threat-model decision. Two blocking threads remain active: the existing-session registry coupling, and a new P1 regression where credential-preserving edits discard the only rotated refresh token. The predefined deferred-auth UI routing concern was investigated and dropped because the client branches on the `predefined-` provider id instead of navigating the OAuth sign-in URL.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
```diff
@@ -0,0 +1,320 @@
...
+            _ = _refreshGate.Release();
+        }
+    }
+
+    /// <inheritdoc />
+    public async Task<OAuthAccessToken> GetAccessTokenAsync(
+        IReadOnlyList<string>? scopes = null,
+        CancellationToken ct = default)
+    {
+        var entry = _entry;
```
gautam-achieveai: Must — `_entry` is captured before `_refreshGate` is acquired. A request can capture the old entry, wait while `UpdateEntry` completes, then acquire the gate and mint/persist a token using the superseded credential. Read `_entry` only after acquiring the gate (and keep all kind/invalidation checks on that locked snapshot).

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        await _persistGate.WaitAsync(ct).ConfigureAwait(false);
+        try
+        {
+            var candidate = _providers.Values
+                .Select(p => p.Entry)
+                .Where(e => !string.Equals(e.Id, entry.Id, StringComparison.Ordinal))
+                .Append(entry)
+                .ToList();
+            await AtomicJsonFile.WriteAsync(_filePath, candidate, JsonOptions, ct).ConfigureAwait(false);
```
gautam-achieveai: Must — The new definition is durably written before the old token record is invalidated, while `UpdateEntry` publishes `_entry` before its fallible `RemoveAsync`. If token removal fails or cancellation arrives, the new definition is active/durable but the old OAuthTokenRecord remains reloadable under the unchanged provider id, including after restart. Version the definition/token record or complete invalidation before publication with a fail-safe commit/reconciliation strategy.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,167 @@
...
+            if (!_providers.TryGetValue(id, out var provider))
+            {
+                return false;
+            }
+
+            var candidate = _providers.Values
+                .Select(p => p.Entry)
+                .Where(e => !string.Equals(e.Id, id, StringComparison.Ordinal))
+                .ToList();
+            await AtomicJsonFile.WriteAsync(_filePath, candidate, JsonOptions, ct).ConfigureAwait(false);
```
gautam-achieveai, line 166: Must — Delete commits the definition removal and removes the provider before `SignOutAsync(ct)`. Cleanup failure/cancellation therefore returns an error after deletion already committed and can leave the minted token on disk. Make cleanup genuinely best-effort with non-request cancellation and retry/logging, or complete cleanup before committing deletion.

src/LmAgentInfra/Controllers/EgressKeysController.cs
```diff
@@ -0,0 +1,253 @@
...
+/// </summary>
+/// <remarks>
+/// SECURITY / THREAT MODEL: like the rest of this local-dev sample (the OAuth sign-in controllers, the
+/// workspaces / providers / conversations endpoints — all unauthenticated) this management API assumes a
+/// trusted local user on a localhost-bound host; it is not a hardened, network-exposed service. It is
+/// deliberately NOT reachable from the sandbox: default-deny egress opens no allow-rule for the app's own
+/// host, so a sandboxed agent cannot call it. GET masks all secret values and nothing here logs them.
+/// An enforced management-auth boundary for the sample as a whole is out of scope for #210.
+/// </remarks>
+[ApiController]
```
gautam-achieveai, line 19: Must — This credential-management CRUD route has no enforced authorization or loopback-only boundary; the surrounding localhost/trusted-user XML comment is not an access control. If the app is reachable beyond the local user, an unauthenticated caller can create, retarget, replace, or delete reusable secrets and widen future sandbox egress. Enforce management authentication/authorization or reject non-loopback callers in code/deployment.

gautam-achieveai (commented):
Re-review of the 4f0fc44..0ab2916 delta: prior fixes are present, but five blocking correctness/security issues remain.

src/LmAgentInfra/Auth/AtomicJsonFile.cs
```diff
@@ -0,0 +1,35 @@
...
+        JsonSerializerOptions options,
+        CancellationToken ct = default)
+    {
+        var dir = Path.GetDirectoryName(filePath);
+        if (!string.IsNullOrEmpty(dir))
+        {
+            _ = Directory.CreateDirectory(dir);
+        }
+
+        var json = JsonSerializer.Serialize(value, options);
```
gautam-achieveai, line 30: Must — Secret-bearing JSON is staged at a predictable `.tmp` path with inherited permissions and no cleanup when write/move is cancelled or fails. The same fixed temp path is also unsafe across processes. Use a unique same-directory temp file, owner-only permissions/ACLs for secret stores, and delete the staging file in `finally` unless replacement succeeds.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
```diff
@@ -62,8 +63,9 @@ public async Task<IActionResult> Evaluate(
 
         // Defense-in-depth: the egress proxy already validates the destination per its rules, but a
         // rules misconfiguration must not turn this endpoint into an open token-minting oracle —
-        // only inject the provider's token toward that provider's own hosts.
-        if (!OAuthProviderHosts.IsAllowed(tokenProvider.ProviderId, body.DestinationHost))
+        // only inject the credential toward that provider's/entry's own hosts. Predefined keys carry
+        // their own (user-entered) host, so they gate on the entry's host, not the managed OAuth list.
+        if (!IsDestinationAllowed(tokenProvider, body.DestinationHost))
```
gautam-achieveai: Must — The predefined-key defense-in-depth gate validates only the destination hostname and ignores `body.DestinationPort`, although the generated rule and documented boundary are HTTPS/443-only. A malformed or misbehaving gateway can request the same host on another port and still receive the credential. Require port 443 for predefined providers and add a regression test.

src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            error = "empty_response";
+        }
+        else
+        {
+            try
+            {
+                using var doc = JsonDocument.Parse(body);
+                var root = doc.RootElement;
+                accessToken = root.TryGetProperty("access_token", out var at) ? at.GetString() : null;
```
gautam-achieveai, line 64: Should — `GetString()` and similar accessors can throw `InvalidOperationException` for valid JSON with unexpected field types (for example `{'access_token':{}}` or `[]`), but only `JsonException` is normalized. Check `JsonValueKind` and return a fixed `invalid_response`/`unparseable_response` error so malformed endpoint responses do not escape the documented result contract.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1536,6 +1542,14 @@ internal static string StripLongPathPrefix(string path)
     internal IReadOnlyList<string> GetAuthProviderIdsForTest() =>
         BuildAuthProviders().Providers?.Select(p => p.Id).ToArray() ?? [];
 
+    /// <summary>
+    /// Test seam: the full <c>auth_providers</c> + <c>network</c> pair the registry would attach, so a
+    /// test can assert the security-sensitive rule shape (host, ports, auth-provider linkage) and the
+    /// webhook endpoint / cache TTL — not just the provider id.
+    /// </summary>
+    internal (IReadOnlyList<SandboxAuthProvider>? Providers, IReadOnlyList<SandboxNetworkRule>? Network) BuildAuthProvidersForTest() =>
```
gautam-achieveai, line 1787: Consider — This is a production method explicitly named/documented solely as a test seam, alongside another `GetAuthProviderIdsForTest` wrapper. Prefer making the builder an internal testable component or using one existing internal seam rather than adding test-only API surface.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyRegistryTests.cs
```diff
@@ -0,0 +1,110 @@
...
+    private static PredefinedKeyEntry Custom(string id, string host = "api.example.com") => new()
+    {
+        Id = id,
+        Host = host,
+        Kind = PredefinedKeyKind.CustomHeaders,
+        Headers = [new PredefinedHeader("X-Key", "v")],
+    };
+
+    [Fact]
+    public async Task Upsert_creates_resolves_and_persists_across_reload()
```
gautam-achieveai, line 71: Should — The changed persistence/update behavior lacks regression coverage for write failure/cancellation, token-store cleanup failure, stale-token reuse after edit, and delete cleanup failure. Add tests that force each boundary to fail and assert the durable definition, live provider, and token record remain fail-safe and consistent.

gautam-achieveai (commented):
Re-review of the latest delta found two remaining security/correctness blockers.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,198 @@
...
+                .Where(e => !string.Equals(e.Id, entry.Id, StringComparison.Ordinal))
+                .Append(entry)
+                .ToList();
+
+            // Invalidate the stale minted token BEFORE the new definition is made durable, so a failed
+            // persist can never leave the new definition on disk while the old token stays reloadable under
+            // the unchanged provider id. Cleanup is best-effort (see TryRemovePersistedTokenAsync).
+            if (existing is not null && credentialChanged)
+            {
+                await TryRemovePersistedTokenAsync(entry.Id).ConfigureAwait(false);
```
gautam-achieveai: Must — token invalidation is still best-effort before the credential update is committed. If `IOAuthTokenStore.RemoveAsync` fails, the new definition is persisted and the live provider is updated while the old access/rotated-refresh token remains under the same `predefined-{id}` key; after restart `LoadCachedFromStoreAsync` can reuse it with the new credential. Make invalidation durable/transactional (or fail the update and retain the old definition), and prevent startup from loading a token tied to a prior credential version.

src/LmAgentInfra/Controllers/EgressKeysController.cs
```diff
@@ -0,0 +1,290 @@
...
+    /// Enforces the local-only boundary: rejects a caller whose remote address is a real, non-loopback IP
+    /// (returns 403). A null remote address — the in-process TestServer and same-host pipelines — is treated
+    /// as local. Returns null when the caller is allowed.
+    /// </summary>
+    private IActionResult? RejectNonLoopback()
+    {
+        // A null HttpContext/remote (in-process TestServer, direct unit construction) is treated as local;
+        // only a real, non-loopback remote IP is rejected.
+        var remote = HttpContext?.Connection?.RemoteIpAddress;
+        if (remote is not null && !System.Net.IPAddress.IsLoopback(remote))
```
gautam-achieveai: Must — the management boundary trusts only `RemoteIpAddress`. Behind a reverse proxy/load balancer, a remote caller can commonly appear to the app as the proxy's loopback address, bypassing this loopback-only check despite the documented guarantee. Bind this route to a loopback-only network path, or configure trusted-proxy forwarding and validate the original client address; do not treat an untrusted loopback peer as proof of a local caller.

src/LmAgentInfra/Controllers/EgressKeysController.cs
gautam-achieveai: Addressed in aa424003. Added enforced loopback-only access **in code** (`RejectNonLoopback`): a real non-loopback remote IP is rejected `403` on all three actions, so this is now an enforced boundary rather than a comment (a null remote — in-process/TestServer — is treated as local). Added a non-loopback-403 test. A full management-auth boundary for the whole sample remains out of scope for #210, consistent with the other unauthenticated sample endpoints.

src/LmAgentInfra/Auth/AtomicJsonFile.cs
gautam-achieveai: Fixed in aa424003. `AtomicJsonFile.WriteAsync` now stages to a **unique** same-directory file (`<file>.<random>.tmp`) and deletes the staging file if the write/move fails — no fixed-path collision across writers/processes, and no secret-bearing partial file left behind. (Owner-only ACLs: the OAuth token store already writes the same plaintext secrets to the same gitignored dir, and the guide documents plaintext-at-rest as acceptable for local dev; ACL hardening for the whole store is a separate item.)

src/LmAgentInfra/Auth/OAuthTokenEndpointClient.cs
gautam-achieveai: Fixed in aa424003. The parse now catches `InvalidOperationException` (e.g. `{"access_token":{}}`) alongside `JsonException`, normalizing both to `unparseable_response`. Added `PostAsync_handles_unexpected_json_field_types`.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: Acknowledged. `BuildAuthProvidersForTest` mirrors the pre-existing `GetAuthProviderIdsForTest` seam on the same class (that pattern predates #210). Extracting the builder into a standalone internal component is a reasonable refactor but broader than this PR — keeping the minimal seam for now.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
gautam-achieveai: Fixed in aa424003. `GetAccessTokenAsync` now acquires `_refreshGate` first and reads `_entry` + the invalidation flag entirely under it, so the kind/credential decision and the mint use one locked snapshot — a concurrent (also-gated) `UpdateEntry` can no longer be interleaved to make a mint use a superseded credential.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Fixed in aa424003. `UpsertAsync` computes credential-change vs the existing entry and, when the credential changed, invalidates the stale persisted token **before** the new definition is written. `UpdateEntry` is now credential-change-aware, so a pure host/header edit keeps the rotated refresh token (this also fixes the P1 rotation regression). Full cross-file definition+token atomicity would need a single record / txn log — out of scope for this local-dev sample; the ordering + best-effort cleanup degrade safely.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Fixed in aa424003. Token cleanup now goes through `TryRemovePersistedTokenAsync`: a non-request `CancellationToken.None` so a disconnecting caller can't strand it, and it swallows + logs failures (never the secret) so a cleanup error never fails the already-committed deletion. A leftover token is harmless — unreachable without its removed definition and overwritten on the next mint. Added a cleanup-failure regression test.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
gautam-achieveai: Fixed in aa424003. The predefined-key defense-in-depth gate now requires `destination_port == 443` in addition to the host match. Added an E2E regression (same host, port 8443 → deny).

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyRegistryTests.cs
gautam-achieveai: Added in aa424003. `PredefinedKeyRegistryTests` now covers: credential-change invalidates the persisted token, a host-only edit **retains** the rotated token, delete cleans up the token, and a token-store cleanup failure does not fail upsert/remove.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
gautam-achieveai: By design. The network **rule** (host reachability) is frozen per session at create time, but injection resolves through the live registry — intentional and safe: editing a credential takes effect immediately for running sessions, and deleting a key makes the webhook fail-**closed** (deny) rather than inject a removed secret. The "applies to new sessions" note refers to the host allow-set; live injection is deliberate. No code change.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Must — Credential-change invalidation is still best-effort: `TryRemovePersistedTokenAsync` swallows token-store failures, then line 131 persists the new definition and line 135 updates the live provider. If removal fails, the old access/rotated-refresh token remains under the same `predefined-{id}` key and can be loaded after restart by the new credential definition. Make invalidation durable/transactional (or fail the update and retain the old definition), and version/bind persisted tokens so startup cannot reuse material from a prior credential.

src/LmAgentInfra/Controllers/EgressKeysController.cs
```diff
@@ -0,0 +1,290 @@
...
+        var removed = await registry.RemoveAsync(id, ct).ConfigureAwait(false);
+        return removed ? NoContent() : NotFound();
+    }
+
+    /// <summary>
+    /// Enforces the local-only boundary: rejects a caller whose remote address is a real, non-loopback IP
+    /// (returns 403). A null remote address — the in-process TestServer and same-host pipelines — is treated
+    /// as local. Returns null when the caller is allowed.
+    /// </summary>
+    private IActionResult? RejectNonLoopback()
```
gautam-achieveai, line 107: Must — The management boundary still trusts only `RemoteIpAddress`. Behind a reverse proxy/load balancer, a remote caller can appear to the application as a loopback proxy address, satisfying this check and exposing secret CRUD. Bind this route to a loopback-only listener/network path, or configure trusted-proxy forwarding and validate the original client address; do not treat an untrusted loopback peer as proof of a local caller.

gautam-achieveai (commented):
Re-reviewed PR 212 across 33 files — 2 Must, 0 Should, 0 Consider.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
gautam-achieveai: [Gautam's review bot] Verified fixed at `70f0309d`. Credential-preserving edits now retain both the cached token and the persisted rotated refresh token, while credential-changing edits invalidate them. The host-only edit regression coverage confirms the rotated token survives. Resolving this thread.

src/LmAgentInfra/Controllers/AuthWebhookController.cs
gautam-achieveai: [Gautam's review bot] Accepted as intentional behavior. Session host reachability remains frozen, while live credential resolution applies updates immediately and fails closed after deletion or a host change. Resolving this thread.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `0ab2916f...70f0309d` found no new non-duplicate defects. The rotated-refresh-token P1 is fixed, and the existing-session behavior is accepted as an intentional fail-closed/live-injection trade-off.

Two already-open Must threads remain active rather than being duplicated: durable credential-change invalidation must prevent an old token from being reloaded after cleanup failure/restart, and loopback-only CRUD must not treat a local reverse proxy as proof that the original caller is local. The merge commit's unrelated `main` changes were excluded from the feature review.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Made reliable in 95eb036b. The credential-change invalidation now runs `_tokenStore.RemoveAsync(..., CancellationToken.None)` **before** the new definition is persisted and does **not** swallow failures — a removal failure aborts the whole upsert (nothing committed or published), so the new definition can never go live with the old access/rotated-refresh token reloadable under the same provider id. Delete cleanup stays best-effort (a deleted definition's orphan token is unreachable). Test: `Credential_change_aborts_when_token_invalidation_fails_but_delete_stays_best_effort`.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Made reliable in 95eb036b. The credential-change invalidation now runs `_tokenStore.RemoveAsync(..., CancellationToken.None)` **before** the new definition is persisted and does **not** swallow failures — a removal failure aborts the whole upsert (nothing committed or published), so the new definition can never go live with the old access/rotated-refresh token reloadable under the same provider id. Delete cleanup stays best-effort (a deleted definition's orphan token is unreachable). Test: `Credential_change_aborts_when_token_invalidation_fails_but_delete_stays_best_effort`.

src/LmAgentInfra/Controllers/EgressKeysController.cs
gautam-achieveai: Hardened in 95eb036b. `RejectNonLoopback` now also rejects (403) any request carrying a forwarding header (`X-Forwarded-For` / `X-Forwarded-Host` / `Forwarded`), so a request proxied through a loopback reverse proxy can no longer satisfy the IP check. Combined with the sample's localhost Kestrel bind and the `RemoteIpAddress` loopback check, direct-local-only is enforced. Test: `Proxied_request_with_forwarding_header_is_rejected_403`. (Full management auth for the whole sample stays out of scope for #210 — every sample endpoint is unauthenticated local-dev.)

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Made reliable in 95eb036b. The credential-change invalidation now runs `_tokenStore.RemoveAsync(..., CancellationToken.None)` **before** the new definition is persisted and does **not** swallow failures — a removal failure aborts the whole upsert (nothing committed or published), so the new definition can never go live with the old access/rotated-refresh token reloadable under the same provider id. Delete cleanup stays best-effort (a deleted definition's orphan token is unreachable). Test: `Credential_change_aborts_when_token_invalidation_fails_but_delete_stays_best_effort`.

src/LmAgentInfra/Controllers/EgressKeysController.cs
gautam-achieveai: Hardened in 95eb036b. `RejectNonLoopback` now also rejects (403) any request carrying a forwarding header (`X-Forwarded-For` / `X-Forwarded-Host` / `Forwarded`), so a request proxied through a loopback reverse proxy can no longer satisfy the IP check. Combined with the sample's localhost Kestrel bind and the `RemoteIpAddress` loopback check, direct-local-only is enforced. Test: `Proxied_request_with_forwarding_header_is_rejected_403`. (Full management auth for the whole sample stays out of scope for #210 — every sample endpoint is unauthenticated local-dev.)

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,200 @@
...
+                .Where(e => !string.Equals(e.Id, entry.Id, StringComparison.Ordinal))
+                .Append(entry)
+                .ToList();
+
+            // Invalidate the stale minted token BEFORE the new definition is made durable. This is a
+            // RELIABLE (not best-effort) step: if removal fails the whole upsert throws before anything is
+            // persisted or published, so we never leave the new definition live with the old access/rotated
+            // token reloadable under the unchanged provider id. CancellationToken.None so a disconnecting
+            // caller can't strand it mid-invalidation.
+            if (existing is not null && credentialChanged)
```
gautam-achieveai: Must — Removing the persisted token before `UpdateEntry` acquires the provider's refresh gate is not sufficient to invalidate it reliably. A concurrent `GetAccessTokenAsync` can already hold that gate, mint with the old entry, and call `SaveAsync` after this removal; `UpdateEntry` then clears only its in-memory cache, leaving the old-credential token persisted under the unchanged `predefined-{id}` key and reloadable after restart. Coordinate token removal and provider mutation through the same provider gate (or add a mutation operation that atomically blocks minting, invalidates the record, and swaps the entry), and add a deterministic concurrent mint/update regression test.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                await _tokenStore.RemoveAsync($"{ProviderIdPrefix}{entry.Id}", CancellationToken.None).ConfigureAwait(false);
+            }
+
+            await AtomicJsonFile.WriteAsync(_filePath, candidate, JsonOptions, ct).ConfigureAwait(false);
+
+            if (existing is not null)
+            {
+                await existing.UpdateEntry(entry, credentialChanged, ct).ConfigureAwait(false);
```
gautam-achieveai: Must — The durable JSON write completes before `UpdateEntry` is awaited with the request cancellation token. If the request is cancelled while waiting for the provider gate, the new definition is on disk but the live provider still has the old definition/cache, so the operation reports cancellation while memory and disk disagree until restart. Once persistence commits, complete the in-memory publication with a non-cancelable/internal commit path, or redesign the transaction so cancellation cannot split durable and live state; add a cancellation-after-persist test.

gautam-achieveai (commented):
Re-reviewed PR #212 across 4 delta files — 2 Must, 0 Should, 0 Consider. The new persistence ordering still permits credential/token state divergence under concurrency and cancellation.

src/LmAgentInfra/Controllers/EgressKeysController.cs
```diff
@@ -0,0 +1,299 @@
...
+    /// Enforces the local-only boundary. Rejects (403) a caller whose remote address is a real, non-loopback
+    /// IP, OR any request that arrived through a proxy (carries a forwarding header) — behind a reverse proxy
+    /// a remote caller can appear to the app as the proxy's loopback address, so the IP check alone is not
+    /// sufficient. A null HttpContext/remote (in-process TestServer, direct unit construction) is treated as
+    /// local. Returns null when the caller is allowed.
+    /// </summary>
+    private IActionResult? RejectNonLoopback()
+    {
+        var headers = HttpContext?.Request?.Headers;
+        var proxied = headers is not null
```
gautam-achieveai, line 110: [Gautam's review bot] **Must — This still does not enforce direct-local-only behind a loopback proxy.** A proxy may omit these exact headers or use a custom forwarding name; the external request then arrives with the proxy's loopback `RemoteIpAddress`, `proxied` remains false, and unauthenticated credential CRUD is allowed. ASP.NET Core's forwarded-headers middleware may also consume `X-Forwarded-*` and move prior values to `X-Original-*`, so checking these three raw header names is not a reliable proxy boundary. Require authorization or isolate this API on a management listener/path that is not reverse-proxied; any trusted-proxy policy must fail closed.

gautam-achieveai (commented):
[Gautam's review bot] **Changes required.** The latest patch improves single-threaded invalidation failure handling, but three Must issues remain:

- The active concurrency thread shows an in-flight old-credential mint can save a token after the registry removes it because invalidation is not coordinated with the provider refresh gate.
- The active cancellation thread shows the durable definition can commit before a canceled `UpdateEntry`, leaving disk and live state divergent.
- The forwarding-header presence check does not establish a direct-local-only boundary behind a loopback proxy; proxies can omit, rename, strip, or have middleware consume those headers.

No additional non-duplicate defects were found in the incremental patch. GitHub does not allow the PR author account to submit a formal REQUEST_CHANGES review, so this is posted as a blocking review comment.

gautam-achieveai (commented):
[Gautam's review bot] **Changes required.** The latest patch improves single-threaded invalidation failure handling, but three Must issues remain:

- The active concurrency thread shows an in-flight old-credential mint can save a token after the registry removes it because invalidation is not coordinated with the provider refresh gate.
- The active cancellation thread shows the durable definition can commit before a canceled `UpdateEntry`, leaving disk and live state divergent.
- The forwarding-header presence check does not establish a direct-local-only boundary behind a loopback proxy; proxies can omit, rename, strip, or have middleware consume those headers.

No additional non-duplicate defects were found in the incremental patch. GitHub does not allow the PR author account to submit a formal REQUEST_CHANGES review, so this is a blocking review comment.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Fixed in e0b23df2. The credential-change edit — invalidate persisted token → persist the new definition → swap the in-memory entry — now runs **entirely inside `PredefinedKeyProvider.ApplyUpdateAsync` under the provider's `_refreshGate`**. So a concurrent `GetAccessTokenAsync`/mint can no longer interleave and re-persist a token from the old credential (both serialize on the same gate), and a failed token removal aborts before anything is written or published. The whole mutation also runs uncancellably (`CancellationToken.None` past the registry persist gate), so a request cancel can't half-apply an edit and leave the definition, token, and live provider diverged. 123 Auth tests green.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Fixed in e0b23df2. The credential-change edit — invalidate persisted token → persist the new definition → swap the in-memory entry — now runs **entirely inside `PredefinedKeyProvider.ApplyUpdateAsync` under the provider's `_refreshGate`**. So a concurrent `GetAccessTokenAsync`/mint can no longer interleave and re-persist a token from the old credential (both serialize on the same gate), and a failed token removal aborts before anything is written or published. The whole mutation also runs uncancellably (`CancellationToken.None` past the registry persist gate), so a request cancel can't half-apply an edit and leave the definition, token, and live provider diverged. 123 Auth tests green.

src/LmAgentInfra/Controllers/EgressKeysController.cs
gautam-achieveai: Fixed in e0b23df2. The credential-change edit — invalidate persisted token → persist the new definition → swap the in-memory entry — now runs **entirely inside `PredefinedKeyProvider.ApplyUpdateAsync` under the provider's `_refreshGate`**. So a concurrent `GetAccessTokenAsync`/mint can no longer interleave and re-persist a token from the old credential (both serialize on the same gate), and a failed token removal aborts before anything is written or published. The whole mutation also runs uncancellably (`CancellationToken.None` past the registry persist gate), so a request cancel can't half-apply an edit and leave the definition, token, and live provider diverged. 123 Auth tests green.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
```diff
@@ -0,0 +1,349 @@
...
+        try
+        {
+            // 1. Invalidate the stale token FIRST (reliable — a failure throws before anything is published).
+            if (credentialChanged)
+            {
+                await _tokenStore.RemoveAsync(ProviderId, CancellationToken.None).ConfigureAwait(false);
+            }
+
+            // 2. Make the new definition durable while still holding the gate (atomic with the swap below).
+            await persistDefinition().ConfigureAwait(false);
```
gautam-achieveai, line 124: Must — `RemoveAsync` has already deleted the persisted token before `persistDefinition()` can fail. If the JSON write throws, `_entry` remains the old definition but its rotated refresh token is gone; after the cached access token expires, the provider may retry the old/original refresh token and become unusable. Preserve/restore the prior token record on persistence failure, or commit through a versioned/transactional protocol that keeps the old definition and token together until the new definition is durable; add a failing-write test.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
```diff
@@ -0,0 +1,197 @@
...
+                return false;
+            }
+
+            var candidate = _providers.Values
+                .Select(p => p.Entry)
+                .Where(e => !string.Equals(e.Id, id, StringComparison.Ordinal))
+                .ToList();
+            await AtomicJsonFile.WriteAsync(_filePath, candidate, JsonOptions, ct).ConfigureAwait(false);
+
+            _ = _providers.TryRemove(id, out _);
```
gautam-achieveai, line 168: Must — deletion removes the provider from the registry and then performs token cleanup, but it never acquires that provider's `_refreshGate` or marks it retired. An already-resolved `PredefinedKeyProvider` can be minting concurrently, save a token after `TryRemovePersistedTokenAsync` runs, and leave credential material orphaned on disk (and potentially still answer a webhook holding the provider reference). Coordinate removal with the provider gate, mark the provider retired before cleanup, and add a deterministic delete-vs-mint regression test.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        await FluentActions.Awaiting(() => provider.GetAccessTokenAsync()).Should().ThrowAsync<InvalidOperationException>();
+        handler.Calls.Should().Be(1);
+
+        // While invalid it short-circuits — no further endpoint calls.
+        await FluentActions.Awaiting(() => provider.GetAccessTokenAsync()).Should().ThrowAsync<InvalidOperationException>();
+        handler.Calls.Should().Be(1);
+
+        // Updating the entry (user re-enters the credential) clears the invalid flag → mint retried.
+        fail = false;
+        await provider.ApplyUpdateAsync(RefreshEntry(), credentialChanged: true, () => Task.CompletedTask);
```
gautam-achieveai, line 155: Should — These changed tests only adapt the call to `ApplyUpdateAsync` and pass a completed no-op persistence callback; they do not exercise either ordering guarantee fixed in this round. Add deterministic barrier-based tests for an in-flight mint racing a credential update and for cancellation after token invalidation begins, asserting the token store, durable definition, and live provider remain consistent.

gautam-achieveai (commented):
Re-review round 06: the two prior update-ordering findings are addressed, but the mutation protocol still has failure and deletion races, and the new concurrency guarantees lack regression coverage.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: By design — the orphan is harmless. Delete removes the definition durably first, so afterward no `network.rules` entry references `predefined-<id>`; a token an in-flight mint might persist just after cleanup is therefore **never injected** (no rule triggers the webhook) and is best-effort cleaned. Draining the provider's refresh gate on delete would add retirement machinery for a token that is already unreachable. In a single-user local-dev sample, deleting a key while a sandbox is mid-mint against that exact key isn't a reachable scenario, so the harmless-orphan outcome is acceptable. Leaving open for a maintainer call.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
gautam-achieveai: Added in 5e53ad88: `ApplyUpdate_when_persist_fails_invalidates_token_but_leaves_the_old_entry` forces the definition persist to fail during a credential-change edit and asserts the deliberate degradation — the token is invalidated (step 1) but the in-memory entry is not swapped, so state stays consistent with the pre-edit definition and the caller sees the failure.

src/LmAgentInfra/Auth/PredefinedKeyProvider.cs
gautam-achieveai: Deliberate, safe tradeoff — not a bug. The definition and the token live in two separate JSON files, so **no ordering is failure-safe across both** without a transaction log / token-versioning (intentionally out of scope for a local-dev sample). I chose remove-token-**first** because its failure mode is SAFE: if the definition write then fails, the in-memory entry is NOT swapped (step 3 never runs), so the next acquisition re-mints from the pre-edit entry credential (consistent with the old definition) and the caller sees the failure — versus the UNSAFE persist-first mode (a stale token reloadable under a new definition, which was the *earlier* finding). Pinned by `ApplyUpdate_when_persist_fails_invalidates_token_but_leaves_the_old_entry` (5e53ad88). The only residual is a refresh-token entry whose original token was already rotated away AND whose definition write fails at that exact instant — a triple-rare window in a single-user local sample. Leaving this open for a maintainer call rather than adding a cross-file transaction to the sample.

gautam-achieveai: Revobot rechecked this at 5e53ad8. This remains a blocking correctness issue: deleting the old record before the fallible definition write can discard a rotated refresh token while the old definition remains live. Please retain/restore the prior record on write failure or use a versioned commit protocol, and test recovery.

src/LmAgentInfra/Auth/PredefinedKeyRegistry.cs
gautam-achieveai: Revobot rechecked this at 5e53ad8. This remains a blocking deletion race: an already-resolved provider is not retired or synchronized with removal, so it can finish minting and save a token after cleanup. The registry dictionary only protects lookup; it does not control references already returned. Retire the provider under its refresh gate before cleanup and reject future acquisitions, with a deterministic delete-vs-mint test.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
gautam-achieveai: Revobot rechecked the new test at 5e53ad8. It verifies only a sequential persistence exception and would not fail on the pre-fix race. Add barrier-controlled coverage for update-vs-mint, cancellation after mutation begins, and delete-vs-mint; also seed a rotated refresh token so the destructive failure mode is exercised.

gautam-achieveai (commented):
Re-review of the latest delta: the new regression test does not resolve the remaining credential lifecycle and concurrency blockers.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
```diff
@@ -0,0 +1,196 @@
...
+        var token = await provider.GetAccessTokenAsync();
+        token.Value.Should().Be("AT2");
+        handler.Calls.Should().Be(2);
+    }
+
+    [Fact]
+    public async Task ApplyUpdate_when_persist_fails_invalidates_token_but_leaves_the_old_entry()
+    {
+        // Deliberate, consistent degradation on the (narrow) persist-failure window of a credential-change
```
gautam-achieveai, line 164: Must — This test explicitly accepts deleting the persisted token before the definition write succeeds. If the old refresh token was rotated, a failed write leaves the old live definition without its only usable persisted refresh token. Preserve/restore the prior token record or use a versioned/journaled commit, then assert recovery with a rotated token.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
```diff
@@ -0,0 +1,196 @@
...
+        // in-memory entry is NOT swapped (step 3 never runs). State stays consistent with the pre-edit
+        // definition — the next acquisition re-mints from the OLD entry credential (safe), and the caller
+        // sees the failure — rather than the unsafe alternative of a stale token under a new definition.
+        var (provider, store, _) = NewProvider(RefreshEntry(), () => "{\"access_token\":\"AT\",\"expires_in\":3600}");
+        _ = await provider.GetAccessTokenAsync();
+        (await store.GetAsync("predefined-r1")).Should().NotBeNull();
+
+        var newEntry = RefreshEntry() with { RefreshToken = "new-rt" };
+        await FluentActions
+            .Awaiting(() => provider.ApplyUpdateAsync(newEntry, credentialChanged: true, () => throw new IOException("persist failed")))
```
gautam-achieveai, line 175: Must — This sequential test does not cover deletion racing an already-resolved provider. `PredefinedKeyRegistry.RemoveAsync` can remove the provider and clean the token while that provider is still minting; it may then save a token after deletion completes. Retire the provider under its refresh gate and add a deterministic delete-versus-mint test.

tests/LmStreaming.Sample.Tests/Auth/PredefinedKeyProviderTests.cs
```diff
@@ -0,0 +1,196 @@
...
+            .Should().ThrowAsync<IOException>();
+
+        provider.Entry.RefreshToken.Should().Be("rt0");           // entry NOT swapped
```
gautam-achieveai, line 178: Should — The added coverage would not fail on the pre-fix update/mint race and omits cancellation-after-mutation and delete-versus-mint scenarios. Add barrier-controlled tests that exercise an in-flight mint, cancellation after mutation begins, and a rotated refresh token.

gautam-achieveai:
[Gautam's review bot] Deep incremental review of `95eb036b...5e53ad88` completed.

Verified fixed: the update-vs-mint race and cancellation-after-persist split are now serialized under the provider refresh gate and run uncancellably after mutation begins. No new non-duplicate defect was found in this increment.

Two active **Must** findings remain after merge:
- Persist failure after token removal can destroy the only server-rotated refresh token while the old definition remains live: https://github.com/achieveai/LmDotnetTools/pull/212#discussion_r3612189617
- Deletion does not retire/synchronize an already-resolved provider, so a concurrent mint can complete and persist credential material after cleanup: https://github.com/achieveai/LmDotnetTools/pull/212#discussion_r3612189623

The new sequential failure test does not exercise rotated-token recovery or the update/delete concurrency interleavings; the existing active coverage thread already captures that gap. The reverse-proxy trust path was untouched by this increment; its thread was already resolved, so this review did not reopen it.

PR #215, opened 2026-07-20
daemon: ADO sub-agent parity + review-pipeline reliability fixes

gautam-achieveai:
Brings the ADO cross-repo-store code-review daemon to full GitHub parity (deep reviews spawn `code-reviewer:*` sub-agents) and lands a series of review-pipeline reliability fixes discovered running both daemons (achieveai/GitHub + mcqdb/ADO) live.

## Highlights (most recent first)
- **KB scope case-collision** (`dbaf1a76`): the extraction agent cased a repo scope inconsistently, splitting entries across `MCQdbDEV/` and `mcqdbdev/`; a case-insensitive checkout collapsed them and dropped entries. `KnowledgeAgent` now reconciles scope case against existing KB dirs.
- **Not-posting bug** (`ec797072`): the review agent reliably writes the review but often skips posting it. After the review turn, a best-effort **post-enforcement turn** makes the agent actually post (agent-owned, provider-aware). Review artifact stays the first turn's text.
- **Context-discovery webhook teardown** (`c0dd6604`): configuring the auth webhook silently enabled context-discovery delivery; the daemon 404'd it and the gateway tore down the sandbox session → per-run `/mcp` 404. Adds a daemon-local `200` discovery endpoint and folds the reviewed repo's `CLAUDE.md`/`AGENTS.md` into the review input host-side.
- **Prior-notes host-side read** (`cf7b7b7c`): removes a boot/per-run session conflict on the pooled path.
- Plus: agent-left-gitlink pool wedge fix, context-exhaustion stream-abort handling, submitted-inline-review posting, unknown-agent-id graceful result, and the #187 sandbox SDK migration folded in.

## Tests
`CodeReviewDaemon.Sample.Tests` green (581), TDD for each fix. Live-verified end-to-end on both daemons (reviews complete + post inline comments; KB retrieval restored).


samples/CodeReviewDaemon.Sample/appsettings.achieveai.json
```diff
@@ -0,0 +1,52 @@
...
+  "Urls": "http://127.0.0.1:5080",
+  "Auth": {
+    "TokenStoreDir": "B:/sandbox-workspaces/crd-oauth-tokens",
+    "Github": {
+      "_comment": "GitHub CLI (gh) first-party OAuth App id + secret (public, safe to embed — same convention as LmStreaming.Sample). Required so SandboxSessionRegistry attaches the github-auth egress rule for the sandbox; the auth-webhook then injects the EXISTING stored token (crd-oauth-tokens/github.json, scopes repo/read:org/workflow/gist) for the review agent's api.github.com calls (code-reviewer:post-pr-review).",
+      "ClientId": "178c6fc778ccc68e1d6a",
+      "ClientSecret": "34ddeff2b558a23d38fba8a6de74f086ede1cc0b",
```
gautam-achieveai, line 9: Must — A GitHub OAuth `ClientSecret` is committed in a tracked operational profile. Treat this credential as exposed: remove it from version control, load it from deployment secret configuration, and rotate it.

samples/CodeReviewDaemon.Sample/Persistence/ReviewStore.cs
```diff
@@ -97,14 +97,32 @@ INSERT INTO repo (provider, normalized_key, display_name, org_or_owner, project,
     /// keying identity on it would spawn a fresh run (and a duplicate review) on the very next poll. A
     /// review is scoped to a commit; a new <c>head_sha</c> is what legitimately starts a new run.
     /// </summary>
+    /// <summary>
+    /// Inserts <paramref name="run"/>, or returns the existing row when this reviewed commit already has
+    /// one. A run's identity is the COMMIT: <c>(repo, pr, head_sha, base_sha, review_kind, variant_id)</c>.
+    /// It deliberately excludes <c>mode</c> and <c>trigger_watermark</c>: <c>mode</c> (post vs collect-only)
```
gautam-achieveai, line 103: Must — Excluding `mode` from identity returns an already-completed collect-only row when posting is later enabled. `PrOrchestrator` short-circuits completed rows before the posting stage, so the existing review can never be delivered. Separate collection completion from delivery completion, or explicitly replay the persisted review when posting is enabled.

samples/CodeReviewDaemon.Sample/Agents/ReviewAgent.cs
```diff
@@ -37,6 +50,31 @@ public async Task<ReviewAgentResult> ReviewAsync(string reviewInput, Cancellatio
...
+                var enforced = await AgentTextCollector
+                    .CollectAsync(_agent, postEnforcementPrompt, cancellationToken)
+                    .ConfigureAwait(false);
+                _logger.LogInformation(
+                    "Post-enforcement turn for run {RunId} completed ({Length} chars).",
+                    enforced.RunId ?? collected.RunId,
+                    enforced.Text.Length
+                );
+            }
+            catch (Exception ex) when (ex is not OperationCanceledException)
```
gautam-achieveai, line 66: Must — This swallows every non-cancellation failure from the post-enforcement turn and still lets the run complete. With host fallback disabled, an authorized review can be retained with no PR delivery and no retry signal. Propagate/persist a delivery failure or require a verified posting receipt before marking the run complete.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -1020,6 +1183,65 @@ private async Task<string> PrependPriorKnowledgeAsync(
...
+            }
+        }
+
+        if (blocks.Count == 0)
+        {
+            return reviewInput;
+        }
+
+        _logger.LogInformation("Prepending reviewed-repo guidance ({Count} file(s)) to the review input.", blocks.Count);
+        return $"## Repository guidance (the reviewed repo's own CLAUDE.md / AGENTS.md)\n\n{string.Join("\n\n", blocks)}\n\n{reviewInput}";
```
gautam-achieveai: Must — PR-controlled `CLAUDE.md`/`AGENTS.md` content is inserted under an authoritative “Repository guidance” heading while the agent has authenticated posting tools. Treat this as explicitly untrusted quoted data that cannot override review/posting rules, and preferably source trusted guidance from the base commit rather than the PR head.

samples/CodeReviewDaemon.Sample/appsettings.achieveai.json
```diff
@@ -0,0 +1,52 @@
...
+      "PublicBaseUrl": "http://127.0.0.1:5080"
+    }
+  },
+  "SandboxGateway": {
+    "BaseUrl": "http://127.0.0.1:3000",
+    "AppId": "code-review-daemon",
+    "WorkspaceBasePath": "B:/sandbox-workspaces/workspaces"
+  },
+  "CodeReviewDaemon": {
+    "EnableCommentPosting": true,
```
gautam-achieveai, line 25: Should — This committed profile enables live PR writes despite its comment claiming collect-only behavior, and the ADO profile has the same setting. Keep committed profiles collect-only and require deployment-only overrides for live posting, or otherwise correct the safety contract and isolate operational configuration from sample files.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -81,15 +81,39 @@ public async Task<PullRequestPage> ListOpenPullRequestsAsync(PrPollRequest reque
...
+                    // is never silently dropped.
+                    DateTimeOffset? updatedAt = null;
+                    if (request.RecencyCutoff is { } cutoff
+                        && createdAt is { } created
+                        && created < cutoff
+                        && !string.IsNullOrEmpty(headSha))
+                    {
+                        updatedAt =
+                            await TryGetLastPushDateAsync(org, project, repo, headSha, cancellationToken)
+                                .ConfigureAwait(false) ?? cutoff;
```
gautam-achieveai: Should — A failed last-push lookup is represented as `updatedAt = cutoff`, conflating an unknown timestamp with an exact boundary timestamp. Keep the signal null and let the recency filter’s explicit unknown-date policy retain the PR, so downstream consumers do not see synthetic activity.

gautam-achieveai (commented):
PR #215 has blocking credential, delivery-state, and prompt-boundary issues; please address the inline findings before merge.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -833,17 +896,39 @@ internal IReadOnlyList<SubmoduleAllowRule> BuildStoreSubmoduleAllowList(ReviewRu
...
+            new(host, RepoPath(repo.RepoName)),
+            new(host, RepoPath("Contracts")),
         };
 
+        // The reviewed repo's OWN first-party submodules (its direct dependencies) are allow-listed
+        // UNCONDITIONALLY — unlike CrossRepoSiblings below, these are the target's own dependency graph
+        // (needed to build/understand it), not store-level siblings, so the fork/public confidentiality gate
+        // does not apply. Still fail-closed: only the explicit configured names are permitted; a submodule an
+        // attacker adds or repoints to any other name/host is denied.
+        foreach (var submodule in _options.ReviewedRepoSubmodules)
```
gautam-achieveai, line 920: Must — `ReviewedRepoSubmodules` is allow-listed unconditionally for fork/public reviews, but the PR-controlled `.gitmodules` can repoint a configured repository name/path to a different private repository or add a new submodule using that configured name. This bypasses the confidentiality gate and can expose private first-party content to an untrusted review session. Bind allowed path+URL to trusted base-branch metadata, or disable these private submodules when trust is not positively established.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -131,6 +156,84 @@ private static string CommitId(JsonElement pr, string property) =>
...
+            value.GetString(),
+            System.Globalization.CultureInfo.InvariantCulture,
+            System.Globalization.DateTimeStyles.RoundtripKind,
+            out var parsed)
+            ? parsed
+            : null;
+
+    /// <summary>
+    /// Fetches the source branch tip commit (<c>GET .../commits/{commitId}</c>) and returns its
+    /// <c>committer.date</c> (falling back to <c>author.date</c>) — the PR's last-push time, ADO's stand-in
```
gautam-achieveai: Must — This treats the head commit's `committer.date`/`author.date` as the PR's last-push time. Commit timestamps are not ref-update/push timestamps, so an old commit recently pushed or force-pushed can be classified as stale and silently excluded by `ApplyRecencyFilter`. Use an actual ref/push activity signal, or retain the PR whenever a reliable push timestamp is unavailable.

samples/CodeReviewDaemon.Sample/Diagnostics/CopilotModelCatalogLogger.cs
```diff
@@ -0,0 +1,129 @@
...
+    private static readonly TimeSpan DiscoveryTimeout = TimeSpan.FromSeconds(30);
+
+    private readonly ILogger<CopilotModelCatalogLogger> _logger;
+
+    public CopilotModelCatalogLogger(ILogger<CopilotModelCatalogLogger> logger)
+    {
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public async Task StartAsync(CancellationToken cancellationToken)
```
gautam-achieveai: Should — `StartAsync` awaits the remote `/models` request, so this best-effort diagnostic can delay hosted-service startup for up to 30 seconds despite the class documentation saying discovery does not block boot. Run discovery after startup in a lifecycle-managed background task, or make it explicitly opt-in.

gautam-achieveai (commented):
Re-review of PR #215: two new trust/correctness blockers and one startup-operability issue remain; prior delivery/config findings are still unresolved.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found two additional workspace-durability defects. The existing recency thread also received a corrective reply because its claimed fix still drops unknown-activity ADO PRs.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,158 @@
...
+        // (an uninitialized stray gitlink shows as clean), so log the best-effort cleanup failure and continue.
+        if (!submodules.Succeeded)
+        {
+            logger?.LogWarning(
+                "Slot hygiene at {StorePath}: `git submodule foreach` cleanup failed (continuing — a re-clone "
+                    + "cannot fix committed content; the status probe still gates cleanliness): {Stderr}",
+                storePath, submodules.Stderr);
+        }
+
+        var status = await git.RunAsync(["-C", storePath, "status", "--porcelain"], storePath, ct)
```
gautam-achieveai: [Gautam's review bot] Must - This cleanliness gate makes every successful warm slot look dirty on its next lease. `PrepareAsync` checks the reviewed submodule out at the PR head, but `StripAsync` only resets the superproject and each submodule to its own current HEAD; it never returns the submodule to the gitlink recorded by the superproject. Consequently `git status --porcelain` reports ` M <submodule>` and this branch forces a full store reclone every time. I reproduced the exact reset/clean/foreach sequence with real Git, and the actual AchieveAiReviews `.gitmodules` has no `ignore` override. Restore the recorded gitlink (or deliberately ignore only submodule commit drift) before using status as the reclone gate, with a real-Git warm-reuse test.

samples/CodeReviewDaemon.Sample/Workspace/ReviewSlotPreparer.cs
```diff
@@ -133,6 +144,26 @@ await RunGitOrThrowAsync(
...
+        // submodule must have actually initialized. Without this a denied/failed init silently proceeds to
+        // the fetch below, which then fails opaquely. But NOT every failure is slot corruption: a TRANSIENT
+        // cause (auth/network/throttle — captured in the denial reason's stderr) must retry the warm store, not
+        // trigger a destructive reclone that can't fix it and would loop (review #180). Only corrupt/unrecognized
+        // failures drive the reclone ladder.
+        if (!outcome.InitializedPaths.Contains(submoduleRelPath, StringComparer.Ordinal))
+        {
+            var reason = outcome.Denied
+                .FirstOrDefault(d => string.Equals(d.Path, submoduleRelPath, StringComparison.Ordinal))?.Reason;
+            if (GitFailureClassifier.Classify(reason) == GitFailureKind.Transient)
```
gautam-achieveai: [Gautam's review bot] Should - This branch maps every `Unknown` init failure to `SlotCorruptException`, even though `GitFailureClassifier` documents Unknown as transient/non-reclone and `RunGitOrThrowAsync` implements that contract. Unmatched auth/403, policy, disk, or other environment failures therefore trigger a destructive full-store reclone that cannot repair them, and repeat on later ContextReady retries. Only `Corrupt` should enter the reclone ladder; add an unknown/auth init-failure test alongside the transient case.

gautam-achieveai:
[Revobot]

# PR Review: daemon: ADO sub-agent parity + review-pipeline reliability fixes

## Summary
- **PR:** #215 (`ado-subagent-parity` → `main`), author: gautam-achieveai
- Total files reviewed: 83
- Findings: **1 Critical, 8 High, 8 Medium, 1 Low**
- Context questions: 0
- Test coverage: Excellent (thorough, well-structured, good regression guards)
- Domain areas touched: Daemon orchestration, agents, workspace/git, persistence, configuration, hosting, framework (LmMultiTurn)

## Strengths
- **Exceptional test coverage** — 29 test files modified/added, 10+ regression-guard tests pinning specific production incidents
- **No debugging artifacts** — zero Console.WriteLine, zero HACK/TODO, zero disabled tests
- **Good structured logging** — named properties, 33 of 37 log calls carry RunId
- **Consistent ConfigureAwait(false)** — zero sync-over-async
- **Correct re-throws** — all `throw;`, zero `async void`
- **Well-designed RetryGovernor** — bounded capacity, clock injection, overflow-safe backoff

---

## 🔴 Critical Issues

**C1 — `ReviewSessionProvisioner.cs:262-268` — [BLOCKER] — DestroyAsync disposes ALL sessions**

`DestroyAsync` iterates ALL sessions via `foreach` over `_bySession` and disposes each one's `IAsyncDisposable` — not just the targeted session. After one `DestroyAsync`, every other active review session loses its gateway connection (`SandboxClient` + `HttpClient` disposed), causing cascading failures.

**Fix:** Replace `foreach` with targeted lookup: resolve `sessionId` from `workspaceId`/`runId`, then `TryRemove` and dispose only that session.

---

## 🟠 High Issues

**H1 — `DaemonReviewStageExecutor.cs` — God Class (1,861 lines, 23 ctor params, 19 fields)**
Owns slot mgmt, review execution, escalation ladders, knowledge/guidance prepend, ADO posting, notes commit, retention publish, sub-agent discovery, tool context assembly. Decompose into `PooledSlotManager`, `ReviewInputAssembler`, `ReviewExecutionOrchestrator`, `ReviewRetentionPublisher`. File a follow-up ticket.

**H2 — `Program.cs` — God Composition Root (742 lines)**
Inline logic (BuildHostGitCredentialsSource, DeriveAdoOrgs, AdoOrgFromUrl, closures) with tricky lifetime management. Extract into named classes. Register `HostGitCommandRunner`/`HostFileSystem` as DI singletons.

**H3 — `SlotHygiene.cs:136, 153` — [BLOCKER] Two bare `catch` blocks**
Swallow ALL exceptions including `OutOfMemoryException`, `AccessViolationException`. Filter to `catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or DirectoryNotFoundException)`. At minimum, add logging.

**H4 — `DaemonReviewStageExecutor.cs:903-904, 1035` — `isAdo` provider check inconsistency**
`BuildStoreSubmoduleAllowList` accepts both "azure-devops" AND "ado". `BuildPromptVariables` only checks "azure-devops". A repo with `provider="ado"` gets `is_ado=false` in prompt variables. Extract shared `IsAdoProvider()` helper.

**H5 — `CopilotModelCatalogLogger.cs` — Unrequested IHostedService**
129 lines, startup latency (30s timeout), 40+ log lines per boot, test file — for a diagnostic that changes no behavior. Remove and replace with a standalone script.

**H6 — `DaemonReviewStageExecutor.cs:1173,1182,1226,1244` — Log messages missing RunId**
Four orphans in `PrependPriorKnowledgeAsync`/`PrependRepoGuidanceAsync`. Thread `runId` into each.

**H7 — `PrOrchestrator.cs:135-152` — `DescribePickReason` no direct tests**
Non-trivial branching (resuming/new PR/new commit/re-review) untested. Add parametrized tests.

**H8 — `RetryGovernor.cs:76,86-89,147-150` — Concurrent access untested**
All 10 tests single-threaded but production is concurrent. `EvictOldestOverCapacity` iterates without lock. Add concurrent tests.

---

## 🟡 Medium Issues

| # | File | Line | Issue |
|---|---|---|---|
| M1 | `SlotRecoveryExceptions.cs` | 7-15 | Two exception types caught identically — collapse into one |
| M2 | `Program.cs` | 460-466 | 3-level nested ternary — use `if`/`return` guards |
| M3 | `DaemonReviewStageExecutor.cs` | 1271-1329 | 4-level nested try-catch — model as loop over steps |
| M4 | `DaemonReviewStageExecutor.cs` | 562-585 | Duplicated 9-argument `PrepareAsync` — extract local function |
| M5 | `ReviewSessionProvisioner.cs` | 237 | `catch(Exception)` too broad in disk-space check |
| M6 | `ReviewStore.cs` | 270-294 | Unbounded `SELECT DISTINCT` — add LIMIT |
| M7 | `ReviewSessionProvisioner.cs` | 82 | `_bySession` no eviction — add size guard |
| M8 | `Program.cs` | 230-245 | Manual `HttpClient` construction — use `IHttpClientFactory` |

## 🟢 Low Issues

| # | File | Line | Issue |
|---|---|---|---|
| L1 | `appsettings.achieveai.json` | 2 vs 25 | Stale safety comment: says "collect-only" but `EnableCommentPosting: true` |

---

## Testing Assessment

**EXCELLENT.** 29 test files, 10+ regression guards pinning specific production incidents (stale-lock, gitlink wedge, path-traversal). Recovery ladder tested end-to-end. Zero fragile tests, zero over-mocked tests.

**Gaps:** `DescribePickReason` (no direct tests), `RetryGovernor` concurrent (untested), `SlotHygiene` gitfile case (untested).

---

1 CRITICAL blocker (C1: `DestroyAsync` cascading session failures) + 8 HIGH findings (architecture, bare catches, provider-check inconsistency, untested concurrency). The code quality within individual classes is high and test coverage is excellent, but the critical correctness bug and the accumulation of high-severity issues require fixes before merge.

gautam-achieveai (commented):
Re-reviewed the two-file test-only delta; no new findings in this round.

tests/ProcessLauncher.Tests/SystemProcessHandleTests.cs
```diff
@@ -16,7 +16,12 @@ private static IProcessHandle LaunchBenign(IReadOnlyList<string>? args = null)
         {
             Agent = CliAgentKind.Codex,
             ExecutableHint = "dotnet",
-            Arguments = args ?? ["--info"],
+            // `--version` (a single short line) rather than `--info`: these tests don't read the redirected
+            // stdout/stderr before calling WaitForExitAsync, so a child whose output exceeds the OS pipe
+            // buffer would block writing and never exit — deadlocking the wait. `--info` grows with the number
+            // of installed SDKs/runtimes and crosses that threshold on developer boxes (it stayed under it on
+            // CI's clean image, so the hang only reproduced locally). `--version` can never fill the buffer.
+            Arguments = args ?? ["--version"],
```
gautam-achieveai, line 24: Should — switching this helper to the very short-lived `dotnet --version` increases the race in `ExitedEvent_FiresExactlyOnce`: the child can exit before the handler is attached, and `SystemProcessHandle` then raises its one-shot event during construction so the late subscriber receives nothing. Use a controllable child that remains alive until after subscription, then signal it to exit, so the exactly-once event test is deterministic.

gautam-achieveai (commented):
Re-review follow-up: the updated process test command introduces an event-subscription race.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -81,15 +81,41 @@ public async Task<PullRequestPage> ListOpenPullRequestsAsync(PrPollRequest reque
...
+                    // the re-review flagged). UpdatedAt is consumed only by this recency keep/drop decision.
+                    DateTimeOffset? updatedAt = null;
+                    if (request.RecencyCutoff is { } cutoff
+                        && createdAt is { } created
+                        && created < cutoff
+                        && !string.IsNullOrEmpty(headSha))
+                    {
+                        updatedAt =
+                            await TryGetLastPushDateAsync(org, project, repo, headSha, cancellationToken)
+                                .ConfigureAwait(false) ?? cutoff;
```
gautam-achieveai: **Must — this reintroduces the synthetic-activity bug.** When the commit metadata lookup fails, assigning `UpdatedAt = cutoff` makes an unknown timestamp look like a real activity exactly at the boundary. Although this currently keeps the PR, downstream recency logic and persisted descriptors can treat it as verified activity and later exclude/reclassify it incorrectly. Preserve `null` on lookup failure and have the recency filter explicitly retain unknown dates (or keep the descriptor out of activity-based decisions).

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -1020,6 +1183,78 @@ private async Task<string> PrependPriorKnowledgeAsync(
...
+            }
+
+            if (!string.IsNullOrWhiteSpace(content))
+            {
+                // SECURITY: this guidance is read from the PR HEAD, so it is attacker-controllable — a hostile
+                // PR could put injection text in its CLAUDE.md/AGENTS.md. Fence each file as quoted DATA, and
+                // neutralize any literal </pr-guidance-file> the content embeds (rewrite it to a bracketed,
+                // non-tag form) so it cannot forge the closing fence and break out of the quoted region.
+                // Belt-and-braces with the "UNTRUSTED, report injection" instruction the block is headed with.
+                var fenced = content.Replace(
```
gautam-achieveai: **Should — the guidance input is still attacker-controlled and unbounded.** Escaping the closing tag fixes the specific delimiter-forgery case, but a PR can supply an arbitrarily large `CLAUDE.md`/`AGENTS.md`; this content is prepended before the review and can exhaust the model context or cause repeated escalation/diff-only degradation. Enforce a bounded byte/character budget (and log/truncate or skip oversized files) before adding the block.

gautam-achieveai (commented):
Re-review of the current delta: two correctness/security issues remain in the changed lines; prior open findings were answered in-thread.

gautam-achieveai (commented):
Re-review of PR #215 across 1 changed file since 3abcaf5d — no new findings in this delta; the process-exit test now uses a controllable long-running child before subscribing.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review found one remaining reliability defect in the current delta: the Windows `timeout` sleeper exits immediately when stdin is redirected, so the process-exit event test is still timing-dependent. Replied on the existing thread; no other new non-duplicate findings.

gautam-achieveai (commented):
Re-review at ee020e5: no new actionable findings in the PR delta; the open Windows process-test thread was rechecked and answered in-thread.

gautam-achieveai (commented):
[Gautam's review bot] Deep incremental review of `1cb6c386...ee020e5` found no new high-confidence, non-duplicate issues in the merge-resolution delta.

Existing thread `3617915765` remains unresolved: the Windows test sleeper still uses `cmd /c timeout /t 30 /nobreak` with redirected stdin, so it can exit immediately and preserve the event-subscription race. The thread already has a head-specific comment, so I did not duplicate it.

tests/CodeReviewDaemon.Sample.Tests/Workspace/ReviewSlotPreparerTests.cs
```diff
@@ -133,12 +133,148 @@ public async Task PrepareAsync_ReturnsThePosixJoinedPreparedCheckoutPaths()
...
+        runner.OnArgvContains(
+            "submodule update --init", new SandboxCommandResult(1, string.Empty, "fatal: clone of submodule failed"));
+        var preparer = new ReviewSlotPreparer(
+            new GitRunner(runner), SeedGitmodules(slot.StorePath), "github", NullLoggerFactory.Instance);
+
+        var act = async () => await preparer.PrepareAsync(
+            slot, CreateRun(), StoreUrl, SubmoduleRelPath, Branch, DefaultBranch, NotesRelPath, BuildPolicy(), CancellationToken.None);
+
+        // It still throws (no silent proceed on a half-inited submodule), but NOT the reclone-driving exception.
+        var thrown = await act.Should().ThrowAsync<Exception>();
```
gautam-achieveai: Should — This assertion accepts any exception other than `SlotCorruptException`, including unrelated regressions such as `NullReferenceException`. The production branch is expected to throw `InvalidOperationException` for an unknown classification; use `ThrowExactlyAsync<InvalidOperationException>()` (and optionally assert the message) so the test fails when the implementation breaks for an unrelated reason.

gautam-achieveai (commented):
Re-reviewed the seven-file delta from ee020e5 to 30e6af8 — production fixes are sound; one test-quality issue remains.

gautam-achieveai (commented):
Re-reviewed the one-file delta since 30e6af8; the prior test-precision finding is fixed and no new findings remain.

gautam-achieveai (commented):
[Gautam's review bot] Reviewed `30e6af87...85e2163e` (deep incremental review). No new findings. The exact-exception assertions correctly address thread `3618554561`: both unknown and transient submodule-init failures now require `InvalidOperationException`, matching the production retry path and preventing unrelated exceptions from passing the tests.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `1cb6c386...ee020e57` found one new non-blocking correctness issue: the shipped dispatcher/reviewer model split executes discovered sub-agents on `SubAgentModelId` but persists their usage under `ReviewModelId`, corrupting the per-model audit breakdown. The only newly updated prior thread is the existing Windows `timeout` sleeper race; it remains active and already has a current comment, so I did not duplicate it.

The head is a clean merge of `main` (predefined egress keys, sub-agent UI/focus, conversation usage accounting, file browsing, and per-session webhook/DeepSeek support). I found no additional high-confidence merge-integration defects after excluding issues that belong solely to the updated target branch.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -81,15 +81,61 @@ public async Task<PullRequestPage> ListOpenPullRequestsAsync(PrPollRequest reque
...
+                    //    `?? cutoff` did, which conflates "unknown" with "active exactly at the cutoff"), and we do
+                    //    NOT fall back to the stale opened-date (which would wrongly DROP a possibly-active PR).
+                    DateTimeOffset? updatedAt = null;
+                    DateTimeOffset? recencyCreatedAt = createdAt;
+                    if (request.RecencyCutoff is { } cutoff
+                        && createdAt is { } created
+                        && created < cutoff
+                        && !string.IsNullOrEmpty(sourceRefName))
+                    {
+                        var lastPush = await TryGetLastPushDateAsync(org, project, repo, sourceRefName, cancellationToken)
```
gautam-achieveai: Must — This awaits one `/pushes` request inside the PR loop, so every old PR in up to 10 pages causes a sequential network round trip. A repository with many old active PRs can make one poll take minutes or hit ADO throttling/timeouts, delaying all reviews. Resolve these lookups with bounded concurrency and/or a cache keyed by source ref (while preserving the poll cancellation bound).

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,182 @@
...
+        //    fetch — and touches only already-initialized, .gitmodules-registered submodules, so it can't hit a
+        //    denied/network submodule and skips a committed embedded gitlink with no .gitmodules URL (the
+        //    PR-11182 wedge). It is not re-clone-gated: a failure leaves the pointer moved, which the status
+        //    probe below then catches as dirty.
+        var reset = await git.RunAsync(["-C", storePath, "reset", "--hard"], storePath, ct).ConfigureAwait(false);
+        var clean = await git.RunAsync(["-C", storePath, "clean", "-ffdx"], storePath, ct).ConfigureAwait(false);
+        var restore = await git.RunAsync(
+                ["-C", storePath, "submodule", "update", "--force"], storePath, ct)
+            .ConfigureAwait(false);
+        if (!restore.Succeeded)
```
gautam-achieveai, line 110: Should — A failed `submodule update --force` is only logged, but the method can still return `Clean` from the later superproject `status --porcelain` check. That status is not a complete proof that every initialized submodule is at its recorded gitlink (for example, submodule ignore settings can hide the difference), so stale submodule state can cross leases. Treat restoration failure as `NeedsReclone`, or explicitly verify each submodule HEAD against its recorded gitlink before returning `Clean`; add a test for a failed restore with empty status.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -81,15 +81,61 @@ public async Task<PullRequestPage> ListOpenPullRequestsAsync(PrPollRequest reque
...
+                    //  - old PR, push date UNKNOWN  -> leave the signal indeterminate (BOTH null) so the filter's
+                    //    "can't date it => keep" path applies. We do NOT fabricate a boundary timestamp (an earlier
+                    //    `?? cutoff` did, which conflates "unknown" with "active exactly at the cutoff"), and we do
+                    //    NOT fall back to the stale opened-date (which would wrongly DROP a possibly-active PR).
+                    DateTimeOffset? updatedAt = null;
+                    DateTimeOffset? recencyCreatedAt = createdAt;
+                    if (request.RecencyCutoff is { } cutoff
+                        && createdAt is { } created
+                        && created < cutoff
+                        && !string.IsNullOrEmpty(sourceRefName))
```
gautam-achieveai: Should — When an old PR has a missing/blank `sourceRefName`, this condition skips the push lookup but leaves `recencyCreatedAt` set to the old creation date. `ApplyRecencyFilter` then drops the PR even though its activity is unknown, unlike the failed-lookup path below which clears the date and keeps it. Treat an unusable source ref as indeterminate (`recencyCreatedAt = null`) and add a regression test for an old PR without `sourceRefName`.

gautam-achieveai (commented):
Re-reviewed the four-file delta since 85e2163 — three actionable correctness and scalability issues remain.

samples/CodeReviewDaemon.Sample/Workspace/Git/ReviewBranchManager.cs
```diff
@@ -229,7 +248,14 @@ CancellationToken cancellationToken
...
         if (!ffOnly.Succeeded)
         {
-            await RunGitAsync(["merge", "--no-edit", remoteBranch], repoRoot, cancellationToken).ConfigureAwait(false);
+            // Not a fast-forward — land a merge commit. `-X theirs` auto-resolves a content conflict on an
+            // accumulated notes file (the notes branch's latest review.md vs a default that advanced under a
+            // concurrent PR's merge) in favour of the notes branch — which is exactly what we want to carry —
+            // instead of leaving a conflicted index that would wedge this checkout on the next sweep. Without
+            // it the same conflict re-throws every cycle and the notes never land.
+            await RunGitAsync(
+                    ["merge", "--no-edit", "-X", "theirs", remoteBranch], repoRoot, cancellationToken)
```
gautam-achieveai, line 257: [Gautam's review bot] **Must — concurrent Knowledge Base merges can overwrite already-landed knowledge.** Each notes branch regenerates KnowledgeBase/_index.jsonl and _toc.md from its own branch snapshot. If branch A lands first and branch B then conflicts, -X theirs takes B's stale bookkeeping/content wholesale; a real Git reproduction leaves both entry files but drops A from _index.jsonl. Merge/rebase the latest default into the notes branch and regenerate bookkeeping on the combined tree, or resolve KB conflicts explicitly instead of applying blanket theirs.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -871,9 +956,26 @@ private static string TargetRemoteUrl(RepoIdentity repo, string provider) =>
...
+        // restart, or a resume after a RetryPending) arrives here with no lease. Without one, BuildToolContextAsync
+        // would fall back to the per-run review-run-{id} mount — a directory that does not exist under the
+        // gateway's read-only workspace base, so the gateway 400s and the review silently degrades to diff-only
+        // with no sub-agents. Re-lease + re-prepare a slot here so the resumed review still runs tool-assisted;
+        // the persisted context's container paths (/workspace/...) are slot-index-independent, so a freshly
+        // leased slot is interchangeable. TryPooledFetchContextAsync returns false for a non-store-submodule
+        // repo, which leaves the existing per-run/diff-only path unchanged.
+        if (UsePooledReview && !_leasedReviews.ContainsKey(run.Id))
+        {
+            await TryPooledFetchContextAsync(run, repo, provider, cancellationToken).ConfigureAwait(false);
```
gautam-achieveai, line 971: [Gautam's review bot] **Should — resumed checkout failures bypass retry governance.** After a restart, a run persisted at ContextReady has no in-memory lease, so this re-runs slot preparation while the orchestrator is executing the Reviewed stage. Persistent prep failures are therefore never recorded by RetryGovernor (which only governs exceptions attributed to ContextReady) and the run retries every poll indefinitely. Classify resumed preparation failures as context failures and apply the same backoff/parking path.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
```diff
@@ -202,6 +202,13 @@ CancellationToken cancellationToken
             return null;
         }
 
+        // Reconcile the scope directory's CASE against the Knowledge Base: the extraction agent is an LLM that
+        // cases a repo scope inconsistently across runs (e.g. 'MCQdbDEV' one run, 'mcqdbdev' the next). Written
+        // verbatim those become two directories a case-sensitive git tracks but a case-insensitive checkout
+        // (Windows) collapses — losing entries and breaking retrieval (observed live on the mcqdb store). Reuse
+        // the first-seen case so every entry for a scope stays in ONE directory.
+        scope = await ReconcileScopeCaseAsync(knowledgeBaseDir, scope, cancellationToken).ConfigureAwait(false);
```
gautam-achieveai, line 210: [Gautam's review bot] **Should — the scope-case fix skips explicit updates.** Reconciliation happens only after the UPDATES path has already been looked up verbatim. On a case-sensitive host, UPDATES: MCQdbDEV/old-name.md misses an existing mcqdbdev/old-name.md; if the refined title slug differs, the fallback creates a second entry instead of updating the named one. Canonicalize the first path segment of updatesRel before the existence check and return that canonical path.

samples/CodeReviewDaemon.Sample/Prompts/daemon-prompts.yaml
```diff
@@ -28,6 +28,55 @@ review:
...
+
+       AUTHENTICATION: you do NOT have, and do not need, a token. The sandbox's egress proxy automatically
+       attaches the bot's authentication to every request you send to the provider's API host, for reads AND
+       writes — just make the HTTP request. `curl` needs nothing. If a tool insists a token be present (e.g.
+       `gh`), set a placeholder first: `export GH_TOKEN=x` (or `GITHUB_TOKEN=x`); the proxy replaces it with
+       the real credential. Never print or ask for a token.
+       {{ if is_ado ~}}
+       This is an Azure DevOps PR — the code-reviewer:post-pr-review skill is GitHub-only, so do NOT use it.
+       Post with `curl` against the ADO REST API (append ?api-version=7.1 to every URL). Base:
+         https://dev.azure.com/{{ ado_org }}/{{ ado_project }}/_apis/git/repositories/{{ ado_repo }}/pullRequests/{{ pr_number }}
```
gautam-achieveai, line 46: [Gautam's review bot] **Should — valid ADO names can make the posting URL malformed.** Azure DevOps permits spaces in project and repository names, but these raw values are interpolated into a curl URL; curl rejects a raw-space URL with exit 3 before either posting turn reaches ADO. Percent-encode both path segments before prompt rendering, or use stable project/repository IDs.

gautam-achieveai (commented):
[Gautam's review bot] Deep live-head review found four additional non-duplicate correctness and reliability issues.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -131,6 +167,158 @@ private static string CommitId(JsonElement pr, string property) =>
...
+    /// </summary>
+    private async Task<(DateTimeOffset? UpdatedAt, DateTimeOffset? RecencyCreatedAt)[]> ResolveRecencySignalsAsync(
+        IReadOnlyList<RawAdoPr> prs,
+        string org,
+        string? project,
+        string repo,
+        DateTimeOffset? cutoff,
+        CancellationToken cancellationToken)
+    {
+        using var gate = new SemaphoreSlim(MaxRecencyLookupConcurrency);
```
gautam-achieveai, line 207: Should — The new bounded-concurrency path is not behaviorally tested: the added recency test exercises only one old PR, so it passes against the prior sequential implementation too. Add a handler-driven test with multiple old PRs that asserts requests overlap, never exceed the six-request cap, preserve PR-to-result ordering, and honor cancellation.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -131,6 +167,158 @@ private static string CommitId(JsonElement pr, string property) =>
...
+        DateTimeOffset? cutoff,
+        CancellationToken cancellationToken)
+    {
+        // Recent PR (or recency off): no lookup; the recent opened-date is the keep signal.
+        if (cutoff is not { } c || pr.CreatedAt is not { } created || created >= c)
+        {
+            return (null, pr.CreatedAt);
+        }
+
+        // Old PR with no usable source ref: recency indeterminate ⇒ keep (both null), like a failed lookup.
```
gautam-achieveai, line 229: Should — Returning `(null, null)` for an old PR with no usable source ref erases the provider-supplied creation date solely to influence `PrPollingService`'s `UpdatedAt ?? CreatedAt` filter. That makes `PullRequestDescriptor.CreatedAt` factually/contextually ambiguous for future consumers. Preserve `CreatedAt` and carry an explicit indeterminate-activity signal (or move this keep-on-unknown policy into a dedicated recency result) so provider data is not overloaded as control flow.

gautam-achieveai (commented):
Re-reviewed the four-file delta since f9c0641; the prior ADO and slot-hygiene correctness issues are fixed, with two remaining maintainability/test-coverage comments.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,184 @@
...
+        //    (`git status` reports it dirty). The `submodule foreach` below only resets each submodule to its
+        //    OWN (PR-head) HEAD, so without this restore the slot would look dirty on every warm lease and be
+        //    needlessly re-cloned. `--force` alone (no --init / --recursive) uses locally-present objects — no
+        //    fetch — and touches only already-initialized, .gitmodules-registered submodules, so it can't hit a
+        //    denied/network submodule and skips a committed embedded gitlink with no .gitmodules URL (the
+        //    PR-11182 wedge). It is not re-clone-gated: a failure leaves the pointer moved, which the status
+        //    probe below then catches as dirty.
+        var reset = await git.RunAsync(["-C", storePath, "reset", "--hard"], storePath, ct).ConfigureAwait(false);
+        var clean = await git.RunAsync(["-C", storePath, "clean", "-ffdx"], storePath, ct).ConfigureAwait(false);
+        var restore = await git.RunAsync(
```
gautam-achieveai, line 88: Must — `git submodule update --force` can perform an implicit remote fetch when the recorded gitlink object is absent. With the new `restore.Succeeded` gate, a transient auth/outage/throttle is classified as slot corruption and triggers destructive full-store re-cloning; `StripAsync` also invokes the same command with `CancellationToken.None`, allowing uncancellable remote I/O while holding the slot. Use `--checkout --force --no-fetch` in both hygiene paths and test the command so missing local objects fail locally and follow the intended recovery/retry path.

gautam-achieveai (commented):
Re-review update: the prior ADO and restore-verdict issues are fixed; one remaining hygiene/recovery issue is tracked inline.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
```diff
@@ -1727,16 +1743,19 @@ private async Task SendToParentAsync(SubAgentState state, string text)
...
+    /// <c>RootConversationId</c> placeholder is re-stamped to the ledger's root by
+    /// <see cref="UsageLedger.RecordUsage"/>.
     /// </summary>
     private UsageRecord BuildDescendantUsageRecord(UsageMessage message, SubAgentState state) =>
         UsageRecordMapper.FromUsageMessage(
             message,
             state.AgentId,
             UsageExecutionKind.SubAgent,
-            state.ModelOverride ?? _parentModelId);
+            state.ModelOverride ?? state.Template.DefaultOptions?.ModelId ?? _parentModelId);
```
gautam-achieveai: Must — This reconstructs the billed model with null-coalescing rather than using the effective model selected during sub-agent creation. `ResolveSubAgentOptions` treats whitespace overrides/template IDs as absent, and the characteristics path can subsequently replace a template model with `_parentModelId` (`CreateSubAgentAsync` lines 1229-1232); this expression can therefore persist a model different from the one that handled the request. Capture the final `defaultOptions?.ModelId` after all provider routing on `SubAgentState` and use that value here (also add tests for whitespace and `UseParentModel=true`).

gautam-achieveai (commented):
Re-review of the two-file usage-attribution delta: one blocking correctness issue remains.

gautam-achieveai (commented):
Re-review of the three-file delta: two correctness issues and two follow-up test/operability issues remain.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        // 5. Cleanliness gate: `rev-parse --git-dir` only proves the repo STRUCTURE is intact, not that the tree
+        //    is CLEAN. A superproject cleanup step that reported failure, a CORRUPT submodule restore (a submodule
+        //    genuinely broken off its recorded gitlink — which `status --porcelain` may not surface when
+        //    submodule-ignore settings hide it), or a working tree still showing tracked/untracked changes
+        //    afterwards, means stale state would cross into the next run — so force a re-clone rather than
+        //    reporting a still-dirty slot as Clean (which would let contamination survive the pool). A merely
+        //    transient restore failure is deliberately NOT gated here (see above).
+        if (!reset.Succeeded || !clean.Succeeded || restoreCorrupt)
```
gautam-achieveai: Must — A failed `submodule update --force` is treated as safe unless its stderr matches `Corrupt`, so the method can proceed to `status --porcelain` and return `Clean` even though the submodule was not restored to the recorded gitlink. The surrounding comment acknowledges that submodule-ignore settings can hide this mismatch, and `ReviewSlotPreparer` proceeds immediately on `Clean`, allowing the prior lease's checkout to contaminate the current review. Return a non-clean/retry-without-reclone outcome for any failed restore, or independently compare every initialized submodule HEAD with the superproject gitlink before returning `Clean`; only report `Clean` after that verification succeeds.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -131,6 +167,158 @@ private static string CommitId(JsonElement pr, string property) =>
...
+                var date = ParseTimestamp(push, "date");
+                if (date is { } d && (latest is null || d > latest))
+                {
+                    latest = d;
+                }
+            }
+
+            return latest;
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
gautam-achieveai: Must — `HttpClient` timeouts commonly surface as `TaskCanceledException`/`OperationCanceledException` even when the caller's cancellation token was not canceled. This filter rethrows all such exceptions, so one timed-out `/pushes` lookup faults `Task.WhenAll` and aborts the whole ADO poll instead of following the documented fail-open behavior (unknown recency keeps the PR). Re-throw only when `cancellationToken.IsCancellationRequested`; otherwise log and return null, and add timeout-vs-caller-cancellation tests.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,195 @@
...
+            logger?.LogWarning(
+                "Slot hygiene at {StorePath}: `git submodule update --force` (restore submodules to the recorded "
+                    + "gitlink) failed ({Classification}); {Action}: {Stderr}",
+                storePath,
+                restoreCorrupt ? "corrupt" : "transient/unknown",
+                restoreCorrupt ? "re-cloning" : "leaving it to the status probe + next lease",
+                restore.Stderr);
+        }
+
+        var submodules = await git.RunAsync(
```
gautam-achieveai, line 128: Should — Once `reset`, `clean`, or a corrupt restore has already made the verdict `NeedsReclone`, the method still runs `submodule foreach` and `rev-parse`. These extra git operations can mutate a store that is already doomed, add latency, and introduce cancellation/error paths without changing the result. Return `NeedsReclone` before the submodule cleanup/probe when those prerequisite results fail.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/AdoPrProviderTests.cs
```diff
@@ -248,4 +269,154 @@ public async Task GetPrState_sends_the_single_pr_request_ado_requires()
...
+
+        using var handler = new ConcurrencyTrackingHandler(prsJson);
+        var provider = new AdoPrProvider(
+            new HttpClient(handler),
+            new FakeOAuthTokenProvider("ado", "ado-token-abc"),
+            LoggerFactory.CreateLogger<AdoPrProvider>());
+
+        var page = await provider.ListOpenPullRequestsAsync(Request(recencyCutoff: cutoff), CancellationToken.None);
+
+        page.PullRequests.Should().HaveCount(oldPrCount);
```
gautam-achieveai, line 372: Should — The bounded-concurrency test proves only that at least two requests overlap and that the peak is <= 6; it does not prove all 20 eligible old PRs receive a `/pushes` lookup. An implementation that processes only a subset would still pass. Track/assert exactly `oldPrCount` push requests and verify each returned descriptor has the expected `UpdatedAt`.

gautam-achieveai (commented):
Re-review of PR 215 across 2 changed files since b06ee14 — no new Must, Should, or Consider findings.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        // 4. Early cleanliness gate. If the superproject reset/clean OR the submodule restore reported failure,
+        //    the store CANNOT be confirmed pristine — re-clone. Do NOT run the remaining ops on a doomed store,
+        //    and do NOT fall through to `status --porcelain`, which may not surface a submodule left off its
+        //    recorded gitlink when submodule-ignore settings hide it. A restore failure is gated even when it
+        //    looks "transient": `submodule update --force` only needs a network fetch when the recorded gitlink
+        //    object is ABSENT locally, which means the warm slot is genuinely stale relative to store main — a
+        //    re-clone is the correct refresh and self-heals once the cause clears (a retry, not an infinite loop;
+        //    corrupt COMMITTED content is handled by the foreach-tolerance in step 5, never by a reclone).
+        if (!reset.Succeeded || !clean.Succeeded || !restore.Succeeded)
```
gautam-achieveai: Must — Any failed `git submodule update --force`, including DNS/TLS/throttling/auth failures, is converted to `NeedsReclone`. `ReviewSlotPreparer` then throws `SlotNeedsRecloneException` and the executor deletes/reclones the persistent store, discarding a healthy warm store and amplifying a remote outage; this also bypasses `GitFailureClassifier`'s transient policy. Preserve the failure classification: return a non-clean retryable result for transient/unknown failures and reserve recloning for classified structural/corrupt failures, with an executor test asserting a transient restore does not call `RecloneStoreAsync`.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,184 @@
...
+        //    prior lease left the reviewed submodule checked out at the PR head, which the superproject sees as a
+        //    moved pointer (`git status` reports it dirty). The `submodule foreach` (step 5) only resets each
+        //    submodule to its OWN (PR-head) HEAD, so without this restore the slot would look dirty on every warm
+        //    lease and be needlessly re-cloned. `--force` alone (no --init / --recursive) uses locally-present
+        //    objects — no fetch in the common case — and touches only already-initialized, .gitmodules-registered
+        //    submodules, so it skips a committed embedded gitlink with no .gitmodules URL (the PR-11182 wedge).
+        var reset = await git.RunAsync(["-C", storePath, "reset", "--hard"], storePath, ct).ConfigureAwait(false);
+        var clean = await git.RunAsync(["-C", storePath, "clean", "-ffdx"], storePath, ct).ConfigureAwait(false);
+        var restore = await git.RunAsync(
+                ["-C", storePath, "submodule", "update", "--force"], storePath, ct)
```
gautam-achieveai: Must — Hygiene restores only top-level submodules here, while the review path recursively initializes submodules. The later recursive `foreach` resets a nested repository to its current HEAD, not the superproject's recorded nested gitlink, so a prior lease can leave a nested submodule at a PR/agent commit; `status --porcelain` may not expose that under submodule-ignore. Restore initialized submodules recursively to their recorded gitlinks (and mirror that in `StripAsync`), then add a nested-submodule regression test.

tests/CodeReviewDaemon.Sample.Tests/Scenarios/AdoPrProviderTests.cs
```diff
@@ -248,4 +269,159 @@ public async Task GetPrState_sends_the_single_pr_request_ado_requires()
...
+        handler.TotalPushes.Should().Be(oldPrCount, "every eligible old PR receives exactly one /pushes lookup");
```
gautam-achieveai, line 373: Should — The changed provider adds a distinct timeout/cancellation branch at `AdoPrProvider.cs:315-327`, but this concurrency test does not exercise it. Add one test where `/pushes` throws `TaskCanceledException` without caller cancellation and assert the poll completes with null recency, plus one where the caller token is canceled and assert cancellation propagates; otherwise either future regression can abort polls or swallow shutdown cancellation unnoticed.

gautam-achieveai (commented):
Round 14 re-review: the restore-path change still has two correctness blockers and one missing regression test.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,186 @@
...
+        //    submodules — checked out at PR-head/agent commits, which the superproject sees as moved pointers
+        //    (`git status` reports dirty). The `submodule foreach` (step 5) only resets each submodule to its OWN
+        //    HEAD, not the recorded (nested) gitlink, so it does NOT fix this. `--recursive --force` (NO --init)
+        //    uses locally-present objects — no fetch in the common case — and touches only already-initialized,
+        //    .gitmodules-registered submodules at every depth, so it skips a committed embedded gitlink with no
+        //    .gitmodules URL (the PR-11182 wedge) and never inits a new/denied submodule.
+        var reset = await git.RunAsync(["-C", storePath, "reset", "--hard"], storePath, ct).ConfigureAwait(false);
+        var clean = await git.RunAsync(["-C", storePath, "clean", "-ffdx"], storePath, ct).ConfigureAwait(false);
+        var restore = await git.RunAsync(
+                ["-C", storePath, "submodule", "update", "--recursive", "--force"], storePath, ct)
```
gautam-achieveai: Must — `git submodule update --recursive --force` can implicitly fetch a missing recorded gitlink for every initialized nested submodule. This runs before `ReviewSlotPreparer` constructs the current run's policy-enforced `SubmoduleInitializer`, through the host runner's broad provider credentials, so a warm slot can contact a retained nested remote outside this review's allow-list (and the recursive expansion increases that scope). Use `--no-fetch` (and ideally `--checkout --force --no-fetch`) in both hygiene paths; let the policy-controlled initializer perform permitted network fetches, and treat missing local objects as a retry/reclone condition. Add a regression test that asserts the no-fetch command shape.

gautam-achieveai (commented):
Round 15 re-review: the recursive hygiene change still has one blocking workspace-boundary issue; the ADO and model-attribution updates are otherwise covered.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        // 4. Early cleanliness gate. If the superproject reset/clean OR the submodule restore reported failure,
+        //    the store CANNOT be confirmed pristine — re-clone. Do NOT run the remaining ops on a doomed store,
+        //    and do NOT fall through to `status --porcelain`, which may not surface a submodule left off its
+        //    recorded gitlink when submodule-ignore settings hide it. A restore failure is gated even when it
+        //    looks "transient": `submodule update --force` only needs a network fetch when the recorded gitlink
+        //    object is ABSENT locally, which means the warm slot is genuinely stale relative to store main — a
+        //    re-clone is the correct refresh and self-heals once the cause clears (a retry, not an infinite loop;
+        //    corrupt COMMITTED content is handled by the foreach-tolerance in step 5, never by a reclone).
+        if (!reset.Succeeded || !clean.Succeeded || !restore.Succeeded)
```
gautam-achieveai: Must — `--no-fetch` correctly prevents hygiene from bypassing the per-review allow-list, but this gate now maps every failed local restore to `NeedsReclone`, including the expected missing-gitlink case and any transient/unknown checkout failure. `ReviewSlotPreparer` turns that verdict into `RecloneStoreAsync`, so a temporary restore problem deletes/reclones the persistent store instead of retrying or handing the permitted fetch to the policy-enforced initializer; the changed test also entrenches this unconditional classification. Preserve the failure kind (for example, a non-clean retryable result for transient/unknown and reclone only for confirmed corruption), and add tests proving a transient/unknown restore does not call `RecloneStoreAsync`.

gautam-achieveai (commented):
Re-review of the two-file hygiene delta: the security bypass is closed, but restore failures are still routed to destructive recloning.

gautam-achieveai (commented):
[Gautam's review bot] Deep incremental review of `742cc342...4c1459d9`: the production `--no-fetch` fix is sound in both hygiene paths, and no new non-duplicate production defect was found. I replied on the existing active security thread because its regression test covers only `EnsureCleanAsync`; `StripAsync` remains unpinned. The separate active thread for the missing real-Git nested-submodule regression remains applicable.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -51,15 +81,23 @@
...
 if (!daemonKeyMissing)
 {
     SandboxCredential.ValidateKeyOrThrow(daemonAppId, daemonAppKey!);
 }
 var daemonCredential = new SandboxCredential(daemonAppId, daemonAppKey ?? string.Empty);
 
+// The already-running sandbox gateway's base URL, resolved once (env overrides config, then the
+// 3000 default). Threaded into every gateway consumer so a profile can set SandboxGateway:BaseUrl
+// and nothing needs CRD_SANDBOX_GATEWAY in env.
+var gatewayBaseUrl = Environment.GetEnvironmentVariable("CRD_SANDBOX_GATEWAY")
```
gautam-achieveai, line 98: Must — This boot `SandboxSessionAdapter` still resolves `CRD_SANDBOX_GATEWAY` independently and defaults to `http://127.0.0.1:8080`, while `gatewayBaseUrl` above resolves environment → profile `SandboxGateway:BaseUrl` → port 3000. With a profile-only gateway setting (or the default), diff-only/legacy paths use a different gateway from the registry/provisioned paths and can fail with connection/session errors. Construct the adapter with the already-resolved `gatewayBaseUrl` and add a profile-only configuration test.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,209 @@
...
+        //    Unknown is "treated as transient"). Either way we do NOT fall through to `status --porcelain`, which
+        //    can hide a submodule left off its recorded gitlink under submodule-ignore settings.
+        if (!reset.Succeeded || !clean.Succeeded)
+        {
+            return HygieneVerdict.NeedsReclone;
+        }
+
+        if (!restore.Succeeded)
+        {
+            var corrupt = GitFailureClassifier.Classify(restore.Stderr) == GitFailureKind.Corrupt;
```
gautam-achieveai: Must — A non-corrupt restore failure returns `NeedsRetry`, but `ReviewSlotPreparer` immediately throws on that verdict before it constructs/runs the policy-enforced `SubmoduleInitializer`. A missing local gitlink object therefore repeats the same `--no-fetch` failure forever and can never reach the permitted fetch path described in the comments; the run is eventually parked even though the store could self-heal. Preserve a distinct missing-object/policy-restore outcome (or otherwise allow the initializer to fetch only the reviewed allow-listed submodule), while keeping auth/network failures retry-only, and add an end-to-end preparer test.

samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs
```diff
@@ -348,6 +366,16 @@ public async Task ReleaseReviewLeaseAsync(long runId, CancellationToken cancella
     {
         if (_slotWorkspace is not null && _leasedReviews.TryRemove(runId, out var lease))
         {
+            // Tear the session down (terminating any lingering sub-agent git child + unmounting) BEFORE the slot
+            // returns to the pool, so a cancelled/failed run — which reaches here via the orchestrator's terminal
+            // finally without running the Posted-stage cleanup — can't leave session-side work racing the next
+            // lease's clean-on-entry on the same store (review #180). Best-effort + idempotent: a no-op when no
+            // session was provisioned, and harmless if the Posted stage already destroyed it.
+            if (_options.EnableToolAssistedReview && _provisioner is not null)
```
gautam-achieveai, line 374: Must — `TryRemove` discards the only lease record before `DestroyAsync` and `ReturnAsync` complete. If session disposal or the pool return throws, the slot is leaked and a later `ReleaseReviewLeaseAsync` cannot retry because the dictionary entry is already gone. Keep ownership until cleanup is complete; independently catch/log teardown failures and guarantee `ReturnAsync` in `finally` (or quarantine the slot when teardown cannot be confirmed), with tests for one-time teardown/return failures.

samples/CodeReviewDaemon.Sample/Orchestration/PrOrchestrator.cs
```diff
@@ -46,35 +55,72 @@ public async Task<ReviewRun> RunAsync(ReviewRun seed, CancellationToken cancella
...
                     await _executor.ExecuteStageAsync(stage, run, cancellationToken);
                 }
                 catch (Exception ex) when (ex is not OperationCanceledException)
                 {
                     _store.UpdateReviewRunState(run.Id, run.Stage, WorkflowStatus.RetryPending, run.PrLifecycleState);
+                    // The RetryGovernor bounds ONLY the ContextReady hot-loop (the stuck-slot case it exists for).
+                    // A failure at a later stage (Reviewed/Judged/Posted) is a different, usually self-healing
+                    // problem — e.g. a Posted-stage lock the next lease's clean-on-entry clears — so it must NOT
+                    // consume the context-retry budget or park otherwise-recoverable work.
+                    if (stage == ReviewStage.ContextReady)
```
gautam-achieveai, line 98: Must — Retry governance is applied only when `stage == ContextReady`. Persistent failures in `Reviewed`, `Judged`, or `Posted` are still marked `RetryPending` and retried on every poll with no backoff or park, so provider/model/posting/retention outages can recreate the hot-loop this feature is meant to eliminate. Apply bounded governance to later stages too (with separate budgets if needed), and test a persistent `Posted` failure.

samples/CodeReviewDaemon.Sample/Orchestration/AdoPrProvider.cs
```diff
@@ -131,6 +167,167 @@ private static string CommitId(JsonElement pr, string property) =>
...
+    private async Task<(DateTimeOffset? UpdatedAt, DateTimeOffset? RecencyCreatedAt)[]> ResolveRecencySignalsAsync(
+        IReadOnlyList<RawAdoPr> prs,
+        string org,
+        string? project,
+        string repo,
+        DateTimeOffset? cutoff,
+        CancellationToken cancellationToken)
+    {
+        using var gate = new SemaphoreSlim(MaxRecencyLookupConcurrency);
+        var tasks = prs.Select(pr => ResolveOneRecencyAsync(pr, gate, org, project, repo, cutoff, cancellationToken));
```
gautam-achieveai, line 208: Should — The semaphore caps simultaneous `/pushes` calls but the provider still repeats one lookup for every old active PR on every poll, even when its head/ref is unchanged. At scale this creates sustained request volume and repeated token/JSON work (and repeats after lookup failures). Cache successful/indeterminate results with a bounded TTL keyed by repo/ref/head SHA, invalidating naturally when the head changes, while preserving caller cancellation.

gautam-achieveai (commented):
Re-review of the current head found four release-blocking recovery/delivery issues and one retry-performance concern; prior restore classification was narrowed but still cannot recover missing local gitlink objects.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        RemoveStaleLocks(Path.Combine(storePath, ".git"));
+        await git.RunAsync(["-C", storePath, "reset", "--hard"], storePath, ct).ConfigureAwait(false);
+        await git.RunAsync(["-C", storePath, "clean", "-ffdx"], storePath, ct).ConfigureAwait(false);
+        // Restore ALL submodule checkouts (top-level + nested) to the recorded gitlink (see EnsureCleanAsync
+        // step 3) so the slot is left pristine for the next lease instead of pinned at this review's PR head.
+        // DenyNetworkArgs (+ --no-fetch) is REQUIRED (same reason as step 3): hygiene must never contact a remote
+        // through the host's broad credentials outside a review's policy-enforced allow-list.
+        await git.RunAsync(
+                ["-C", storePath, .. DenyNetworkArgs,
```
gautam-achieveai, line 186: Should — The new transport guard is applied to `EnsureCleanAsync`, but this `StripAsync` call has no behavioral test proving the success-path cleanup also cannot clone/fetch a deinitialized or missing-object submodule. Add a real-git regression covering `StripAsync` (including its failure handling), not only the fake-runner argv assertion, so the two hygiene paths cannot drift.

gautam-achieveai (commented):
Re-review round 18: transport denial is improved, but the missing-gitlink recovery path remains unresolved; one cleanup-path regression test is still missing.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+            if (GitFailureClassifier.Classify(restore.Stderr) == GitFailureKind.Corrupt)
+            {
+                logger?.LogWarning(
+                    "Slot hygiene at {StorePath}: submodule restore failed with CORRUPTION; re-cloning: {Stderr}",
+                    storePath, restore.Stderr);
+                return HygieneVerdict.NeedsReclone;
+            }
+
+            logger?.LogInformation(
```
gautam-achieveai, line 120: Must — A non-corrupt restore failure is now treated as safe and the slot continues, but this method has no current-run policy context. If a previously initialized submodule is denied by the next run, `SubmoduleInitializer` records the denial without removing its worktree, so the denied repository remains readable from the shared store. Do not return a usable/clean verdict until denied submodules are deinitialized or policy-aware verification proves they are absent; add a two-lease allow-then-deny regression test.

samples/CodeReviewDaemon.Sample/Workspace/SlotHygiene.cs
```diff
@@ -0,0 +1,244 @@
...
+                storePath, submodules.Stderr);
+        }
+
+        // Status probe. `--ignore-submodules=all`: submodule state (a moved/stale/missing pointer or uncommitted
+        // content) is deliberately NOT gated here — the review re-checks-out every submodule with permitted
+        // fetches, so submodule state between leases can't cross into it, and gating on it would drive the exact
+        // warm-slot re-clone churn this path exists to avoid (and mask the DenyNetworkArgs-blocked restore as a
+        // spurious reclone). Only leftover SUPERPROJECT (non-submodule) state means contamination that survives.
+        var status = await git.RunAsync(
+                ["-C", storePath, "status", "--porcelain", "--ignore-submodules=all"], storePath, ct)
```
gautam-achieveai: Must — `--ignore-submodules=all` hides exactly the state this shared-slot hygiene must validate. If `submodule foreach` fails for a non-corrupt reason (for example permissions/read-only files), the code logs and proceeds, while `ReviewSlotPreparer` only force-checks the reviewed repo and does not remove arbitrary untracked files in submodules. A later lease can therefore see prior-review files or a checkout at the wrong gitlink. Fail closed except for the explicitly recognized embedded-gitlink case, or clean/verify the reviewed tree with a non-ignored status check before declaring the slot usable.

tests/CodeReviewDaemon.Sample.Tests/Workspace/ReviewSlotPreparerTests.cs
```diff
@@ -133,12 +133,171 @@ public async Task PrepareAsync_ReturnsThePosixJoinedPreparedCheckoutPaths()
...
+        // NON-corruptly (a missing local object / deinit'd submodule). EnsureCleanAsync proceeds (Clean) — the
+        // review re-establishes submodules with permitted fetches — so PrepareAsync must NOT throw
+        // SlotNeedsRecloneException (which the executor turns into a destructive delete + RecloneStoreAsync).
+        runner.OnArgvContains(
+            "submodule update --recursive --no-fetch",
+            new SandboxCommandResult(1, string.Empty, "fatal: Unable to checkout 'deadbeef' in submodule path 'repos/X'"));
+        var preparer = new ReviewSlotPreparer(
+            new GitRunner(runner), SeedGitmodules(slot.StorePath), "github", NullLoggerFactory.Instance);
+
+        var thrown = await Record.ExceptionAsync(async () => await preparer.PrepareAsync(
```
gautam-achieveai, line 221: Should — This regression test accepts either success or any exception other than `SlotNeedsRecloneException`, so it passes even when preparation fails downstream and does not establish the claimed recovery behavior. Assert `PrepareAsync` succeeds and verify the reviewed/nested submodule state (prefer a real-git fixture with a failed local restore), so the test would fail on an unsafe or incomplete recovery.

gautam-achieveai (commented):
Re-review of the four-file hygiene delta: the retry-loop change still permits shared-slot submodule state to cross policy and cleanup boundaries.

tests/CodeReviewDaemon.Sample.Tests/Workspace/SlotHygieneTests.cs
```diff
@@ -0,0 +1,391 @@
...
+    [Fact]
+    public async Task EnsureClean_does_not_clone_a_deinitialized_submodule_over_the_network()
+    {
+        // REAL-GIT regression for the implicit-clone hole (an argv-only test cannot detect a clone): a prior lease
+        // can leave a submodule registered (URL in .git/config) with its worktree + .git/modules gitdir removed.
+        // `submodule update` (even --no-fetch) then drives `submodule--helper clone` → git clone through the host's
+        // broad credentials, outside the review's allow-list. Hygiene must NOT clone it (DenyNetworkArgs denies all
+        // transports) — it proceeds (Clean, a non-corrupt restore failure is non-fatal), never cloning.
+        var runner = NewHostGitRunner();
+        var (super, sub) = await SetupDeinitializedSubmoduleStoreAsync(runner);
```
gautam-achieveai, line 277: **Should** — This real-git assertion does not prove `StripAsync` supplied the network-denial guard. The fixture uses a local `file://` submodule, and `GitRunner` already applies `protocol.file.allow=never`, so the submodule remains uncloned even if `StripAsync` loses its `DenyNetworkArgs`. Use a transport not independently blocked by `GitRunner` (for example, a loopback HTTP server that records requests), or assert the exact `submodule update` argv includes the full `protocol.*.allow=never` policy; make the test fail when the `StripAsync` guard is removed.

gautam-achieveai (commented):
Round 20 re-review: the delta adds coverage, but the new StripAsync security regression test is not behaviorally isolated and can pass without the production guard.

PR #217, opened 2026-07-20
Per-session webhook secrets, DeepSeek provider support, and Actions-log egress hosts

### LINKED ISSUE #216: Per-session webhook secrets, DeepSeek provider support, and Actions-log egress hosts
Bundles a few related fixes that landed together on this branch:

- **Per-session webhook secrets**: replace the single process-wide `AuthSharedSecret` (which silently 401s every in-flight sandbox session's egress-auth/context-discovery webhooks on every app restart) with `SessionSecretStore`, a per-session secret persisted to disk. Restart-survival and cross-session isolation both fall out of this for free.
- **Anthropic-compatible provider families**: `AnthropicCompatProviders` + `ProviderRegistry` additions let any Anthropic-compatible vendor (e.g. DeepSeek) be added to the model dropdown purely via env vars (`ANTHROPIC_COMPAT_PROVIDERS` + per-family `_ANTHROPIC_URL`/`_APIKEY`/`_MODELS`), no code change needed per vendor/model.
- **GitHub Actions run-log egress hosts**: `OAuthProviderHosts` was missing two hops in GitHub's Actions run-log download redirect chain (`results-receiver.actions.githubusercontent.com` and the per-shard blob host), so egress-gated sandbox sessions couldn't download workflow run logs.
- Assorted fixes found while stabilizing the full-solution test suite: a `ProcessLauncher.Tests` hang (unread redirected stdout/stderr pipe blocking child exit), a client-side race where the mount-time conversation list refresh could silently drop a just-created unsaved conversation, and a stale tool name in a sandbox git-clone E2E test.

gautam-achieveai:
Closes #216

## Summary

- **Per-session webhook secrets** (`SessionSecretStore`): replaces the single process-wide `AuthSharedSecret` that silently 401'd every in-flight sandbox session's egress-auth/context-discovery webhooks on every app restart. Secrets now persist to disk per session id, giving restart-survival and cross-session isolation. Both webhook controllers (`AuthWebhookController`, `ContextDiscoveryController`) and `SandboxSessionRegistry` (LmStreaming.Sample + CodeReviewDaemon.Sample) move to the new store; `AuthSharedSecret` is deleted.
- **Anthropic-compatible provider families** (`AnthropicCompatProviders`, `ProviderRegistry`): any Anthropic-compatible vendor (e.g. DeepSeek) can now be added to the model dropdown purely through env vars (`ANTHROPIC_COMPAT_PROVIDERS` + per-family `_ANTHROPIC_URL`/`_APIKEY`/`_MODELS`) — no code change needed per vendor or model.
- **GitHub Actions run-log egress hosts** (`OAuthProviderHosts`): added the two missing hops in GitHub's Actions run-log download redirect chain (`results-receiver.actions.githubusercontent.com` and the per-shard blob host) so egress-gated sandbox sessions can download workflow run logs.
- **Sandbox repro script**: added `workspace-agent-sandbox-repro.mjs`, a reusable Playwright script for reproducing Workspace Agent sandbox issues.
- Three smaller fixes found while stabilizing the full-solution test suite:
  - `ProcessLauncher.Tests` hang — tests waited on process exit without draining the redirected stdout/stderr pipes; `dotnet --info`'s output is close to the pipe buffer capacity, so the unread pipe blocked the child's write and the wait never completed.
  - Client-side race in `useConversations.ts` — the mount-time conversation list refresh could overwrite a just-created, not-yet-persisted conversation if the fetch resolved after the local add.
  - Stale `sandbox-Bash` tool name in `SandboxGitCloneInstructionChainTests` — the gateway is the sole MCP server in that scenario, so its tools are exposed under their natural (unprefixed) names.

## Test plan

- [x] `dotnet test LmDotnetTools.sln --no-build` — full solution green, modulo two known environment-specific, non-deterministic failures unrelated to this branch's changes (confirmed via 100/100 recent CI runs never hitting either): a tight 5s `TaskCompletionSource` wait in `LmMultiTurn.Tests.DeferredToolExecutionTests` under heavy local concurrent load, and `SandboxGitCloneInstructionChainTests`'s `[SkippableFact]` gateway-adoption path (only runs against a live, externally-managed sandbox gateway + GitHub sign-in; always skips in CI).
- [x] New `SessionSecretStoreTests`, `AnthropicCompatProvidersTests`, `ProviderRegistryTests`, `WebToolRegistrationPolicyTests` cover the new behavior.


src/LmAgentInfra/Auth/SessionSecretStore.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// <summary>
+    /// Maps a session id to its on-disk file path, sanitizing the id into a safe file name.
+    /// </summary>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="sessionId"/> is null/whitespace or contains no characters
+    /// that survive sanitization (which prevents path traversal via separators or <c>..</c>).
+    /// </exception>
+    private string GetFilePath(string sessionId)
+    {
+        var fileName = SanitizeSessionId(sessionId);
```
gautam-achieveai: Must — This sanitization is lossy: distinct valid session IDs that differ only by discarded characters or case can resolve to the same `.secret` path. A later session can overwrite/read the other session's secret, violating the documented cross-session authentication isolation. Encode the full ID reversibly or use a collision-resistant digest, and add collision-pair tests.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1022,6 +1032,12 @@ private async Task<SandboxSession> CreateSessionAsync(
             // destroy, discovery) resolve the SAME credential via CredentialFor, not the process
             // default (relevant once M2 lets callers create under a per-caller credential).
             _sessionCredentials[session.SessionId] = effectiveCredential;
+            // Persist BEFORE returning: a webhook call for this session id must never race ahead of
+            // its secret hitting disk. If this throws, let it propagate — the surrounding
+            // Lazy<Task<SandboxSession>> single-flight fault/retry already handles exceptions in this
+            // exact window; swallowing it would instead silently produce a session whose webhooks can
+            // never authenticate.
+            await _sessionSecretStore.SaveAsync(session.SessionId, sessionSecret, ct).ConfigureAwait(false);
```
gautam-achieveai: Must — The reverse and credential maps are published before `SaveAsync` completes. If persistence is cancelled or fails after the gateway create succeeds, the lazy creation faults but the remote session and these maps remain; a retry can create another session while discovery/liveness sees stale state. Roll back exact map entries and best-effort destroy the newly-created gateway session before rethrowing, with a failure-injection test.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-egress-auth-test.mjs
```diff
@@ -0,0 +1,203 @@
...
+//      real GitHub "Requires authentication" body means network egress works but token injection is
+//      NOT happening. A connect/resolve/timeout failure means network is locked down or misconfigured.
+async (page) => {
+  const BASE = 'http://127.0.0.1:5050';
+  const WORKSPACE_ID = '192a3465-67a1-4945-9323-44c2168aeb2b'; // "LmDotnetTools"
+  const PROVIDER_ID = 'test'; // "Test (Mock)"
+  const MODE_ID = 'workspace-agent'; // "Workspace Agent"
+
+  const bashCommand =
+    "env | grep -i -E 'proxy|ssl_cert' ; echo '---' ; curl -sS -w '\\nHTTP_STATUS:%{http_code}\\n' https://api.github.com/user --max-time 30";
```
gautam-achieveai: Must — `env | grep -i -E 'proxy|ssl_cert'` prints complete HTTP_PROXY/HTTPS_PROXY values into tool output, and those values can contain proxy credentials. Check only whether the variables are set or redact their values before emitting diagnostic output.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -838,6 +843,7 @@ private void InvalidateSession((string WorkspaceId, string AppId) key, SandboxSe
         {
             DecrementSessionRefAndMaybeDispose(credential);
         }
+        await _sessionSecretStore.RemoveAsync(session.SessionId, ct).ConfigureAwait(false);
```
gautam-achieveai: Should — Invalidation removes the main maps and secret but does not clear `_subAgentBindings`, `_sessionThreads`, or `_discoverySeen`, unlike explicit destruction. A gateway-evicted session therefore retains session-ID-indexed state until registry disposal. Centralize lifecycle eviction and clear every per-session collection.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1116,6 +1133,7 @@ public async ValueTask DisposeAsync()
             }
 
             await DestroySessionAsync(entry.Value.Result, CancellationToken.None).ConfigureAwait(false);
+            await _sessionSecretStore.RemoveAsync(entry.Value.Result.SessionId, CancellationToken.None).ConfigureAwait(false);
```
gautam-achieveai: Should — `DisposeAsync` now awaits secret-file removal inside the session loop without isolating filesystem failures. One `RemoveAsync` exception aborts disposal before later sessions, clients, and transports are cleaned up, contradicting the best-effort contract. Make secret cleanup best-effort and guarantee final resource cleanup in a `finally` path.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs
```diff
@@ -0,0 +1,72 @@
+// workspace-agent-sandbox-repro.mjs — single-call repro for the "Workspace Agent" mode session
+// creation failure against a REAL sandbox gateway + REAL Copilot provider (no mocks). Run with:
+//
+//   browser_run_code_unsafe({ filename: "samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs" })
```
gautam-achieveai, line 4: Should — This committed script is explicitly live-only, hardcodes an environment workspace UUID/provider, and sends a state-changing prompt to a real Copilot/sandbox environment. Remove it from shipped sample code or parameterize it with explicit confirmation/dry-run safeguards and document the manual-only lifecycle.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
```diff
@@ -0,0 +1,174 @@
...
+//   4. A second, meaningfully different instruction-chain prompt (multi-turn + reasoning + parallel
+//      tool calls) sent in the SAME conversation streams to completion and renders thinking-pills,
+//      more tool-call pills, and final assistant text — proving multi-turn context works.
+//   5. One screenshot is captured for the record.
+//
+// Prompts are taken verbatim from PromptExamples.md ("Calculate Tool" and "Multiturn Example").
+// Do NOT drive this with snapshot->act->screenshot loops. Assert only DETERMINISTIC, browser-observable
+// state (DOM/data-testid).
+async (page) => {
+  const HTTPS_URL = 'https://lmstreaming.bhakars.internal/';
```
gautam-achieveai: Should — The script hardcodes a private hostname and later writes the screenshot to a machine-specific `B:/...` path, so it is not portable and can write into an unexpected checkout. Read URL/output from environment or arguments and use a portable temporary/output location.

tests/LmStreaming.Sample.Tests/Auth/AuthWebhookControllerForwarderTests.cs
```diff
@@ -102,10 +102,18 @@ private static AuthWebhookController CreateController(
         IAuthResolutionPolicy policy,
         IAuthWebhookForwarder forwarder)
     {
-        var authOptions = new AuthOptions { Webhook = new WebhookOptions { GatewaySharedSecret = Secret } };
+        var authOptions = new AuthOptions();
+        var sessionSecretStore = new SessionSecretStore(
```
gautam-achieveai, line 106: Consider — Per-session store construction/seeding is repeated across many test files. A shared test store fixture/helper could own temporary-directory cleanup and seeding, reducing drift when the store format or setup changes.

gautam-achieveai (commented):
Reviewed PR 217 across 58 files — 3 Must, 4 Should, 1 Consider.

src/LmAgentInfra/Auth/OAuthProviderHosts.cs
```diff
@@ -10,7 +10,18 @@ internal static class OAuthProviderHosts
...
+        // redirect chain (api.github.com -> results-receiver -> a per-shard SAS-signed blob URL).
+        // The blob host is exact-listed (not *.blob.core.windows.net) since that suffix covers
+        // every Azure Blob Storage tenant, not just GitHub's; if GitHub rotates to a new shard,
+        // this will need the new host added.
+        ["github"] = [
+            "github.com",
+            "api.github.com",
+            "codeload.github.com",
+            "results-receiver.actions.githubusercontent.com",
+            "productionresultssa7.blob.core.windows.net",
```
gautam-achieveai: [Gautam's review bot] **[P1] Keep redirect egress separate from OAuth injection.** Every host in this list is placed on the `github` network rule with `authProvider: "github-auth"`, and the webhook returns the GitHub access token in an `Authorization` header for every non-`github.com` destination. These additions therefore send the user's GitHub token to the Actions receiver and Azure SAS blob request even though the redirect URL is already authorized. The exact `sa7` entry also blocks redirects to any other results shard; GitHub documents `*.blob.core.windows.net` for log/artifact/cache transfers. Please add network-only rules for the redirect chain (with appropriately strict host/path constraints) and keep OAuth injection limited to GitHub API/git hosts.

gautam-achieveai: [gautam-achieveai's bot] Fixed in 38802893. Split `OAuthProviderHosts`: the `github` provider list now holds only the OAuth-injected API/git hosts (`github.com`, `api.github.com`, `codeload.github.com`). The Actions run-log redirect-chain hosts moved to a new **network-only** allow rule (id `github-egress`, no `authProvider`) sourced from `GithubEgressOnlyHosts` = [`results-receiver.actions.githubusercontent.com`, `*.blob.core.windows.net`], so the gateway permits egress there without ever calling the auth webhook — the GitHub token is never injected to those hosts. Since they're out of the injectable `github` list, the webhook's `IsAllowed` gate also refuses token injection to them (defense-in-depth). Used the documented `*.blob.core.windows.net` suffix rather than pinning the `sa7` shard (safe: this rule injects no credential). Kept the rule host+443-scoped to match the existing rule shape; didn't add path constraints since the redirect paths aren't stable/documented and removing token injection is the actual fix. Regression test: `Github_token_injection_scoped_to_api_hosts_only_redirect_chain_is_network_only`.

src/LmAgentInfra/Auth/SessionSecretStore.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in c8d49916. Replaced the lossy `SanitizeSessionId` (lowercase + drop everything outside [a-z0-9_-]) with a SHA-256 hex digest of the id's exact UTF-8 bytes (`ToFileNameStem`). Every distinct session id now maps to a distinct 64-char stem, so ids that differed only by case or by a stripped character can no longer share a `.secret` file — cross-session isolation is restored. The hex output is still inherently path-traversal-proof (no separators/dots). Added collision-pair tests `PreviouslyCollidingSessionIds_NowMapToDistinctSecrets` proving "Session-A"/"session-a" and "session.a"/"sessiona" now map to distinct files and never cross-match.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 9ba5d97c. In `CreateSessionAsync`, the reverse/credential map publish + `SaveAsync` are now a single try/catch unit. On any failure (incl. cancellation) it best-effort destroys the just-created gateway session FIRST — while `_sessionCredentials` still holds the entry, so `CredentialFor` resolves the owner app id for the DELETE (a default/foreign id would 404 and leak it) — then rolls back exactly the entries this attempt added (exact key+value removal, so a concurrent recreation is never clobbered) before rethrowing. The existing `finally` still releases the client refcount. Failure-injection test `Create_rolls_back_maps_and_destroys_remote_session_when_secret_persist_fails` (asserts the DELETE was issued + maps are clean) plus `Create_retries_into_a_fresh_session_after_a_rolled_back_attempt`.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-egress-auth-test.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 5c7c71d3. The diagnostic Bash command now pipes the proxy/ssl_cert env grep through `sed -E 's/=.*/=<redacted>/'`, so it emits only `NAME=<redacted>` — the set/unset precondition (HTTP_PROXY/HTTPS_PROXY present?) is still diagnosable, but the raw values (which can carry proxy credentials) never land in tool output. Header comment updated to document the redaction.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 9ba5d97c. Extracted a shared `EvictSessionStateAsync` helper that clears EVERY per-session-id collection — `_sessionsById`, `_subAgentBindings`, `_sessionThreads`, `_discoverySeen`, `_sessionCredentials` (with client-refcount release) and the persisted secret. Both `InvalidateSessionAsync` (liveness-evict path) and `DestroyWorkspaceSessionAsync` (explicit-teardown path) now funnel through it, so a gateway-evicted session no longer retains sub-agent/thread/discovery state until disposal, and the two lifecycle paths can't diverge if a new per-session collection is added later. Destroy still issues the gateway DELETE before calling the helper (so the credential is still resolvable). Test `Invalidation_on_recreate_clears_all_per_session_state`.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 9ba5d97c. `DisposeAsync` now wraps the per-session `_sessionSecretStore.RemoveAsync` in a try/catch (log + continue), so one filesystem failure no longer aborts disposal of later sessions. The session-loop + collection clears are additionally wrapped in a try/**finally**, and the per-credential client / transport / shared `HttpClient` disposal moved into that finally, so final resource cleanup is guaranteed even if a teardown step throws.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 5c7c71d3 (kept the script — it's this PR's headline repro). It's now **DRY-RUN by default**: it exercises the workspace/provider/mode selection wiring but does NOT send the state-changing prompt unless `LMSTREAMING_REPRO_CONFIRM` is explicitly opted in (1/true/send/yes). The workspace UUID / provider / mode / URL / prompt are all parameterized via env vars (former values kept as documented defaults), so nothing environment-specific is hardcoded. Added a header comment documenting it as a manual-only, live-environment diagnostic (not part of the automated suite) with the safety/opt-in lifecycle.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 5c7c71d3. The URL is now read from env vars (`LMSTREAMING_HTTPS_URL` / `LMSTREAMING_URL`) with the former values as documented defaults instead of a hardcoded private hostname, and the screenshot is written to a portable location — `path.join(os.tmpdir(), 'smoke-test-egress-result.png')`, overridable via `LMSTREAMING_OUTPUT_DIR` — instead of the machine-specific `B:/...` path, so it can never write into an unexpected checkout. The resolved path is returned in the result for visibility.

tests/LmStreaming.Sample.Tests/Auth/AuthWebhookControllerForwarderTests.cs
gautam-achieveai: [gautam-achieveai's bot] Won't Fix (this iteration). The repeated setup is a benign 3-line inline store construction (`new SessionSecretStore(Path.Combine(Path.GetTempPath(), "lmstreaming-test-secrets", Guid.NewGuid().ToString("N")), NullLogger<>.Instance)`) with no cleanup logic of its own — the temp dirs are left to OS temp reclamation, matching every sibling test file. A shared helper that also owns temp-dir cleanup (the value the suggestion is really after) pushes toward an `IDisposable`/fixture pattern, and a pure factory helper would only save ~2 lines/site while touching ~10 files — churn that broadens this security-focused PR's blast radius for marginal gain. Deferring per the repo's "three similar lines beats a premature abstraction" guidance; happy to do it as a standalone test-hygiene cleanup PR if desired.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -826,18 +832,40 @@ private void InvalidateSession((string WorkspaceId, string AppId) key, SandboxSe
...
+        _ = _sessionThreads.TryRemove(session.SessionId, out _);
+        _ = _discoverySeen.TryRemove(session.SessionId, out _);
         // Drop the session's credential and release its client refcount (evicting+disposing the per-
         // credential client when this was its last session). Guarded on the TryRemove so an already-
         // invalidated session never double-decrements.
         if (_sessionCredentials.TryRemove(session.SessionId, out var credential))
         {
             DecrementSessionRefAndMaybeDispose(credential);
         }
+        await _sessionSecretStore.RemoveAsync(session.SessionId, ct).ConfigureAwait(false);
```
gautam-achieveai: Must — Eviction removes the in-memory credential/session state before awaiting `RemoveAsync` with the caller's cancellable token. If cancellation occurs while waiting for the store gate or deleting the file, the gateway session is considered evicted but its file-backed secret remains valid, so a later callback can still authenticate a supposedly dead session. Complete secret revocation with `CancellationToken.None` after the committed eviction, and make filesystem failure best-effort/logged rather than allowing partial cleanup.

gautam-achieveai (commented):
Re-review of PR #217 at 5c7c71d: the earlier collision/rollback/eviction/disposal and diagnostic portability issues are partly addressed, but cancellation-safe teardown and diagnostic/configuration data handling still need fixes.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1075,22 +1135,16 @@ public async Task DestroyWorkspaceSessionAsync(string workspaceId, CancellationT
...
-            _ = _sessionThreads.TryRemove(session.SessionId, out _);
-            _ = _discoverySeen.TryRemove(session.SessionId, out _);
 
-            // Destroy BEFORE dropping the credential: DestroySessionAsync resolves the session's
+            // Destroy BEFORE evicting per-session state: DestroySessionAsync resolves the session's
             // creating credential via CredentialFor(sessionId), so the DELETE must carry the owner's
-            // X-Sbx-App-Id (a foreign/default id would be rejected 404, leaking the gateway session).
+            // X-Sbx-App-Id (a foreign/default id would be rejected 404, leaking the gateway session) —
+            // and CredentialFor reads _sessionCredentials, which EvictSessionStateAsync clears.
             await DestroySessionAsync(session, ct).ConfigureAwait(false);
```
gautam-achieveai: Must — This method removes the `_sessions` entry before teardown, then passes the caller token into both destruction and eviction. A cancellation can therefore forget the session while leaving the remote sandbox and/or secret file behind; `DestroySessionAsync` also swallows cancellation, so the caller cannot reliably retry. Check cancellation before removing the cache entry, then perform committed teardown with `CancellationToken.None` (and isolate cleanup failures) so remote and persisted state are not orphaned.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
```diff
@@ -1107,43 +1161,67 @@ public async ValueTask DisposeAsync()
...
+                remainingClients = [.. _clients.Values];
+                _clients.Clear();
+            }
 
-        foreach (var entry in remainingClients)
-        {
-            entry.Client.Dispose();
-            entry.Transport.Dispose();
-        }
+            foreach (var entry in remainingClients)
```
gautam-achieveai: Should — The new `finally` guarantees this block is entered, but one exception from `Client.Dispose()` or `Transport.Dispose()` aborts the loop and skips later transports and `_httpClient.Dispose()`. That still violates the documented best-effort disposal contract. Wrap each per-entry disposal independently and put shared HTTP-client disposal in a nested `finally`.

samples/LmStreaming.Sample/Services/AnthropicCompatProviders.cs
```diff
@@ -0,0 +1,80 @@
...
+                    "Skipping Anthropic-compatible family {FamilyKey}: one or more of its _ANTHROPIC_URL/_APIKEY/_MODELS env vars is not set.",
+                    familyKey
+                );
+                continue;
+            }
+
+            var modelNames = modelNamesRaw.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
+            foreach (var modelName in modelNames)
+            {
+                var id = Slugify(modelName);
```
gautam-achieveai, line 76: Should — Slugification is not collision-safe: model names such as `kimi-2.5` and `kimi-2-5` generate the same ID, and identical model names from two families collide as well. `ProviderRegistry` silently skips the later entry, making a configured model unavailable (or routing the shared ID through the wrong family). Include the family in the ID or detect duplicate generated IDs and fail/log clearly; add same- and cross-family collision tests.

samples/LmStreaming.Sample/Program.cs
```diff
@@ -1686,6 +1706,28 @@ private static IStreamingAgent CreateAnthropicAgent(ILoggerFactory loggerFactory
...
 
+    /// <summary>
+    ///     Creates an agent for a discovered Anthropic-compatible provider-family model (e.g. DeepSeek),
+    ///     using that model's own base URL/API key rather than fixed env vars — see
+    ///     <see cref="AnthropicCompatProviders.DiscoverFromEnv"/>.
+    /// </summary>
+    private static IStreamingAgent CreateAnthropicCompatAgent(AnthropicCompatModel model, ILoggerFactory loggerFactory)
+    {
+        Log.Information(
+            "Creating {FamilyKey} agent with base URL: {BaseUrl}",
```
gautam-achieveai: Should — `BaseUrl` is environment-controlled and is logged verbatim at Information. If an operator supplies URL user-info or query credentials, this persists secret material in application logs. Log only a validated/sanitized origin (or omit the URL), and avoid logging query/user-info components.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-egress-auth-test.mjs
```diff
@@ -0,0 +1,209 @@
...
+  const BASE = 'http://127.0.0.1:5050';
+  const WORKSPACE_ID = '192a3465-67a1-4945-9323-44c2168aeb2b'; // "LmDotnetTools"
+  const PROVIDER_ID = 'test'; // "Test (Mock)"
+  const MODE_ID = 'workspace-agent'; // "Workspace Agent"
+
+  // `sed 's/=.*/=<redacted>/'` collapses each matched `NAME=value` line to `NAME=<redacted>` so the
+  // presence of HTTP_PROXY/HTTPS_PROXY/SSL_CERT* is diagnosable WITHOUT leaking any (possibly
+  // credential-bearing) proxy value into the captured tool output.
+  const bashCommand =
+    "env | grep -i -E 'proxy|ssl_cert' | sed -E 's/=.*/=<redacted>/' ; echo '---' ; curl -sS -w '\\nHTTP_STATUS:%{http_code}\\n' https://api.github.com/user --max-time 30";
```
gautam-achieveai: Should — Redacting proxy environment variables fixes the prior leak, but this command still emits the complete authenticated GitHub `/user` response. The tool pill and returned diagnostics can therefore contain profile EUII. Use `curl -o /dev/null -w 'HTTP_STATUS:%{http_code}'` (or otherwise discard the body) and assert only status/reachability.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs
```diff
@@ -0,0 +1,99 @@
...
+    await tid(`mode-option-${MODE_ID}`).click();
+    const modeLabel = await tid('mode-selector-button').textContent();
+    record('mode-selected', !!(modeLabel && modeLabel.trim()), { modeId: MODE_ID, modeLabel });
+
+    // SAFETY GATE: the prompt below is state-changing (it asks the live agent to init a git repo +
+    // worktree inside the workspace). Only send it when the caller has explicitly opted in.
+    if (!CONFIRM_SEND) {
+      record('dry-run-skipped-send', true, {
+        note: 'DRY-RUN: selection wiring exercised; state-changing prompt NOT sent. Set LMSTREAMING_REPRO_CONFIRM=1 to send.',
+        promptPreview: PROMPT.slice(0, 80),
```
gautam-achieveai: Consider — The dry-run now safely gates the state-changing prompt, but it still returns the first 80 characters of arbitrary operator-supplied prompt text, and the live path returns complete assistant/error text. Those diagnostics can contain repository data, credentials, or EUII. Return only metadata such as prompt length, terminal status, and boolean presence flags.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
```diff
@@ -0,0 +1,187 @@
...
+//
+// Portability: nothing here is hardcoded to one machine. Override via environment variables:
+//   - LMSTREAMING_HTTPS_URL  (default 'https://lmstreaming.bhakars.internal/') — the Traefik/HTTPS host.
+//   - LMSTREAMING_URL        (default 'http://127.0.0.1:5050/')                — the direct Kestrel origin.
+//   - LMSTREAMING_OUTPUT_DIR (default the OS temp dir)                         — where the screenshot lands.
+async (page) => {
+  const os = await import('node:os');
+  const path = await import('node:path');
+
+  const HTTPS_URL = process.env.LMSTREAMING_HTTPS_URL || 'https://lmstreaming.bhakars.internal/';
```
gautam-achieveai: Consider — The script is parameterized, but its default remains the environment-specific `lmstreaming.bhakars.internal` hostname while the comment claims portability. Require the HTTPS URL, use a local portable default, or skip that attempt when the variable is unset.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `9f266a1f...5c7c71d3` found no additional high-confidence defects beyond the eight active threads already posted on this head.

Verified addressed and resolved: session-secret filename collisions, create/persist rollback, stale invalidation state, disposal secret-removal isolation, GitHub Actions redirect egress/token separation, proxy-value redaction, dry-run gating, and portable screenshot output. The remaining cancellation-safe teardown, disposal isolation, provider-ID/logging, and diagnostic-data findings stay active and should be addressed before merge.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: `EvictSessionStateAsync` now completes the persisted-secret revocation under `CancellationToken.None` after the in-memory eviction is committed, and wraps the `RemoveAsync` in try/catch (log + continue). Once the maps/credential are cleared the session is unreachable, so a cancellation between that point and the file delete — or a filesystem failure — can no longer leave a valid on-disk webhook secret behind for an already-evicted session that a later callback could authenticate. The now-unused `ct` parameter was dropped since eviction is an always-run-to-completion operation (its two callers updated accordingly).

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: `DestroyWorkspaceSessionAsync` now calls `ct.ThrowIfCancellationRequested()` BEFORE removing each entry from `_sessions` (the point of no return), then runs the gateway DELETE and per-session state eviction under `CancellationToken.None`. A cancellation is therefore observed while the cache entry is still intact (leaving it retryable), and once teardown is committed it can no longer forget the session while orphaning the remote sandbox session or its secret file. `DestroySessionAsync` isolates its own gateway failures and the secret removal inside `EvictSessionStateAsync` is best-effort, so one cleanup step failing never skips the rest. Updated the method doc to state the new cancellation semantics.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: each per-entry `Client.Dispose()` and `Transport.Dispose()` is now wrapped in its own try/catch (log + continue) inside the loop, so one throwing disposal no longer aborts the loop and skips the remaining transports. The shared `_httpClient.Dispose()` moved into a nested `finally` around that loop, so it always runs even if a per-entry disposal throws — honoring the documented best-effort disposal contract.

samples/LmStreaming.Sample/Services/AnthropicCompatProviders.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: took the "detect duplicate generated ids and log clearly" option (rather than family-prefixing every id, which would change the existing dropdown/provider-id contract and every persisted thread's provider id). `DiscoverFromEnv` now tracks minted ids in a `Dictionary` keyed with `StringComparer.OrdinalIgnoreCase` — the same comparer `ProviderRegistry` indexes by — and on a collision it logs a clear warning naming both colliding models/families and skips the later entry (first configured entry wins) instead of letting the registry silently drop it. This covers both the same-family slug collision (`kimi-2.5` vs `kimi-2-5`) and the cross-family same-name collision uniformly. Added two tests (`DiscoverFromEnv_DropsSameFamilySlugCollision_KeepingFirstAndWarning`, `DiscoverFromEnv_DropsCrossFamilyIdCollision_KeepingFirstAndWarning`) asserting the dedup outcome and the emitted warning via a capturing `ILoggerFactory`.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: `CreateAnthropicCompatAgent` now parses `BaseUrl` and logs only the validated `scheme://host:port` origin (built explicitly from `Uri.Scheme`/`Host`/`Port`, so no user-info or query components are included), with an `(invalid base URL)` placeholder when it fails to parse. Operator-supplied URL credentials (`user:pass@host` or `?token=...`) can therefore no longer be persisted into application logs.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-egress-auth-test.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed: the diagnostic curl now runs `curl -sS -o /dev/null -w 'HTTP_STATUS:%{http_code}\n' https://api.github.com/user --max-time 30`, discarding the response body so only the status code reaches the tool pill and returned diagnostics — the authenticated GitHub `/user` profile JSON (and any 401 error body) is EUII and is no longer captured. Updated the surrounding comments that previously described asserting on the response body.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed: the dry-run now returns `promptLength` instead of the first 80 characters of the operator-supplied prompt, and the live path returns only metadata about the terminal state (`result`, `hasError`, `hasAssistantText`, plus `errorBannerLength`/`assistantTextLength`) instead of the full assistant/error text. Repository data, credentials, or EUII in the prompt or the live agent's output can therefore no longer leak through the returned diagnostics.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
gautam-achieveai: [gautam-achieveai's bot] Fixed: `LMSTREAMING_HTTPS_URL` no longer defaults to the environment-specific `lmstreaming.bhakars.internal` host. It is now optional with no default; when unset the HTTPS attempt is skipped entirely and the script goes straight to the portable Kestrel origin (`LMSTREAMING_URL`, default `http://127.0.0.1:5050/`). When it is set, that host is tried first with fallback as before. Updated the header/portability comments to match, so no environment-specific hostname is baked in.

tests/LmStreaming.Sample.Tests/Auth/AuthWebhookControllerForwarderTests.cs
gautam-achieveai: [Gautam's review bot] Accepted as deferred. This is test-hygiene cleanup rather than a correctness issue in #216, and extracting it across the suite would unnecessarily broaden this security-focused change. Keeping the current setup is reasonable for this PR.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review complete for `9f266a1f...5c7c71d3`: no additional high-confidence findings.

Verified the OAuth/egress separation, collision-resistant session-secret filenames, create rollback and lifecycle state cleanup, and diagnostic-script hardening. The deferred test-helper suggestion is non-blocking and has been accepted for this PR. No existing resolved threads were reopened and no duplicate comments were added.

gautam-achieveai (commented):
Re-reviewed the 7-file delta since 5c7c71d; prior lifecycle and secret-handling findings are addressed, with one remaining diagnostic URL exposure concern.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
```diff
@@ -0,0 +1,199 @@
...
+
+  try {
+    // 1. Navigate. When LMSTREAMING_HTTPS_URL is set, prefer that Traefik/HTTPS hostname and fall back
+    //    to the direct Kestrel origin on failure. When it is UNSET, skip the HTTPS attempt entirely and
+    //    go straight to the portable Kestrel origin — no environment-specific host is hardcoded.
+    if (HTTPS_URL) {
+      try {
+        await page.goto(HTTPS_URL, { waitUntil: 'domcontentloaded', timeout: 20000 });
+        await tid('chat-input-textarea').waitFor({ timeout: 15000 });
+        urlUsed = HTTPS_URL;
```
gautam-achieveai: Should: `urlUsed` is returned as the complete operator-controlled URL, and the nearby fallback error path also interpolates the raw URL and browser exception. If either configured URL contains user-info or a query token, this diagnostic result can persist it in the captured output. Return only a sanitized origin (scheme/host/port) and structured error metadata (for example error name), never the raw URL or exception string.

gautam-achieveai: [gautam-achieveai's bot] Fixed: `urlUsed` now holds only a sanitized scheme/host/port origin (via a new `sanitizeOrigin` helper), and the nearby fallback error path plus the outer catch capture only the structured error name (`error.name`) instead of the raw URL or exception string — so an operator-supplied URL carrying user-info/query tokens (EUII) or a navigation exception embedding the raw URL can no longer persist in the captured output. Commit 1718920d.

gautam-achieveai (commented):
Re-reviewed PR 217 across 1 file changed since the previous review — no new Must, Should, or Consider findings.

src/LmAgentInfra/Sandbox/SandboxSessionRegistry.cs
gautam-achieveai: [Gautam's review bot] **Still blocking:** the cancellation fix runs deletion to completion, but the filesystem-failure path is not fail-closed. RemoveAsync exceptions are swallowed after the registry state is evicted, while AuthWebhookController authenticates solely through SessionSecretStore.MatchesAsync and never checks that the session is still live. The undeleted session-id/secret pair therefore remains able to authorize token injection after teardown. Please persist a revocation/tombstone state that MatchesAsync checks (or otherwise make revocation durable before committing eviction), and cover a forced-delete-failure callback.

samples/LmStreaming.Sample/Services/AnthropicCompatProviders.cs
gautam-achieveai: [Gautam's review bot] **Residual collision:** seenIds only compares Anthropic-compatible models with each other. ProviderRegistry still silently skips a compat model whose ID matches an existing built-in or discovered Copilot ID (uilder.ContainsKey(model.Id)), so configurations such as openai or a Copilot model name can still vanish without the warning promised by this fix. Detect collisions against the complete provider catalog (or log/fail in the registry) and add built-in/Copilot collision coverage.

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [Gautam's review bot] **Residual URL leak:** the Information log is sanitized, but the same raw model.BaseUrl is passed to AnthropicClient, which logs _baseUrl/ExplicitBaseUrl and the constructed request URL verbatim at Trace (AnthropicClient.cs:55-59, 111-112, 199-200). Enabling trace diagnostics therefore still persists user-info/query credentials. Sanitize at the client logging boundary as well, so every call site is protected.

samples/LmStreaming.Sample/playwright-scripts/workspace-agent-sandbox-repro.mjs
gautam-achieveai: [Gautam's review bot] **Residual diagnostic leak:** the outer catch still records String(e.stack || e). A failed page.goto(BASE) includes the operator-controlled URL in Playwright's stack/message, so user-info or query tokens can still escape through the returned steps. Return only structured metadata such as error name/phase and a sanitized origin.

samples/LmStreaming.Sample/playwright-scripts/smoke-test-egress.mjs
gautam-achieveai: [Gautam's review bot] **Residual diagnostic leak:** navigation exceptions are sanitized, but the console and pageerror listeners still collect full text and append it verbatim to the returned rrors. The configured page can emit its URL or arbitrary application data through console.error/an uncaught exception, bypassing sanitizeOrigin. Record only counts/categories or structured error names, not raw page-provided strings.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review found five residual defects in the latest remediation commits. The blocking issue is that a failed secret-file deletion still leaves an evicted session secret accepted by the auth webhook, preserving a token-minting credential. Four non-blocking fixes remain incomplete: provider-ID collision detection excludes built-in/Copilot IDs, raw compat URLs still reach AnthropicClient trace logs, and both Playwright diagnostics retain raw exception/page-error output. I replied in the existing threads and reopened them to preserve discussion history.

PR #224, opened 2026-07-25
Revobot: dedup re-reviews and add review-quality guardrails

gautam-achieveai:
## Summary
Makes Revobot (the CodeReviewDaemon reviewer) post only genuinely **NEW** findings and act as a responsive PR participant, instead of re-posting a full review every run. Provider-agnostic (GitHub + ADO).

## What changed
**Delta-awareness (host-side).** The daemon fetches the PR's existing review comments through each provider's `IReviewCommentPublisher.ListExistingReviewCommentsAsync` and prepends them to the review input, so dedup is deterministic rather than left to the agent.

**Review/posting guardrails** (in `daemon-prompts.yaml` + the prepended block):
- Review the FULL PR, every changed file.
- Never re-post a finding that already exists as an ACTIVE comment.
- Consider comments from ALL authors (other bots, humans, prior bot names).
- Answer any question directed at Revobot, in-thread.
- Judge thread resolution from the conversation (LLM), not just the provider status flag.
- Split the prepended block into "past reviews" vs "new since last review".

**Refactor.** Behavior-preserving cleanup of the rendering (prose → named const, single `IsNew` predicate, simplified render cap).

## Provider specifics
- **ADO** maps thread status → active/resolved, reads `publishedDate`, carries thread id.
- **GitHub** reads `created_at`/`submitted_at` and `in_reply_to_id ?? id` for thread grouping; defaults to active (REST can't tell resolved for review comments — conservative).

## Testing
- 604 `CodeReviewDaemon.Sample.Tests` green.
- Live-verified on the achieveai daemon reviewing real PR #222: dedup block injected, agent posted **zero** new/empty reviews, correct "No new findings" verdict.

## Related
- Fixes #223
- The GitHub empty-review-spam root cause (per-comment posting) is fixed separately in the shared `code-reviewer:post-pr-review` skill (findings batched into one review).

## Base note
Branch sits a few commits behind `origin/main` (unrelated wi214 work); changes are daemon-only — no expected conflicts.

PR #226, opened 2026-07-26
fix(daemon): stop empty-review spam from GitHub reply guidance

gautam-achieveai:
## What

Fixes false GitHub reply guidance in the CodeReviewDaemon prompt that recreated the empty-review spam this pipeline is meant to eliminate — the last review comment on the now-merged #224.

## The bug

`daemon-prompts.yaml` told the review agent that answering an open thread via `POST /pulls/{pr}/comments/{comment_id}/replies` *"does NOT create a new review, so it is fine."* That is false.

GitHub wraps **every** write to a review-comment endpoint — both standalone `POST /pulls/{pr}/comments` **and** the `/comments/{id}/replies` endpoint — in its own submitted, empty-bodied `COMMENTED` review. On #224 the six in-thread replies produced **six empty reviews**, correlated 1:1 by `pull_request_review_id` — the exact spam the surrounding guardrail forbids.

Second defect on the same guidance: the replies endpoint **cannot** answer a PR-conversation issue comment, because those ids are issue-comment ids, not review-comment ids.

There is no GitHub endpoint that posts a true threaded reply without a review wrapper: the batched `POST /pulls/{pr}/reviews` accepts only `path/line/side/body` (no `in_reply_to`), and the only wrapper-free write on a PR is `POST /issues/{pr}/comments`.

## The fix

- **`daemon-prompts.yaml`** — drop the false "replies are safe" claim; add `/comments/{id}/replies` to the FORBIDDEN list beside standalone `/comments`; route an answer to an open review thread into the **same single batched `POST /reviews`** (an inline `comments[]` entry anchored to the thread's file+line), and route a PR-conversation issue-comment answer to the wrapper-free `POST /issues/{pr}/comments` (or the review summary body).
- **API-level contract test** (`GitHubReviewCommentPublisherTests`) — asserts the host publisher's write surface targets `/issues/{pr}/comments` only and **never** `/replies` or standalone `/pulls/{pr}/comments`. This is the finding's explicit requirement: prompt-text assertions cannot catch this regression, so the invariant is pinned at the HTTP level.
- **Prompt-model test** (`DaemonAgentFactoryTests`) — the rendered GitHub guidance no longer contains the false claim, flags `/replies` as empty-review spam, and routes issue-comment answers through the issues endpoint.

## Testing

- TDD RED→GREEN: the prompt-model test failed on the false guidance before the fix, passes after.
- Full `CodeReviewDaemon.Sample.Tests`: **613 passed** (was 611 + 2 new).
- `dotnet format whitespace` clean; husky format-verify passed on commit.

Addresses the outstanding `Must` finding from #224 (review comment 3651349038).

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Prompts/daemon-prompts.yaml
+++ b/samples/CodeReviewDaemon.Sample/Prompts/daemon-prompts.yaml
@@ -75,15 +75,21 @@ review:
          Never post multiple reviews, never one review per finding, and NEVER submit an EMPTY review (no comments
          AND no meaningful body). If you have nothing new to post, post NOTHING at all.
        - FORBIDDEN — never create standalone review comments with POST .../pulls/{{ pr_number }}/comments (the
-         one-call-per-finding endpoint). GitHub wraps EVERY standalone review comment in its OWN empty-bodied
-         review, so that endpoint turns N findings into N empty reviews — the exact spam we are eliminating. The
-         ONLY write that carries findings is the single POST .../pulls/{{ pr_number }}/reviews above.
+         one-call-per-finding endpoint) and never reply with POST .../pulls/{{ pr_number }}/comments/{comment_id}/replies.
+         GitHub wraps EVERY write to either review-comment endpoint in its OWN submitted, empty-bodied review, so a
+         standalone comment OR a thread reply each turns into one empty review — the exact spam we are eliminating.
+         The ONLY write that carries findings is the single POST .../pulls/{{ pr_number }}/reviews above.
        - Do NOT use the code-reviewer:post-pr-review skill to POST on GitHub: it posts findings one-at-a-time via
          POST .../pulls/{{ pr_number }}/comments and produces exactly those empty reviews. Assemble all findings
          into the one batched POST /reviews call above and make it yourself.
-       To answer an OPEN review thread, reply IN-THREAD — the replies endpoint below ADDS to an existing thread
-       and does NOT create a new review, so it is fine (it is not the forbidden standalone-comment endpoint):
-         POST /repos/{{ gh_owner }}/{{ gh_repo }}/pulls/{{ pr_number }}/comments/{comment_id}/replies  body {"body":"<reply>"}
+       To ANSWER an OPEN review thread (a prior finding you are replying to), do NOT use the /replies endpoint —
+       it makes an empty review (see above). Instead fold your answer into the SAME single batched POST /reviews:
+       add a comments[] entry anchored to that thread's file+line whose body opens by naming the thread it answers
+       (e.g. "Re: <finding> —"). A PR-CONVERSATION issue comment (a question in the PR's main conversation, not on a
+       code line — its id is an issue-comment id, not a review-comment id, so it CANNOT be answered on any review
+       endpoint) is answered EITHER by a line in the POST /reviews summary body OR by ONE wrapper-free comment:
+         POST /repos/{{ gh_owner }}/{{ gh_repo }}/issues/{{ pr_number }}/comments  body {"body":"<answer>"}
+       That issues endpoint is a PR-conversation comment, not a review, so it never creates an empty review.
        {{ end ~}}
        On a re-review, the review input includes a "## Already posted on this PR" section: the existing
        discussion FROM ALL AUTHORS (other bots, humans, and you), grouped into threads (a finding + its replies)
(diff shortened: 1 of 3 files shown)


samples/CodeReviewDaemon.Sample/Prompts/daemon-prompts.yaml
```diff
@@ -75,15 +75,21 @@ review:
...
+         The ONLY write that carries findings is the single POST .../pulls/{{ pr_number }}/reviews above.
        - Do NOT use the code-reviewer:post-pr-review skill to POST on GitHub: it posts findings one-at-a-time via
          POST .../pulls/{{ pr_number }}/comments and produces exactly those empty reviews. Assemble all findings
          into the one batched POST /reviews call above and make it yourself.
-       To answer an OPEN review thread, reply IN-THREAD — the replies endpoint below ADDS to an existing thread
-       and does NOT create a new review, so it is fine (it is not the forbidden standalone-comment endpoint):
-         POST /repos/{{ gh_owner }}/{{ gh_repo }}/pulls/{{ pr_number }}/comments/{comment_id}/replies  body {"body":"<reply>"}
+       To ANSWER an OPEN review thread (a prior finding you are replying to), do NOT use the /replies endpoint —
+       it makes an empty review (see above). Instead fold your answer into the SAME single batched POST /reviews:
+       add a comments[] entry anchored to that thread's file+line whose body opens by naming the thread it answers
```
gautam-achieveai: [Gautam's review bot] **Must - this does not reply to the existing thread.** `POST /pulls/{pr}/reviews` `comments[]` has no `in_reply_to`, so anchoring a new comment at the old path/line (even with `Re:`) creates a new top-level thread. `GitHubReviewCommentPublisher.ThreadIdOf` will group it under its own ID, leaving the original thread active and contradicting the earlier requirement to reply in-thread; an outdated original line can also make the whole batch return 422. GitHub's GraphQL `addPullRequestReviewThreadReply` accepts the existing thread ID and an optional pending-review ID, so replies can retain thread identity while being submitted in one non-empty review. Please use that path (or remove the in-thread contract explicitly) and test the agent-owned posting flow, not the unchanged host summary fallback.

gautam-achieveai (commented):
[Gautam's review bot] Reviewed PR #226 across 3 files - 1 Must. Request changes.

samples/CodeReviewDaemon.Sample/Prompts/daemon-prompts.yaml
gautam-achieveai: Confirmed and fixed in f9375bfa. You're right on all counts: the batched `POST /reviews` `comments[]` array has no `in_reply_to`, so anchoring an entry at the old path/line opens a new top-level thread (`ThreadIdOf` groups it under its own id, original stays open), and an outdated line 422s the whole batch.

Rather than add a GraphQL posting surface to the REST-only agent flow, we removed the in-thread contract: `daemon-prompts.yaml` now states plainly that `comments[]` cannot reply in-thread and carries NEW in-diff findings only. A prior-thread answer goes in the `POST /reviews` summary body (one line per thread, naming it), and a PR-conversation issue-comment answer goes to the wrapper-free `POST /issues/{pr}/comments` — both ride the same single non-empty review, no empty-review wrapper, no 422 from stale anchors.

Covered by a new prompt-model test (`ReviewProfile_Prompt_DoesNotClaimBatchedCommentsCanReplyInThread`) asserting the flawed 'anchored to that thread's file+line' instruction is gone and that `comments[]` is documented as non-replying — this exercises the agent-owned posting guidance, not the host summary fallback. 614 tests green.

gautam-achieveai: [Gautam's review bot] **Must — the rendered prompt still requires the forbidden in-thread path.** The new summary-body route is overridden by the same prompt's lines 35–37 and 108 (`REPLY` / `reply in-thread`), lines 61 and 70–71 (`body` is one short summary and a body-only review is forbidden), plus `DaemonReviewStageExecutor.ExistingCommentsGuidance` lines 1292–1296, which still says `ANSWER it as an in-thread reply` and is explicitly asserted by its test. On a re-review the agent can therefore still choose `/replies` and recreate the empty-review spam; when a thread answer is the only update, the intended meaningful body-only review is explicitly disallowed. Please make the GitHub contract consistent end-to-end: condition shared reply guidance by provider, route GitHub thread answers to the review body, allow a meaningful body-only review for answer-only runs, update `ExistingCommentsGuidance` and its test, and cover both answer-plus-findings and answer-only rendered prompts.

gautam-achieveai (commented):
[Gautam's review bot] Re-review: 1 Must remains; the GitHub posting guidance is still internally contradictory.

PR #231, opened 2026-07-29
feat: add lifecycle and approval hooks to MultiTurn agents

### LINKED ISSUE #227: Add lifecycle and approval hooks to MultiTurn agents
## Summary

Add a shared, versioned lifecycle/approval hook contract across `LmMultiTurn`
and sandbox owners. Trusted in-process components and authenticated S2S
subscribers receive complete final-state events for runs, turns, delayed tool
results, loaded context, and sandbox creation. Host-routed model tool calls can
be held for asynchronous approval before execution.

The V1 observational events are `RunStarted`, `ContextLoaded`,
`TurnCompleted`, `ToolCompleted`, `RunCompleted`, and `SandboxCreated`.
Observational delivery is bounded, best-effort, and gap-detectable; it never
waits on subscriber or network work and is not an authoritative audit log.

## Value

Applications and daemons can construct an external conversation record from
delivered complete messages, correlate activity across runs and sandboxes,
capture usage/model facts and loaded instructions, and implement extensions
without changing each provider loop. Approval services can enforce human or
service policy before sensitive host-owned tool execution.

### Shared contract and compatibility

- [ ] A new dependency-neutral `net8.0;net9.0` contracts package contains only
      lifecycle/approval DTOs, canonical JSON serialization metadata, and
      in-process observer/approval abstractions. `LmMultiTurn`, `Sandbox`, and
      `LmAgentInfra` consume it without introducing circular or ASP.NET-to-core
      dependencies.
- [ ] Existing public/protected signatures remain available. No required member
      is added to an existing public interface and no existing positional record
      or constructor is changed in place; retained overloads and no-op defaults
      preserve hook-disabled behavior.
- [ ] API-compatibility checks against the latest release and N-1 precompiled
      consumers pass on `net8.0` and `net9.0`.
- [ ] Each event envelope has stable wire names, `schema_version` major,
      immutable `event_id`, stable `event_type`, `source_stream_id`,
      `source_sequence`, `producer_epoch`, occurrence timestamp, and applicable
      existing correlation IDs (`ThreadId`, `RunId`, `GenerationId`,
      `ParentRunId`, input/receipt IDs, `ToolCallId`, sandbox `SessionId`).
- [ ] Agent lifecycle events use `thread:{ThreadId}` as the canonical source
      stream so runs, delayed results, loaded context, and child runs have one
      atomic order. Sandbox lifecycle events use `sandbox:{SessionId}` and carry
      thread/workspace correlations when known; no process-wide cross-tenant
      sequence is introduced.
- [ ] Each webhook delivery additionally has immutable `delivery_id`,
      subscriber-specific `delivery_sequence`, and attempt ID. Source sequence
      orders canonical lifecycle events; delivery sequence is assigned after
      filtering and exposes subscriber-specific loss.
- [ ] Event and delivery IDs/sequences are assigned once before fan-out/drop and
      reused across retries. Concurrent publication is monotonic within the
      typed source stream; source sequence allocation is atomic per stream and
      producer restart uses a new epoch.
- [ ] Registration negotiates supported lifecycle and approval protocol majors;
      delivery/approval starts only when a compatible major exists.
- [ ] Unknown top-level events are acknowledged and ignored. Unknown nested
      message/tool/outcome/enum values preserve their discriminator and raw
      payload without rejecting the containing event. Unknown approval decisions
      fail closed.
- [ ] A published field matrix defines every V1 wire field's JSON type,
      requiredness, nullability, and absent/null/empty semantics. Fields added
      within V1 are optional; tightening requiredness/nullability requires a new
      major.
- [ ] One source-generated serializer contract owns property names,
      discriminators, enum encoding, timestamps, null handling, and UTF-8 bytes.
      Producers, webhook delivery, signature verification, and fixtures use the
      same contract; retries reuse the identical signed body.

### Canonical lifecycle facts

- [ ] Raw, Claude, Codex, and Copilot loops use one provider-neutral turn
      finalization seam. It receives causal/accepted/injected inputs, final joined
      messages, final usage, exact requested/provider-observed model identity,
      non-deferred tool outcomes, and terminal outcome.
- [ ] Existing run-ledger/terminal state is extended—not replaced with a
      hook-local store—to retain ordered `GenerationId`s, terminal outcome,
      pending deferred IDs, and the minimal tool/run summary used by
      `RunCompleted`, including restart reconciliation.
- [ ] Exactly one `RunStarted` is produced per run with assignment, parent/cause,
      accepted input IDs when applicable, and provider/model metadata. Its causal
      input may be a non-user message.
- [ ] Exactly one `TurnCompleted` is produced per `GenerationId` after stream
      joining, final usage, and non-deferred host-owned tool completion. It
      contains the turn's causal message delta, complete assistant messages, tool
      calls/results, stable message order, raw final usage, exact requested and
      provider-observed model values, and outcome.
- [ ] `TurnCompleted` never contains text, reasoning, or tool update fragments or
      duplicate synthesized final messages. Missing observed model values remain
      absent/null rather than copied, inferred, or replaced by sentinels.
- [ ] Partial/error/cancelled turns expose only complete messages available before
      termination and an explicit stable outcome; incomplete fragments are never
      represented as complete history.
- [ ] Exactly one `RunCompleted` freezes at the terminal boundary and contains
      terminal outcome/error code, ordered turn IDs/count, tool-call
      summary/references, canonical raw usage snapshot, and pending deferred IDs.
      Later work is represented only by separately correlated events/runs.

### Delayed tool results

- [ ] Every first successful delayed/deferred resolution produces one
      `ToolCompleted` with the complete result and originating `ThreadId`,
      `RunId`, `GenerationId`, and `ToolCallId`. Duplicate resolution creates no
      additional event.
- [ ] If the originating run is terminal, each delayed result independently
      queues one fresh child run as soon as that agent's execution slot is
      available. Child runs remain serialized per agent and start in delayed
      result resolution order.
- [ ] The child run has a fresh `RunId`/`GenerationId`, the originating `RunId` as
      `ParentRunId`, `WasForked=false`, and immutable resume-cause metadata
      referencing/snapshotting the resolved result. The result is not appended
      twice.
- [ ] Source ordering proves `ToolCompleted(origin)` precedes the child's full
      `RunStarted -> TurnCompleted -> RunCompleted` sequence. The child cause is
      a `ToolCallResultMessage`, not a fabricated user message, and the provider
      request contains the resolved result through canonical history.

### Loaded context

- [ ] Boot and mid-session context preserve typed provenance through system
      prompt/`NotifyMessage` composition: discovery kind, normalized
      sandbox/workspace-relative source path/name, truncation, dedup identity,
      and boot versus mid-session mode.
- [ ] `ContextLoaded` is emitted only at, or provably tied to, the provider-request
      snapshot that contains the exact rendered context block. Discovery,
      formatting, enqueue, or inventory alone does not count as loaded.
- [ ] Boot ordering establishes `RunStarted -> ContextLoaded -> first
      TurnCompleted`. Mid-session context appears exactly once as its existing
      non-user `NotifyMessage` in the next applicable provider request and turn.
- [ ] Queued-then-cancelled context, rediscovery of the same dedup identity, and
      plugin/skill/subagent inventory not injected into model context emit no
      false `ContextLoaded`.

### Tool approval

- [ ] Approval covers only model-requested executions routed through the common
      LmDotnetTools executor: local functions, host-routed MCP calls, and sandbox
      operations exposed through that executor. Direct SDK and provider/CLI
      remote execution are explicitly outside V1.
- [ ] Static filtering runs before approval. Opening a gate freezes the canonical
      argument bytes/hash used by execution, matching approver set, tool identity,
      correlation, owner scope, and UTC expiry.
- [ ] Matching approvers are invoked concurrently under one monotonic end-to-end
      deadline and all must explicitly allow. Any deny, missing approver,
      overload, timeout, or hook error fails closed; owner/shutdown cancellation
      remains cancellation.
- [ ] Pending approvals are bounded per owner and process by count/bytes and have
      one atomic state transition: `Pending -> Allowed | Denied | Expired |
      Cancelled`. A terminal transition atomically replaces pending state with a
      bounded tombstone retained for a documented idempotency window.
      Tombstones have per-owner/process count/byte limits and a defined stable
      response after expiry/eviction. Tool execution count is always zero or one.
- [ ] Versioned S2S approval requests are signed and decisions arrive through an
      authenticated asynchronous API. The first valid decision wins; an identical
      retry returns the same result. Conflicting, wrong-ID/hash/scope, expired,
      cancelled, replayed, or late decisions return stable conflict/expired
      outcomes and never execute the tool. A decision is accepted only from the
      exact frozen approver identity/subscription selected when the gate opened;
      another approval-capable actor in the same scope cannot substitute.
- [ ] Approval denials/failures expose stable safe codes (for example denied,
      hook-error, timeout, overload, missing) and non-sensitive reasons. Raw
      exceptions, stacks, headers, bodies, and URIs remain internal.

### S2S authorization and privacy

- [ ] Shared webhook security is extracted from existing implementations before
      lifecycle endpoints are added: HMAC raw-body signing/verification,
      timestamp/delivery replay protection, safe destination validation, key
      rotation, and constant-time comparison.
- [ ] Authenticated APIs support register, rotate-secret, and revoke for HTTPS
      subscriptions. Subscription scope is derived server-side from existing
      credential/app/workspace ownership; a dedicated management capability and
      exact owner scope are required, and ungranted wildcard scopes are rejected.
- [ ] Scope/authorization is checked at registration, enqueue, each retry,
      decision, and revocation. Delivery/content authorization is rechecked
      immediately before the first and every subsequent outbound attempt.
      Rotation reauthorizes management capability/scope. One scope cannot
      inspect, decide, or receive another scope's data.
- [ ] Full content is the default only for principals granted the dedicated
      full-content lifecycle capability. Approval decisions require a separate
      capability and bind an opaque authenticated actor/service identity to the
      approval record.
- [ ] Producer-controlled structural credentials are removed before
      serialization: auth/cookie headers, host-injected tokens, configured
      secrets, environment/vault values, connection strings, private keys, and
      signing secrets. Arbitrary user/model/context/tool content may itself
      contain sensitive data and is intentionally sent to trusted full-content
      subscribers.
- [ ] Endpoint signing secrets are returned only at create/rotate, stored
      encrypted and write-only, redacted from logs/responses, and support a
      bounded previous-key overlap. Revocation atomically prevents new sends,
      cancels active/queued delivery and retries, invalidates selected pending
      approval gates before they can allow execution, and erases current/previous
      secrets when their valid overlap ends.
- [ ] Callback URLs require HTTPS and configured egress host/port authorization;
      URLs with userinfo, fragments, credential-bearing query parameters, or
      unapproved internal destinations are rejected before persistence.
      Destination addresses are resolved and validated on every connection
      attempt to resist DNS rebinding; redirects are disabled or each hop is
      revalidated against the same policy.
- [ ] Context events use logical sandbox/workspace-relative paths. Public events
      and ordinary logs exclude host paths, raw exceptions/provider responses,
      callback credentials, and non-allowlisted private marketplace metadata.

### Sandbox lifecycle

- [ ] `SandboxCreated` is emitted once only after the registry's committed
      creation boundary for an actual gateway session. Concurrent callers produce
      one event; cache hits, create failures, and rolled-back attempts produce
      none.
- [ ] Sandbox recreation emits one event after replacement commit with a fresh
      `SessionId` and explicit replaced-session correlation; the old cache entry
      is unusable.
- [ ] The gateway/SDK create result is extended to report confirmed loaded
      plugin/skill/agent identifiers and versions atomically. If a compatible
      gateway cannot provide them, the event marks inventory unavailable with an
      explicit reason; requested/preview data is never labeled confirmed.
- [ ] Sandbox payloads include only allowlisted runtime/workspace/capability and
      confirmed inventory fields; they exclude credentials, environment values,
      manifests/content, private repository/publisher metadata, command
      arguments, install paths, and host paths.
- [ ] Sandbox lifecycle delivery cannot cross credential/app/workspace ownership
      boundaries.

### Resource-bounded delivery and failure semantics

- [ ] Observational publication is a no-throw boundary that performs bounded
      materialization, canonical serialization, filtering, and non-blocking
      enqueue work only; it never invokes subscriber code or network I/O.
- [ ] Each event projection is serialized once into an immutable capped UTF-8
      buffer shared across matching subscribers. Per-owner and process-wide
      subscription, queued-item, queued-byte, and pending-delivery limits bound
      aggregate memory and fan-out.
- [ ] Publication meets a documented latency budget at maximum supported payload
      and subscriber counts.
- [ ] Events are never silently truncated. Materialization/serialization failure,
      oversize, queue full/closed, revoked authorization, subscriber timeout,
      retry exhaustion, and shutdown drop the complete event/delivery with a
      stable reason and observable sequence gap.
- [ ] In-process and webhook callbacks have bounded deadlines. A non-cooperative
      in-process callback is quarantined and not invoked again; abandoned task
      faults are observed, while the documented .NET limitation prevents forced
      termination.
- [ ] Observers and approvers execute through a process-wide bounded bulkhead.
      Timed-out/abandoned executions remain charged against owner/process task and
      byte budgets until actual completion. Quarantine cannot be bypassed through
      re-registration; when the orphan budget is exhausted, new observational
      work is dropped and new approval work fails closed. Late approval results
      can never execute tools.
- [ ] Webhook retry policy defines retryable transport/status classes, maximum
      attempts, total deadline, jittered backoff, and bounded `Retry-After`.
      Non-retryable authorization/signature/ordinary 4xx failures make one
      attempt; retries reuse event/delivery identity and body. Delayed retries do
      not occupy a delivery worker; bounded per-subscriber concurrency and
      documented ordering rules prevent one failing delivery from indefinitely
      blocking fresh events.
- [ ] Shutdown atomically stops admission, cancels approval/network work, drains
      to a configured deadline, classifies remaining drops, releases shared
      buffers/pending state, and completes without invoking subscribers afterward.
- [ ] Failure diagnostics use direct logger/metric sinks only and cannot recurse
      through lifecycle hooks. Metrics use low-cardinality event/outcome/reason
      dimensions; restricted logs use opaque IDs and never payloads, signatures,
      headers, bodies, secrets, or full URLs.

### Verification

- [ ] Golden JSON and sender/receiver compatibility tests cover minimal/maximal
      V1, N/N-1 protocol negotiation, unknown future event/message/enum values,
      absent/null/empty fields, exact signed bytes, and net8/net9 symmetry.
- [ ] Parameterized raw/Claude/Codex/Copilot contract tests use equivalent
      scripted streams and assert one final-only `TurnCompleted` per generation,
      no fragments/duplicates, canonical ordering, and one terminal event for
      success, error, cancellation, interruption, and max-turn cases.
- [ ] Delayed-result tests deterministically release multiple results in chosen
      orders and assert one serialized child run per result, correct source order,
      non-user causes, parent/child IDs, no duplicate result/run/provider call,
      and no folding into an active child run.
- [ ] Context tests prove the exact rendered boot/mid-session block is present in
      the provider request and emitted once; queued/cancelled, rediscovered, and
      inventory-only content produces no event.
- [ ] Approval tests use controllable time/barriers to cover unanimous allow,
      deny/final-allow races, cancellation/final-allow races, expiry boundary,
      overload, hook errors, identical/conflicting retries, wrong
      ID/hash/scope, and zero-or-one execution.
- [ ] Approval TOCTOU tests pause after canonical argument capture, mutate or
      rebind the original caller-owned arguments, then approve; execution receives
      exactly the frozen approved bytes or does not execute.
- [ ] HTTP tests assert 401/403 authorization, raw-body signature integrity,
      timestamp window, replay, key rotation/cutoff, revocation, callback
      validation, cross-scope denial, and no secret disclosure.
- [ ] Delivery tests cover exact item/byte/subscriber limits, concurrent
      publication, shared-buffer release, retry classes/budget, blocked
      subscriber isolation, source/delivery gaps, producer epoch restart, and
      bounded shutdown using virtual time/barriers rather than sleeps.
- [ ] Sandbox tests cover concurrent create, cache hit, failure/rollback,
      recreation, inventory status, and cross-scope isolation.
- [ ] With no hooks or approval policy configured, provider requests, persisted
      conversation history, tools, usage, and terminal outcomes match baseline
      behavior except generated hook IDs/timestamps.

## Out of Scope

- Guaranteed delivery, durable outbox/replay, and historical backfill.
- Approval of provider/CLI-executed remote tools.
- Cross-provider usage normalization or pricing changes tracked by #116/#196.
- Marketplace/plugin code registering privileged hooks.
- Streaming-fragment/message-mutation hooks and mutable run summaries.
- Broad subscription inspection/control-plane APIs beyond register, rotate, and
  revoke.
- Out-of-process plugin execution, advanced DLP, and tamper-evident/WORM audit.

## Related Work

#78, #95, #116, #138, #148, #161, #171, #187, #191, #196, #216

gautam-achieveai:
## Summary

Closes #227.

Adds provider-neutral lifecycle observation and fail-closed tool approval across the MultiTurn stack, including:

- a dependency-neutral `LmLifecycle` wire-contract package with canonical source-generated JSON, compatibility fixtures, field semantics, and API-surface gates;
- final-state lifecycle events for raw, Claude, Codex, and Copilot loops, delayed tool results as serialized child runs, exact loaded-context reporting, and committed sandbox lifecycle/inventory facts;
- fail-closed approval over frozen executable argument bytes, concurrent approvers with unanimous allow, bounded pending/tombstone state, exact approver-subscription identity, and asynchronous authenticated decisions;
- default-disabled S2S subscription/delivery infrastructure with server-derived ownership, content/capability gates, bounded queues/retries/quarantine/shutdown, HMAC signing, key rotation, and revocation that abandons queued/retrying work;
- callback egress hardening against SSRF and DNS rebinding, including connect-time address validation, redirect refusal, and IDN/punycode canonicalization;
- responsive Codex/Copilot transport dispatch while approvals are pending.

## Design records

Adds and indexes ADRs 0002–0008 covering the wire contract, fail-closed approval, delayed child runs, S2S delivery, workflow observation ordering, settling-seam observation, and asynchronous provider dispatch.

## Verification

- Clean full-solution build: **0 errors, 0 warnings**
- Repository-wide CSharpier check: **passed**
- Husky `format-verify`: **passed**
- Final unfiltered solution tests: **6,022 passed, 0 failed, 8 intentional skips** across 30 TRX files
- `LmAgentInfra.Tests`: **362/362 passed**
- `LmLifecycle.Tests`: **107/107 passed** on both `net8.0` and `net9.0`
- Live #227 acceptance-criteria audit: **64/64 passed**
- Multi-lens local review plus scoped re-review: all Must/Should findings addressed; no new Important findings

## Notes

The test command uses a process-local valid-length dummy `ANTHROPIC_API_KEY` because three mock-backed Anthropic tests validate key length during client construction. No credential was written to source or configuration.


src/LmMultiTurn/DelayedResultCoordinator.cs
```diff
@@ -0,0 +1,446 @@
...
+                pending = null;
+                inFlightFingerprint = entry.ResolvingFingerprint;
+                return false;
+            }
+
+            entry.Resolving = true;
+            entry.ResolvingFingerprint = fingerprint;
+            pending = new ResolvingDeferral(
+                entry,
+                RequiresChildRun: entry.Parked,
```
gautam-achieveai: [Gautam's review bot] **P1: Re-read the parked state when the resolution commits.** `TryBeginResolve` snapshots `entry.Parked` here, but the claim remains in `_entries` across several awaited durable/publication operations. During that window `TryPark` can set this same entry to parked and end the run. `CompleteResolve` then uses this stale `RequiresChildRun = false`, returns no cause, and `ResolveClaimedAsync` schedules no loop wake, so the resolved result is never shown to the provider. Please decide child-run ownership from the entry's current parked state under the `CompleteResolve` lock (and add a resolve-vs-park interleaving test).

src/LmAgentInfra/Lifecycle/RemoteApprovalStore.cs
```diff
@@ -0,0 +1,719 @@
...
+            );
+            return Unknown();
+        }
+
+        // An allow settles nothing until it is the last one outstanding. A deny skips the ballot
+        // entirely: one approver refusing is already the answer, and counting it would only postpone
+        // a call that is not going to run either way.
+        if (
+            WireOutcomes.IsAllowed(decision.Decision)
+            && ticket.RecordAllow(submitter) != RemoteApprovalBallot.Unanimous
```
gautam-achieveai, line 505: [Gautam's review bot] **P1: A duplicate allow racing the winning commit is reported as still waiting for another approver.** Once an allow completes the frozen ballot, concurrent retries from that same approver get `Duplicate`; this branch treats that like an incomplete ballot and can return `Recorded`/null before `TryDecide` publishes the standing result. With one approver there is nobody else to wait for, yet the controller maps this to HTTP 202, violating the documented idempotent standing-outcome contract. The included `Exactly_one_of_many_racing_decisions_settles_the_request` test currently fails deterministically on this path. Duplicate submissions need to await/re-read the in-flight settlement and return its outcome rather than an artificial pending state.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found two merge-blocking concurrency defects: delayed tool results can be stranded when resolution races run parking, and remote approval retries can return an incorrect pending response while the request is settling.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `7c0d809...8f088927` found no new issues. Both previously reported P1 races are resolved: delayed-result completion now re-reads the parked state under the coordinator lock and schedules the required child run, while completed approval ballots route duplicate allows through the one-shot settlement path to return the standing outcome. The new deterministic race coverage passes, and the `0.1.4` sandbox image pin restores the live-contract check.

gautam-achieveai:
[Revobot]

Reviewed PR #231 across 8 files — 2 Must, 1 Should, 0 Consider.

### Must — durable continuation is lost when parking races resolution
`src/LmMultiTurn/DelayedResultCoordinator.cs:354-355`  
The durable resolution write at `MultiTurnAgentLoop.cs:1320-1324` stores `pending.ChildRunId`, which is null when resolution began while the run was active. If the run parks before `CompleteResolve`, this code creates a child-run ID only in memory. A process failure before that queued cause runs leaves the result durably resolved with no recoverable continuation; restart recovery only restores unresolved (`IsDeferred`) calls.

Persist the late-created continuation identity atomically and recover pending continuation work on restart. Add an end-to-end persistence/restart test for this interleaving.

### Must — cancellation can suppress the only delayed-result wake-up
`src/LmMultiTurn/MultiTurnAgentLoop.cs:1445-1447`, `1758`, `1771-1784`  
A committed delayed cause schedules its wake-up using the external resolver’s cancellation token. If the request is cancelled while the bounded channel write is pending, `WriteLoopWakeAsync` exits through the cancellation branch without clearing `_wakeScheduled`. The cause remains queued, but the loop is not woken and no later resolution can repair it when this was the final deferred result.

Schedule committed causes with the loop’s own lifetime token (or an internal uncancellable enqueue) and reset `_wakeScheduled` for cancelled writes. Cover a resolver-token cancellation immediately after commit.

### Should — store-side cancellation is misreported as caller cancellation
`src/LmMultiTurn/MultiTurnAgentLoop.cs:1326-1329`  
Every `OperationCanceledException` from the lifecycle-store write is returned as `ResolveToolCallOutcome.Cancelled`, even if the supplied `ct` was not cancelled. A store timeout/internal cancellation can therefore be reported as caller cancellation without proving that the resolution was untouched.

Restrict this path to `when (ct.IsCancellationRequested)` and classify other cancellation failures as store failures.

The previously reported approval-ballot and late-park live-process races are resolved. Temporary-code review and `git diff --check` found no issues. Targeted tests could not run because `dotnet` is unavailable in the sandbox.

**Verdict: Request changes.**

🔎 Full review conversation: http://localhost:5051/?threadId=thread-7c42a0ea980d4702b33988402f5d5650&focus=1

gautam-achieveai:
[Revobot follow-up]

Addressed the later review at commit `f99ad2e8`:

- persisted/adopted late child-run identities before history mutation, with in-memory/file/SQLite parity and restart recovery;
- made committed wake-ups loop-lifetime-bound and reset the wake latch when mid-run polling consumes a sentinel;
- distinguished caller cancellation from lifecycle-store cancellation;
- gated delayed child provider execution on a durable run-start marker to prevent duplicate recovery;
- preserved source compatibility for existing `IRunLifecycleStore` implementations and prevented legacy-store retry stalls;
- corrected NotFound versus contradiction diagnostics.

Verification: `LmMultiTurn.Tests` 737/737, clean full-solution build (0 errors/0 warnings), whitespace/format hook passed, mutation RED→GREEN coverage for each race/failure path, and final independent scoped re-review reports no open Important findings.

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -949,85 +1291,524 @@ public Task ResolveToolCallAsync(
...
+
+        if (cause != null)
+        {
+            if (pending.ChildRunId == null)
+            {
+                // Minted later still: the run parked during the history write, after the attempt
+                // above had already found it live. Nothing can be refused now — the result is in
+                // history and the reservation is retired — so this attaches what it can and says so
+                // plainly when it cannot.
+                await ReportLateChildRunAsync(toolCallId, cause.ChildRunId);
```
gautam-achieveai, line 1513: [Gautam's review bot] **P1: Persist the late child-run identity before resolved history can become durable.** `AttachChildRunBeforeHistoryAsync` samples `Parked` only once. If the requesting run parks after that check but while `ReplacePersistedAsync` is awaited, `CompleteResolve` mints the child ID, but this attachment runs only after the resolved history entry has already been persisted. A process failure in between leaves history non-deferred while the lifecycle record still has `ChildRunId == null`; restart recovery filters that record out, and history restoration only re-registers deferred results, so the provider continuation is silently lost. Make the child-name write and history replacement atomic, or recover the `resolved history + terminal requesting run + no child ID` state, and add a deterministic park-during-history-write crash/restart test.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review found one remaining merge-blocking crash-consistency defect. The prior wake-cancellation and store-cancellation findings are addressed, but the delayed-continuation durability guarantee still has a residual window.

PR #243, opened 2026-08-01
refactor(daemon): remove in-process MultiTurnAgent, move 100% to S2S

gautam-achieveai:
## Summary

Remove the in-process `MultiTurnAgentLoop` path from the CodeReviewDaemon. The daemon now runs **100% over the S2S REST API** against the LmStreaming review host — no direct dependency on `MultiTurnAgent` concrete types.

### Deleted (−2,042 lines)
- `LiveReviewAgentLoopFactory.cs` — in-process Copilot Anthropic/OpenAI loop assembly
- `InProcessReviewSubAgentCompletionSource.cs` — direct `SubAgentManager` state reads
- `ToolScopedReviewLoop.cs` — MCP client disposal wrapper
- `DiscoveredSubAgentTemplateBuilder.cs` — local sub-agent template discovery/mapping
- Project references: `OpenAIProvider`, `AnthropicProvider`, `McpMiddleware`
- All tests exercising the removed in-process path

### Simplified
- `ReviewToolContext` — removed `SubAgentOptions` field (host owns sub-agents)
- `DaemonReviewStageExecutor` — removed `BuildSubAgentOptionsAsync`, template builder field, provider agent factory
- `ReviewLoopSubAgentSurface` — removed `LiveLoopSurface` (only S2S surface remains)
- `Program.cs` — removed `LiveReviewAgentLoopFactory` registration, `else` branch, `Func<IStreamingAgent>` factory; S2S is now the only path with a fail-fast guard

### Config
- Added `UseS2SReviewAgent`, `LmStreamingBaseUrl`, `LmStreamingProviderId`, etc. to `achieveai` and `mcqdb` profiles
- Fixed MCQdb auth webhook port (5081→5082, from prior commit)
- Updated sub-agent intelligence tiers in Development settings

### Kept
- `GithubCopilotProvider` reference (used by `CopilotModelCatalogLogger` for diagnostic model catalog discovery)
- `IMultiTurnAgent` interface (contract only, via transitive through LmAgentInfra)
- `S2SReviewAgent` + `LmStreamingS2SClient` (the S2S adapter that replaces in-process)

## Testing

- **746 daemon tests pass** (0 failures)
- Daemon boots clean on both `achieveai` and `mcqdb` profiles in S2S mode
- Review host capabilities verified (messageIdempotency + spawnSuppression)


gautam-achieveai:
[Revobot]

# PR Review: PR #243

**Verdict: REQUEST_CHANGES**

Reviewed head `5f5c8a4aa1f50be738adb6da804960a021a2f77e` against `ddd4d029`.

## Findings

### 1. [BLOCKER] HIGH — Default daemon configuration cannot start

**File:** `samples/CodeReviewDaemon.Sample/Program.cs:279-290`  
**Related configuration:** `samples/CodeReviewDaemon.Sample/appsettings.json:30-36`

The base configuration leaves `UseS2SReviewAgent` false and does not define `LmStreamingBaseUrl` or the other required S2S settings. Consequently, a normal launch without `--review achieveai` or `--review mcqdb` throws before the host is built:

```csharp
if (!daemonOptions.UseS2SReviewAgent)
{
    throw new InvalidOperationException(...);
}
```

This breaks default/inert deployments and any caller that previously relied on the base configuration to boot and expose its routes.

**Suggestion:** Either make the base configuration a valid S2S configuration, or preserve a non-throwing inert/default startup path. Add startup coverage using the actual base configuration rather than only tests that override the required values.

---

### 2. [BLOCKER] HIGH — S2S silently discards per-run model selection

**File:** `samples/CodeReviewDaemon.Sample/Agents/S2SReviewAgentLoopFactory.cs:101-106`

`Create` accepts `modelId`, but the S2S implementation always provisions `_options.LmStreamingProviderId`. Any caller-specific model selection is therefore silently ignored. This changes the behavior of model-specific review arms such as variants or overflow escalation, and any other path that supplies a per-call model.

The factory currently creates every hosted conversation with the same provider:

```csharp
_options.LmStreamingProviderId
```

**Suggestion:** Extend the S2S provisioning contract to carry a validated per-run model/provider selection, or explicitly reject unsupported per-run overrides. Do not silently collapse distinct configured arms onto one provider.

---

### 3. [BLOCKER] HIGH — Configured conversation persistence is disconnected

**File:** `samples/CodeReviewDaemon.Sample/Program.cs:271-275`

The PR still constructs a `FileConversationStore` when `ConversationStorePath` is configured, but the resulting instance is not registered or passed to the S2S factory/host after the in-process factory was removed. The configured persistence option therefore has no effective runtime behavior.

This is especially misleading because both production profiles continue to advertise conversation-store paths.

**Suggestion:** Remove the obsolete daemon-side construction and document that persistence is owned by the LmStreaming host, or add an explicit S2S persistence configuration/API and verify it with an end-to-end test.

---

### 4. [BLOCKER] HIGH — S2S HTTP requests can outlive the review deadline

**Files:**
- `samples/CodeReviewDaemon.Sample/Program.cs:306-313`
- `samples/CodeReviewDaemon.Sample/Agents/LmStreamingS2SClient.cs:389-391`
- `samples/CodeReviewDaemon.Sample/Agents/S2SReviewAgent.cs:376-414`

The S2S client is built with a raw `HttpClient` using its default timeout, while the review agent separately tracks an absolute review deadline. Individual status or control requests are not clipped to the remaining review budget.

A blocked request can therefore continue beyond the configured review deadline. Its `TaskCanceledException`/`OperationCanceledException` can also be indistinguishable from caller shutdown, bypassing the normal retry-governance path and blocking the daemon’s serial polling work.

**Suggestion:** Create a linked per-request cancellation token with a timeout equal to the remaining review deadline, and distinguish an internally imposed request timeout from caller cancellation. Ensure the resulting failure is routed through the normal retry/pending-state mechanism.

---

### 5. HIGH — S2S wiring is not exercised by startup tests

**Files:**
- `samples/CodeReviewDaemon.Sample/Program.cs:300-367`
- `tests/CodeReviewDaemon.Sample.Tests/Scenarios/RouteExposureTests.cs:20-39`

The route-exposure tests verify that the host can build, but they do not resolve `IReviewAgentLoopFactory`, `LmStreamingS2SClient`, `S2SReviewWorkspacePreparer`, or `IReviewSubAgentCompletionSource`. A stale registration or missing S2S dependency could therefore pass startup tests and fail only when the first review is created.

**Suggestion:** Resolve and assert the S2S services from the test host, and add a scripted first-review test using a fake S2S HTTP handler.

---

### 6. MEDIUM — Supported non-S2S executor paths lost regression coverage

**File:** `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:317-363`  
**Deleted test:** `tests/CodeReviewDaemon.Sample.Tests/Orchestration/ReviewToolContextBuildTests.cs`

The PR deletes coverage for still-present non-S2S behavior, including:

- successful session provisioning;
- provisioning failure degrading to diff-only;
- disabled tool-assisted review not consulting the provisioner;
- cancellation propagation.

The S2S tests do not cover these executor paths.

**Suggestion:** Port the relevant tests to the new `ReviewToolContext` shape, particularly the provisioning failure, disabled-path, and cancellation cases.

---

### 7. MEDIUM — Only one profile’s S2S configuration is validated

**Files:**
- `samples/CodeReviewDaemon.Sample/appsettings.achieveai.json:24-30`
- `samples/CodeReviewDaemon.Sample/appsettings.mcqdb.json:24-30`
- `tests/CodeReviewDaemon.Sample.Tests/Scenarios/DaemonProfileConfigurationTests.cs:8-30`

The new profile test checks only the MCQdb listener/webhook origin. It does not validate the required S2S settings or inspect the `achieveai` profile.

A profile can therefore drift into an invalid S2S configuration while the existing test remains green.

**Suggestion:** Load and validate both profiles, including `UseS2SReviewAgent`, S2S base URL, provider ID, marketplace, skill-support requirement, and workspace-base configuration. Boot each profile through the test host where practical.

---

## Verification

- `git diff --check ddd4d029..5f5c8a4aa1f50be738adb6da804960a021a2f77e` passed.
- The .NET build/tests could not be run because `dotnet` is unavailable in the review environment.
- No additional exception-handling finding was substantiated by the exception-handling review.
- No context questions were required.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-63b6245bf0464f1cae72b2fea3bf29ac&focus=1

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -266,141 +264,108 @@
...
-// also resolve its SharedAgentFactory (Task 12) — the same Copilot-backed agent every review loop and any
-// discovered sub-agent are driven by — without standing up a second provider agent.
-builder.Services.AddSingleton<LiveReviewAgentLoopFactory>(sp => new LiveReviewAgentLoopFactory(
-    sp.GetRequiredService<ILoggerFactory>(),
-    daemonOptions,
-    conversationStore,
-    sp.GetService<MultiTurnLifecycleServices>()));
+// The daemon drives all reviews over the S2S REST API against a running LmStreaming.Sample review host.
+// UseS2SReviewAgent must be on — the in-process path has been removed.
+if (!daemonOptions.UseS2SReviewAgent)
```
gautam-achieveai, line 279: [Gautam's review bot] **[P1] Do not recycle a slot while its retained hosted conversation can still access it.** Making S2S mandatory means every production review now follows the lifecycle where `PostAsync` intentionally leaves the review-host sandbox/conversation alive for the deep link, but immediately strips and returns the mounted pooled slot; the terminal failure path also returns the slot without stopping the hosted run, and `S2SReviewAgent.DisposeAsync` is a no-op. A still-running cancelled turn can therefore race the next lease, and after a review-host restart the persisted conversation can recreate an agent from the same workspace ID and operate on whichever PR now occupies that slot. Tear down/detach the sandbox before returning the slot and make retained links transcript-only, or retain an immutable per-review workspace. Please add cancellation and post-restart slot-reuse coverage.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -266,141 +264,108 @@
...
+        retentionWindow,
+        sp.GetRequiredService<ILogger<DeepLinkRetentionSweeper>>()));
 }
 
-builder.Services.AddSingleton<Func<IStreamingAgent>>(sp =>
-    sp.GetRequiredService<LiveReviewAgentLoopFactory>().SharedAgentFactory);
+builder.Services.AddSingleton<IReviewAgentLoopFactory>(sp =>
+{
+    var store = sp.GetRequiredService<ReviewStore>();
+    return new S2SReviewAgentLoopFactory(
```
gautam-achieveai, line 360: [Gautam's review bot] **[P2] Preserve the model override used by overflow recovery.** This is now the only factory, but `S2SReviewAgentLoopFactory.Create` ignores its `modelId` argument and always provisions `LmStreamingProviderId`. `RunPrimaryReviewAsync` still catches context exhaustion and retries with `OverflowEscalationModelId` (the live profiles configure Luna -> Terra), so the logged escalation actually reruns Luna and can exhaust again. Map requested model IDs to hosted providers, or remove the ineffective escalation/configuration path instead of silently claiming a larger-model retry.

samples/CodeReviewDaemon.Sample/appsettings.mcqdb.json
```diff
@@ -13,7 +13,7 @@
     },
     "TokenStoreDir": "B:/sources/LmDotnetTools/.run/oauth-tokens",
     "Webhook": {
-      "PublicBaseUrl": "http://127.0.0.1:5081"
+      "PublicBaseUrl": "http://127.0.0.1:5082"
```
gautam-achieveai, line 16: [Gautam's review bot] **[P2] Configure the callback on the process that owns the S2S sandbox session.** In S2S mode LmStreaming creates the sandbox session, mints its session secret, and builds the ADO auth-provider endpoint from its own `Auth:Webhook:PublicBaseUrl`; the provision request does not forward this daemon URL. The checked-in LmStreaming development settings still point callbacks at port 5000 while this profile targets the review host on 5051, and its token store is independent. Consequently this 5082 change and the new test validate a daemon callback that hosted review egress does not use. Add a review-host profile/override for the 5051 callback and token store, and test the auth-provider emitted by an actual S2S-created session.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -266,141 +264,108 @@
...
 builder.Services.AddSingleton<IDiscoveredItemsSource>(sp =>
     new RegistryDiscoverySource(sp.GetRequiredService<SandboxSessionRegistry>()));
-builder.Services.AddSingleton<DiscoveredSubAgentTemplateBuilder>();
 
-// Optional conversation persistence: when ConversationStorePath is set, every review agent's
-// MultiTurnAgentLoop (primary + sub-agents) persists its full message history — Skill loads, sub-agent Task
-// dispatches, tool results — to <path>/<threadId>/messages.json for after-the-fact audit (DuckDB-queryable).
-// Unset ⇒ conversations are streamed and discarded, exactly as before.
+// Optional conversation persistence: when ConversationStorePath is set, the S2S review host persists its
+// full message history. The daemon retains the path for log auditing references.
```
gautam-achieveai, line 272: [Gautam's review bot] **[P2] Remove or reconnect the now-dead conversation-store setting.** Before this refactor, the `FileConversationStore` below was passed to `LiveReviewAgentLoopFactory`; that consumer is deleted, and the local variable is now never registered or used. LmStreaming persists to its own fixed `AppContext.BaseDirectory/conversations` path, so the configured `achieveai-conversations`/`mcqdb-conversations` directories no longer contain the advertised audit history. Either configure persistence explicitly on the review host or remove the daemon option/profile entries and update the operational documentation.

samples/CodeReviewDaemon.Sample/Program.cs
```diff
@@ -266,141 +264,108 @@
...
-    daemonOptions,
-    conversationStore,
-    sp.GetService<MultiTurnLifecycleServices>()));
+// The daemon drives all reviews over the S2S REST API against a running LmStreaming.Sample review host.
+// UseS2SReviewAgent must be on — the in-process path has been removed.
+if (!daemonOptions.UseS2SReviewAgent)
+{
+    throw new InvalidOperationException(
+        "UseS2SReviewAgent must be true; the in-process review path (LiveReviewAgentLoopFactory) has been removed. "
+        + "Set CodeReviewDaemon:UseS2SReviewAgent=true and configure LmStreamingBaseUrl.");
```
gautam-achieveai, line 283: [Gautam's review bot] **[P2] Fail startup when `LmStreamingProviderId` is missing.** The new mandatory-path error tells operators to enable S2S and configure only the base URL, while the provider ID is also required and has no `ValidateOnStart`/options validator. An empty value therefore boots a healthy polling daemon, prepares a PR workspace, and fails only when `S2SReviewAgentLoopFactory.Create` is reached. Validate the provider alongside `LmStreamingBaseUrl` so this S2S-only daemon actually fails fast on all required settings.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found one workspace-isolation issue and four concrete configuration/runtime regressions. The daemon test project passes, but the current tests do not exercise these S2S ownership and deployment boundaries.

samples/LmStreaming.Sample/ClientApp/package.json
```diff
@@ -23,5 +23,8 @@
     "vite": "^7.3.1",
     "vitest": "^4.0.18",
     "vue-tsc": "^2.2.0"
+  },
+  "allowScripts": {
+    "esbuild@0.27.2": true
   }
```
gautam-achieveai, line 29: [Gautam's review bot] **[P3] Do not represent this as an install-script allow-list when npm ignores it.** This client uses npm (`package-lock.json`, `npm ci` in the project/CI), and npm does not consume a top-level `allowScripts` field. There is also no LavaMoat allow-scripts dependency/configuration, so esbuild's install script continues to run unconditionally and this entry only creates a false security signal. Remove it, or wire a script-gating mechanism supported by the package manager and verify the install path.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review found one new package-configuration issue. The workspace response guard is safe, but this iteration does not address any of the five existing daemon findings, which remain active.

PR #245, opened 2026-08-02
Enable hierarchy-wide agent collaboration and bounded delegation

### LINKED ISSUE #244: Enable hierarchy-wide agent collaboration and bounded delegation
# Enable hierarchy-wide agent collaboration and bounded delegation

## Summary
Add a collaboration-scoped agent directory that connects every agent created under one root conversation, including ordinary subagents, nested descendants, and WorkflowAgent controllers/delegates.

Agents use this shared collaboration surface to:

- `GetAgents` — view the full agent hierarchy and identify themselves.
- `SendMessage` — asynchronously send a typed, correlated message to any agent in the same collaboration.
- `CheckAgents` — check one or more agents' current status without waiting.
- `GetAgentTranscript` — inspect another agent's permitted prompt/message history under a configurable visibility policy.
- `WaitForAgents` — allow a parent to wait only for its direct children, with child Questions interrupting that wait so the parent can help.
- `Agent` — create nested subagents only while below the configured delegation-depth limit.

Contact permission and transcript-read permission are deliberately separate: every agent in a collaboration can contact every other agent, while prompt/message visibility is configurable.

## Value
Complex tasks can be delegated recursively while agents coordinate without blocking one another. Stable agent/message IDs make cross-agent communication reliable, a full hierarchy makes collaborators discoverable, configurable transcript access supports both supervised and open team cultures, and an interruptible parent wait prevents a blocked child from being ignored.

## Delivery Approach
This is one WorkItem and remains one integrated feature. Implementation must proceed through ordered internal phases in the same branch/PR so each foundation is proven before dependent behavior is enabled:

1. Reconcile and complete the plural `CheckAgents`/`WaitForAgents` contract.
2. Add configurable nested delegation and root-wide quantity limits.
3. Add the read-only collaboration registry and `GetAgents`.
4. Add correlated `Question`/`Response` messaging.
5. Add any-to-any routing plus `DelegateTask`/`TaskUpdate`/`Steer` and inbox backpressure.
6. Add configurable transcript access.
7. Add Question-driven parent-wait interruption.

Each phase must leave the solution building and its focused tests passing. The final feature is complete only when every phase and the end-to-end scenarios pass; no phase is a separate WorkItem or separately required PR.

## Internal Architecture Boundaries
The root collaboration capability must be composed from focused services rather than one all-purpose manager:

- **Registry:** live hierarchy, canonical identity, ownership routes, status, depth, and lifecycle registration.
- **Router:** target resolution, bounded inbox admission, and delivery through the target's owning loop/manager.
- **Message ledger:** trusted message IDs, open/closed Question and DelegateTask correlation, idempotency, and progress/final-response rules.
- **Transcript policy:** configurable authorization for prompt/message reads.
- **Wait broker:** direct-child wait observation and Question interruption.

These services share the trusted collaboration ID and may be packaged together, but their state and responsibilities must remain independently testable.

### Agent identity and purpose metadata
- [ ] Agent metadata separates five concepts: canonical `agent_id` for routing, addressable `name`, technical `agent_type`/template, effective `role` describing responsibility, short instance `description` describing the current delegation, and full private `task`/prompt.
- [ ] Each template declares whether its role is `Fixed` or `Customizable`. A specialized template supplies a trusted fixed role; the runtime rejects a conflicting model-supplied role rather than silently relabeling the agent.
- [ ] `general-purpose` and other customizable-role templates require the `Agent` call to supply a concise, non-blank instance role. Missing roles produce a recoverable validation result and no agent is created; roles are never inferred from arbitrary task text.
- [ ] Every `Agent` call requires a concise, non-blank `description` explaining what this instance is doing. The existing declared-but-unused `description` argument is wired through spawn, state, snapshots, persistence, and hierarchy output rather than replaced with a duplicate field.
- [ ] `role` and `description` are intentionally collaboration-visible directory metadata. The `Agent` tool description warns callers not to include secrets, private prompt content, or customer data; length limits and validation keep them short. The full `task` remains governed by transcript-visibility policy.
- [ ] WorkflowAgent controllers and delegates receive equivalent effective role/description metadata from trusted workflow/controller and node/task definitions; existing `WorkflowTask.Label`/node titles are reused where appropriate rather than introducing parallel labels.
- [ ] Runtime registration is atomic with agent creation: an agent is not admitted to the collaboration directory unless canonical ID, effective role, effective description, parent, depths, and delivery route are all valid. Restart preserves the original instance metadata.
- [ ] Tests prove specialized-role derivation, general-purpose/custom-role validation, conflicting-role rejection, required description handling, workflow mapping, persistence/restart, hierarchy projection, and that role/description never appear in content-bearing logs.

### Collaboration scope and directory
- [ ] Every root conversation creates a trusted collaboration ID propagated to all ordinary agents, nested descendants, WorkflowAgent controllers, and workflow delegates; agents from another collaboration cannot be discovered, contacted, inspected, or used as wait targets.
- [ ] A root-scoped live directory registers and unregisters each agent at lifecycle boundaries and records its canonical agent ID, owning delivery route, parent ID, readable name, kind, status, structural/display depth, and delegation-budget depth.
- [ ] Cross-manager delivery resolves through the root directory and delegates injection/restart to the target's owning manager or loop, preserving the existing target lifecycle lock and continuation rules.
- [ ] `GetAgents` returns a versioned, cycle-safe full hierarchy for the caller's collaboration, including ordinary agents and explicit WorkflowAgent controller/delegate nodes.
- [ ] `GetAgents` clearly marks the current/calling agent and returns canonical ID, readable name, technical agent type/template, effective role, instance description, parent ID, kind, status, structural depth, delegation depth, and whether transcript details are readable for every visible node.
- [ ] Canonical IDs are authoritative. Names are display labels and may be used only when they resolve uniquely within the collaboration; ambiguous or reassigned names return a recoverable result requiring a canonical ID.
- [ ] Directory cleanup and retention are bounded so disposed agents do not remain live routing targets indefinitely; completed retained agents are clearly distinguished from live agents.

### Configurable transcript visibility
- [ ] Directory visibility/contact does not imply transcript visibility. `GetAgentTranscript` is the explicit operation for reading an agent's permitted context rather than embedding prompts or history in `GetAgents`.
- [ ] The root configuration supports at least `Ancestors` and `Open` transcript-visibility modes: `Ancestors` permits an agent's parents and all higher ancestors to inspect it; `Open` permits every agent in the same collaboration to inspect every other agent. `Open` is an explicit trust-mode choice and is never enabled implicitly.
- [ ] The registry maintains a reliable collaboration-scoped ancestor path/index so `Ancestors` checks do not depend on an unbounded persisted-thread scan.
- [ ] The visibility decision is enforced by trusted collaboration/lineage data on every transcript request and cannot be overridden by model arguments.
- [ ] An allowed transcript read can expose the configured system/assigned-task prompts, ordinary user/assistant messages, agent-message envelopes, tool calls/results, final output, and errors. Hidden/internal reasoning and secrets excluded by existing message/security policy are never exposed.
- [ ] A denied read returns metadata explaining that the agent exists but its transcript is not readable; it does not leak prompt/message snippets through errors, logs, directory fields, or status summaries.
- [ ] Root policy defines whether a runtime change applies immediately to existing agents; the same policy result is used consistently by tools, APIs, and UI projections.

### Typed and correlated asynchronous messages
- [ ] `SendMessage` requires `target`, `content`, and schema-enforced `msg_type` with exactly `Question`, `DelegateTask`, `TaskUpdate`, `Steer`, or `Response`; `TaskUpdate` and `Response` additionally require `in_response_to`. Invalid combinations return recoverable validation results.
- [ ] Sender identity, collaboration ID, and a new stable message ID are assigned by trusted runtime context and cannot be supplied or overridden by the model.
- [ ] Message IDs use a collision-resistant, persistence-safe format, survive transcript/store round trips, and are distinct from run, generation, tool-call, and persistence-row IDs.
- [ ] `SendMessage` is always non-blocking and has no legacy `run_in_background` choice. After acceptance it returns at least `accepted`, `message_id`, canonical target ID, and a precise delivery disposition; acceptance means queued/accepted for delivery, not that the target has read or processed it.
- [ ] Every target has a configurable bounded inbox. When admission limits are reached, `SendMessage` returns an observable recoverable backpressure result; messages are never silently dropped, and one sender cannot bypass the bound through concurrent sends.
- [ ] Inbox processing has deterministic ordering and message-type behavior: accepted messages preserve admission order, while `Steer` may affect active work without silently replacing queued `Question` or `DelegateTask` messages.
- [ ] Any agent can send to any canonical target in the same collaboration, regardless of transcript-read permission. Cross-collaboration targets are rejected.
- [ ] A running target receives the message in its active loop. `Question`, `DelegateTask`, and `Response` may resume an idle/completed target through existing continuation behavior. `Steer` only redirects active work and is rejected for a target with no active work. `TaskUpdate` is valid only for an open delegated task.
- [ ] Delivered content uses safely escaped attributes/body and the shape `<agent-message message-id="..." from="SenderName" from-agent-id="..." type="Question|DelegateTask|TaskUpdate|Steer|Response" [in-response-to="..."]>...</agent-message>`.
- [ ] `Question` appends `<reply-instruction in-reply-to="QuestionMessageId" reply-to="CanonicalSenderId" reply-label="SenderName" reply-tool-name="SendMessage" reply-msg-type="Response"/>`.
- [ ] `DelegateTask` appends a task reply instruction identifying the delegated message ID, canonical requester, `TaskUpdate` as the progress-message type, and `Response` as the final-message type.
- [ ] `Steer`, `TaskUpdate`, and `Response` do not request another reply and never append a reply instruction.
- [ ] A `Question` accepts exactly one correlated `Response`, which closes it. A `DelegateTask` accepts zero or more correlated `TaskUpdate` messages followed by exactly one correlated final `Response`, which closes it.
- [ ] `TaskUpdate.in_response_to` must reference an open `DelegateTask`; `Response.in_response_to` must reference an open Question or DelegateTask addressed to that responder. Unknown, mismatched, or closed references return recoverable results.
- [ ] Retried delivery with the same message identity is idempotent; a correction after closure starts a new message rather than silently replacing the closed response.
- [ ] The `SendMessage` tool and parameter descriptions explain all five message types, asynchronous acknowledgement semantics, correlation fields, and which messages permit progress/final replies.
- [ ] Agent messages are untrusted agent-originated input and can never grant user approval, alter permissions/configuration, or impersonate a user message.

### Status and parent-child waiting
- [ ] Reconcile this feature with the existing plural-agent-control design before changing the tool surface: one canonical `CheckAgents` and one canonical `WaitForAgents` contract are implemented, and the legacy singular/blocking paths are migrated rather than duplicated.
- [ ] `CheckAgents` accepts one or more collaboration-visible targets, remains non-blocking, and uses the established statuses `queued`, `running`, `completed`, `error`, `stopped`, and `not_found`.
- [ ] `WaitForAgents` accepts only a parent's direct children, supports the agreed batch `all`/`any` and timeout behavior where applicable, and reuses non-destructive observation; ending or interrupting a wait never cancels a child or consumes/releases an extra child concurrency permit.
- [ ] When a direct child sends a `Question` to a parent whose active `WaitForAgents` includes that child, the Question interrupts that wait exactly once and returns an explicit `question_received` outcome containing the child ID and Question message ID.
- [ ] The interrupted-wait result states that the child needs help, that the child was not completed or cancelled, and that the parent should answer with `SendMessage(Response, in_response_to: ...)` and call `WaitForAgents` again afterward if it has nothing else to do.
- [ ] A child Question is delivered to the parent even when no wait is active; wait interruption is an additional wake signal, not the delivery mechanism.
- [ ] Completion, Question, timeout, and cancellation races have deterministic single outcomes without losing other events: if Question wins, completion remains observable by the next wait; if completion wins, the independently queued Question remains in the parent's inbox.
- [ ] Multiple child Questions retain distinct message IDs and can be delivered/queued without overwriting one another. Responding does not automatically re-arm the parent's wait.
- [ ] No sibling or arbitrary cross-hierarchy wait capability is introduced.

### Configurable bounded delegation
- [ ] One trusted root option configures maximum ordinary subagent depth. Root is delegation depth `0`; every ordinary `Agent` spawn adds one; the option is validated and configurable without model control.
- [ ] An agent receives `Agent` only below the configured maximum. At the limit it retains permitted collaboration tools but cannot create another child.
- [ ] The default maximum preserves today's one-level ordinary-subagent behavior.
- [ ] Depth is passed explicitly through construction/restart paths, is independent of optional lifecycle-observability services, and remains unchanged when an agent restarts.
- [ ] Each nested parent owns an independent local `SubAgentManager`, concurrency gate, queue, and copied mutable template source while also registering its nodes/routes with the root collaboration directory.
- [ ] Recursive child options preserve ordinary templates, limits, and model selection but clear workflow-controller-only spawn-name/model-selection gates and redundant external inheritable-tool merges.
- [ ] `StartWorkflowAgent` is a zero-cost orchestration hop: a controller started by a caller at depth `d` has delegation depth `d`; its ordinary delegate is `d + 1`. Caller depth comes from trusted per-call context, not a model argument or long-lived manager state.
- [ ] Workflow control/launch tools remain excluded from delegates and descendants at every configured depth; greater ordinary-agent depth never permits nested workflow launches.
- [ ] Structural/display depth and delegation-budget depth are separate fields so WorkflowAgent controllers can appear in `GetAgents` without incorrectly consuming delegation budget.
- [ ] Existing per-manager concurrent/queued-agent limits remain enforced, and a configurable root-wide total-agent limit prevents breadth explosion across nested managers.
- [ ] Root-wide agent-cap reservation/release is concurrency-safe across simultaneous nested spawns, queueing, restart, failure, and disposal; capacity cannot be overbooked or leaked.

### Security, observability, compatibility, and verification
- [ ] The implementation documents an identifier glossary and data-flow diagram covering collaboration ID, canonical agent ID, thread/run/generation IDs, tool-call ID, agent-message ID, and persistence-row ID so each is generated, persisted, and correlated at the correct boundary.
- [ ] Agent-message and reply-instruction construction share the existing envelope escaping/sanitization behavior so content cannot close an envelope, inject a forged message/reply instruction, or override trusted IDs.
- [ ] Logs contain collaboration ID, source/target canonical IDs, message ID/type, correlation ID, lifecycle outcome, and content length as applicable, but never raw prompt/message/response content.
- [ ] Persisted message conversion preserves agent message IDs and correlation. Recovery behavior for outstanding Questions, directory entries, pending delivery, and interrupted waits is explicit; unsupported in-flight recovery fails clearly rather than misrouting or duplicating messages.
- [ ] Existing sample/API hierarchy projections use the same collaboration/depth semantics or document their compatibility boundary; no bounded global thread scan is used for live routing.
- [ ] Automated tests cover collaboration isolation; full-tree/current-agent directory output; ambiguous names; configurable transcript access and EUII protection; all exact envelope shapes; ID uniqueness/persistence/correlation/idempotency; running/idle delivery; Steer rejection; acknowledgements; any-to-any routing; CheckAgents states; parent-wait interruption and races; depth ceilings/restart; WorkflowAgent zero-depth behavior; root-wide quantity limits; recursive option isolation; cleanup/retention; concurrency accounting; and workflow-tool exclusion.

## Out of Scope
- Blocking `SendMessage` request/response calls
- Sibling or arbitrary-agent waits
- Broadcast/multicast messaging
- Cross-collaboration or process-global routing
- Exposing hidden/internal reasoning
- Treating agent messages as user approval
- Nested WorkflowAgent launches
- Guaranteed replay of in-flight deliveries after an unsupported process failure

gautam-achieveai:
## Summary

- add opt-in, root-scoped collaboration for ordinary and workflow agents
- support bounded nested delegation, trusted roles/descriptions, hierarchy discovery, and root-wide capacity limits
- add typed asynchronous Question, DelegateTask, TaskUpdate, Steer, and Response messaging with correlation, FIFO delivery, and lifecycle cleanup
- add plural status/wait tools, direct-child Question interruption, policy-checked transcript access, persistence compatibility, and client hierarchy/message rendering
- harden concurrency, privacy, raw agent-thread access, workflow resume, and sender/target retirement races
- fix the pre-existing FileConversationStore input-acceptance mutation race discovered during full verification

## Key Decisions

- collaboration is opt-in; absent options preserve legacy tool schemas and behavior
- one root-owned directory and message ledger span all managers in a collaboration, while each manager retains local lifecycle ownership
- workflow controllers are visible structural nodes but consume no delegation depth
- agent-to-agent sends are always non-blocking and bounded; waits remain direct-child-only
- transcript visibility is configured once per collaboration, excludes reasoning, and is enforced through trusted agent identities
- in-flight collaboration state is memory-only; persisted DTO changes are additive and backward-compatible

## Verification

- Release solution build: 0 warnings, 0 errors
- LmCore: 920/920
- LmMultiTurn: 1,043/1,043
- LmWorkflow: 377/377
- LmStreaming.Sample.Tests: 1,175 passed, 1 platform skip
- LmStreaming.Sample WebSocket E2E: 90/90
- Client Vitest: 750/750 across 61 files
- Client type-check: passed
- Browser E2E with current SPA: all #244 scenarios passed in both provider modes; full suite 49 passed, 4 skipped, 1 external sandbox clone failure
- External browser failure is tracked upstream: achieveai/SandboxedOstoolsMcpServer#161
- Independent specification, quality, concurrency, lifecycle, and coverage reviews completed; all verified HIGH/MEDIUM findings resolved

Fixes #244

### DIFF
--- a/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
+++ b/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
@@ -23,6 +23,26 @@ public enum AgentTranscriptOutcome
     Allowed,
 }
 
+/// <summary>
+///     The content-free reason codes a transcript read answers with when there is no hierarchy to read —
+///     the counterpart of <see cref="TranscriptAccessReasons"/>, which covers refusals.
+/// </summary>
+/// <remarks>
+///     Named once so the HTTP route and the in-agent <c>GetAgentTranscript</c> tool report the same
+///     vocabulary for the same outcome. They previously diverged (the tool said
+///     <c>hierarchy_unavailable</c> where the route said <c>collaboration_unavailable</c>), which makes a
+///     denial impossible to reason about from either side and impossible to document as one contract.
+///     Like every code on this path these carry no name, thread, or task.
+/// </remarks>
+public static class AgentTranscriptReasons
+{
+    /// <summary>The conversation itself is not known to this host.</summary>
+    public const string UnknownThread = "unknown_thread";
+
+    /// <summary>The host never enabled collaboration, so there is no hierarchy to read across.</summary>
+    public const string CollaborationUnavailable = "collaboration_unavailable";
+}
+
 /// <summary>The result of a transcript read, in the vocabulary both the API and the tool map from.</summary>
 public sealed record AgentTranscriptResult
 {
--- a/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
+++ b/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
@@ -1,0 +1,255 @@
+using AchieveAi.LmDotnetTools.LmAgentInfra.Agents;
+using AchieveAi.LmDotnetTools.LmMultiTurn;
+using AchieveAi.LmDotnetTools.LmMultiTurn.Collaboration;
+using AchieveAi.LmDotnetTools.LmMultiTurn.Persistence;
+using AchieveAi.LmDotnetTools.LmMultiTurn.SubAgents;
+using LmStreaming.Sample.Models;
+
+namespace LmStreaming.Sample.Services;
+
+/// <summary>How a transcript read resolved.</summary>
+public enum AgentTranscriptOutcome
+{
+    /// <summary>The conversation itself is not known to this host.</summary>
+    UnknownThread,
+
+    /// <summary>The host never enabled collaboration, so there is no hierarchy to read across.</summary>
+    CollaborationUnavailable,
+
+    /// <summary>The reader may not see this agent; <see cref="AgentTranscriptResult.DenialCode"/> says why.</summary>
+    Denied,
+
+    /// <summary>The read was permitted and <see cref="AgentTranscriptResult.Messages"/> carries the transcript.</summary>
+    Allowed,
+}
+
+/// <summary>The result of a transcript read, in the vocabulary both the API and the tool map from.</summary>
+public sealed record AgentTranscriptResult
+{
+    /// <summary>How the read resolved.</summary>
+    public required AgentTranscriptOutcome Outcome { get; init; }
+
+    /// <summary>
+    ///     A content-free <see cref="TranscriptAccessReasons"/> code when the read was denied. It carries
+    ///     no name, thread, or task: a refusal must not disclose what it is refusing.
+    /// </summary>
+    public string? DenialCode { get; init; }
+
+    /// <summary>The row that was read, when the read was allowed.</summary>
+    public SubAgentSummary? Agent { get; init; }
+
+    /// <summary>The normalized, reasoning-free transcript, when the read was allowed.</summary>
+    public IReadOnlyList<PersistedMessage> Messages { get; init; } = [];
+}
+
+/// <summary>
+///     The single trusted view of one conversation's agent hierarchy (#244), and the single place a
+///     cross-agent transcript read is authorized.
+/// </summary>
+/// <remarks>
+///     <para>
+///         Both readers of the hierarchy go through here: the REST endpoints
+///         (<c>/subagents</c>, <c>/agents/{id}/transcript</c>) and the in-agent
+///         <c>GetAgentTranscript</c> tool. That is the point — the row a reader was told is readable is
+///         the row that gets authorized and opened, with no second, subtly different derivation in
+///         between that could drift from the first.
+///     </para>
+///     <para>
+///         The service holds no state of its own; it is a projection over the live pool, the durable tab
+///         index, and the store, and is cheap to construct per call site.
+///     </para>
+/// </remarks>
+public sealed class AgentHierarchyService(
+    MultiTurnAgentPool agentPool,
+    WorkflowRunRegistry workflowRunRegistry,
+    IConversationStore store)
+{
+    /// <summary>
+    ///     Assembles one conversation's agent rows: the live sub-agent/workflow snapshot, unioned with the
+    ///     durable tab index, projected through the collaboration hierarchy when it is enabled.
+    /// </summary>
+    /// <param name="threadId">The conversation whose hierarchy is being read.</param>
+    /// <param name="viewerAgentId">
+    ///     The agent the answer is for, or null for the conversation root (the human at the keyboard). It
+    ///     only affects the viewer-scoped flags; it can never widen what the listing contains.
+    /// </param>
+    /// <param name="ct">Cancellation for the cold-path store lookup.</param>
+    /// <returns>The rows, whether the thread exists at all, and the live collaboration when there is one.</returns>
+    public async Task<(IReadOnlyList<SubAgentSummary> Rows, bool IsKnown, AgentCollaborationSetup? Collaboration)>
+        BuildAsync(string threadId, string? viewerAgentId, CancellationToken ct)
+    {
+        var summaries = new List<SubAgentSummary>();
+        var isLive = agentPool.TryGet(threadId, out var agent) && agent is not null;
+        var loop = isLive ? agent as MultiTurnAgentLoop : null;
+
+        // Agent-tool sub-agents (the historical /subagents contents) — LIVE-ONLY: they live on the main
+        // conversation loop's SubAgentManager, so they're gone after a restart until the loop is rehydrated.
+        if (loop?.SubAgentManager is { } subAgentManager)
+        {
+            summaries.AddRange(subAgentManager.ListAgents().Select(ToSummary));
+        }
+
+        // StartWorkflowAgent runs + their delegates. The live snapshot (when the WorkflowManager is present)
+        // is write-through-persisted to a small on-disk index, and the response is the union of live ∪
+        // persisted (live wins) — so completed workflow/delegate tabs SURVIVE A SERVER RESTART that evicts
+        // the in-memory manager. Delegate transcripts already persist as subagent-{id} threads, so a
+        // persisted tab replays read-only.
+        var workflowTabs = new List<SubAgentSummary>();
+        if (isLive && workflowRunRegistry.TryGet(threadId, out var workflowManager) && workflowManager is not null)
+        {
+            workflowTabs.AddRange(
+                workflowManager.ListRuns()
+                    .Select(r => new SubAgentSummary
+                    {
+                        AgentId = r.WorkflowId,
+                        Kind = SubAgentSummary.WorkflowTabKind,
+                        Name = r.Objective,
+                        Template = "workflow",
+                        Task = r.Objective,
+                        Status = r.Status,
+                        // The controller thread is conversation-scoped (workflow-{id}-{conversationId}); use the
+                        // run's real ThreadId so the ⚙ tab opens the ACTUAL persisted thread, not a stale
+                        // reconstruction. Fall back to the legacy shape only for a run with no scoped id.
+                        ThreadId = r.ThreadId ?? $"workflow-{r.WorkflowId}",
+                        LastActivityUtc = r.LastActivityUtc ?? r.StartedUtc,
+                    })
+            );
+
+            foreach (var run in workflowManager.ListRuns())
+            {
+                workflowTabs.AddRange(workflowManager.ListRunDelegates(run.WorkflowId).Select(ToSummary));
+            }
+        }
+
+        // Write-through: fold this live snapshot into the durable index (upsert, never deletes). The
+        // Agent-tool rows join the index only once collaboration is on — persisting them unconditionally
+        // would start surfacing restart-surviving sub-agent tabs in a host that never opted in.
+        var persistable = loop?.Collaboration is null
+            ? workflowTabs
+            : new List<SubAgentSummary>([.. workflowTabs, .. summaries]);
+        if (persistable.Count > 0)
+        {
+            workflowRunRegistry.PersistTabs(threadId, persistable);
+        }
+
+        // Merge live tabs with the persisted index (live wins on Kind+AgentId), so a restart that
+        // evicted the in-memory runs still surfaces them from disk.
+        var merged = new Dictionary<(string Kind, string AgentId), SubAgentSummary>();
+        foreach (var tab in workflowRunRegistry.GetPersistedTabs(threadId))
+        {
+            merged[(tab.Kind, tab.AgentId)] = tab;
+        }
+
+        foreach (var tab in workflowTabs.Concat(summaries))
+        {
+            merged[(tab.Kind, tab.AgentId)] = tab;
+        }
+
+        IReadOnlyList<SubAgentSummary> rows = [.. merged.Values];
+        if (loop?.Collaboration is { } collaboration)
+        {
+            rows = AgentHierarchyProjection.Project(
+                rows,
+                collaboration.Directory.Snapshot(),
+                viewerAgentId ?? collaboration.AgentId,
+                collaboration.Options.TranscriptVisibility);
+        }
+
+        // A live conversation, or one with persisted tabs, always answers 200. Otherwise the
+        // conversation is idle (evicted from the pool, or reopened but not yet messaged this session)
+        // and has no children to project — but it may still be a KNOWN thread on disk. Distinguish
+        // "known but idle with no sub-agents" (→ empty 200, the common case: a plain chat you reopened)
+        // from a genuinely unknown thread (→ 404) by consulting the store. Without this, every idle
+        // conversation with no sub-agents gets a spurious 404 and the client's sub-agent panel logs
+        // "Failed to list sub-agents" on every 3s poll. The store is only touched on this cold path,
+        // so the live hot path is unchanged.
+        var isKnown = isLive
+            || merged.Count > 0
+            || await store.LoadMetadataAsync(threadId, ct) is not null;
+
+        return (rows, isKnown, loop?.Collaboration);
+    }
+
+    /// <summary>
+    ///     Reads one agent's transcript on behalf of <paramref name="viewerAgentId"/>, applying the
+    ///     collaboration's transcript policy and excluding reasoning from what is returned.
+    /// </summary>
+    /// <remarks>
+    ///     Authorization is the projection's own <c>isReadable</c> — the very flag the listing published
+    ///     to this same reader — so there is no second decision here that could disagree with the first.
+    /// </remarks>
+    /// <param name="threadId">The conversation the hierarchy belongs to.</param>
+    /// <param name="agentId">The target, by either identifier a row publishes (tab id or node id).</param>
+    /// <param name="viewerAgentId">The reader, or null for the conversation root.</param>
+    /// <param name="ct">Cancellation for the store reads.</param>
+    public async Task<AgentTranscriptResult> ReadTranscriptAsync(
+        string threadId,
+        string agentId,
+        string? viewerAgentId,
+        CancellationToken ct)
+    {
+        var (rows, isKnown, collaboration) = await BuildAsync(threadId, viewerAgentId, ct);
+        if (!isKnown)
+        {
+            return new AgentTranscriptResult { Outcome = AgentTranscriptOutcome.UnknownThread };
+        }
+
+        if (collaboration is null)
+        {
+            // The feature is off (or the loop is not live), so the hierarchy this read needs does not
+            // exist. Reported as absence rather than refusal — there is nothing here to be refused.
+            return new AgentTranscriptResult { Outcome = AgentTranscriptOutcome.CollaborationUnavailable };
+        }
+
+        var row = AgentHierarchyProjection.Find(rows, agentId);
+        if (row is null || !row.IsReadable)
+        {
+            return new AgentTranscriptResult
+            {
+                Outcome = AgentTranscriptOutcome.Denied,
+                DenialCode = DenialCode(row, collaboration, viewerAgentId, agentId),
+            };
+        }
+
+        var messages = await store.LoadMessagesAsync(row.ThreadId, ct);
+        return new AgentTranscriptResult
+        {
+            Outcome = AgentTranscriptOutcome.Allowed,
+            Agent = row,
+            Messages = TranscriptProjection.Normalize(messages, excludeReasoning: true),
+        };
+    }
+
+    /// <summary>
+    ///     The content-free reason a transcript read was refused. An agent this conversation has never
+    ///     heard of is reported as an unknown target, which is also exactly what a caller outside the
+    ///     collaboration is told about an agent it may not know exists.
+    /// </summary>
+    private static string DenialCode(
+        SubAgentSummary? row,
+        AgentCollaborationSetup collaboration,
+        string? viewerAgentId,
+        string agentId) =>
+        row is null
+            ? TranscriptAccessReasons.UnknownTarget
+            : collaboration.Bundle
+                .EvaluateTranscriptAccess(viewerAgentId ?? collaboration.AgentId, row.AgentNodeId ?? agentId)
+                .Reason;
+
+    /// <summary>Projects one sub-agent snapshot (an Agent-tool spawn or a workflow delegate) to a tab row.</summary>
+    private static SubAgentSummary ToSummary(SubAgentSnapshot s) =>
+        new()
+        {
+            AgentId = s.AgentId,
+            Kind = SubAgentSummary.SubAgentTabKind,
+            Name = s.Name,
+            Template = s.TemplateName,
+            Task = s.Task,
+            Status = s.Status.ToString().ToLowerInvariant(),
+            ThreadId = s.ThreadId,
+            LastActivityUtc = s.LastActivityUtc,
+            EffectiveModelId = s.EffectiveModelId,
+            EffectiveModelIntelligence = s.EffectiveModelIntelligence,
+            ModelSelectionSource = s.ModelSelectionSource,
+        };
+}
--- a/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
+++ b/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
@@ -156,11 +156,18 @@ public sealed class AgentHierarchyService(
         // with unknown_target even for its legitimate root ancestor. Enrich() only stamps the structural
         // fields (WithCollaboration), never the viewer-scoped IsCurrent/IsReadable pair, so nothing here
         // bakes in one reader's answer for every future one.
+        //
+        // Enrich() alone only ever touches a tab THIS conversation already knew about. An ordinary
+        // descendant owned by a child's own SubAgentManager (a grandchild the model spawned through a
+        // spawn of its own) has no tab here at all — it exists only in the shared directory snapshot —
+        // so Enrich() never sees it and it never reaches the index. Project() below still surfaces it for
+        // THIS live answer (via its own unmatched-node pass), which is why the gap only shows up after a
+        // restart: PersistableRowsFor also folds in AgentHierarchyProjection.UnmatchedDescendantRows so
+        // that same descendant is written through now, before any particular viewer is known, exactly
+        // like Project would show it today.
         var persistable = loop?.Collaboration is null
             ? workflowTabs
-            : AgentHierarchyProjection.Enrich(
-                [.. workflowTabs, .. summaries],
-                loop.Collaboration.Directory.Snapshot());
+            : PersistableRowsFor([.. workflowTabs, .. summaries], loop.Collaboration.Directory.Snapshot());
         if (persistable.Count > 0)
         {
             workflowRunRegistry.PersistTabs(threadId, persistable);
@@ -291,6 +298,27 @@ public sealed class AgentHierarchyService(
             : collaboration.Bundle
                 .EvaluateTranscriptAccess(viewerAgentId ?? collaboration.AgentId, row.AgentNodeId ?? agentId)
                 .Reason;
+
+    /// <summary>
+    ///     Every row write-through persistence should stamp for one snapshot: <paramref name="tabs"/>
+    ///     enriched with their matching node's hierarchy metadata, plus a row for every node in
+    ///     <paramref name="nodes"/> that none of those tabs already accounts for — an ordinary descendant
+    ///     owned by a child's own <c>SubAgentManager</c>, invisible to this conversation's own tabs but
+    ///     present in the shared directory (see the remarks at the call site in <see cref="BuildAsync"/>).
+    /// </summary>
+    /// <remarks>
+    ///     The two halves can never collide: <see cref="AgentHierarchyProjection.Enrich"/> only ever
+    ///     touches a tab already present in <paramref name="tabs"/>, and
+    ///     <see cref="AgentHierarchyProjection.UnmatchedDescendantRows"/> only ever produces a row for a
+    ///     node none of those tabs matched.
+    /// </remarks>
+    private static IReadOnlyList<SubAgentSummary> PersistableRowsFor(
+        IReadOnlyList<SubAgentSummary> tabs,
+        IReadOnlyList<AgentDirectoryEntry> nodes) =>
+        [
+            .. AgentHierarchyProjection.Enrich(tabs, nodes),
+            .. AgentHierarchyProjection.UnmatchedDescendantRows(tabs, nodes),
+        ];
 
     /// <summary>Projects one sub-agent snapshot (an Agent-tool spawn or a workflow delegate) to a tab row.</summary>
     private static SubAgentSummary ToSummary(SubAgentSnapshot s) =>
--- a/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
+++ b/samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
@@ -211,7 +211,15 @@ public sealed class AgentHierarchyService(
         var subAgentRosterRelevant = !isLive || loop is { Collaboration: null, SubAgentManager: not null };
         if (subAgentRosterRelevant)
         {
-            foreach (var node in await GetOrScanPersistedSubAgentChildrenAsync(threadId, ct))
+            // The cache entry is keyed on this owner alongside threadId (see SubAgentScanCoverageCache's
+            // remarks): the live SubAgentManager instance when one covers this thread, or the shared
+            // NoLiveManager sentinel otherwise. A mode/provider switch, pool eviction+reopen, or restart
+            // rehydration each construct a brand-new SubAgentManager, so the owner here changes and the
+            // cache is invalidated automatically — no recreation call site needs to remember to do it.
+            var owner = loop is { Collaboration: null, SubAgentManager: { } liveSubAgentManager }
+                ? liveSubAgentManager
+                : SubAgentScanCoverageCache.NoLiveManager;
+            foreach (var node in await GetOrScanPersistedSubAgentChildrenAsync(threadId, owner, ct))
             {
                 merged[(node.Kind, node.AgentId)] = node;
             }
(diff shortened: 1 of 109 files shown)


gautam-achieveai:
[Revobot]

## Re-Review Summary: PR #245

**Self-contribution**: Of my 0 new findings, 0 are inside changes made to satisfy my previous review; this round converges on verified closure.

### Intent Check
- Goal coverage: **SOLVED**
- Solution direction: **RIGHT_BALLPARK**

### Previous Issues Resolution Status

| ID | Issue | Previous impact | Status | Closure evidence |
|---|---|---:|---|---|
| F-006 | An in-flight persisted scan could write back after `Forget` and resurrect a deleted/reused hierarchy roster | Medium / blocker | **VERIFIED** | `AgentHierarchyService` captures a cache epoch before scanning; `RecordRecovered` and `Forget` serialize under one gate; `Forget` advances the epoch. The deterministic service test at `AgentHierarchyServiceTests.cs:658-755` proves stale writeback is rejected and a reused-ID request rescans without the stale child. Capacity-1 ABA and cross-thread cases are also covered. |

All earlier review threads were already resolved and remain closed.

### New Issues Found

None in `18ad5eaa..964e33cf`.

### Verification
- GitHub **Build and test**: passed
- GitHub **browser-e2e**: passed
- `git diff --check 18ad5eaa..964e33cf`: passed
- Focused runtime tests were not rerun in this sandbox because `dotnet` is unavailable; CI supplies the runtime evidence.

### Verdict

**APPROVE**

The stated problem is solved, the solution remains in the right ballpark, the final blocker meets its original required outcome and objective done-when condition, and no new delta risk was found. This comment records the recommendation only; no approval review was cast without user confirmation.

### Blocks merge / shortest path to approval
- None.

### Follow-up issues
- None.

<details>
<summary>Review state (machine-readable)</summary>

#### Review Intent
```json
{
  "statedProblem": "Enable hierarchy-wide agent collaboration and bounded delegation",
  "acceptanceCriteria": [
    "root-scoped hierarchy and canonical identity",
    "bounded delegation and inbox admission",
    "typed correlated asynchronous messaging",
    "policy-checked transcript visibility",
    "persistence/restart correctness",
    "legacy behavior preserved when collaboration is disabled"
  ],
  "explicitNonGoals": [],
  "deliveredApproach": "Root-scoped collaboration directory, router, ledger, transcript policy, wait integration, workflow/sample/UI support; final delta hardens persisted hierarchy scan caching with a cache-wide forget epoch.",
  "goalCoverage": "SOLVED",
  "solutionDirection": "RIGHT_BALLPARK",
  "evidence": [
    "GitHub issue #244",
    "PR description",
    "passing Build and test check",
    "passing browser-e2e check",
    "delta review 18ad5eaa..964e33cf"
  ]
}
```

#### Active Review Threads
```json
[]
```

#### Closed Thread Archive
```json
{
  "closedThreadArchiveOmittedCount": 0,
  "closedThreadArchive": [
    {
      "findingId": "F-001",
      "threadId": "PRRT_kwDOOPysWM6Vz7d6",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-001:CLOSE:resolved:0"
    },
    {
      "findingId": "F-002",
      "threadId": "PRRT_kwDOOPysWM6Vz7gL",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-002:CLOSE:resolved:0"
    },
    {
      "findingId": "F-003",
      "threadId": "PRRT_kwDOOPysWM6Vz7gd",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-003:CLOSE:resolved:0"
    },
    {
      "findingId": "F-004",
      "threadId": "PRRT_kwDOOPysWM6V1ACd",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-004:CLOSE:resolved:0"
    },
    {
      "findingId": "F-005",
      "threadId": "PRRT_kwDOOPysWM6V1mjj",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-005:CLOSE:resolved:0"
    },
    {
      "findingId": "F-006",
      "threadId": "PRRT_kwDOOPysWM6V39Ux",
      "status": "CLOSED",
      "blocker": true,
      "closedAt": "2026-08-26T22:55:02.971151Z",
      "lastCompletedActionId": "F-006:CLOSE:964e33cf4b0b7c168ed574d35230f5ff162d50bc:1"
    }
  ]
}
```

</details>

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
```diff
@@ -0,0 +1,275 @@
...
+    public async Task<(IReadOnlyList<SubAgentSummary> Rows, bool IsKnown, AgentCollaborationSetup? Collaboration)>
+        BuildAsync(string threadId, string? viewerAgentId, CancellationToken ct)
+    {
+        var summaries = new List<SubAgentSummary>();
+        var isLive = agentPool.TryGet(threadId, out var agent) && agent is not null;
+        var loop = isLive ? agent as MultiTurnAgentLoop : null;
+
+        // Agent-tool sub-agents (the historical /subagents contents) — LIVE-ONLY: they live on the main
+        // conversation loop's SubAgentManager, so they're gone after a restart until the loop is rehydrated.
+        if (loop?.SubAgentManager is { } subAgentManager)
```
gautam-achieveai, line 110: [Gautam's review bot] **High:** The flat `/subagents` route now gets ordinary Agent-tool children only from this live manager. Before this PR it also reconstructed direct children from persisted `SubAgentProvenance`; with collaboration disabled (the checked-in default), lines 147-149 deliberately do not persist these rows to the new index. After pool eviction or restart, a finished conversation therefore loses all ordinary sub-agent tabs even though their transcripts/provenance remain on disk. Restore the bounded persisted-provenance union for the flat path, with live rows winning.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
```diff
@@ -295,6 +490,12 @@ public async Task<string> SpawnAsync(
         ct.ThrowIfCancellationRequested();
         ObjectDisposedException.ThrowIf(Volatile.Read(ref _disposeStarted) != 0, this);
 
+        // Admission to the collaboration happens BEFORE the concurrency permit and before the defer
+        // queue: capacity and delegation depth are root-wide invariants, so a spawn that the
+        // collaboration will not accept must never occupy a local slot or sit in the queue. No-op when
+        // collaboration is off.
+        AdmitToCollaboration(agentId, effectiveName, templateName, template, role, description);
```
gautam-achieveai, line 497: [Gautam's review bot] **High:** Collaboration admission now happens here before the local defer queue, but the pre-start foreground cancellation exits at lines 786-823 only remove queue indexes and cancel the waiter. They never retire `_admissions`, so the root-wide capacity lease and a permanently `queued` directory entry survive until manager disposal. Repeated cancelled queued spawns can exhaust `MaxTotalAgents`. Retire the admission on every pre-start cancellation path and add a collaboration-enabled full-gate regression test.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        // Write-through: fold this live snapshot into the durable index (upsert, never deletes). The
+        // Agent-tool rows join the index only once collaboration is on — persisting them unconditionally
+        // would start surfacing restart-surviving sub-agent tabs in a host that never opted in.
+        var persistable = loop?.Collaboration is null
+            ? workflowTabs
+            : new List<SubAgentSummary>([.. workflowTabs, .. summaries]);
+        if (persistable.Count > 0)
+        {
+            workflowRunRegistry.PersistTabs(threadId, persistable);
```
gautam-achieveai, line 174: [Gautam's review bot] **Medium:** `persistable` contains raw summaries without collaboration ID, node kind, parent, or ancestry; enrichment only happens later at lines 169-176. After restart, the fresh directory contains only the root, so a retained child row cannot rebuild a node via `ToNodeRecord()` and the checked transcript route fails closed with `unknown_target` for its legitimate root ancestor. Project/enrich the live rows before writing them to the durable index, and cover `BuildAsync -> persist -> restart -> transcript read`.

gautam-achieveai (commented):
[Gautam's review bot] **Verdict: REQUEST_CHANGES.** Deep review of !245 against #244 found three regressions: a collaboration-capacity leak when queued foreground spawns are cancelled, loss of persisted ordinary sub-agent tabs from the default flat roster after restart/eviction, and durable hierarchy rows written before collaboration enrichment. The existing active retired-sender test-coverage thread remains unresolved and is not duplicated here.

gautam-achieveai (commented):
[Gautam's review bot] **Verdict: REQUEST_CHANGES.** The incremental fixes address all four prior findings, but restart persistence still omits nested collaboration descendants.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
```diff
@@ -0,0 +1,350 @@
...
+        // the viewer-scoped flags). A row persisted in its raw, unenriched shape carries no
+        // AgentNodeId/AgentKind/CollaborationId/ancestry, so SubAgentSummary.ToNodeRecord() returns null
+        // for it once the live node that used to back it is gone (e.g. after a restart the fresh
+        // directory holds only the root), and a transcript read for a retained child then fails closed
+        // with unknown_target even for its legitimate root ancestor. Enrich() only stamps the structural
+        // fields (WithCollaboration), never the viewer-scoped IsCurrent/IsReadable pair, so nothing here
+        // bakes in one reader's answer for every future one.
+        var persistable = loop?.Collaboration is null
+            ? workflowTabs
+            : AgentHierarchyProjection.Enrich(
```
gautam-achieveai: [Gautam's review bot] **Medium:** This persists only `workflowTabs` plus the root manager's direct `summaries`. A grandchild is owned by the child's `SubAgentManager`, so it exists only as an unmatched node in the shared directory and is appended later by `Project`, after persistence. After restart the direct-child provenance scan cannot recover it, the fresh directory contains only the root, `/subagents` drops the descendant, and an authorized root transcript read returns `unknown_target`. Persist the full structural projection (including unmatched non-root directory nodes) before viewer flags, and add a root -> child -> grandchild restart test covering both listing and transcript access.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [Gautam's review bot] Resolved in 0884eef0: the flat path now bounded-scans persisted SubAgentProvenance for direct children before overlaying persisted and live rows, with live rows winning. The added restart/default-collaboration test covers this regression.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
gautam-achieveai: [Gautam's review bot] Resolved in 0884eef0: every pre-start foreground cancellation path now retires the collaboration admission, releasing the root-wide lease and removing the queued directory entry. The full-gate collaboration regression test covers the leak.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [Gautam's review bot] Resolved in 0884eef0 for persisted tab rows: Enrich now stamps structural hierarchy metadata before PersistTabs, and the fresh-directory restart test verifies an authorized root transcript read. A separate nested-descendant persistence gap is called out in the new review.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: restored bounded persisted-provenance reconstruction for idle/evicted conversations while gating the scan off the live polling path. Live rows still win, and the browser hierarchy now preserves the real nested agent identity. Verified by the pool-eviction regression, live-store no-scan test, and both hierarchy browser provider modes.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: every pre-start cancellation and shutdown queue-drain path now goes through CancelQueuedSpawn, which retires the collaboration admission and releases root capacity idempotently. Added caller-cancellation and manager-disposal coverage; LmMultiTurn passed 1,046/1,046.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed: live collaboration rows are enriched before persistence, while viewer-scoped flags remain request-local. Added direct Enrich match/no-match/mixed tests and restart transcript recovery coverage. Persisted hierarchy authorization now survives a fresh directory without widening access.

gautam-achieveai:
[gautam-achieveai's bot] Fixed the review-summary blocker: added a real SubAgentToolProvider regression for retired senders, asserting invalid_sender and the actionable 
o longer active response. The subsequent formal re-review also found and resolved the live hierarchy identity regression, hot-path store scan, direct Enrich coverage, shutdown-drain consistency, and silent scan truncation. Fresh verification: LmMultiTurn 1,046/1,046; sample tests 1,184 passed + 1 platform skip; hierarchy browser 2/2; Release build 0 warnings/0 errors.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [Gautam's review bot] This remains unresolved at `d0ae381c`. `Enrich` still maps only workflow tabs plus the root manager's direct summaries; unmatched directory descendants are appended later by `Project`, after `PersistTabs`, so a root -> child -> grandchild is still absent after restart and root transcript lookup returns `unknown_target`. This requires `MaxDelegationDepth >= 2` (the shipped default is 1), but deeper bounded delegation is a supported #244 configuration. Please keep this thread open and add a configured-depth persistence/restart regression.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
```diff
@@ -0,0 +1,366 @@
...
+        // disk. Folded in first so a live row (added below) always wins on a match, restoring the
+        // pre-#244 flat-listing contract.
+        //
+        // COLD PATH ONLY: gated on there being no live loop for this thread. When the loop is live, the
+        // collaboration-enabled write-through above (or the live SubAgentManager snapshot) already
+        // accounts for every child that matters, and restart recovery for a collaboration-enabled host
+        // goes through the enriched persisted WorkflowRunRegistry tabs a few lines down, not this scan.
+        // Without this gate, every live 3-second sub-agent poll and transcript read paid for a bounded but
+        // still expensive multi-page store scan on every single call.
+        if (loop is null)
```
gautam-achieveai: [Gautam's review bot] **Medium:** `loop is null` is not equivalent to a cold conversation. Live Codex/Claude/Copilot pool entries are different `IMultiTurnAgent` implementations, so they enter this branch and scan up to 2,000 thread rows on every hierarchy poll. In the opposite direction, after restart a default collaboration-off `MultiTurnAgentLoop` is re-created with an empty `SubAgentManager`; this branch is skipped, so persisted ordinary children that were visible while cold disappear as soon as the parent becomes live. Gate the scan on whether the live state actually covers the persisted roster—not on the concrete loop type—and add regressions for both a live CLI loop and a rehydrated collaboration-off loop with a persisted child.

gautam-achieveai (commented):
[Gautam's review bot]

## Verdict

**REQUEST_CHANGES**

The shutdown queue drain, direct-row enrichment, placeholder identity, and scan-cap observability fixes are sound. Two hierarchy persistence issues remain:

1. The existing nested-descendant thread is still unresolved when `MaxDelegationDepth >= 2`: unmatched directory descendants are projected only after persistence.
2. The new `loop is null` cold-path gate both scans live CLI-backed conversations on every poll and hides persisted children after a default collaboration-off conversation is rehydrated.

Focused hierarchy/transcript tests (39) and collaboration integration tests (45) pass, but neither remaining scenario is covered.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 92a94cfb: the pre-persistence structural snapshot now unions enriched tabs with unmatched non-root directory descendants using the same root/workflow-controller exclusions as live projection. Added a real root→child→grandchild restart test proving listing, ancestry/depth, readability, and transcript access survive a fresh registry/directory. Verified sample tests 1,185 passed + 1 platform skip, hierarchy browser 2/2, Release build 0 warnings/0 errors.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [Gautam's review bot] Still unresolved at `92a94cfb`. This incremental commit fixes nested-descendant persistence but does not change the `if (loop is null)` gate at line 193 or add either missing regression. Live CLI-backed `IMultiTurnAgent` implementations still scan up to 2,000 stored threads on every hierarchy poll, while a rehydrated collaboration-off `MultiTurnAgentLoop` still skips provenance recovery and hides persisted ordinary children. Please keep this thread active.

gautam-achieveai (commented):
[Gautam's review bot]

**Verdict: REQUEST_CHANGES.** The nested root -> child -> grandchild persistence finding is fixed at `92a94cfb`: the structural write-through now includes unmatched ordinary directory descendants, and the new restart test covers listing, ancestry/depth, readability, and transcript authorization. No new regression was found in this delta.

The existing scan-gate thread remains unresolved: `if (loop is null)` still makes live CLI-backed agents pay the persisted-thread scan and makes rehydrated collaboration-off `MultiTurnAgentLoop` instances hide persisted children. Focused hierarchy/projection/transcript tests pass (40/40).

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 9dbc4a0c: cold provenance recovery is now based on whether live state covers the ordinary child roster, not on concrete loop type. Live CLI/non-MultiTurn agents skip ListThreads; rehydrated collaboration-off loops with an empty manager recover persisted children; collaboration-on and populated live managers stay on their indexed/live paths. Added RED→GREEN tests for both directions. Verified sample tests 1,187 passed + 1 platform skip, hierarchy browser 2/2, Release build 0 warnings/0 errors.

gautam-achieveai: [Gautam's review bot] The concrete-type mismatch is fixed, but this thread is still unresolved because `ListAgents().Count == 0` is not a stable coverage signal. A rehydrated collaboration-off loop initially recovers its persisted children, but after it spawns one new child the manager becomes nonempty, the scan stops, and every pre-restart child disappears because ordinary collaboration-off rows are not in `WorkflowRunRegistry`. Conversely, a genuinely childless live loop keeps an empty manager forever and now scans up to 2,000 stored threads on every hierarchy request. Cache/retain the recovered roster (or track lifetime coverage explicitly) and always union it with live rows, with live winning. Please add regressions for the empty-to-populated transition and repeated childless requests.

gautam-achieveai (commented):
[Gautam's review bot]

**Verdict: REQUEST_CHANGES.** The final remediation fixes the concrete `MultiTurnAgentLoop` type gate, but the replacement empty-manager heuristic leaves the existing thread unresolved. Rehydrated collaboration-off loops drop all recovered pre-restart children after their first new spawn, while genuinely childless live loops rescan up to 2,000 persisted threads on every hierarchy request. The new tests cover only the initial empty-manager call, not either lifecycle case.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 244b4af6, 263f1402, and 18ad5eaa: recovered provenance rosters are retained per thread and live-manager generation, unioned before live rows so live wins, invalidated naturally on every manager reset, evicted on deletion, and bounded by MaxPersistedHierarchyEntries. Added RED→GREEN coverage for repeated childless requests, empty→populated transition, two reset cycles with an intervening spawn, failed/concurrent scans, deterministic bounded eviction, concurrent multi-key writes, deletion/reuse, and real DI capacity binding. Fresh local verification: sample tests 1,206 passed + 1 platform skip, hierarchy browser 2/2, Release build 0 warnings/0 errors.

gautam-achieveai:
[gautam-achieveai's bot] Final re-review delta is pushed at 18ad5eaa. Independent architecture, performance, coverage, and absolute-final reviews found no remaining HIGH/MEDIUM issues after lifecycle-aware cache invalidation, bounded retention, deletion eviction, and concurrent eviction/DI tests. Local evidence: sample tests 1,206 passed + 1 platform skip; hierarchy browser 2/2; SPA build passed; Release build 0 warnings/0 errors; diff check clean. Awaiting fresh CI and post-push review feedback.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
```diff
@@ -0,0 +1,446 @@
...
+        object owner,
+        CancellationToken ct)
+    {
+        if (scanCoverageCache.TryGetRecovered(threadId, owner, out var cached))
+        {
+            return cached;
+        }
+
+        var scanned = await ScanPersistedSubAgentChildrenAsync(threadId, ct);
+        scanCoverageCache.RecordRecovered(threadId, owner, scanned);
```
gautam-achieveai, line 396: [Gautam's review bot] **Medium:** `Forget` is not ordered against this write-back. A `/subagents` request can miss the cache and begin the persisted scan, then `DELETE` can remove the conversation and call `Forget`, and finally this continuation can recreate the entry under `NoLiveManager`. Later cold reads keep serving that resurrected roster until capacity eviction, defeating the delete/reuse invalidation this patch adds. Capture a per-thread generation before scanning, increment it in `Forget`, and record only if the generation is unchanged; add a deterministic scan/delete interleaving test.

gautam-achieveai (commented):
[Gautam's review bot]

**Verdict: REQUEST_CHANGES**

The owner-keyed cache resolves the prior empty-manager lifecycle defect, and that thread is correctly resolved. One new Medium issue remains: `Delete -> Forget` can race an already-running persisted scan, whose unconditional write-back recreates the deleted cache entry. The inline comment includes the required generation/epoch fix and regression scenario.

samples/LmStreaming.Sample/Services/AgentHierarchyService.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 381b7a8c and hardened in 964e33cf: scans capture a cache-wide monotonic write epoch before I/O, Forget increments it and removes the entry, and stale writeback is rejected if any deletion occurred. The final global epoch avoids the capacity-eviction ABA of per-thread tombstones and keeps retention strictly bounded. Added deterministic delete/scan interleaving, capacity-1 churn, cross-thread invalidation, repeated Forget, thread reuse, explicit-epoch API, and integration coverage. Fresh local verification: sample tests 1,218 passed + 1 platform skip, hierarchy browser 2/2, Release build 0 warnings/0 errors, diff check clean.

gautam-achieveai:
[gautam-achieveai's bot] Global write-epoch hardening is pushed at 964e33cf. Independent architecture and concurrency stress reviews found no HIGH/MEDIUM issues. Local verification: sample tests 1,218 passed + 1 platform skip; SPA build passed; hierarchy browser 2/2; Release build 0 warnings/0 errors; committed delta clean. Awaiting fresh CI and post-push review feedback.

PR #248, opened 2026-08-02
feat(proxy): compose GitHub-first MCP web tools

### LINKED ISSUE #247: Expose GitHub-first web tools through the Proxy MCP endpoint
## Summary

Expose `web_search` and `web_fetch` through the **CopilotAnthropicProxy.Sample MCP tool surface only** (`/mcp` and `/mcp/readonly`).

For each exact tool name, preserve GitHub Copilot's MCP definition and call path when GitHub advertises it. When GitHub omits that name, expose the existing Jina-backed implementation only when `JINA_API_KEY` is configured and valid.

This builds on the Jina web-tool implementation completed in #113.

## Value

MCP clients using the Proxy app receive dependable web tools without configuring a second MCP endpoint, while retaining GitHub's native capabilities whenever available. Existing transparent proxy behavior and security controls remain intact.

### Included

- The Proxy app's `/mcp` and `/mcp/readonly` MCP tool listing and call routing.
- Reuse of the existing Jina provider, contracts, validation, sanitization, and output limits.
- Focused Proxy app tests, documentation, structured logging, and conditional live verification.

### Excluded

- Anthropic native/server tools.
- `/v1/messages` and `/v1/responses` behavior or translation.
- `LmStreaming.Sample` tool registration.
- General-purpose `McpServer.AspNetCore` changes.
- Mid-call backend failover.

## Availability Rules

| GitHub advertises exact name | Valid `JINA_API_KEY` | Exposed implementation |
|---|---|---|
| Yes | Either | GitHub definition and call path, unchanged |
| No | Yes | Existing Jina-backed implementation |
| No | No | Tool omitted |

The rules apply independently to `web_search` and `web_fetch`, so mixed catalogs are valid.

## Delivery Phases

1. Reuse the existing Jina implementation inside the Proxy app without creating a second Jina client or duplicating validation and security logic.
2. Compose supported JSON `tools/list` responses while preserving GitHub-owned definitions.
3. Route locally advertised `tools/call` requests consistently with the catalog and filter context the client received.
4. Preserve transparent behavior for unsupported transport shapes and unrelated MCP traffic.
5. Add lifecycle cleanup, structured observability, security/regression tests, documentation, and an authenticated smoke test when credentials are available.

## Acceptance Criteria

- [ ] Changes are limited to the Proxy app's `/mcp` and `/mcp/readonly` MCP tool surface and its focused tests/documentation.
- [ ] At most one definition for each exact name, `web_search` and `web_fetch`, is advertised.
- [ ] A GitHub-advertised definition wins and remains structurally unchanged.
- [ ] A missing name is supplied by Jina only when `JINA_API_KEY` is nonblank and existing web-tool configuration is valid.
- [ ] Calls use the backend represented by the latest supported catalog returned for the same endpoint, GitHub session, and `X-MCP-*` filter context.
- [ ] A GitHub-owned call failure remains a GitHub error and never triggers Jina failover.
- [ ] GitHub `tools/list` HTTP, transport, and JSON-RPC failures remain failures; they are not replaced with a Jina-only catalog.
- [ ] SSE-framed catalogs, paginated catalogs, JSON-RPC batch arrays, and unrelated MCP methods pass through unchanged without recording local ownership.
- [ ] When Jina is disabled or invalid, the MCP body remains byte-transparent because no composition is required.
- [ ] `/mcp` and `/mcp/readonly` maintain independent routing state, and GitHub readonly/tool-filter restrictions remain authoritative.
- [ ] Session routing state is removed after MCP DELETE and GitHub session-expiry responses.
- [ ] Local call results preserve MCP `isError`, and client cancellation reaches the Jina request.
- [ ] Existing SSRF/private-target blocking, header-injection protection, untrusted-content framing, output caps, and secret-safe error handling remain enforced.
- [ ] `JINA_API_KEY` never appears in tool definitions, results, errors, logs, or GitHub-bound traffic.
- [ ] Locally handled list/call paths retain structured request observability and identify the selected backend without logging tool arguments, URLs, queries, response bodies, or credentials.
- [ ] An unexpected composition failure cannot break an otherwise valid GitHub catalog: list composition logs safely and returns the original upstream response; local-call failures return a bounded MCP/JSON-RPC error rather than an unhandled HTTP 500.
- [ ] Concurrent catalog requests remain isolated; each response is composed from its own GitHub result, with the latest completed supported catalog becoming the routing snapshot.
- [ ] The Proxy app's added dependency footprint is reviewed, including the transitive SQLite dependency currently carried by `Misc`; no duplicate Jina implementation is introduced.

## Verification Gates

- [ ] Focused RED→GREEN tests cover catalog precedence, mixed ownership, key/config gating, routing provenance, failure behavior, pass-through shapes, lifecycle cleanup, readonly isolation, cancellation, and structured errors.
- [ ] Existing Proxy MCP transport tests remain green.
- [ ] Existing Jina secret-hygiene, SSRF-validation, and output-bound tests remain green.
- [ ] Proxy build and tests complete with zero warnings/errors and CSharpier passes.
- [ ] When GitHub and Jina credentials are available, a real MCP client initializes a Proxy session, lists tools, and successfully calls each advertised web tool according to its advertised schema.

## Implementation Notes

Detailed implementation sequencing is captured locally in `docs/superpowers/plans/2026-08-02-mcp-web-tools.md`; the approved design is in `docs/superpowers/specs/2026-08-02-mcp-web-tools-design.md`.

gautam-achieveai:
## Summary

Expose GitHub-first `web_search` and `web_fetch` tools through the Proxy app's existing `/mcp` and `/mcp/readonly` endpoints.

GitHub's exact-name MCP tool definition and call path remains authoritative when available. When GitHub omits a name, the proxy adds the existing hardened Jina implementation only with valid Jina configuration.

## Changes

- Add narrow MCP `tools/list` composition and local `tools/call` routing.
- Preserve transparent forwarding for unrelated, SSE, paginated, batch, and sessionless MCP traffic.
- Respect `X-MCP-Tools`, `X-MCP-Exclude-Tools`, lockdown, session, endpoint, and filter-context boundaries.
- Reuse existing Jina validation, SSRF protection, cancellation, output limits, and secret-safe error handling.
- Add focused routing, transport, lifecycle, failure, cancellation, and security coverage.
- Document MCP availability rules, configuration, limitations, and cleanup behavior.

## Testing

- CSharpier: clean
- Proxy build: 0 warnings, 0 errors
- Proxy tests: 359/359 passed
- Focused Jina/security tests: 106/106 passed
- Publish-size impact: +900,987 bytes from the `Misc` dependency, documented in the README
- Live MCP smoke probe: skipped because `JINA_API_KEY` was unavailable in this session

## Key Decisions

- Scope is limited to the Proxy app's `/mcp` and `/mcp/readonly` surfaces.
- GitHub exact-name tools always win; Jina only fills missing names.
- No mid-call failover between GitHub and Jina.
- Client MCP filters remain authoritative and cannot be bypassed by local injection.
- Unsupported MCP shapes remain byte-transparent.
- Relay and composition share one GitHub transport, timeout, error, and body-cap path.

## Dependency Impact

The Proxy now references `Misc` to reuse its Jina tooling and security implementation. This transitively adds SQLite native assets and increased the measured framework-dependent publish output by approximately 0.9 MB. No SQLite service is constructed by the Proxy.

## Related Issues

Fixes #247

### DIFF
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -239,9 +239,14 @@ internal sealed class McpToolComposition
         var isJson = string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase);
         var isSse = string.Equals(mediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
         var hasBoundedSseBody = isSse && upstream.Content.Headers.ContentLength is { } sseLength && sseLength <= maxBodyBytes;
-        if (!upstream.IsSuccessStatusCode || (!isJson && !hasBoundedSseBody))
+        if (!upstream.IsSuccessStatusCode)
         {
             return null;
+        }
+
+        if (!isJson && !hasBoundedSseBody)
+        {
+            return new McpComposedList(null, endpoint, sessionId, McpSnapshotAction.Remove, Generation: generation);
         }
 
         var original = await ProxyHttp.ReadCappedBytesAsync(
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -164,10 +164,9 @@ internal sealed class McpToolComposition
         }
 
         var mediaType = upstream.Content.Headers.ContentType?.MediaType;
-        if (
-            !upstream.IsSuccessStatusCode
-            || !string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase)
-        )
+        var isJson = string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase);
+        var isSse = string.Equals(mediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
+        if (!upstream.IsSuccessStatusCode || (!isJson && !isSse))
         {
             Invalidate();
             return null;
@@ -181,9 +180,24 @@ internal sealed class McpToolComposition
         }
 
         ReplaceContent(upstream, original, mediaType);
+        var jsonBytes = original;
+        string? ssePrefix = null;
+        string? sseSuffix = null;
+        if (isSse)
+        {
+            var sse = Encoding.UTF8.GetString(original);
+            if (!TryReadSingleSseMessage(sse, out ssePrefix, out var json, out sseSuffix))
+            {
+                Invalidate();
+                return original;
+            }
+
+            jsonBytes = Encoding.UTF8.GetBytes(json);
+        }
+
         try
         {
-            var root = JsonNode.Parse(original) as JsonObject;
+            var root = JsonNode.Parse(jsonBytes) as JsonObject;
             var result = root?["result"] as JsonObject;
             var tools = result?["tools"] as JsonArray;
             if (
@@ -224,7 +238,10 @@ internal sealed class McpToolComposition
             }
 
             upstream.Headers.ETag = null;
-            var composed = Encoding.UTF8.GetBytes(root!.ToJsonString());
+            var composedJson = root!.ToJsonString();
+            var composed = isSse
+                ? Encoding.UTF8.GetBytes(ssePrefix + composedJson + sseSuffix)
+                : Encoding.UTF8.GetBytes(composedJson);
             _logger.LogInformation(
                 "MCP tools listed endpoint={Endpoint} backend={Backend} status={Status} localToolCount={LocalToolCount}",
                 endpoint,
@@ -254,6 +271,50 @@ internal sealed class McpToolComposition
 
     private static string ToolCallId(JsonNode id) =>
         id is JsonValue value && value.TryGetValue<string>(out var text) ? text : id.ToJsonString();
+
+    private static bool TryReadSingleSseMessage(
+        string sse,
+        out string prefix,
+        out string json,
+        out string suffix
+    )
+    {
+        prefix = string.Empty;
+        json = string.Empty;
+        suffix = string.Empty;
+        var newline = sse.Contains("\r\n", StringComparison.Ordinal) ? "\r\n" : "\n";
+        var terminator = newline + newline;
+        if (!sse.EndsWith(terminator, StringComparison.Ordinal))
+        {
+            return false;
+        }
+
+        var lines = sse[..^terminator.Length].Split(newline, StringSplitOptions.None);
+        var dataIndexes = lines
+            .Select((line, index) => (line, index))
+            .Where(item => item.line.StartsWith("data:", StringComparison.Ordinal))
+            .Select(item => item.index)
+            .ToArray();
+        if (dataIndexes.Length != 1 || lines.Any(string.IsNullOrWhiteSpace))
+        {
+            return false;
+        }
+
+        var dataIndex = dataIndexes[0];
+        var dataLine = lines[dataIndex];
+        var separatorLength = dataLine.StartsWith("data: ", StringComparison.Ordinal) ? 6 : 5;
+        json = dataLine[separatorLength..];
+        prefix = string.Join(newline, lines[..dataIndex]);
+        if (dataIndex > 0)
+        {
+            prefix += newline;
+        }
+        prefix += dataLine[..separatorLength];
+        suffix = dataIndex + 1 < lines.Length
+            ? newline + string.Join(newline, lines[(dataIndex + 1)..]) + terminator
+            : terminator;
+        return true;
+    }
 
     private static void ReplaceContent(HttpResponseMessage upstream, byte[] body, string? mediaType)
     {
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -97,8 +97,7 @@ internal sealed class McpToolComposition
         var endpoint = context.Request.Path.Value;
         if (request["params"] is not JsonObject parameters)
         {
-            await WriteInvalidParamsAsync(context, request["id"]!);
-            return true;
+            return false;
         }
 
         var name = Text(parameters["name"]);
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -5,6 +5,13 @@ using System.Text.Json.Nodes;
 using AchieveAi.LmDotnetTools.LmCore.Messages;
 using AchieveAi.LmDotnetTools.LmCore.Middleware;
 
+internal sealed record McpComposedList(
+    byte[] Body,
+    string? Endpoint,
+    string? SessionId,
+    McpToolSnapshot? Snapshot
+);
+
 internal sealed class McpToolComposition
 {
     private readonly McpJinaToolCatalog _catalog;
@@ -132,12 +139,14 @@ internal sealed class McpToolComposition
         return true;
     }
 
-    public async Task<byte[]?> ComposeListAsync(
+    public async Task<McpComposedList?> ComposeListAsync(
         HttpContext context,
         JsonObject request,
         HttpResponseMessage upstream,
         long maxBodyBytes,
-        CancellationToken cancellationToken
+        CancellationToken cancellationToken,
+        CancellationTokenSource idleCts,
+        TimeSpan idleTimeout
     )
     {
         if (!IsEnabled || !IsToolsList(request))
@@ -172,11 +181,17 @@ internal sealed class McpToolComposition
             return null;
         }
 
-        var original = await ProxyHttp.ReadCappedBytesAsync(upstream.Content, maxBodyBytes, cancellationToken);
+        var original = await ProxyHttp.ReadCappedBytesAsync(
+            upstream.Content,
+            maxBodyBytes,
+            cancellationToken,
+            idleCts,
+            idleTimeout
+        );
         if (original is null)
         {
             Invalidate();
-            return [];
+            return new McpComposedList([], endpoint, sessionId, null);
         }
 
         ReplaceContent(upstream, original, mediaType);
@@ -189,7 +204,7 @@ internal sealed class McpToolComposition
             if (!TryReadSingleSseMessage(sse, out ssePrefix, out var json, out sseSuffix))
             {
                 Invalidate();
-                return original;
+                return new McpComposedList(original, endpoint, sessionId, null);
             }
 
             jsonBytes = Encoding.UTF8.GetBytes(json);
@@ -211,7 +226,7 @@ internal sealed class McpToolComposition
             )
             {
                 Invalidate();
-                return original;
+                return new McpComposedList(original, endpoint, sessionId, null);
             }
 
             var githubNames = tools
@@ -226,15 +241,14 @@ internal sealed class McpToolComposition
                 tools.Add(local.Definition.DeepClone());
             }
 
-            _snapshots.Set(
-                endpoint,
-                sessionId,
-                new McpToolSnapshot(injectable.Select(tool => tool.Name).ToHashSet(StringComparer.Ordinal), headerContext)
+            var snapshot = new McpToolSnapshot(
+                injectable.Select(tool => tool.Name).ToHashSet(StringComparer.Ordinal),
+                headerContext
             );
 
             if (injectable.Count == 0)
             {
-                return original;
+                return new McpComposedList(original, endpoint, sessionId, snapshot);
             }
 
             upstream.Headers.ETag = null;
@@ -249,13 +263,25 @@ internal sealed class McpToolComposition
                 (int)upstream.StatusCode,
                 injectable.Count
             );
-            return composed;
+            return new McpComposedList(composed, endpoint, sessionId, snapshot);
         }
         catch (Exception ex) when (ex is JsonException or InvalidOperationException)
         {
             Invalidate();
             _logger.LogWarning(ex, "MCP tool-list composition failed; returning the original upstream response");
-            return original;
+            return new McpComposedList(original, endpoint, sessionId, null);
+        }
+    }
+
+    public void PublishSnapshot(McpComposedList composed)
+    {
+        if (
+            composed.Snapshot is not null
+            && !string.IsNullOrWhiteSpace(composed.Endpoint)
+            && !string.IsNullOrWhiteSpace(composed.SessionId)
+        )
+        {
+            _snapshots.Set(composed.Endpoint, composed.SessionId, composed.Snapshot);
         }
     }
 
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -68,7 +68,7 @@ internal sealed class McpToolComposition
 
         var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
         var endpoint = context.Request.Path.Value;
-        var requestId = request["params"]?["requestId"];
+        var requestId = (request["params"] as JsonObject)?["requestId"];
         if (
             string.IsNullOrWhiteSpace(endpoint)
             || string.IsNullOrWhiteSpace(sessionId)
@@ -80,7 +80,14 @@ internal sealed class McpToolComposition
 
         if (_localCalls.TryGetValue((endpoint, sessionId, requestId.ToJsonString()), out var cancellation))
         {
-            cancellation.Cancel();
+            try
+            {
+                cancellation.Cancel();
+            }
+            catch (ObjectDisposedException)
+            {
+                // The local call completed between lookup and cancellation; treat it as already finished.
+            }
         }
 
         return false;
@@ -217,20 +224,11 @@ internal sealed class McpToolComposition
         var endpoint = context.Request.Path.Value;
         var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
 
-        void Invalidate()
-        {
-            if (!string.IsNullOrWhiteSpace(endpoint) && !string.IsNullOrWhiteSpace(sessionId))
-            {
-                _snapshots.Remove(endpoint, sessionId);
-            }
-        }
-
         var mediaType = upstream.Content.Headers.ContentType?.MediaType;
         var isJson = string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase);
         var isSse = string.Equals(mediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
         if (!upstream.IsSuccessStatusCode || (!isJson && !isSse))
         {
-            Invalidate();
             return null;
         }
 
@@ -243,7 +241,6 @@ internal sealed class McpToolComposition
         );
         if (original is null)
         {
-            Invalidate();
             return new McpComposedList([], endpoint, sessionId, McpSnapshotAction.Remove);
         }
 
@@ -256,7 +253,6 @@ internal sealed class McpToolComposition
             var sse = Encoding.UTF8.GetString(original);
             if (!TryReadSingleSseMessage(sse, out ssePrefix, out var json, out sseSuffix))
             {
-                Invalidate();
                 return new McpComposedList(original, endpoint, sessionId, McpSnapshotAction.Remove);
             }
 
@@ -278,7 +274,6 @@ internal sealed class McpToolComposition
                 || !McpToolSnapshotStore.TryBuildHeaderContext(context.Request.Headers, out var headerContext)
             )
             {
-                Invalidate();
                 return new McpComposedList(original, endpoint, sessionId, McpSnapshotAction.Remove);
             }
 
@@ -320,7 +315,6 @@ internal sealed class McpToolComposition
         }
         catch (Exception ex) when (ex is JsonException or InvalidOperationException)
         {
-            Invalidate();
             _logger.LogWarning(ex, "MCP tool-list composition failed; returning the original upstream response");
             return new McpComposedList(original, endpoint, sessionId, McpSnapshotAction.Remove);
         }
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -1,0 +1,264 @@
+using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using AchieveAi.LmDotnetTools.LmCore.Messages;
+using AchieveAi.LmDotnetTools.LmCore.Middleware;
+
+internal sealed class McpToolComposition
+{
+    private readonly McpJinaToolCatalog _catalog;
+    private readonly McpToolSnapshotStore _snapshots;
+    private readonly ILogger _logger;
+
+    public McpToolComposition(
+        McpJinaToolCatalog catalog,
+        McpToolSnapshotStore snapshots,
+        ILoggerFactory loggerFactory
+    )
+    {
+        _catalog = catalog;
+        _snapshots = snapshots;
+        _logger = loggerFactory.CreateLogger("CopilotAnthropicProxy.McpTools");
+    }
+
+    public bool IsEnabled => _catalog.IsEnabled;
+
+    public static bool TryParseSingleRequest(ReadOnlySpan<byte> body, out JsonObject? request)
+    {
+        request = null;
+        try
+        {
+            request = JsonNode.Parse(body) as JsonObject;
+            return request is not null;
+        }
+        catch (JsonException)
+        {
+            return false;
+        }
+    }
+
+    public static bool IsToolsList(JsonObject request) => Text(request["method"]) == "tools/list";
+
+    public static bool IsToolsCall(JsonObject request) => Text(request["method"]) == "tools/call";
+
+    public async Task<bool> TryHandleCallAsync(HttpContext context, JsonObject request)
+    {
+        if (!IsEnabled || !IsToolsCall(request) || !request.ContainsKey("id") || request["id"] is null)
+        {
+            return false;
+        }
+
+        var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
+        var endpoint = context.Request.Path.Value;
+        var name = Text(request["params"]?["name"]);
+        if (
+            string.IsNullOrWhiteSpace(sessionId)
+            || string.IsNullOrWhiteSpace(endpoint)
+            || string.IsNullOrWhiteSpace(name)
+            || !_snapshots.TryGet(endpoint, sessionId, out var snapshot)
+            || !snapshot.LocalToolNames.Contains(name)
+            || !_catalog.Tools.TryGetValue(name, out var tool)
+        )
+        {
+            return false;
+        }
+
+        if (
+            McpToolSnapshotStore.HasToolFilterHeaders(context.Request.Headers)
+            && (
+                !McpToolSnapshotStore.TryBuildHeaderContext(context.Request.Headers, out var contextHeaders)
+                || !string.Equals(contextHeaders, snapshot.HeaderContext, StringComparison.Ordinal)
+            )
+        )
+        {
+            return false;
+        }
+
+        var stopwatch = Stopwatch.StartNew();
+        ToolHandlerResult result;
+        try
+        {
+            var arguments = request["params"]?["arguments"]?.ToJsonString() ?? "{}";
+            result = await tool.Handler(
+                arguments,
+                new ToolCallContext { ToolCallId = ToolCallId(request["id"]!) },
+                context.RequestAborted
+            );
+        }
+        catch (OperationCanceledException) when (context.RequestAborted.IsCancellationRequested)
+        {
+            return true;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Local MCP tool {ToolName} failed unexpectedly", name);
+            result = ToolHandlerResult.FromError("The local MCP tool failed unexpectedly.");
+        }
+
+        var payload = result switch
+        {
+            ToolHandlerResult.Resolved resolved => resolved.Payload,
+            ToolHandlerResult.Deferred => new ToolHandlerResultPayload(
+                "Deferred tool execution is not supported by this MCP endpoint.",
+                IsError: true
+            ),
+            _ => new ToolHandlerResultPayload("The local MCP tool returned an unsupported result.", IsError: true),
+        };
+
+        var response = new JsonObject
+        {
+            ["jsonrpc"] = "2.0",
+            ["id"] = request["id"]!.DeepClone(),
+            ["result"] = new JsonObject
+            {
+                ["content"] = new JsonArray(new JsonObject { ["type"] = "text", ["text"] = payload.Text }),
+                ["isError"] = payload.IsError,
+            },
+        };
+        var bytes = Encoding.UTF8.GetBytes(response.ToJsonString());
+        context.Response.StatusCode = StatusCodes.Status200OK;
+        context.Response.ContentType = "application/json";
+        await context.Response.Body.WriteAsync(bytes, context.RequestAborted);
+        _logger.LogInformation(
+            "MCP tool handled endpoint={Endpoint} backend={Backend} tool={ToolName} status={Status} elapsedMs={ElapsedMs} isError={IsError}",
+            endpoint,
+            "jina-local",
+            name,
+            StatusCodes.Status200OK,
+            stopwatch.ElapsedMilliseconds,
+            payload.IsError
+        );
+        return true;
+    }
+
+    public async Task<byte[]?> ComposeListAsync(
+        HttpContext context,
+        JsonObject request,
+        HttpResponseMessage upstream,
+        long maxBodyBytes,
+        CancellationToken cancellationToken
+    )
+    {
+        if (!IsEnabled || !IsToolsList(request))
+        {
+            return null;
+        }
+
+        var endpoint = context.Request.Path.Value;
+        var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
+        if (
+            sessionId is null
+            && upstream.Headers.TryGetValues("Mcp-Session-Id", out var responseSessionIds)
+        )
+        {
+            sessionId = responseSessionIds.FirstOrDefault();
+        }
+
+        void Invalidate()
+        {
+            if (!string.IsNullOrWhiteSpace(endpoint) && !string.IsNullOrWhiteSpace(sessionId))
+            {
+                _snapshots.Remove(endpoint, sessionId);
+            }
+        }
+
+        var mediaType = upstream.Content.Headers.ContentType?.MediaType;
+        if (
+            !upstream.IsSuccessStatusCode
+            || !string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase)
+        )
+        {
+            Invalidate();
+            return null;
+        }
+
+        var original = await ProxyHttp.ReadCappedBytesAsync(upstream.Content, maxBodyBytes, cancellationToken);
+        if (original is null)
+        {
+            Invalidate();
+            return [];
+        }
+
+        ReplaceContent(upstream, original, mediaType);
+        try
+        {
+            var root = JsonNode.Parse(original) as JsonObject;
+            var result = root?["result"] as JsonObject;
+            var tools = result?["tools"] as JsonArray;
+            if (
+                root?["error"] is not null
+                || result is null
+                || tools is null
+                || !string.IsNullOrWhiteSpace(Text(result["nextCursor"]))
+                || string.IsNullOrWhiteSpace(endpoint)
+                || string.IsNullOrWhiteSpace(sessionId)
+                || !McpToolSnapshotStore.TryBuildHeaderContext(context.Request.Headers, out var headerContext)
+            )
+            {
+                Invalidate();
+                return original;
+            }
+
+            var githubNames = tools
+                .OfType<JsonObject>()
+                .Select(tool => Text(tool["name"]))
+                .Where(name => name is not null)
+                .Cast<string>()
+                .ToHashSet(StringComparer.Ordinal);
+            var injectable = _catalog.SelectInjectable(context.Request.Headers, githubNames);
+            foreach (var local in injectable)
+            {
+                tools.Add(local.Definition.DeepClone());
+            }
+
+            _snapshots.Set(
+                endpoint,
+                sessionId,
+                new McpToolSnapshot(injectable.Select(tool => tool.Name).ToHashSet(StringComparer.Ordinal), headerContext)
+            );
+
+            if (injectable.Count == 0)
+            {
+                return original;
+            }
+
+            upstream.Headers.ETag = null;
+            var composed = Encoding.UTF8.GetBytes(root!.ToJsonString());
+            _logger.LogInformation(
+                "MCP tools listed endpoint={Endpoint} backend={Backend} status={Status} localToolCount={LocalToolCount}",
+                endpoint,
+                "github+jina-local",
+                (int)upstream.StatusCode,
+                injectable.Count
+            );
+            return composed;
+        }
+        catch (Exception ex) when (ex is JsonException or InvalidOperationException)
+        {
+            Invalidate();
+            _logger.LogWarning(ex, "MCP tool-list composition failed; returning the original upstream response");
+            return original;
+        }
+    }
+
+    public void RemoveSnapshot(string endpoint, string? sessionId)
+    {
+        if (!string.IsNullOrWhiteSpace(sessionId))
+        {
+            _snapshots.Remove(endpoint, sessionId);
+        }
+    }
+
+    private static string? Text(JsonNode? node) => node is JsonValue value && value.TryGetValue<string>(out var text) ? text : null;
+
+    private static string ToolCallId(JsonNode id) =>
+        id is JsonValue value && value.TryGetValue<string>(out var text) ? text : id.ToJsonString();
+
+    private static void ReplaceContent(HttpResponseMessage upstream, byte[] body, string? mediaType)
+    {
+        upstream.Content.Dispose();
+        upstream.Content = new ByteArrayContent(body);
+        upstream.Content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue(mediaType ?? "application/json");
+    }
+}
--- a/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
+++ b/samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
@@ -18,7 +18,8 @@ internal sealed record McpComposedList(
     string? Endpoint,
     string? SessionId,
     McpSnapshotAction SnapshotAction,
-    McpToolSnapshot? Snapshot = null
+    McpToolSnapshot? Snapshot = null,
+    long Generation = 0
 );
 
 internal sealed class McpToolComposition
@@ -206,6 +207,15 @@ internal sealed class McpToolComposition
         return true;
     }
 
+    public long CaptureListGeneration(HttpContext context, JsonObject request)
+    {
+        var endpoint = context.Request.Path.Value;
+        var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
+        return IsToolsList(request) && !string.IsNullOrWhiteSpace(endpoint) && !string.IsNullOrWhiteSpace(sessionId)
+            ? _snapshots.Generation(endpoint, sessionId)
+            : 0;
+    }
+
     public async Task<McpComposedList?> ComposeListAsync(
         HttpContext context,
         JsonObject request,
@@ -213,7 +223,8 @@ internal sealed class McpToolComposition
         long maxBodyBytes,
         CancellationToken cancellationToken,
         CancellationTokenSource idleCts,
-        TimeSpan idleTimeout
+        TimeSpan idleTimeout,
+        long generation
     )
     {
         if (!IsEnabled || !IsToolsList(request))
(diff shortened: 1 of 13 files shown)


gautam-achieveai:
[Revobot]

# Code Review: `achieveai/LmDotnetTools#248`

**Commit reviewed:** `cc21abc9290185ea28f061d5d7ac9ee81ff209bb`  
**Verdict:** **REQUEST CHANGES**

## Findings

### [BLOCKER] Injected MCP schemas may be serialized with an invalid `type` shape

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpJinaToolCatalog.cs:69-74`

`FunctionContract.GetJsonSchema()` returns a schema whose `Type` property is `Union<string, IReadOnlyList<string>>`. The PR serializes that object with the default `JsonSerializer`, but the repository’s required union converter is registered in `JsonSchemaValidator.SchemaSerializationOptions`, not globally.

Consequently, the advertised MCP `inputSchema.type` can be emitted as a CLR-shaped object instead of the JSON Schema string/array representation. MCP clients may reject the injected tools or fail to invoke them.

Use the repository’s schema serializer options:

```csharp
JsonNode.Parse(
    JsonSerializer.Serialize(schema, JsonSchemaValidator.SchemaSerializationOptions)
)
```

Add a wire-level test asserting that the injected schema contains `"type": "object"` plus the expected `properties` and `required` values.

---

### [BLOCKER] Sessionless requests can receive locally injected tools

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:148-155`

When the request has no `Mcp-Session-Id`, the implementation adopts an ID returned by the upstream response:

```csharp
if (sessionId is null && upstream.Headers.TryGetValues("Mcp-Session-Id", out var responseSessionIds))
{
    sessionId = responseSessionIds.FirstOrDefault();
}
```

That permits a sessionless request to receive Jina-injected tools and stores a routing snapshot even though the client did not present a session identifier that can safely bind later calls to that catalog. This contradicts the design’s explicit sessionless behavior and can create incorrect local routing.

Only compose and store a local snapshot when the incoming request already contains a nonblank `Mcp-Session-Id`. Add a regression test for a sessionless request whose upstream response includes a session header.

---

### [BLOCKER] MCP snapshot state has no eviction or capacity bound

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolSnapshotStore.cs:10-18`

The singleton dictionary retains one entry per endpoint/session pair and removes entries only after DELETE or an upstream 404. Abandoned clients that disconnect without either request leave snapshots resident for the lifetime of the process.

The header provenance string can be up to 8 KB, and session IDs have no explicit length limit. A long-running process can therefore accumulate stale state and memory pressure.

Add an explicit bounded lifecycle policy, such as:

- absolute/sliding expiration;
- a maximum entry or memory limit;
- a session-ID length limit;
- periodic eviction.

Retain DELETE/404 removal as an optimization.

---

### [HIGH] Structurally invalid local calls are dispatched as tool errors instead of JSON-RPC invalid-params errors

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:47-87`

Local dispatch validates only that `params.name` is a nonblank string. A locally owned call with malformed parameters—such as a scalar or array `arguments` value—is serialized and passed to the handler. The result is then returned as an MCP `isError: true` tool result rather than a JSON-RPC `-32602 Invalid params` error.

Validate the intercepted request shape before invoking the handler and return a JSON-RPC error with the original ID for structurally invalid parameters. Keep this separate from valid tool-level validation failures, which may correctly be represented by `isError: true`; `arguments` should not be rejected merely because it is omitted if the MCP contract permits omission.

---

### [MEDIUM] `tools/list` response content metadata is lost on the pass-through path

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:176-183, 258-263`

`ReplaceContent` runs for every supported successful JSON `tools/list` response, including catalogs where no local tool is injected. It disposes the original content and recreates it with only the bare media type:

```csharp
upstream.Content.Headers.ContentType =
    new MediaTypeHeaderValue(mediaType ?? "application/json");
```

This drops content-type parameters such as `charset` and other content metadata, so the supposedly transparent path is no longer wire-preserving.

Avoid replacing `upstream.Content` when returning the original bytes. When composition is required, preserve the complete original content headers and update only metadata that must change for the new body.

---

### [MEDIUM] Catalog composition creates multiple full-size in-memory representations

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:176-227`  
**Related:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2683-2689`

The composition path can simultaneously retain:

- the `MemoryStream` backing buffer;
- the returned original `byte[]`;
- a `JsonNode` object graph;
- a UTF-16 string from `ToJsonString()`;
- a second UTF-8 serialized `byte[]`.

With the default 32 MB proxy cap, concurrent large catalog requests can cause substantially higher transient memory usage and LOH/GC pressure. The path also fully buffers successful catalogs even when no local tool is injected.

Use a smaller MCP-specific catalog limit, avoid the DOM/string/byte[] round trip where practical, and return the original bytes without reparsing or replacement when no injection is needed.

---

### [MEDIUM] All MCP POST bodies are parsed into a DOM when Jina is enabled

**File:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2583-2595`  
**Related:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:27-38`

Once local tools are enabled, every MCP POST is first buffered and parsed into a `JsonNode`, including `initialize`, notifications, unrelated methods, and requests that will be forwarded unchanged. The raw body is already retained for forwarding, so these requests temporarily hold both the byte buffer and a parsed object graph.

Use a lightweight top-level method scan first, and materialize a JSON object only for single-object `tools/list` and `tools/call` requests. This preserves the transparent path’s lower allocation cost.

## Testing and verification

The PR adds useful focused tests, but the following important behaviors remain uncovered:

- sessionless catalog responses with an upstream session header;
- successful local `web_fetch` execution;
- MCP catalog body-cap behavior;
- replacement of an existing snapshot after a later `tools/list`;
- the reverse mixed-ownership case where GitHub owns `web_fetch` and Jina supplies `web_search`;
- schema serialization on the actual MCP wire.

`git diff --check` passed. Build and test execution was unavailable because the review environment does not contain the `dotnet` executable (`/bin/sh: dotnet: not found`).

No files were modified and no provider comments were posted.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-60069d30116b40c1bd14e6e93b84e140&focus=1

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,264 @@
...
+        if (
+            !upstream.IsSuccessStatusCode
+            || !string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase)
+        )
+        {
+            Invalidate();
+            return null;
+        }
+
+        var original = await ProxyHttp.ReadCappedBytesAsync(upstream.Content, maxBodyBytes, cancellationToken);
```
gautam-achieveai: [Gautam's review bot] **[Medium] Reset the idle deadline while buffering `tools/list`.** The linked token passed here was armed before `SendAsync`, but `ReadCappedBytesAsync` never resets it after successful reads. A healthy chunked catalog that keeps making progress but takes longer than the configured idle interval is therefore cancelled and returned as 504. This contradicts the reset-per-read behavior used by `ProxyHttp.CopyBodyAsync` and described in `ProxyMcp`. Please reset the deadline after each buffered read (or make the capped reader explicitly idle-aware).

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,264 @@
...
+                .Where(name => name is not null)
+                .Cast<string>()
+                .ToHashSet(StringComparer.Ordinal);
+            var injectable = _catalog.SelectInjectable(context.Request.Headers, githubNames);
+            foreach (var local in injectable)
+            {
+                tools.Add(local.Definition.DeepClone());
+            }
+
+            _snapshots.Set(
```
gautam-achieveai: [Gautam's review bot] **[Medium] Publish the snapshot only after the composed catalog is delivered.** This `Set` runs before `Program.cs` writes `composedBody`. With concurrent `tools/list` requests, snapshot order follows upstream buffering rather than completed downstream responses: request A can be delivered after B while routing remains based on B. A downstream write failure also leaves an unseen catalog authoritative. That violates #247's requirement that routing follow the latest completed catalog. Return the candidate snapshot and commit it only after the response write succeeds.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            return false;
+        }
+
+        var stopwatch = Stopwatch.StartNew();
+        ToolHandlerResult result;
+        try
+        {
+            var arguments = request["params"]?["arguments"]?.ToJsonString() ?? "{}";
+            result = await tool.Handler(
```
gautam-achieveai, line 154: [Gautam's review bot] **[Medium] MCP cancellation notifications cannot cancel intercepted local calls.** The local handler only receives the original HTTP request's `RequestAborted`. A later `notifications/cancelled` POST is forwarded to GitHub, but GitHub never received this intercepted request, so the Jina operation continues and may still produce a response. MCP receivers should stop processing cancelled requests and free their resources. Track in-flight local calls by endpoint/session/request ID and cancel the linked token when the matching notification arrives.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found three medium-severity correctness gaps: buffered tool-list reads turn the idle timeout into a total-transfer timeout, routing snapshots are published before clients receive the catalog, and MCP cancellation notifications do not reach intercepted Jina calls. The focused MCP tests pass, but none covers these scenarios.

Context checked: #247 acceptance criteria and binding corrections, the approved design/implementation documents, existing Proxy/Jina architecture, official MCP cancellation behavior, and all changed production/test files.

GitHub does not allow this account to request changes on its own PR, so this is submitted as a blocking review comment rather than an approval.

gautam-achieveai:
[gautam-achieveai's bot] Live MCP verification completed after copying JINA_API_KEY via a secret-safe script from the main checkout's ignored LmStreaming .env into the Proxy app's ignored .env. The live GitHub MCP server returned 	ools/list as one vent: message SSE block, which exposed a gap in the initial implementation. Added bounded SSE catalog composition and reran verification: build 0 warnings/errors, Proxy tests 360/360, focused Misc tests 106/106. Real proxy probe advertised exactly one web_search and one web_fetch; both calls returned HTTP 200 with isError: false.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Incremental re-review:** this is still unresolved, and `7903e65c` broadens the impact to the live GitHub shape. `text/event-stream` `tools/list` replies now enter `ReadCappedBytesAsync`, while the previous revision relayed SSE through `CopyBodyAsync`, which resets the idle deadline after every successful read. A healthy SSE catalog that keeps making progress but takes longer than the configured idle interval can now be returned as 504. Please make the capped reader reset the idle deadline after each successful read before using it for SSE composition.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental re-review of `7903e65c`: REQUEST CHANGES.**

The new single-message SSE composition works for the live GitHub framing, and I found no separate high-confidence defect in the parser/reconstruction logic. However, it does not address any prior finding. I replied on the existing idle-timeout thread because this commit now routes live SSE catalogs through the same non-resetting capped reader, broadening that defect rather than warranting a duplicate comment.

Still unresolved: idle-deadline reset, snapshot publication before downstream delivery, MCP cancellation for intercepted calls, schema union serialization, sessionless snapshot adoption, unbounded snapshot state, invalid-params handling, content metadata loss, catalog memory amplification, and parsing every MCP POST into a DOM.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in ecfe2958. The bounded catalog reader now resets the existing idle CTS after every successful read, preserving reset-per-progress behavior for JSON and SSE catalog composition. Full verification: Proxy tests 360/360 and focused Misc tests 106/106.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in ecfe2958. Composition now returns a candidate snapshot with the body; ProxyMcp publishes it only after the downstream write completes successfully. Failed or cancelled writes no longer make an unseen catalog authoritative.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] By design for #247. The approved scope intercepts only tools/list and locally-owned tools/call; all notifications remain transparent, and local Jina cancellation is tied to the originating HTTP RequestAborted token (covered by a live cancellation test). Supporting a later notifications/cancelled POST would require a new endpoint/session/request-id in-flight registry and expand the narrow composer into notification routing. That architecture is outside this item; this thread is being resolved as an accepted limitation rather than silently widening scope.

gautam-achieveai:
[Revobot]

## New finding

### [Medium] Malformed `tools/call` parameters can produce an unhandled 500

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:61`

```csharp
var name = Text(request["params"]?["name"]);
```

The code indexes `params` as an object without first verifying its JSON kind. A targeted request such as:

```json
{"jsonrpc":"2.0","id":1,"method":"tools/call","params":[]}
```

causes the `JsonNode` string indexer to throw `InvalidOperationException` before the handler’s exception boundary is reached. The request therefore becomes an unhandled ASP.NET error instead of the documented JSON-RPC `-32602` invalid-params response preserving the original ID.

Validate that `params` is a `JsonObject` before reading `name` and `arguments`, and return the specified `-32602` response for malformed intercepted calls.

**Verification note:** No sub-agents were dispatched. The focused test command could not run because the sandbox has no `dotnet` executable.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-f66e6658f00d4d48a2ddc2a6cf6a657c&focus=1

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Incremental re-review: only the publication half is fixed.** `ecfe2958` still calls `Invalidate()` inside `ComposeListAsync`, so an existing snapshot is removed before `Program.cs` writes the replacement response body. If that write is cancelled or fails, the client still has its previously delivered catalog while subsequent calls are incorrectly forwarded without that catalog's local ownership. Please represent list processing as a deferred set/remove/no-op mutation and apply it only after successful downstream delivery; add failed-write and concurrent-list coverage.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Incremental re-review: this accepted limitation conflicts with #247's explicit criterion that client cancellation reaches Jina.** `RequestAborted` only covers aborting the original HTTP call; it does not handle a later MCP `notifications/cancelled` request for an intercepted operation. The MCP specification says receivers should stop processing, free resources, and not send a response. Either route cancellation to in-flight local calls, or explicitly revise #247 and track the protocol gap before merging.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental verdict: changes still required.** `ecfe2958` fixes progress-aware catalog buffering and defers snapshot publication, but snapshot removal remains eager, so failed response delivery can invalidate routing before the client receives a replacement catalog. The cancellation resolution also conflicts with #247's stated acceptance criterion. The current malformed non-object `params` finding is valid but duplicates the original structural invalid-params feedback, so I did not create another thread.

Validation: Proxy tests 360/360, focused MCP tests 29/29, focused Misc/Jina web-tool tests 112/112, and both requested diff checks passed.

gautam-achieveai:
[gautam-achieveai's bot] Addressed the latest blocking review in 87f5bb52: malformed non-object params now return JSON-RPC -32602; sessionless requests never adopt an upstream session for local injection; snapshot set/remove mutations are applied only after successful downstream delivery; and MCP notifications/cancelled now cancel matching in-flight local Jina calls by endpoint/session/request id. Verification: build 0 warnings/errors, Proxy tests 362/362, focused Misc tests 106/106.

gautam-achieveai:
[Revobot]

## Review: REQUEST CHANGES

### [Medium] Preserve transparent forwarding for calls that are not locally owned

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:64-73`

`TryHandleCallAsync` returns a local `-32602` response for any `tools/call` request with non-object `params` as soon as local Jina support is enabled:

```csharp
if (request["params"] is not JsonObject parameters)
{
    await WriteInvalidParamsAsync(context, request["id"]!);
    return true;
}
```

This validation occurs before checking the endpoint, session snapshot, requested tool name, or whether the tool was locally injected. Therefore, with `JINA_API_KEY` configured, malformed calls for GitHub-owned tools, unknown tools, or sessions with no local snapshot are intercepted instead of being forwarded unchanged. That violates the documented behavior that GitHub-owned and unknown calls remain transparent and can also alter upstream error semantics solely due to enabling Jina.

Only return the local `-32602` response after identifying a matching local snapshot and locally advertised tool. Otherwise return `false` so the existing relay forwards the original request to GitHub.

No sub-agents were dispatched for this review.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-263f7c6efa414cb8920eecfe98fcb129&focus=1

gautam-achieveai:
[gautam-achieveai's bot] Fixed the post-87f5bb52 transparency finding in 69f46119. Non-object params are now passed through unless local ownership can be established; malformed arguments for a confirmed local tool still return JSON-RPC -32602. Added both regression cases. Verification: build 0 warnings/errors, Proxy tests 363/363, focused Misc tests 106/106.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 4: snapshot removal is still eager.** `87f5bb52` adds `McpSnapshotAction.Remove`, but every invalidation path still calls `Invalidate()`, which executes `_snapshots.Remove(...)` before the replacement response is written. The deferred action therefore duplicates an already-completed mutation rather than deferring it. The unsupported-response path returns `null`, and the body-cap path returns an empty body that `Program.cs` turns into 502 before `PublishSnapshot`, so those removals can never be delivery-gated. A failed/cancelled downstream write still leaves the client using its previous catalog while routing has already discarded it. Remove the eager mutation and apply set/remove only after the corresponding response body is written; paths that do not deliver a replacement catalog must retain the prior snapshot.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,433 @@
...
+    public bool TryHandleCancellation(HttpContext context, JsonObject request)
+    {
+        if (Text(request["method"]) != "notifications/cancelled")
+        {
+            return false;
+        }
+
+        var sessionId = context.Request.Headers["Mcp-Session-Id"].FirstOrDefault();
+        var endpoint = context.Request.Path.Value;
+        var requestId = request["params"]?["requestId"];
```
gautam-achieveai: [Gautam's review bot] **[Medium] Make cancellation notifications safe under malformed-input and completion races.** `request["params"]?["requestId"]` assumes `params` is an object, so a fire-and-forget notification such as `{"method":"notifications/cancelled","params":[]}` throws instead of being ignored. There is also a normal completion race: `TryGetValue` can return a CTS, the call can then remove and dispose it, and `Cancel()` throws `ObjectDisposedException`. Both cases can turn a cancellation POST into an unhandled 500, while MCP requires malformed/unknown/already-completed cancellations and timing races to be handled gracefully. Require object-shaped params and coordinate the registry entry lifetime (or safely treat disposal as an already-completed no-op); add notification-level tests for malformed params and cancellation racing completion.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental verdict: changes still required.** `87f5bb52` fixes sessionless upstream-session adoption and implements cancellation routing for well-formed matching notifications, but snapshot removal remains eager rather than delivery-gated; I replied on the existing lifecycle thread. I also found a distinct cancellation robustness defect: malformed notifications and normal completion/disposal races can throw and return 500 instead of being ignored gracefully.

Revobot's active finding about validating malformed `tools/call` params before establishing local ownership is valid and was not duplicated. The idle-timeout fix remains resolved.

Context reviewed: #247's routing, transparency, cancellation, and concurrent-completion criteria; saved background from prior iterations; the approved design/plan; official MCP cancellation and Streamable HTTP behavior; the incremental and full diffs; all live review threads; and focused MCP composition tests. Review agent used: `pr-248-final-review` (`code-review`, GPT-5.6 Sol). GitHub does not allow this account to request changes on its own PR, so this is submitted as a blocking comment review.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,432 @@
...
+        var mediaType = upstream.Content.Headers.ContentType?.MediaType;
+        var isJson = string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase);
+        var isSse = string.Equals(mediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
+        if (!upstream.IsSuccessStatusCode || (!isJson && !isSse))
+        {
+            Invalidate();
+            return null;
+        }
+
+        var original = await ProxyHttp.ReadCappedBytesAsync(
```
gautam-achieveai, line 252: [Gautam's review bot] **[Medium] Map buffered `tools/list` transport failures to a JSON-RPC 502.** This read occurs after upstream headers and outside the `HttpRequestException` catch around `SendAsync`. If `ReadAsStreamAsync` or `ReadAsync` throws `IOException`/`HttpRequestException` while buffering a JSON or SSE catalog, `Program.cs` catches only cancellation around `ComposeListAsync`, so the exception escapes as a bare 500. The transparent relay maps the same pre-body failure to a protocol-shaped 502 in `CopyBodyAsync`. Catch only these transport faults around composition and return the MCP gateway error; add a `ThrowingStream` `tools/list` regression.

gautam-achieveai (commented):
[Gautam's review bot] **Deep incremental review: changes are still required.** The incremental diff is empty because this head was already reviewed, so I rechecked the full diff and live thread state. `69f46119` correctly fixes transparent forwarding for malformed calls whose local ownership cannot be established.

The previously reported schema serialization defect, eager snapshot removal, and cancellation-notification race remain unresolved. I found one additional distinct transport-error path inline: a dropped `tools/list` body during composition can escape as a bare 500 instead of the proxy's JSON-RPC-shaped 502. Existing resolved threads were left resolved, and the active cancellation thread remains active.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in f8061b67. Snapshot set/remove is now represented as a deferred action and applied only after the corresponding downstream body write succeeds. All eager invalidation calls were removed; failed/cancelled delivery retains the prior catalog ownership.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in f8061b67. Cancellation notifications now require object-shaped params and malformed notifications pass through safely. The lookup/cancel disposal race treats ObjectDisposedException as an already-completed no-op. Added malformed-notification coverage.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in f8061b67. IOException/HttpRequestException while buffering JSON or SSE tools/list bodies now map to the Proxy's JSON-RPC-shaped 502 instead of escaping as 500. Added a ThrowingStream regression. Verification: build 0 warnings/errors, Proxy tests 365/365, focused Misc tests 106/106.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 6: delivery gating is fixed, but failed refreshes still discard the last delivered catalog.** After a successful list establishes local ownership, an HTTP-200 JSON-RPC error or malformed `tools/list` body returns `McpSnapshotAction.Remove`. `Program.cs` writes that failure body and then publishes the removal, even though the client received no replacement catalog and can still rely on its previously delivered tool list. This is the prior thread's explicit "paths that do not deliver a replacement catalog must retain the prior snapshot" case. Use `None` for failed refreshes, reserve `Remove` for a successfully delivered catalog that no longer grants local ownership, and add successful-list → error-list → local-call coverage.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 6: the transport fault is now mapped to 502, but the generated body is not a valid JSON-RPC response for this request.** `ComposeListAsync` is running for an already parsed `tools/list`, so its request ID is known, yet `WriteMcpErrorAsync` emits only `jsonrpc` and `error`. JSON-RPC 2.0 requires a response to carry the same `id`; MCP's no-ID HTTP-error allowance is for notification/response input the server could not accept. Pass through `rpcRequest["id"]` for this body-read failure (and the analogous timeout/cap paths) and assert that an input ID such as `42` is preserved.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,426 @@
...
+
+        var mediaType = upstream.Content.Headers.ContentType?.MediaType;
+        var isJson = string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase);
+        var isSse = string.Equals(mediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase);
+        if (!upstream.IsSuccessStatusCode || (!isJson && !isSse))
+        {
+            return null;
+        }
+
+        var original = await ProxyHttp.ReadCappedBytesAsync(
```
gautam-achieveai, line 252: [Gautam's review bot] **[Medium] Do not buffer a POST SSE exchange to EOF before forwarding any event.** `ReadCappedBytesAsync` waits for the entire `text/event-stream` body, and only then does `TryReadSingleSseMessage` decide whether it can compose the catalog. Streamable HTTP permits the server to send JSON-RPC requests or notifications before the final response; if one of those requests waits for the client's reply, the client never sees it and this proxy eventually returns 504 despite a valid MCP exchange. The observed one-event GitHub response is not a protocol guarantee. Parse/forward SSE incrementally and rewrite only the matching final list response, or relay SSE transparently when incremental composition is not supported.

samples/CopilotAnthropicProxy.Sample/Program.cs
```diff
@@ -2611,17 +2756,25 @@ await WriteMcpErrorAsync(
...
-                linked,
-                logger,
-                WriteMcpErrorAsync,
-                isSse,
-                keepAliveInterval
-            );
+            if (composedList is not null)
+            {
+                await ctx.Response.Body.WriteAsync(composedList.Body, ctx.RequestAborted);
+                composition.PublishSnapshot(composedList);
```
gautam-achieveai, line 2771: [Gautam's review bot] **[Medium] Prevent an in-flight list from resurrecting a terminated session snapshot.** A DELETE or upstream 404 removes the snapshot at lines 2677-2686, but an earlier concurrent `tools/list` can finish buffering afterward and publish its candidate here. The proxy then continues handling local tools for a session the upstream has terminated, bypassing the required 404/new-session flow. Track a per-session generation/tombstone and publish only if no terminal mutation occurred after composition began; add a delayed-list versus DELETE/404 regression.

gautam-achieveai:
[Gautam's review bot] **Incremental deep review of `f8061b67`: REQUEST CHANGES.**

The malformed-cancellation and disposed-CTS race is addressed. Two claimed fixes remain partial: failed `tools/list` refreshes still remove the last delivered snapshot, and proxy-generated body-read failures omit the known JSON-RPC request ID. I also posted two new medium findings: POST SSE exchanges are buffered to EOF before any event is forwarded, and an in-flight list can republish local routing after DELETE/404 terminates the session.

The previously reported injected-schema serialization blocker also remains active and was not duplicated here. Review used the saved !248 history, full and incremental diffs, official JSON-RPC/MCP transport specifications, a feature-context research pass, and an independent code-review pass. The existing Proxy suite passes 365/365; the reported scenarios are not covered.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 1311b6e3. Failed JSON-RPC/malformed refreshes now use no snapshot mutation and preserve the last delivered catalog; only a successfully delivered replacement catalog changes ownership.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 1311b6e3. Timeout, body-cap, and buffered transport-failure errors now include a deep clone of the known tools/list JSON-RPC request ID. The shared generic error writer still omits IDs only when no request ID is known.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Addressed in 1311b6e3 with a bounded-by-design split: only finite Content-Length SSE catalogs at or below the cap enter composition. Chunked/multi-event-capable POST SSE remains on the existing incremental relay, so server requests/notifications cannot deadlock behind buffer-to-EOF. This retains live GitHub single-event composition without claiming general SSE aggregation.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 1311b6e3. List composition captures a per-session generation before the upstream request; DELETE/404 increments the generation. Snapshot publication uses compare-by-generation, so an earlier in-flight list cannot resurrect terminated session state.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 7: failed refreshes are fixed, but successful pass-through replacements still retain stale ownership.** `ComposeListAsync` returns `null` for successful chunked/multi-event SSE and `McpSnapshotAction.None` for paginated or otherwise uncomposed successful catalogs. `Program.cs` delivers those responses without a deferred removal. After an earlier list established local ownership, the latest successfully delivered catalog can therefore omit that local definition while later calls still dispatch to Jina. `None` is correct for a failed refresh; a successfully delivered uncomposed replacement needs a delivery-gated `Remove`, including only after a streaming copy completes. Please add local-list → paginated/chunked-list → call coverage.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: [Gautam's review bot] **Iteration 7: the generation guard still races publication.** `SetIfGeneration` reads `_generations` and writes `_snapshots` as separate operations on separate dictionaries. A list can observe generation 0, then DELETE/404 increments to 1 and removes the snapshot, and the list can still write afterward—resurrecting the terminated session exactly as reported. Make the generation comparison and snapshot update one atomic per-session operation (lock or CAS over a single state), and add the delayed-list versus DELETE/404 regression.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 7: buffered failures now preserve the ID, but known-ID coverage is still incomplete.** `rpcRequest` is parsed before `SendAsync`, yet the header idle-timeout and connection-failure catches call `WriteMcpErrorAsync` without `rpcRequest["id"]`. The incremental relay callback also discards that known ID, so a pre-first-byte SSE body failure that can still be converted to a JSON-RPC error is emitted without correlation. Pass the parsed ID through every proxy-generated error path where it is known, and cover both header and relay failures.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental deep review of `1311b6e3`: REQUEST CHANGES.**

Three medium correctness gaps remain. Successful pass-through `tools/list` replacements can retain stale local ownership; the generation check and snapshot write are not atomic, so DELETE/404 can still be followed by stale resurrection; and several proxy-generated failures still omit an already parsed JSON-RPC request ID.

Of the four new replies, the failed-refresh preservation is only partially complete, buffered error-ID preservation is only partially complete, the unknown-length SSE deadlock is addressed, and the session-generation fix remains race-prone. I replied only to the existing discussions, so no duplicate threads were opened. Review used the saved !248 history, full and incremental diffs, a feature-context research pass, and an independent code-review pass. The existing Proxy tests pass 365/365 and focused MCP tests pass 34/34, but the reported interleavings and refresh sequence are not covered.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in cc6fbcdd. Failed refreshes remain no-op, while a successfully delivered uncomposed replacement now carries a deferred Remove action. Chunked SSE removal is applied only after CopyBodyAsync completes successfully.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in cc6fbcdd. Parsed request IDs now flow through header timeout, token acquisition, connection failure, buffered timeout/cap/transport failure, and the incremental relay error callback. Generic errors still omit ID only when no request was parsed.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in cc6fbcdd. Generation and snapshot now live in one per-session state guarded by a single lock. SetIfGeneration compares and writes atomically; DELETE/404 increments generation and clears under the same lock, preventing resurrection.

gautam-achieveai:
[gautam-achieveai's bot] Addressed the post-1311b6e3 review in cc6fbcdd: union-aware schema serialization; failed-refresh retention versus successful pass-through removal; atomic per-session generation/snapshot state; complete known-ID propagation across proxy-generated failures; bounded finite SSE composition with chunked SSE incremental relay. Verification: build 0 warnings/errors, Proxy tests 365/365, focused Misc tests 106/106. The original requests for global snapshot eviction and lower-allocation parsing remain documented, accepted design constraints for this loopback single-developer proxy rather than blocking defects.

gautam-achieveai:
[Revobot]

## Review: REQUEST CHANGES

### [Medium] Preserve exception details when logging MCP boundary failures

**File:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2656-2657`  
**File:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2668-2669`

The new MCP forwarding catches log only `ex.Message`:

```csharp
logger.LogError("Copilot token acquisition failed: {Reason}", ex.Message);
logger.LogError("Upstream MCP connection failed: {Reason}", ex.Message);
```

This discards the exception object, including its stack trace and inner exception details. These are precisely the boundary failures that operators will need to diagnose, and the resulting generic JSON-RPC responses do not preserve that context.

Log the exception through the `ILogger` overload while keeping the client-facing message generic:

```csharp
logger.LogError(ex, "Copilot token acquisition failed while forwarding an MCP request.");
logger.LogError(ex, "Upstream MCP connection failed while forwarding an MCP request.");
```

The performance-focused review found no additional high-confidence findings. The focused test command could not be executed in this environment because the .NET SDK was unavailable (`dotnet: not found`).

🔎 Full review conversation: http://localhost:5051/?threadId=thread-e0cd1ccab9d44ff3b4bf0e477e28bb81&focus=1

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 8: the stale-ownership fix is still incomplete.** The new deferred `Remove` applies only to 2xx responses that are neither JSON nor bounded SSE. Valid multi-event SSE and paginated results still return `None`, while a final continuation request with `params.cursor` and no `nextCursor` is composed as if that page were the complete catalog. The raw path also publishes `Remove` whenever `CopyBodyAsync` returns, but that helper returns normally after client aborts, idle/read failures, and downstream write cancellation, so partial or failed delivery can clear ownership. Please distinguish valid-but-uncomposable replacements from failed refreshes, keep every cursor-bearing page raw, and apply a generation-aware clear only after the relay reports successful completion.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: [Gautam's review bot] **Iteration 8: compare/write is atomic now, but ordinary catalog removal breaks completion order.** `PublishSnapshot(Remove)` calls the same `Remove` used for DELETE/404, which increments the terminal generation. If pass-through list B clears first and an earlier-started composed list A finishes later, A is the latest delivered catalog but its `SetIfGeneration(0, ...)` is rejected because B advanced the generation to 1. Reserve generation increments for session termination; use a generation-checked clear that does not advance the DELETE/404 epoch for ordinary catalog replacement.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [Gautam's review bot] **Iteration 8: known request IDs are still configuration-gated.** `TryParseSingleRequest` remains inside `composition.IsEnabled`, so with no valid Jina configuration `rpcRequest` stays null even for a valid JSON-RPC request with an ID. The newly updated header-timeout, token, connection, and relay error paths therefore still emit `id: null` in transparent/no-Jina mode. Parse request metadata independently of local composition enablement, and gate only local injection/dispatch.

gautam-achieveai (commented):
[Gautam's review bot] ## Review: REQUEST CHANGES

Iteration 8 fixes the original atomic compare/write race and propagates IDs through the edited call sites when composition is enabled, but three blocking lifecycle gaps remain:

1. Successful uncomposed catalogs still preserve stale ownership in valid SSE/pagination paths, final cursor pages can be composed as standalone catalogs, and failed raw relays can clear ownership.
2. Ordinary catalog clearing increments the DELETE/404 generation, so a later-delivered concurrent catalog can be incorrectly rejected.
3. Request parsing is still gated by Jina enablement, so known IDs remain missing from proxy errors in transparent/no-Jina mode.

I replied to the existing discussions rather than opening duplicates. The separate Revobot exception-logging finding also remains active.

gautam-achieveai:
[gautam-achieveai's bot] Fixed the post-cc6fbcdd logging finding in 960fb3c5. MCP token-acquisition and upstream-connection catches now use ILogger exception overloads, preserving stack and inner-exception context while client-facing JSON-RPC errors remain generic. Verification: build 0 warnings/errors, Proxy tests 365/365, focused Misc tests 106/106.

gautam-achieveai:
[Revobot]

## Review: REQUEST CHANGES

### [Medium] Evict removed MCP session state

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolSnapshotStore.cs:10,67-75`

`Remove` clears the snapshot but leaves the `(endpoint, sessionId)` entry in the singleton `_states` dictionary permanently. Every distinct session ID observed by the proxy is therefore retained after DELETE or upstream 404. A long-running proxy handling session churn can grow memory monotonically.

Add coordinated eviction for inactive, snapshot-free states while preserving generation checks for in-flight catalog requests.

### [Medium] Bound concurrent full-body catalog composition

**File:** `samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs:252-323`

A `tools/list` composition can retain several large representations simultaneously: the buffered body, the `MemoryStream` backing data, the parsed `JsonNode` tree, serialized JSON, and the composed byte array. SSE composition additionally creates full-body strings and extracted byte arrays. Multiple concurrent responses near the configured body cap can therefore create substantial LOH pressure and transient allocation spikes.

Limit concurrent compositions and avoid retaining multiple full-body representations where practical.

### [Medium] Apply the idle timeout to composed downstream writes

**File:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2768-2772`

The composed response is written using `ctx.RequestAborted` only:

```csharp
await ctx.Response.Body.WriteAsync(composedList.Body, ctx.RequestAborted);
```

Unlike the normal relay path, this does not use the linked token containing the configured idle timeout. If a client remains connected but stops consuming the response, the write can remain blocked beyond the idle deadline while retaining the composed body and upstream response. Use the linked cancellation token and handle cancellation consistently with `CopyBodyAsync`.

### [Medium] Handle downstream transport failures during translated SSE writes

**File:** `samples/CopilotAnthropicProxy.Sample/Program.cs:2051-2068`

`DispatchAsync` catches cancellation around `Response.WriteAsync` and `FlushAsync`, but not `IOException` or `ObjectDisposedException`. A downstream reset or response disposal can therefore escape after the translated response has started, producing an unhandled request failure instead of cleanly terminating the relay.

Catch the expected downstream transport exceptions, log them with the exception object at debug level, and terminate the relay without attempting to append an error envelope after the response has started.

### Verification and review coverage

The focused test command could not run because the environment has no .NET SDK (`dotnet: not found`).

The exception-handling and performance reviews completed. The test-coverage review and review-grader sub-agents failed before producing results; no findings were inferred from those failures.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-fcbb351125e442ae9bcac6e72c3a2225&focus=1

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
```diff
@@ -0,0 +1,448 @@
...
+                // The local call completed between lookup and cancellation; treat it as already finished.
+            }
+        }
+
+        return false;
+    }
+
+    public async Task<bool> TryHandleCallAsync(HttpContext context, JsonObject request)
+    {
+        if (!IsEnabled || !IsToolsCall(request) || !request.ContainsKey("id") || request["id"] is null)
```
gautam-achieveai, line 99: [Gautam's review bot] **[Medium] Preserve MCP protocol-version enforcement for intercepted calls.** This path can return a local Jina result before the request reaches GitHub, but it never validates Mcp-Protocol-Version. Once a local snapshot exists, a call with an explicitly invalid or unsupported version therefore succeeds locally, whereas the MCP Streamable HTTP specification requires HTTP 400. The transparent path lets GitHub enforce that boundary; interception changes it. Record the negotiated/list protocol version in snapshot provenance and forward or reject a local call when its explicit version is incompatible, with regression coverage.

gautam-achieveai (commented):
[Gautam's review bot] **Changes still required.** The 960fb3c5 logging fix is correct and preserves exception context. The three iteration-8 lifecycle/request-ID findings remain unchanged, and this review found one additional protocol-boundary issue inline. GitHub does not allow the PR author account to submit REQUEST_CHANGES, so this is posted as a blocking comment review.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in b24d8f63. Cursor-bearing requests remain raw; successful uncomposed replacements carry a generation-checked clear applied only after successful relay completion. Failed refreshes remain no-op. Ordinary catalog clear no longer advances the DELETE/404 termination generation.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in b24d8f63. Request metadata is now parsed independently of Jina enablement, so all proxy-generated error paths preserve known JSON-RPC IDs even in byte-transparent/no-key mode.

samples/CopilotAnthropicProxy.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in b24d8f63. Session state now has atomic generation-aware Set and Clear operations under one per-session lock. Ordinary catalog replacement clears without advancing the termination epoch; DELETE/404 alone increments it.

samples/CopilotAnthropicProxy.Sample/Mcp/McpToolComposition.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in b24d8f63. Snapshots record the list request's Mcp-Protocol-Version. An intercepted call with an explicit incompatible version now falls through to GitHub for protocol enforcement; matching or omitted versions retain local routing. Build 0 warnings/errors, Proxy tests 365/365, Misc tests 106/106.

PR #249, opened 2026-08-03
feat: add browser-hosted client tools to LmStreaming.Sample

### LINKED ISSUE #246: Add browser-hosted client tools to LmStreaming.Sample
## Summary
Add a generic attached-client tool bridge to `LmStreaming.Sample` so `MultiTurnAgentLoop` can invoke typed capabilities implemented by its browser client over the existing bidirectional WebSocket.

Support two execution modes and prove each with a reference tool:

1. **Blocking request/response — `AskUserQuestion`:** publish a structured question invocation, persist it as a deferred tool call, and park agent progress until a browser returns a correlated answer, Skip result, or the conversation is cancelled. Calls remain discoverable while no browser is connected and survive refresh, reconnect, and process restart.
2. **Non-blocking notification — `NotifyClient`:** persist and publish the existing `NotifyMessage`, return the tool result immediately, and do not start another agent run or wait for client delivery.

`AskUserQuestion` follows the proven interaction semantics from `B:\sources\Hitl_MCP`: one question or a batch of one to four; stable option values with labels and optional descriptions; single- or multi-selection; optional custom Other input; optional preview content for single-select choices; and explicit Skip. Batch questions render as a stepper. The result uses a consistent array of selected values per question and a separately named Other-text field.

The first implementation targets `MultiTurnAgentLoop` modes only. Claude, Codex, and Copilot CLI-backed loops do not advertise these tools because their tool execution does not support externally resolved deferred calls. Office/Word/Excel adapters are follow-up consumers of the generic bridge, not part of this item.

## Value
Users of `LmStreaming.Sample` get first-class human collaboration instead of relying on the model to ask questions in generic assistant text. The model receives typed, correlated answers, and users receive accessible controls that remain correct across reconnects.

The two reference tools establish both attached-client execution modes needed by the planned client-extensibility work in #241: operations that wait for the user and notifications that do not block agent progress.

### Generic client-tool channel
- [ ] The existing browser WebSocket protocol supports discriminated server-to-client client-tool invocation frames and client-to-server result frames, with thread ID, tool-call ID, tool name, execution mode (`blocking` or `notification`), and structured payload.
- [ ] Inbound client-tool result frames are distinguished from ordinary chat frames before deserialization; malformed or unsupported frames receive a typed client-visible error and are not silently treated as chat text.
- [ ] Client-tool results are accepted only for the frame's conversation/thread and matching pending tool-call ID; a result cannot resolve a call from another thread.
- [ ] The browser reconnects or retains its WebSocket while pending client work exists even when the agent run is parked and `IsInProgress` is false.
- [ ] A newly connected or refreshed browser reconstructs pending blocking client tools from persisted deferred-call history and renders them without a new pending-question store.
- [ ] Existing WebSocket bounds, single-writer send gating, typed error-frame conventions, cancellation, and EUII-safe logging remain in force for client-tool frames.

### Blocking `AskUserQuestion`
- [ ] `AskUserQuestion` is registered in every eligible `MultiTurnAgentLoop` chat mode and is not registered or advertised in Claude, Codex, or Copilot CLI-backed modes.
- [ ] The tool contract requires human-facing context and accepts either one question or a batch of one to four questions. Each question has text/header, one or more options, `allowMultiple`, and `allowOther`; each option has a label, stable value, optional description, and optional preview.
- [ ] Invalid input—including zero questions, more than four questions, no options, duplicate option values within a question, or an empty required context—is rejected synchronously as a structured tool error before any deferred placeholder is persisted.
- [ ] Missing option values received at the client boundary fall back to their labels for compatibility without dropping the full question; server-generated contracts still emit explicit stable values.
- [ ] A valid call returns `ToolHandlerResult.Deferred()`, is persisted through the existing deferred tool-call path, and remains pending across browser disconnect, process restart, and history restoration.
- [ ] The client identifies the persisted deferred placeholder through `IsDeferred` and renders an interactive question pill while unresolved; it renders the canonical answer as read-only after live resolution or history hydration.
- [ ] Single-select questions render mutually exclusive choices. Multi-select questions render independent choices. Both optionally accept custom Other text. A response must contain at least one selected option or non-empty Other text.
- [ ] Single-select options may show a markdown preview panel; preview content is rendered safely, and multi-select questions do not show the preview panel.
- [ ] A batch renders one to four questions in a stepper, preserves selections while navigating, and allows submission when at least one question has an answer; unanswered questions remain explicitly represented in the result.
- [ ] Skip is an explicit action and resolves the call with a distinct structured skipped result; it is not represented as an empty answer.
- [ ] Answer results use a consistent shape: one answer entry per question with question index, selected values as an array, optional Other text, and response type. Request context and response Other text use different field names.
- [ ] Submitting an answer calls the existing deferred resolution API through the WebSocket bridge. `Resolved` resumes the agent once; an identical retry is treated as `Duplicate`; a different second answer is `Conflict`; missing or cancelled calls return stable typed outcomes.
- [ ] When multiple browser tabs race, the losing tab receives a clear already-answered state, refreshes the canonical stored result, and renders the pill read-only.
- [ ] The question has no automatic timeout in this increment. It remains pending until answered, skipped, or explicitly cancelled; the UI does not show a cosmetic expiry that leaves the server call unresolved.
- [ ] Cancelling the conversation/run closes pending question UI and prevents a late answer from resuming the cancelled work.

### Sub-agent questions
- [ ] `AskUserQuestion` remains inheritable by sub-agents.
- [ ] A pending question from any descendant is surfaced in the parent conversation with the originating agent identity and a control that navigates to the originating transcript; it cannot remain discoverable only inside an unopened sub-agent tab.
- [ ] The question remains interactive in its originating transcript and resolves the originating tool call, preserving thread/run/generation/tool-call correlation.

### Parked conversation behavior
- [ ] Ordinary chat submitted while a blocking client tool is pending follows the existing parked-input queue behavior and cannot be interpreted as a tool answer.
- [ ] The composer clearly indicates that ordinary chat will be queued until the blocking client tool resolves.
- [ ] Mode and provider switching are blocked while any blocking client-tool call is pending, with a stable explanatory conflict response; the pending call cannot be silently discarded or orphaned.

### Non-blocking `NotifyClient`
- [ ] `NotifyClient` is registered in every eligible `MultiTurnAgentLoop` chat mode and uses a stable `NotifyKinds` value.
- [ ] A valid call immediately returns a resolved tool result and persists/publishes an existing `NotifyMessage` for the client, including source tool-call correlation and structured display fields.
- [ ] The tool-handler path exposes and reuses `MultiTurnAgentLoop`'s existing history persistence and subscriber publication primitives; it does not introduce a parallel client-notification channel.
- [ ] Notification delivery is best-effort live and durable through normal conversation hydration: a disconnected browser sees it after reconnect/reload without blocking the agent turn.
- [ ] Emitting `NotifyClient` from an active run does not enqueue a new user input, start a redundant run, or cause the model to respond to its own notification.
- [ ] Existing `NotificationPill` rendering, provider mapping, ordering, merge-key disambiguation, and legacy context-notification compatibility continue to work.

### Verification and compatibility
- [ ] Exact-scenario tests prove an offline `AskUserQuestion` persists, appears after a browser connects, resolves over WebSocket, and causes exactly one continuation after restart.
- [ ] Browser E2E tests cover single-select, multi-select with Other text, a multi-question stepper, Skip, refresh while pending, two-tab answer races, sub-agent global surfacing/navigation, queued ordinary chat, and mode/provider switch blocking.
- [ ] Contract tests cover malformed frames, invalid question schemas, thread/tool-call mismatches, `Resolved`/`Duplicate`/`Conflict`/`NotFound`/`Cancelled` outcomes, and ensure free-text chat cannot resolve a question.
- [ ] A notification E2E proves the tool completes immediately, the pill renders live and after hydration, and no extra agent run is created.
- [ ] Existing chat, deferred-tool, notification, sub-agent, reconnect, and tool-pill tests remain green without changing persisted conversation data.
- [ ] Documentation states that this is an `LmStreaming.Sample` single-identity reference implementation over `MultiTurnAgentLoop`, not the multi-tenant authorization/extension contract planned by #235/#238/#241 and not the durable escalation/takeover state machine planned by #239/#240.

## Out of Scope
- Word, Excel, Office.js, or general canvas adapters.
- Claude/Codex/Copilot CLI-loop client-tool parity.
- Multi-tenant client authorization, client capability leases, callback/webhook registration, or the #238 platform outbox.
- Human-assist escalation, claims, takeover, rota management, or automatic timeout/sweeper behavior from #239/#240.
- Tool approval policy from #227; client-tool execution may compose with approval but does not replace its fail-closed allow/deny gate.
- A second notification store or delivery protocol alongside `NotifyMessage` and conversation history.

gautam-achieveai:
## Summary

- add `AskUserQuestion` as a durable deferred browser-hosted tool with structured single/batch question UI
- add `NotifyClient` as an immediate persisted notification without scheduling another model turn
- carry correlated results over the existing root and sub-agent WebSockets, including reconnect, duplicate/conflict, cancellation, and lifecycle handling
- surface descendant questions globally while keeping answers scoped to the originating agent transcript
- preserve collaboration message rendering and exclude browser tools from CLI/read-only workflow-controller loops

## Key decisions

- deferred conversation history remains the source of truth; no separate question store or outbox
- outbound questions remain ordinary tool-call/result messages; only `client_tool_result` is a new inbound frame
- thread and sub-agent identity come from the owning WebSocket route, never from client payload
- `NotifyClient` uses `NotifyMessage`, history persistence, and subscriber publication without `SendAsync`

## Verification

- `dotnet build LmDotnetTools.sln`: 0 warnings, 0 errors
- Vue/Vitest: 66 files, 834 tests passed
- Vue type-check: passed
- `LmMultiTurn.Tests`: 1,074 passed
- Browser E2E: 56 passed, 5 expected skips
- reusable whole-flow Playwright script: 13/13 steps passed
- independent post-rebase conflict audit completed; confirmed findings fixed and reverified

Fixes #246


gautam-achieveai:
[Revobot]

# Review: `achieveai/LmDotnetTools#249`

**Reviewed head:** `74452d95e680fe3bd49edc92bd64f0085265f093`  
**Verdict:** **REQUEST CHANGES**

## Findings

### [HIGH] [BLOCKER] Preserve the previous public constructor for binary compatibility

**File:** `src/LmMultiTurn/MultiTurnAgentLoop.cs:197-220`

`MultiTurnAgentLoop` is part of the packable `AchieveAi.LmDotnetTools.LmMultiTurn` NuGet package, but the PR changes its existing public constructor by appending:

```csharp
bool includeAskUserQuestionTool = true,
bool includeNotifyClientTool = true
```

Optional parameters only affect source compilation; they do not preserve the previous CLR constructor metadata. An already-compiled consumer of the prior package will still attempt to invoke the old constructor and can fail with `MissingMethodException` after upgrading.

Please retain the prior constructor overload and route it through a shared implementation. Add a separate overload or options object for the new tool-registration controls. If this is intentionally a major-version API break, it should be explicitly documented and versioned accordingly.

## Review coverage

- The workflow restriction behavior is covered by `tests/LmWorkflow.Tests/WorkflowControllerToolRestrictionTests.cs:25-51`; that test would fail before this change because the controller would expose the two browser tools.
- Exception-handling review found no additional concrete findings.
- Temporary-code review found no debugging artifacts, hardcoded credentials, or accidental test helpers.
- Wire/persistence schema review found no additional schema compatibility issues beyond the public constructor break.
- Performance review identified WebSocket teardown and notification-persistence concerns in broader code, but the WebSocket registry issues were not introduced by this PR. The notification-persistence concern was not retained as a separate finding because its impact depends on the existing out-of-band notification/run-ID contract and was not sufficiently established against the requested change.
- The targeted .NET tests could not be executed because `dotnet` is unavailable in the review environment.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-b9a803257301404ca0b823f700a246ba&focus=1

src/LmMultiTurn/SubAgents/SubAgentManager.cs
```diff
@@ -2996,17 +3029,51 @@ private async Task HandleRunCompletionAsync(
...
 
             var resultText =
                 $"<sub-agent name=\"{state.TemplateName}\" " +
                 $"id=\"{state.AgentId}\">\n" +
-                $"[Completed] Task: {state.Task}\n" +
+                $"{statusTag} Task: {state.Task}\n" +
                 $"Result: {result}\n" +
                 $"</sub-agent>";
 
             _ = state.TryCompleteWithResult(result);
```
gautam-achieveai, line 3115: [Gautam's review bot] **[High] Keep a blocking child pending while `AskUserQuestion` is deferred.** `Agent`/`SendMessage` default to `run_in_background: false`, and those paths await `state.Completion`; this line resolves that latch with the placeholder, so the parent continues as though the delegation finished. After the human answers, the child's real completion cannot replace the already-set latch, and foreground children have `NotifyParentOnCompletion == false`, so the final answer is lost. The child has also already been marked terminal and its owned provider disposed earlier in this method. Detect the parked question before terminal disposal/gate release, emit the descendant notification without settling `Completion`, and let the resumed run perform the sole final completion. Please add a foreground `AskUserQuestion -> answer -> final text` regression test.

gautam-achieveai (commented):
[Gautam's review bot] Found one new high-severity sub-agent lifecycle defect. The existing public-constructor binary-compatibility blocker reported on this same head also remains unresolved.

src/LmMultiTurn/SubAgents/SubAgentManager.cs
gautam-achieveai: [Gautam's review bot] Resolved. `HandleRunCompletionAsync` now takes a precomputed `awaitingQuestion` flag (`HasPendingAskUserQuestionAsync`, evaluated by the monitor loop before the gate-release decision, so the two never disagree). When `awaitingQuestion` is true, the method returns early: it never calls `TryCompleteWithResult`, never runs the terminal-disposal path (`BeginTerminalDisposalAsync`/`PersistTerminalStateAsync`/`SyncCollaborationStatus`/`DisposeOwnedProviderAgentAsync`), and never releases the concurrency gate — it only fires the `#246` descendant-question notification and, if `NotifyParentOnCompletion`, relays an `[AwaitingAnswer]`-tagged message. `state.Completion` is left unresolved, so a foreground `Agent`/`SendMessage` caller keeps blocking on `state.Completion.Task` until the real answer's run reaches the genuinely-terminal branch and performs the sole `TryCompleteWithResult` with the real final text.

Added `SpawnAsync_Foreground_WhenParkedOnAskUserQuestion_BlocksThenReturnsRealAnswerAfterResolution` in `tests/LmMultiTurn.Tests/SubAgentManagerTests.cs`, which drives the actual foreground (`runInBackground: false`) path end-to-end: spawns, waits for the child to park on `AskUserQuestion`, asserts the foreground task is still pending and `Peek` reports `"running"` (not terminal), asserts a second spawn under `maxConcurrent: 1` queues rather than starts (proving the parked child still holds its concurrency permit), resolves the deferred call via `MultiTurnAgentLoop.TryResolveToolCallAsync`, then asserts the foreground task unblocks with the real final text and `Peek` then reports `"completed"`.

RED-proven: stashing this fix reproduced the exact predicted failure (`foregroundTask.IsCompleted` was `true` immediately, with the placeholder). Full suite green: `LmMultiTurn.Tests` 1080/1080, `SubAgentManagerTests` 45/45.

gautam-achieveai:
[Gautam's review bot] Resolved. Restored the exact prior CLR constructor signature (`MultiTurnAgentLoop(IStreamingAgent, FunctionRegistry, string, string?, GenerateReplyOptions?, int, int, int, IConversationStore?, ILogger<MultiTurnAgentLoop>?, SubAgentOptions?, MutableSubAgentTemplateSource?, ILoggerFactory?, bool, TriggerOptions?, IPricingResolver?, IUsageSink?, MultiTurnLifecycleServices?, MultiTurnLifecycleServices?, AgentCollaborationSetup?, Func<NotifyMessage, CancellationToken, ValueTask>?)`) with its original parameter order and defaults, so an already-compiled consumer of the packable `AchieveAi.LmDotnetTools.LmMultiTurn` library still binds to the same overload after upgrading. It forwards (`: this(...)`) to a new designated constructor that inserts `includeAskUserQuestionTool`/`includeNotifyClientTool` as required (non-optional) parameters right after `threadId` — required, so it can never be ambiguous with the restored legacy overload — and every other code path (including `WorkflowSession`, which already calls this with `includeAskUserQuestionTool: false, includeNotifyClientTool: false`) routes through that same designated constructor. Both overloads share one implementation body; nothing is duplicated.

Added `tests/LmMultiTurn.Tests/MultiTurnAgentLoopConstructorCompatibilityTests.cs` with reflection-based regression coverage: confirms the prior public constructor still exists with the same parameter types/order, confirms only the original three leading parameters are required, confirms its defaults are unchanged, confirms the new tool-control overload exists with required flags and is unambiguous against the legacy one, and confirms exactly two public constructors exist in total (guarding against silent future drift). Also verified `tests/LmWorkflow.Tests/WorkflowControllerToolRestrictionTests.cs` still passes, confirming the workflow controller can still disable both tools.

Full suite green: `LmMultiTurn.Tests` 1080/1080, `LmWorkflow.Tests` 377/377.

gautam-achieveai:
[Revobot]

# Re-Review Summary: PR #249

**Reviewed head:** `c6dfa50ed3ba5b95db3e7503787520ba2d86ba8e`  
**Verdict:** **REQUEST CHANGES**

## New Finding

### [MEDIUM] Error frames omit the diagnostic message

**File:** `samples/LmStreaming.Sample/WebSocket/ChatWebSocketManager.cs`  
**Location:** `SendClientToolResultErrorAsync` (around line 1250)

`SendClientToolResultErrorAsync` serializes only `$type`, `toolCallId`, and `code`, omitting the `message` field. However, the browser callback contract expects `(toolCallId, code, message)`, and `wsClient` substitutes `"Unknown error"` when the message is absent.

As a result, real server rejections—including `invalid`, `not_found`, `conflict`, `store_failed`, and `cancelled` outcomes—are displayed to users as `"Unknown error"`, hiding actionable diagnostics. The added client tests do not catch this because they inject synthetic error frames that already contain a `message`.

**Suggestion:** Pass a safe diagnostic message through each error path, including malformed-frame, unavailable-agent, and outcome-mapping paths, and serialize it in the error frame. Use a generic fallback when no detailed message should be exposed.

## Review Coverage

No sub-agent findings were available; no sub-agents were dispatched. Existing unresolved findings from prior reviews were not duplicated, including the public-constructor compatibility issue and the previously discussed sub-agent lifecycle issue.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-449398fa71ba4d61ac9885f726289335&focus=1

src/LmMultiTurn/MultiTurnAgentLoop.cs
```diff
@@ -200,7 +219,121 @@ public MultiTurnAgentLoop(
         IUsageSink? externalUsageSink = null,
         MultiTurnLifecycleServices? lifecycleServices = null,
         MultiTurnLifecycleServices? subAgentLifecycleServices = null,
-        AgentCollaborationSetup? collaboration = null)
+        AgentCollaborationSetup? collaboration = null,
+        Func<NotifyMessage, CancellationToken, ValueTask>? descendantQuestionSink = null)
```
gautam-achieveai: [Gautam's review bot] **[High] Restore the constructor signature that was actually released.** This overload still appends descendantQuestionSink, but the constructor at base 6ce59df4 ends at collaboration; the sink was introduced inside this PR. Already-compiled consumers therefore target the 20-parameter CLR signature, while this head exposes only 21- and 23-parameter signatures, so upgrade can still fail with MissingMethodException. MultiTurnAgentLoopConstructorCompatibilityTests masks this by including the PR-only Func<NotifyMessage, CancellationToken, ValueTask> in PriorConstructorParameterTypes. Keep an overload matching the base/released signature exactly (routing its sink as null), or explicitly version/document the binary break and correct the test baseline.

gautam-achieveai (commented):
[Gautam's review bot] Re-review of c6dfa50: the parked-child AskUserQuestion lifecycle fix is correct, but the constructor compatibility fix remains incomplete. The existing medium WebSocket diagnostic-message finding remains unresolved and was not duplicated.

gautam-achieveai:
[Gautam's review bot] Resolved. `SendClientToolResultErrorAsync` now serializes a `message` field alongside `$type`/`toolCallId`/`code`, populated by a new `SafeClientToolResultErrorMessage(code)` switch that maps every outcome code (`invalid`, `not_found`, `conflict`, `store_failed`, `cancelled`) to a stable, code-derived diagnostic string — never server internals (no exception text, no store paths). Any code not in that set still falls back to `"Unknown error"`.

Added coverage in `wsClient.test.ts`: one test proving the client's own `"Unknown error"` fallback still applies when a frame omits `message` entirely (reproducing the old buggy shape), and a parameterized test asserting the exact safe message is passed through for each of the five codes the server can now produce, so a future regression that drops `message` for any one code specifically would fail here.

Commit: e4e201b6.

src/LmMultiTurn/MultiTurnAgentLoop.cs
gautam-achieveai: [Gautam's review bot] Resolved. Confirmed the compat overload had genuinely drifted: it forwarded a caller-supplied `descendantQuestionSink` as an unreleased 21st parameter, and `MultiTurnAgentLoopConstructorCompatibilityTests` had its baseline (`PriorConstructorParameterTypes`) updated to include that same `Func<NotifyMessage, CancellationToken, ValueTask>` type instead of catching the break — so the test was passing for the wrong reason.

Restored the overload to the exact 20-parameter shape released on `main` (ending at `AgentCollaborationSetup? collaboration`, no `descendantQuestionSink` at all), routing `descendantQuestionSink: null` to the designated constructor internally. `SubAgentManager`'s one call site that needs a non-null sink now goes through the designated (tool-control) overload instead, since that parameter no longer exists on the back-compat one.

Corrected the compatibility test: removed the extra type from `PriorConstructorParameterTypes`, changed the defaults assertion to require the overload has NO `descendantQuestionSink` parameter at all (rather than asserting a null default on a parameter that shouldn't exist), and expanded the designated-overload assertion to also require the `descendantQuestionSink` parameter with the correct type/optionality.

RED-proven against the reverted baseline before reapplying the fix; full suite green after: `LmMultiTurn.Tests` 1080/1080. Also fixed four downstream tests that relied on `SubAgentManager.ObserveCompletionAsync` to detect a spawned child parking on `AskUserQuestion` — since the fix that keeps a parked question non-terminal, that completion is deliberately never observed while parked, so they now poll `MultiTurnAgentLoop.GetDeferredToolCallsAsync` directly.

Commit: ebee49da.

gautam-achieveai (commented):
[Gautam's review bot] **APPROVED** — deep incremental review of `c6dfa50e...ebee49da` found no remaining high-confidence defect.

The released 20-parameter `MultiTurnAgentLoop` constructor is now restored exactly, the designated overload preserves descendant-question routing and workflow tool restrictions, and `client_tool_result_error` frames now include stable code-derived diagnostic messages. The earlier foreground sub-agent `AskUserQuestion` lifecycle fix is not regressed.

Context checked against #246: durable/reconnect-safe deferred questions, exactly-once continuation, descendant socket scoping, pending-question switch guards, immediate notifications, and exclusion from non-browser loops. Both prior inline threads are resolved; no duplicate findings were posted.

Review agent: `incremental-reviewer-6` (`code-review`, Claude Opus 4.8). Focused regression coverage passed: 6 `LmMultiTurn.Tests`, 15 `LmStreaming.Sample.Tests`, and 24 client WebSocket tests.

GitHub does not permit the PR author account to submit an `APPROVE` review, so this approval is recorded as a comment review.

PR #250, opened 2026-08-03
fix(daemon): make the daemon author the per-PR notes artifacts

gautam-achieveai:
## Problem

Every PR directory in the cross-repo review store held nothing but `review.md`.

The intended contents — a PR context sheet and per-review-agent findings — were requested from the review agent by a prompt directive in `daemon-prompts.yaml`: *"You MUST persist this review as two files … written BEFORE you produce your final answer."*

The hosted agent never complied. Across five live review threads (~680 messages) `Write` and `Edit` appear **zero** times. A directive the model may decline is not a guarantee, and the daemon already holds every fact those files need.

## Change

Authorship moves into the daemon.

- `AwaitSubAgentSettlementAsync` now returns the settled **roster** instead of only its rendered inventory. The synthesis prompt still receives just `ToSafeInventory()`; the artifacts need the full nodes — above all the agent ids the inventory deliberately strips. Discarding the roster there is what left the directories bare.
- The barrier stashes a `ReviewNotesArtifactContext` per run. Both commit gates — pooled lease (`CommitPooledNotesAsync`) and host retention (`PublishToReviewBotAsync`) — consume it, so a run produces the same PR directory shape whichever path it took.
- `ReviewNotesArtifactBuilder` writes the context sheet and one findings file per review agent, pulling each reviewer's own words from the agent directory.

## Agent2Agent

Transcript access uses the directory the review host **already publishes over HTTP**, rather than reaching for `AgentCollaborationDirectory` (which lives in the host process and deliberately never hands out reach):

| Half | Route | Daemon interface |
|---|---|---|
| Roster | `GET /api/conversations/{id}/subagents?recursive=true` | `IReviewSubAgentCompletionSource` |
| Transcript | `GET /api/conversations/{id}/agents/{agentId}/transcript` | `IReviewAgentTranscriptSource` |

The key already matches: `AgentHierarchyProjection.Find` accepts exactly the `AgentId` a `ReviewSubAgentNode` carries. The daemon reads with no `viewer`, so the host authorizes it as the root agent — an ancestor of every descendant. Reasoning is stripped host-side and never crosses the wire.

The two interfaces are kept **separate on purpose**, mirroring the collaboration directory's own split: knowing an agent exists and being able to read what it said are different privileges, and the barrier needs only the first. `IReviewAgentTranscriptSource` is optional at the call site — a review host predating the transcript route costs artifact detail, never the review.

Both interfaces resolve to **one** `S2SReviewSubAgentCompletionSource` instance (registered concrete-first in `Program.cs`); two factory registrations would have silently given the barrier and the artifact builder separate objects.

## Untrusted content

Transcript bodies are model- and tool-produced text from agents the daemon does not control. They can carry ANSI escapes, unbalanced markdown fences, and text shaped like tool-call markers — this is not hypothetical, `PRs/lmdotnettools-222` in the review store received four artifacts whose **filenames** carried fake `assistant to=functions.Write` delimiters.

Transcript text therefore reaches a file only through `UntrustedTranscriptText`, which fences it, strips ANSI escapes, and neutralizes tool-call-shaped tokens. It is never concatenated into a prompt.

*(The four affected blobs have been quarantined separately in the review store — renamed into `_quarantine/` with a README, nothing deleted. They were inert to the review loop: `IsPriorNotesFile` requires a trailing `.md` and the corrupt names end in `{`.)*

## Failure behavior

Nothing in artifact building may fail a commit — a review that produces only `review.md` is worse than one with thin artifacts, so every failure path logs and returns what it has.

The absent-context path logs at **Warning** deliberately: that outcome looks identical to the old silent breakage, and it should be loud on the first occurrence rather than discovered a week later.

## Testing

`dotnet test tests/CodeReviewDaemon.Sample.Tests` — **747 passed, 0 failed**.

Two `DaemonAgentFactoryTests` cases were updated: they asserted the prompt names `PR_Context_NN.md` / `PR_Findings_NN.md`, which is now the inverse of the contract. They assert absence, and the first was renamed to drop `_and_write_convention`.

---

# Second commit: knowledge extraction never wrote a single entry

Same theme — *the daemon must actually write things* — and the same failure shape: a prompt directive the hosted model declined, failing silently.

## Problem

Across all of August 2026, on **both** daemons, knowledge extraction produced **zero** entries. Eleven runs logged `emitted no usable SCOPE/TITLE markers; nothing written`.

On the S2S path the extraction turn runs as a hosted **`workspace-agent`** conversation, and the extraction profile's system prompt (1,389 chars) is delivered only as a **system-prompt appendix** to the host's much larger mode prompt. The mode prompt mandates sandbox tool use for every operation, task memory under `Memory/tasks/`, and an action summary when done — and it won. The model answered like a coding agent: prose, or a review verdict.

That reply is non-empty and does not start with `NO_KNOWLEDGE`, so the decline gate passed it. It carries no markers, so `ResolveTargetRelPathAsync` returned null and the entry was dropped with only a warning.

Evidence, timestamp- and RunId-correlated: 11 logged failures across both daemons' JSONL; four hosted conversations titled `Review PR #<n> — Knowledge Extraction Agent` on the review host, PR #249's thread carrying `runId d48d700477504d739d8512d14d888f74` matching the logged failure at `2026-08-03T15:27:11.172Z`; all four assistant replies conversational with no markers; the provision log showing `SystemPromptChars = 1389`, `ModeId = workspace-agent`.

**The parser was never at fault.** The "a blank line breaks the contiguous-marker run" theory is refuted — July replies parsed correctly. `ParseEntry` and `TryParseMarker` are exonerated.

## Change

- **`daemon-prompts.yaml`** gains `knowledge-extraction: v1.1`. v1.0 is left fully intact; `PromptReader.GetPrompt` resolves `"latest"` to the highest `Version`-parseable key, so v1.1 is selected with **no C# change**. It opens by overriding the mode, states the no-tools rule first, and pins the FIRST line of the reply.
- **`KnowledgeAgent`** restates the output contract at the **end of the user turn**, where the mode prompt cannot outrank it — the appendix loses, the last line of the message does not.
- On a reply with no usable markers, **exactly one** corrective same-thread turn. Same thread, so the notes the model already read stay in context and only the output shape has to change. Exactly one on purpose: a model locked into the wrong mode will not conform on retry N, and a loop would burn the review budget to prove it.

`NO_KNOWLEDGE` or an empty reply on **either** turn remains a quiet gate — "not every PR contributes" is a correct answer, not a failure.

## Failure behavior

A still-non-conforming reply now logs a bounded, whitespace-collapsed **300-char** prefix. The notes it derives from are attacker-influenceable PR content, so the preview is capped rather than dumped; but this failure stops being silent, which is what let it run for a month.

## Testing

`dotnet test tests/CodeReviewDaemon.Sample.Tests` — **758 passed, 0 failed** (754 before; the four additions are the new facts).

Four RED-then-GREEN facts in `KnowledgeAgentTests`:
- the contract is restated in the user turn;
- one nudge on a non-conforming reply, then the entry is written;
- a nudged reply that declines gates quietly, writing nothing;
- at most one nudge — a still-bad reply writes nothing, with exactly two inputs sent.

The RED run failed in exactly the predicted shape: a single `UserInput` ending at `## Existing Knowledge Base table of contents (_toc.md)\n(empty)` with no contract, and the warning `emitted no usable SCOPE/TITLE markers; nothing written`.

## Not in this PR

The other extraction failure family is untouched and still open: 237 skips reading `notes branch no longer exists on origin (already merged on an earlier sweep); nothing to extract` (mcqdb 152, achieveai 85), including why the same branches repeat every sweep without being marked done.

### DIFF
--- a/samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
+++ b/samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
@@ -22,6 +22,61 @@ internal sealed class KnowledgeAgent
     /// <summary>The gate sentinel: the extraction agent replies with this when nothing durable is worth writing.</summary>
     private const string NoKnowledgeSentinel = "NO_KNOWLEDGE";
 
+    /// <summary>
+    /// The output contract, restated at the <b>end of the user turn</b> — the last thing the model reads.
+    /// <para>
+    /// On the S2S path this extraction runs as a hosted <c>workspace-agent</c> conversation, and the
+    /// extraction profile's system prompt is only an <i>appendix</i> to the host's mode prompt. That mode
+    /// prompt says to use sandbox tools for all operations, keep task memory, and summarize what changed —
+    /// and it won. Every August 2026 run on both daemons answered with an action summary or a review
+    /// verdict, so every run parsed to no markers and wrote nothing. Restating the contract here puts it
+    /// where the mode prompt cannot outrank it.
+    /// </para>
+    /// </summary>
+    private const string OutputContract = """
+
+
+        ## REQUIRED OUTPUT CONTRACT (overrides any workspace or mode instructions)
+
+        This is a data-extraction turn, not a workspace task. Everything you need is already in this
+        message. Do not use tools. Do not read, write, create, or edit any file. Do not record task
+        memory. Do not describe actions you took.
+
+        Reply with the entry itself and nothing else. The FIRST line of your reply must be exactly one of:
+
+          NO_KNOWLEDGE
+          ## SCOPE: <one-word area, e.g. system, testing, provider>
+
+        When you emit `## SCOPE:`, follow it with:
+
+          ## TITLE: <short entry title>
+          ## TAGS: <comma-separated tags>
+          ## UPDATES: <existing KnowledgeBase <scope>/<slug>.md path — only when revising that entry>
+
+        then a blank line, then the entry body in Markdown.
+
+        If this PR contributed nothing durable and generalizable, reply with the single line
+        NO_KNOWLEDGE. That is a correct and expected answer — prefer it over prose.
+        """;
+
+    /// <summary>
+    /// The single corrective turn sent when the first reply carried no usable markers. Same thread, so the
+    /// notes the model already read stay in context and only the output shape has to change.
+    /// </summary>
+    private const string ContractNudge = """
+        Your previous reply did not follow the output contract, so nothing could be written.
+
+        Reply again with the entry only — no prose, no action summary, no tool use. The FIRST line must be
+        exactly `NO_KNOWLEDGE` or `## SCOPE: <area>`, followed by `## TITLE:` and `## TAGS:` lines, a blank
+        line, and the entry body.
+
+        If these notes hold nothing durable and generalizable, reply with the single line NO_KNOWLEDGE.
+        """;
+
+    /// <summary>Characters of a non-conforming reply carried into the log — bounded because the notes it
+    /// derives from are attacker-influenceable PR content.</summary>
+    private const int ReplyPreviewChars = 300;
+
     private readonly IMultiTurnAgent _agent;
     private readonly ISandboxFileSystem _fileSystem;
     private readonly ILogger<KnowledgeAgent> _logger;
@@ -67,7 +122,7 @@ internal sealed class KnowledgeAgent
 
         // Gate: an empty or NO_KNOWLEDGE reply means nothing durable — write nothing, leave the KB as-is.
         var text = collected.Text?.TrimStart() ?? string.Empty;
-        if (text.Length == 0 || text.StartsWith(NoKnowledgeSentinel, StringComparison.Ordinal))
+        if (IsDecline(text))
         {
             _logger.LogInformation(
                 "Knowledge run {RunId} produced no durable knowledge (gate); Knowledge Base left unchanged.",
@@ -82,11 +137,49 @@ internal sealed class KnowledgeAgent
         // scope+slug path (which is a create, or an in-place update when that file already exists).
         var targetRelPath = await ResolveTargetRelPathAsync(knowledgeBaseDir, parsed, cancellationToken)
             .ConfigureAwait(false);
+
+        if (targetRelPath is null)
+        {
+            // The extraction profile's system prompt rides as an APPENDIX to the host's much larger
+            // workspace-agent mode prompt on the S2S path, and it loses: the model answers like a coding
+            // agent (action summary, task memory, review verdict) instead of emitting the contract. One
+            // corrective same-thread turn recovers it — same thread, so the notes it already read stay in
+            // context. Exactly one: a model locked into the wrong mode will not conform on retry N, and a
+            // loop would burn the review budget to prove it.
+            _logger.LogWarning(
+                "Knowledge run {RunId} replied without SCOPE/TITLE markers; sending one corrective turn. "
+                    + "Reply began: {ReplyPrefix}",
+                collected.RunId,
+                Preview(text)
+            );
+
+            collected = await AgentTextCollector
+                .CollectAsync(_agent, ContractNudge, cancellationToken)
+                .ConfigureAwait(false);
+            text = collected.Text?.TrimStart() ?? string.Empty;
+
+            if (IsDecline(text))
+            {
+                _logger.LogInformation(
+                    "Knowledge run {RunId} produced no durable knowledge (gate, after nudge); "
+                        + "Knowledge Base left unchanged.",
+                    collected.RunId
+                );
+                return null;
+            }
+
+            parsed = ParseEntry(text);
+            targetRelPath = await ResolveTargetRelPathAsync(knowledgeBaseDir, parsed, cancellationToken)
+                .ConfigureAwait(false);
+        }
+
         if (targetRelPath is null)
         {
             _logger.LogWarning(
-                "Knowledge run {RunId} emitted no usable SCOPE/TITLE markers; nothing written.",
-                collected.RunId
+                "Knowledge run {RunId} emitted no usable SCOPE/TITLE markers after the corrective turn; "
+                    + "nothing written. Reply began: {ReplyPrefix}",
+                collected.RunId,
+                Preview(text)
             );
             return null;
         }
@@ -140,7 +233,27 @@ internal sealed class KnowledgeAgent
         _ = builder.Append(string.IsNullOrWhiteSpace(index) ? "(empty)" : index);
         _ = builder.Append("\n\n## Existing Knowledge Base table of contents (_toc.md)\n");
         _ = builder.Append(string.IsNullOrWhiteSpace(toc) ? "(empty)" : toc);
+        _ = builder.Append(OutputContract);
         return builder.ToString();
+    }
+
+    /// <summary>
+    /// True when the reply is the gate answer: the <see cref="NoKnowledgeSentinel"/>, or nothing at all.
+    /// Reaching this on the corrective turn is a success, not a failure — "not every PR contributes".
+    /// </summary>
+    private static bool IsDecline(string text) =>
+        text.Length == 0 || text.StartsWith(NoKnowledgeSentinel, StringComparison.Ordinal);
+
+    /// <summary>
+    /// A bounded, single-line prefix of a reply, so a non-conforming extraction stops being a silent
+    /// failure without letting untrusted note-derived content flood the daemon log.
+    /// </summary>
+    private static string Preview(string text)
+    {
+        var collapsed = string.Join(' ', text.Split((char[]?)null, StringSplitOptions.RemoveEmptyEntries));
+        return collapsed.Length <= ReplyPreviewChars
+            ? collapsed
+            : string.Concat(collapsed.AsSpan(0, ReplyPreviewChars), "…");
     }
 
     /// <summary>
--- a/samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
+++ b/samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
@@ -93,14 +93,20 @@ internal sealed class KnowledgeAgent
     /// giving it the PR's accumulated <paramref name="notesInput"/> plus the existing index/ToC so it can
     /// update a related entry rather than duplicate it. The agent applies the <b>durable-knowledge gate</b>:
     /// when it replies with the <see cref="NoKnowledgeSentinel"/> (the common case — "not every PR
-    /// contributes"), this returns <c>null</c> and writes nothing. Otherwise it parses the agent's header
+    /// contributes"), this returns <see cref="KnowledgeExtractionOutcome.Declined"/> and writes nothing.
+    /// Otherwise it parses the agent's header
     /// markers (<c>## SCOPE/TITLE/TAGS/UPDATES</c>), resolves create-vs-update against
     /// <paramref name="repoRoot"/>, injects daemon-owned YAML frontmatter deterministically
     /// (<c>updated</c> = <paramref name="todayUtc"/>, <c>sourcePrs</c> merges the existing set with
     /// <paramref name="sourcePrRef"/>), writes the layered entry, then regenerates <c>_index.jsonl</c> and
     /// <c>_toc.md</c> from the entries actually present.
+    /// <para>
+    /// A reply that still carries no usable markers after the one corrective turn is
+    /// <see cref="KnowledgeExtractionOutcome.Failed"/>, <b>not</b> a decline: it is a lost extraction the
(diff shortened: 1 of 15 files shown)


gautam-achieveai:
[Revobot]

# Code Review: PR #250

**Verdict: Request changes**

## Findings

### [HIGH] Artifact context is removed before the notes commit succeeds

**File:** `samples/CodeReviewDaemon.Sample/Orchestration/DaemonReviewStageExecutor.cs:2701-2729`  
**Also used at:** `DaemonReviewStageExecutor.cs:2650-2653`, `2780-2782`

`BuildDaemonNotesArtifactsAsync` removes the per-run context from `_artifactContexts` before `CommitNotesAsync` completes:

```csharp
if (!_artifactContexts.TryRemove(run.Id, out var context))
```

If the subsequent commit or push fails and the commit stage is retried without rerunning the review barrier, the context is gone. The retry therefore falls through to the `review.md`-only path and permanently omits the new context/findings artifacts. This undermines the existing retry/reconcile behavior for notes persistence.

Keep the context until the commit is durably successful, or persist the artifact context durably and make artifact generation idempotent across retries.

---

### [HIGH] Opaque transcript payloads can copy sensitive data into the review store

**File:** `samples/CodeReviewDaemon.Sample/Agents/LmStreamingS2SClient.cs:646-674`  
**Consumed at:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:379-406`

When a persisted message JSON does not contain a string `text` or `content` property, `ExtractTranscriptBody` returns the entire raw `messageJson`:

```csharp
return messageJson;
```

The artifact builder then fences and persists that value for every transcript entry. Opaque payloads include tool calls, tool results, system messages, and provider-specific message shapes; these can contain credentials, request data, URLs, or other user-sensitive content. ANSI stripping and marker neutralization do not constitute secret or EUII redaction.

Only copy explicitly approved message types/fields into durable PR artifacts, or apply a dedicated sensitive-data redaction policy. Tool arguments and opaque provider payloads should not be persisted by default.

---

### [MEDIUM] Raw transcript HTTP error bodies can be persisted

**File:** `samples/LmStreaming.Sample/Controllers/ConversationsController.cs:536-561`  
**Client handling:** `samples/CodeReviewDaemon.Sample/Agents/LmStreamingS2SClient.cs:616-620`  
**Artifact handling:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:362-369`

A non-array transcript response causes the client to include the complete response body in the exception:

```csharp
$"Agent transcript response was not a JSON array ... Body: {body}"
```

The artifact builder catches that exception and writes `ex.Message` into the committed findings file. An error envelope may contain stack traces, absolute paths, request details, or accidentally echoed sensitive values. The artifact sanitizer removes control characters but does not redact this content.

Use a bounded, generic artifact failure message and log only a redacted diagnostic/status identifier. Do not include the raw response body in committed notes.

---

## Testing

The PR adds no tests for `ReviewNotesArtifactBuilder`, transcript parsing, transcript failure handling, or retrying artifact commits. The affected test command could not be run in this environment because the `dotnet` executable is unavailable (`exit 127`).

Tests should cover:

- commit/push failure followed by retry;
- opaque and tool-call transcript payloads;
- error responses containing sensitive-looking fields;
- ANSI/control-character and Markdown-fence containment;
- successful and partially failed per-agent transcript reads.

## Review coverage note

The completed sub-agent review specifically corroborated the transcript payload and raw error-body leakage risks. No other sub-agent areas were available for this synthesis.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-bb64bfa975594eee9fa640208e9168b7&focus=1

gautam-achieveai:
[Gautam's review bot]

**Additional findings — request changes**

### [HIGH] A transcript timeout bypasses the best-effort guards and aborts retention

**Files:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:352-370`, `DaemonReviewStageExecutor.cs:2714-2729`

Both new guards exclude every `OperationCanceledException`. On .NET 5+, however, an `HttpClient` timeout is reported as `OperationCanceledException` with an inner `TimeoutException`; this client keeps the default 100-second timeout (`Program.cs:309-313`). A slow transcript request therefore escapes `RenderFindingsAsync` and `BuildDaemonNotesArtifactsAsync`. Because both retention paths build artifacts before calling `CommitNotesAsync` (`DaemonReviewStageExecutor.cs:2650-2657`, `2780-2790`), one timed-out enrichment read prevents even `review.md` from being retained.

Catch timeout cancellation separately from caller cancellation (for example, only rethrow when the caller token is canceled), and preferably give transcript enrichment its own short bounded timeout so it cannot become load-bearing.

---

### [MEDIUM] Agent names can forge columns in the daemon-authored manifest

**File:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:125-139,466-481`

`UntrustedTranscriptText.Inline` claims to make model-derived text safe for a Markdown table cell, but it does not escape or remove `|`. `SubAgentManager` retains an explicitly supplied agent name verbatim, and the artifact builder interpolates that name directly into the dispatched-agent table. A model-supplied name containing pipes can therefore shift or replace the rendered Template/Status/Findings-file columns, including making an errored reviewer appear completed in the daemon-authored context sheet.

Escape or remove Markdown table delimiters in `Inline` before using its result in table rows.

---

### [MEDIUM] Findings have no aggregate history budget

**Files:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:31-35,275-307`, `DaemonReviewStageExecutor.cs:1413-1428`, `Prompts/daemon-prompts.yaml:67-70`

The 200,000-character limit is per agent, not per round. The host allows up to 32 collaboration agents, so one round can persist roughly 6.4 million characters of findings. Every `PR_Context_*` and `PR_Findings_*` file from every prior round is then listed into `prior_files`, and the re-review prompt instructs the next agent to read them all first. The history therefore grows without a cross-round bound and can exhaust the review context before the agent reaches the current diff.

Add a whole-round/history budget or pass only a compact context artifact into subsequent rounds instead of every raw transcript artifact.

gautam-achieveai:
[Revobot]

## Review findings

### [MEDIUM] Untrusted metadata can forge Markdown table structure in committed notes

**File:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:131-144, 377-386, 620-635`

`UntrustedTranscriptText.Inline` removes newlines and backticks, but it does not escape Markdown table delimiters (`|`) or other Markdown structure. Several values that can originate from the review host or agent roster are inserted directly into table cells:

- `node.Template`
- `node.FailureCode`
- `findings[i].Label`, derived from `node.Name`
- repository/provider metadata

An agent or host-controlled name such as `architecture | Injected | text` can add columns or rows to the daemon-authored context/findings files. Those files are read back into later review prompts, so this allows untrusted metadata to impersonate daemon-authored structure and inject misleading context.

Escape Markdown table content before appending it, at minimum replacing `\` and `|` (and ensuring line breaks are removed), or avoid Markdown tables for untrusted values. Add a regression test covering a roster name/template/failure code containing `|`.

## Coverage note

No sub-agents were dispatched; this review is based on direct inspection of the PR diff and repository context.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-27dd699e351d4e9c8bd1b1523bad377c&focus=1

gautam-achieveai:
[Gautam's review bot]

**Request changes — one new finding**

### [HIGH] Oldest-first budgeting drops the authoritative verdict from the lead findings file

**File:** `samples/CodeReviewDaemon.Sample/Orchestration/ReviewNotesArtifactBuilder.cs:478-518,547-559`

`IsFindingsBearing` filters only on `MessageType`, so it retains user-role `TextMessage` rows. The root transcript is returned chronologically and contains the initial review prompt, provisional answer, synthesis prompt, and only then the authoritative final review. `AppendRetainedEntries` writes from the oldest row forward and stops once the builder has reached 12,000 characters.

Because each earlier entry may consume up to 6,000 characters, a full review prompt plus a sufficiently long provisional answer causes the loop to stop before it reaches the synthesis prompt or final verdict. The resulting `PR_Findings_*_00_lead-reviewer.md` can therefore contain the diff/prompt and provisional output while omitting the review it claims to preserve. Specialist files have the same problem because their initial task is also a user `TextMessage`.

This sends stale inputs rather than conclusions into the next review round and the knowledge extractor. Exclude non-reviewer roles and preserve the newest assistant entries (or reserve budget for the final assistant turn). Add a regression test using the real `user prompt → provisional assistant → user synthesis prompt → final assistant verdict` ordering and assert that the final verdict survives truncation.

The previously posted active findings were rechecked and should remain open; they are not repeated here.

samples/CodeReviewDaemon.Sample/Agents/KnowledgeAgent.cs
```diff
@@ -22,6 +22,61 @@ internal sealed class KnowledgeAgent
...
+        ## REQUIRED OUTPUT CONTRACT (overrides any workspace or mode instructions)
+
+        This is a data-extraction turn, not a workspace task. Everything you need is already in this
+        message. Do not use tools. Do not read, write, create, or edit any file. Do not record task
+        memory. Do not describe actions you took.
+
+        Reply with the entry itself and nothing else. The FIRST line of your reply must be exactly one of:
+
+          NO_KNOWLEDGE
+          ## SCOPE: <one-word area, e.g. system, testing, provider>
```
gautam-achieveai, line 48: [Gautam's review bot] **[MEDIUM] Keep the overriding contract aligned with the canonical KB schema**

This text is deliberately appended last so it overrides the hosted mode prompt, but it now tells the model that `testing` and `provider` are valid scopes. The v1.1 profile defines the only canonical choices as `system` or the repository directory name, while `ResolveTargetRelPathAsync` accepts any path-safe segment. A model following this last instruction can therefore persist entries under `KnowledgeBase/testing/` or `KnowledgeBase/provider/`, fragmenting repo knowledge into noncanonical scopes.

The corrective `ContractNudge` has the same drift: it restates the required fields but omits optional `## UPDATES`, so a retry following it exactly cannot select an existing entry whose path differs from the newly generated title slug. Use one shared contract for both turns—`system` or the actual repo scope, plus `## UPDATES` when refining—and add a corrective-turn update regression test.

gautam-achieveai (commented):
[Gautam's review bot]

**Iteration 3 deep review — one new finding**

Iteration 3 correctly targets the production failure mode: the hosted `workspace-agent` prompt was outranking knowledge extraction, so this adds v1.1, a user-turn contract, and one same-thread corrective turn. One new contract defect remains below.

The seven previously reported artifact/transcript findings remain active and unaddressed; this iteration does not touch those paths. Targeted knowledge-agent tests pass (16/16).

samples/CodeReviewDaemon.Sample/Orchestration/PrLifecycleSweeper.cs
```diff
@@ -210,22 +226,12 @@ private async Task<bool> ResolveMergedAsync(ReviewedPr pr, CancellationToken can
 
         // Layer-2 (design §1): distill durable knowledge from the PR's accumulated notes BEFORE the notes
         // branch merges into the default branch, so the same merge carries the new/updated entry into main.
-        // Extraction failure (agent error, IO) is logged and swallowed — it must NEVER block the merge/delete
-        // (design §6: a capability gap degrades, never fails the lifecycle).
-        if (_extractKnowledgeAsync is not null)
+        if (_extractKnowledgeAsync is not null && !await TryExtractKnowledgeAsync(pr, cancellationToken).ConfigureAwait(false))
```
gautam-achieveai, line 229: [Gautam's review bot] **[LOW] Repair the shared checkout before retrying extraction**

A `Failed` outcome returns here before `MergeToDefaultAsync`, but that method is also the only clean-on-entry repair for this shared checkout (`merge --abort` + `reset --hard`; `ReviewBranchManager.cs:278-289`). If a prior non-FF merge leaves an unresolved index, `KnowledgeExtractionCommitter`'s `checkout -B` fails and is classified as `Failed`. Attempts 1 and 2 then repeat a checkout that cannot succeed because the repair remains deferred; only attempt 3 reaches the cap, runs the merge, and heals the index. This is bounded rather than a permanent wedge, but it wastes two sweep cycles and can emit the final "knowledge is lost" warning even though the extraction agent was never invoked.

Run the checkout cleanup before this early return, or distinguish checkout/index failures from retryable agent/commit/push failures. Add an integration-style regression covering `conflicted index -> extraction checkout failure -> next sweep repair`.

gautam-achieveai (commented):
[Gautam's review bot]

**Iteration 4 deep review — one low-severity retry defect**

The new outcome model correctly preserves notes after transient extraction failures, but checkout-state failures currently consume retry attempts before the only repair step runs. The eight previously reported threads remain active and unaddressed by this commit.

PR #252, opened 2026-08-04
feat(sample): mirror conversation transcripts into the workspace as JSONL

gautam-achieveai:
Fixes #251

Every conversation that has a sandbox workspace now writes its own message history into `.conversations/` inside that workspace, as append-only JSONL an agent can read with the tools it already has — `Read`, `Grep`, `Bash`, DuckDB, whatever it likes.

```
.conversations/
├── .gitignore                                  # "*" — never committed by an agent working in the repo
├── refactor-the-parser-7kzf.jsonl              # {slug(title)}-{shortThreadId}.jsonl
└── refactor-the-parser-7kzf_agents/
    └── researcher-sq7p.jsonl                   # {agentName}-{shortAgentId}.jsonl
```

Each line is one JSON object: `uid`, `parent_uid`, `ts`, `role`, `message_type`, `agent_name`, `message_json`. Flushed at each turn boundary.

## Key decisions

**Full fidelity, reasoning included.** Every other transcript reader in the sample strips reasoning, because every other reader is *cross-agent*. This one is not — it is the conversation reading its own record inside its own workspace. `TranscriptProjection`'s remarks now name this fourth caller explicitly so the next person doesn't "fix" it back.

**Ordering is by parent pointers, not an ordinal.** `uid` / `parent_uid`, so a line's position is a property of the line rather than of the flush that wrote it. `uid` is a truncated SHA-256 of the persisted message id — derived, so the same message always gets the same uid, and a re-flush produces byte-identical lines.

**`message_json` is an opaque string.** The reader owns the schema. The mirror never parses or reinterprets a provider payload, so a provider adding a field can't break the writer.

**Appends are genuine appends.** Stage a temp file, `cat` it onto the destination, remove the temp — one `sh -c` with positional parameters, and no data is ever interpolated into the script. Retitling a conversation is `mv`. Nothing is ever read-modify-rewritten, so the file never briefly ceases to exist and a reader never sees a truncated file.

**Every failure path leaves the watermark unadvanced.** The watermark advances *only* on exit code 0, so a failed append means the next flush recomputes the same suffix and writes it again. The worst outcome is a duplicated tail on retry — never silent loss. That is the deliberate shape for "works in 95% of cases without extreme durability."

**The store lags the stream.** History is persisted on an unawaited task that `CompleteRunAsync` never joins, so `RunCompletedMessage` can arrive before the last message is on disk. A flush reads twice and compares message-**id sequences** — not counts, which would miss a reorder — bounded at three attempts, then defers whole.

[...]

Issue #251, opened 2026-08-03
Mirror conversation transcripts into the sandbox workspace as JSONL


gautam-achieveai:
## Summary

Mirror each conversation's message history into its sandbox workspace as JSONL, so any other
conversation sharing that workspace can read what has already been done there.

Design validated against **Claude Code's own transcript format** (`~/.claude/projects/`, 1020 files,
93 sessions, measured with DuckDB — see `prior-art-claude-code-transcript-format.md`), which solves
this exact problem in production.

**Layout** — workspace-relative, parent directories created by `mkdir -p` in the same command as the
write:

```
.conversations/
  {slug(title)}-{shortThreadId}.jsonl          # the main conversation
  {slug(title)}-{shortThreadId}_agents/
    {agentName}-{shortAgentId}.jsonl           # one file per sub-agent
```

Sub-agents get their own files rather than an inline flag: in the reference corpus sub-agent volume
**exceeded the main conversation** (26 MB vs 12 MB for one session), so a single flat log would be
dominated by delegate chatter.

**Trigger.** A subscriber on the existing message stream (`MultiTurnAgentBase.SubscribeAsync`)
flushes when it sees `RunCompletedMessage` (published by `CompleteRunAsync`, `:2083`). This is a real
run/turn boundary, needs no lifecycle wiring, and because deferred tool results resolve as child runs
(ADR 0004) their late results flush too.

**The flush must not run inline on the subscription thread.** `PublishToSubscriber`
(`MultiTurnAgentBase.cs:1444-1457`) *drops* a subscriber whose channel fills — removing it from
fan-out and completing its enumerator permanently. Awaiting a sandbox `PUT` inside the
`await foreach` would silently stop transcript writing for the rest of a busy run. Hand the flush
off; do not block the enumerator.

**Write — genuine append, not whole-file replacement.** The typed file API is **PUT / GET / LIST
only; there is no append, delete, move, or rename** (verified across `SandboxClient.Files.cs`; the
only `DeleteAsync` is session teardown, `Lifecycle.cs:115`). But `SandboxClient.ExecuteAsync`
(`Command.cs:78`) gives a shell, and the shell has all four.

So each flush loads the thread via `IConversationStore.LoadMessagesAsync`, takes **only the lines
added since the last flush**, and writes them in two calls:

1. `PUT .conversations/.tmp/{shortThreadId}.part` — the new lines only.
[...]

PR #263, opened 2026-08-05
fix(transcript): attach the mirror on every agent branch, drain flushes FIFO, refuse aliased paths — with a real browser E2E

gautam-achieveai:
Follow-up to #252 (issue #251). Three fixes found while babysitting that PR, plus a real end-to-end proof that the main one works.

## The fixes

| Commit | What it fixes |
|---|---|
| `57bf4200` | **P2** — the flush scheduler drained in `HashSet` slot order, not FIFO |
| `458dbca1` | **P1** — the mirror attached on one agent-construction branch, not all of them |
| `acc96e35` | a transcript path that acquires a *second name* (hard link) is now refused, not just a symlinked one |

### P1 (`458dbca1`) — what is and isn't proven

`MultiTurnAgentPool` builds agents down several branches, and only one of them reached the single
`Attach` call. Providers routed through the others mirrored **nothing**.

Being precise about the coverage claim, because it is easy to overstate:

- **Two branches are covered by test** — `WorkspaceTranscriptMirrorAttachCompositionTests` drives
  those directly.
- **Four more are covered by a structural argument**: attachment moved to the single point every
  branch funnels through, so they cannot individually regress without that point changing.

That is **not** six tested branches, and this PR does not claim it is. The structural half is only as
good as the funnel staying a funnel — which is exactly what the E2E below checks from the outside.

### The bug's real signature

Worth stating because it is not "no transcript appears". Turn 1 on an attaching provider already
created the file. Switching that same conversation to a non-attaching provider left the file in
place and simply **stopped adding to it** — a silent hole in the middle of a live transcript, with
no error anywhere. Any check that only asserts "a file exists" passes on the broken build.

## Availability traded for containment (`acc96e35`) — read this before merging

This one deliberately makes the feature **less available** to make it safer, and the tradeoff should
be a decision, not a surprise.

A transcript destination is refused if it is a symlink **or if it bears more than one name** (a hard
link). The consequence:

> **A conversation whose transcript leaf acquires a second name stops mirroring, and stays stopped
> until a human removes the link.** It does not recover on its own, and no amount of restarting
> helps.

That is not a limitation waiting to be polished away — it is forced by the filesystem. A hard link is
**symmetric**: the two names are equally real, and the filesystem **does not record which one came
first**. So on finding two names there is no way to tell the original from the intruder, which means
there is no safe automatic repair. Deleting "the other one" is a coin flip that can destroy the
user's file; following it writes the conversation's bytes to a location nobody chose.

Refusing is the only option that cannot lose data or leak it somewhere unintended. It is also why the
guard **detects but never repairs**, and why the check walks every **ancestor** directory rather than
just the leaf — a symlinked `.conversations` redirects every file beneath it, including the
`.gitignore` that is supposed to be covering them, while the leaf itself tests perfectly clean
because at that point it does not exist yet.

Detection cost, for the record: `[ -L ]` cannot see a hard link at all, so the link count comes from
`ls -dn` field 2 (POSIX-specified; `-n` implies `-l`). `stat` was rejected (the format string is not
portable — `%h` GNU vs `%l` BSD) and POSIX `find … -links 1` is worse: the guard is a compile-time
`const string` shared by **both** executors, and the test surface is a Windows host where `sh`
resolves `find` to `C:\WINDOWS\system32\find.exe`, which prints `FIND: Parameter format not correct`
and no path — indistinguishable from a legitimate refusal. That was not reasoned about, it was run:
it made the guard refuse four pre-existing symlink cases that require a successful write. `ls -dn` is
POSIX and has no Windows name collision. Both alternatives were tried; neither shipped.

### Verified on the production surface, not just the host

The guard text was run inside `llm-sandbox-agent:latest` — **six of six behaviours as specified**
(absent → allow, nlink 1 → allow, nlink 2 → refuse, other end of the same link → refuse, directory →
refuse, symlink → caught by the ancestor walk). Shell is `/usr/bin/dash`, not bash; coreutils is GNU
9.1 with no busybox.

Two readings sharpen the severity claim, and both **reduce** it:

- **`protected_hardlinks=1` and enforced**, proven with a valid control: uid 1000 linking a
  root-owned file it can read but not write fails `EPERM`, while linking a file it owns succeeds
  (`exit=0`, `-rw-r--r-- 2 1000 1000`).
- **The writer runs as the agent's own uid by construction, not by coincidence.** Both mutation paths
  in `SandboxSessionRegistry` (`:1638` file-PUT, `:1655` `ExecuteWorkspaceCommandAsync`) resolve the
  same `sessionId` → same credential → same container, and `SandboxCommand` has exactly three members
  — `Arguments`, `WorkingDirectory`, `OperationId`. There is **no user, uid, or privilege field on the
  type at all**, so no mechanism exists for the writer to run as anything else.

**So this is not a confused deputy.** The agent can only link files it already owns, which it can
already write directly. The guard's value is containment and hygiene — keeping the transcript from
landing inside a git-tracked file — not privilege escalation. It ships as defense-in-depth, and the
severity is stated that way rather than inflated.

The guard **fails closed** when `ls` succeeds but prints something unparseable. The obvious objection
— "a missing binary would then disable the feature forever" — was checked and does not apply: GNU
coreutils 9.1 is present, there is no busybox, and `ls` is the least removable external a POSIX shell
has. The axis it actually fails closed on is *"the link count could not be read"*, where proceeding
would mean writing while admitting you don't know what you're writing to.

Both PUT paths are guarded **before** the bytes land (`ConversationTranscriptWriter.cs:1541` ahead of
`:1549`, same shape at `:1830` for the `.gitignore`), because a script-side check alone would be
useless — the PUT completes before any script runs. The in-script `solo "$2"` is the second layer,
catching an alias planted in the window between the guard and the splice.

## The end-to-end proof (`42a3771d`)

The unit tests verify the seam. They cannot verify that the mirror survives a real provider switch on
a real conversation writing through a real gateway — which is the exact thing P1 broke. So this adds
`playwright-scripts/transcript-mirror.mjs`: one self-contained script that drives the whole flow and
returns a structured verdict.

It exercises all three attach paths **in a single conversation**:

1. a provider that always attached → turn 1, `test-anthropic`, workspace-agent mode
2. named background sub-agents → turn 1 spawns `mirror-alpha` / `mirror-beta`
3. a CLI provider that did **not** attach → same thread switched to `codex-mock`, turn 2

Because turn 1 creates the file, a P1 regression appears as **MARKER1 present, MARKER2 absent**. The
verdict distinguishes `P1_HOLDS` / `P1_REGRESSED` / `FAILED_NOTHING_MIRRORED` rather than collapsing
to pass/fail.

**Result: `P1_HOLDS`, 21/21 steps, zero failures — green on two consecutive independent runs**, the
second against a `.conversations/` directory that already held the first run's transcript.

| Claim | Evidence |
|---|---|
| P1 holds | MARKER1 **and** MARKER2 in the same file; 13319 → 21773 bytes across the provider switch |
| P2 holds | 19/19 lines parsed, 0 unparsable, 0 duplicate `uid`s, MARKER1 line 0 → MARKER2 line 15 |
| Sub-agent fan-out | `mirror-alpha-….jsonl` + `mirror-beta-….jsonl`, both markers present |
| Filename contract | `transcript-mirror-e2e-<stamp>-<shortThreadId>.jsonl` |
| Binding survived | workspace intact across mode→`default` and provider→`codex-mock` |
| RAW reasoning fidelity | `ReasoningMessage` rows present, counted by discriminator |

### Three things the script gets right only because the first run got them wrong

- **Gates the switch on `GET {id}/run-state`, not on the UI going idle.** UI-idle is not run-idle:
  when a turn spawns background sub-agents the stop button hides while `GetRunStateInfo` stays
  `IsInProgress`, and the switch guards read *exactly that state* — so switching on the UI proxy
  earns a 409. The poll now reads the same signal the guard reads, so it cannot disagree with it.
- **Polls for both sub-agent files.** Each child flushes on its own schedule, so a single-shot
  listing catches only the first and reads as "the second was never mirrored". Observed once,
  against a perfectly healthy mirror.
- **Matches this run's transcript by its own slug prefix.** `.conversations/` accumulates one file
  per conversation, so "the first `.jsonl` in the listing" silently becomes a *previous* run's file
  on any re-run, whose marker carries a stale stamp.

Reading the bytes back **through the app's own file-browser route** is load-bearing, not a
convenience: the mirror writes into the sandbox workspace through the gateway, and this deployment
pins that gateway to a different machine, so a host-filesystem check reads as "no file" and reports a
false regression. `files/download` is used rather than `preview` because `FilePreviewPolicy` excludes
dot-directories by design.

RAW reasoning fidelity is asserted off the `message_type` discriminator, **never a substring** — a
`/"reasoning"/` scan matches the User row's own `message_json`, which echoes the prompt's
`"reasoning":{"length":30}` back verbatim, and so reports hits on a broken mirror too.

## Known gap, not fixed here: state lines are dead code

The E2E surfaced this and it is worth a decision rather than silence. `WorkspaceTranscriptLine.ForState`
(title / mode / provider) is specified, serialized, and unit-tested — and has **zero production
callers**. A run that set the title, switched mode, *and* switched provider produced 43/43
`type:"message"` lines and **0** `type:"state"`.

It is not a one-line fix. `ResolveStartIndexAsync` re-derives the entire transcript on every flush and
locates a watermark `uid` inside that recomputed list, which requires every line to be **replayable**.
`ForState` seeds its uid with the wall clock, so a state line is not reproducible on the next flush →
the watermark goes missing → the whole history re-appends, every flush. Dropping the timestamp does
not rescue it either: at the head, state changes are never appended; at the tail, new messages are
never appended.

A correct fix needs a persisted, timestamped state-change history (`Properties["mode"]` keeps only the
current value), plus `RecoverWatermarkAsync` skipping state lines when reading the file tail. That
type lives in `src/`, and the locked decision for this feature is **zero `src/` changes** — so it is
filed rather than smuggled in here.

Closes nothing on its own; follow-up to #251.

### DIFF
--- a/samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs
+++ b/samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs
@@ -1,0 +1,698 @@
+// transcript-mirror.mjs — single-call Playwright E2E driver AND verifier for the WORKSPACE
+// CONVERSATION TRANSCRIPT MIRROR (#251 / PR #252) and its two follow-up fixes on
+// `wt3-transcript-followup`.
+//
+//   browser_run_code_unsafe({ filename: "<path>/transcript-mirror.mjs" })
+//
+// WHAT THIS PROVES
+//   (a) a provider that ALWAYS attached           → `test-anthropic` + workspace-agent mode, turn 1
+//   (c) a named background sub-agent              → turn 1 spawns `mirror-alpha` / `mirror-beta`
+//   (b) a CLI/mock provider that DID NOT attach   → same thread switched to `codex-mock`, turn 2
+//       ^^^ THIS IS THE P1 PROOF (458dbca1). Before P1 the CLI provider branches in Program.cs
+//       returned before the single Attach call, so those agents mirrored nothing. Because turn 1
+//       already created the file, the bug's signature is: file exists, contains MARKER1, and MARKER2
+//       is ABSENT — a silent hole in the middle of a live transcript.
+//   P2 (57bf4200, FIFO flush drain) is checked best-effort: no duplicate `uid`s, and MARKER1's row
+//       precedes MARKER2's row. A drain that re-ordered or replayed a queued flush shows up here.
+//
+// HOW IT VERIFIES — read this before changing the assertions.
+// The mirror writes into the SANDBOX WORKSPACE through the gateway, so the bytes are NOT on the
+// browser side. Earlier drafts therefore punted the verdict to a host-filesystem PowerShell script.
+// That is WRONG on this deployment: `samples/LmStreaming.Sample/.env` pins
+// `SandboxGateway__BaseUrl=http://192.168.11.139:3000` with `AutoSpawn=false`, and 192.168.11.139 is
+// a DIFFERENT machine (this host is 192.168.11.20; the route to .139 leaves via the Tailscale
+// interface, and the two gateways report different `active_sessions`). The transcript lands on
+// whichever machine the configured gateway runs on, so a host-path check on the wrong machine reads
+// as "no file" — a FALSE P1 REGRESSED.
+//
+// So this script reads the transcript back through the app's own file browser
+// (`FileBrowserController`, `[Route("api/conversations/{threadId}/files")]`), which goes through the
+// SAME gateway the writer used and is therefore correct whether that gateway is local or remote.
+// `disk-checks.ps1` remains useful only as a second opinion when the gateway is the local one.
+//   • `files?path=…`          → listing (200 + `{state:"no_session_yet"}` before a sandbox exists)
+//   • `files/download?path=…` → raw bytes. Use THIS, not `preview`: `FilePreviewPolicy` excludes
+//     anything under a dot-directory outright, so `preview` returns `{previewable:false,
+//     reason:"excluded"}` for `.conversations/*.jsonl` BY DESIGN. Not a bug — do not "fix" it.
+//   • `[InboundS2SAuth]` is marker-gated (it enforces only when `X-S2S-Auth` / `X-Sbx-App-Id` is
+//     present), so plain same-origin `fetch` from the page is allowed. All API calls run in-page.
+//
+// KNOWN INCONCLUSIVE PATH: the controller resolves a path one component at a time against the
+// gateway's listing, so reaching `.conversations` needs the ROOT listing to include dot entries. The
+// UI can navigate there (that is why the preview exclusion exists), but this script does not assume
+// it — step `workspace root listing` records every root entry name, and if `.conversations` never
+// resolves the verdict is INCONCLUSIVE_NOT_LISTABLE, never REGRESSED. Fall back to disk-checks.ps1.
+//
+// HARD PREREQUISITE: a LIVE SANDBOX GATEWAY. With no workspace binding every flush returns
+// TranscriptFlushOutcome.Unavailable and writes nothing, silently — indistinguishable from the bug.
+// The script fails fast at `workspace bound before send` rather than producing a false negative.
+//
+// Prompt: PromptExamples.md → "Sub-Agent Tabs (center-pane, colored)" → "Two named background
+// sub-agents, text-only (transcript-mirror fixture)". Nested chains are built with JSON.stringify so
+// the escaping is correct by construction — never hand-escape these.
+async (page) => {
+  // PORTS — there is NO launchSettings.json in samples/LmStreaming.Sample, so nothing overrides the
+  // binding except `samples/LmStreaming.Sample/.env`, which Program.cs:66 loads via FindEnvFile.
+  // FindEnvFile walks up from **AppContext.BaseDirectory** (the bin dir), NOT the shell cwd, so the
+  // launch directory cannot change which .env wins. That file sets
+  // ASPNETCORE_URLS=http://0.0.0.0:5055. `:5000` is the bare framework default you only get without
+  // it; `:5050`/`:5098`/`:5273` are isolated instances. Probe rather than pin — put your origin
+  // first to skip the probe.
+  const CANDIDATES = [
+    'http://localhost:5055',
+    'http://localhost:5050',
+    'http://localhost:5000',
+    'http://localhost:5098',
+    'http://localhost:5273',
+    'http://localhost:5173',
+  ];
+  const WORKSPACE_NAME = 'transcript-e2e';
+  const WORKSPACE_DIR = 'transcript-e2e'; // single segment — separators are stripped by the sanitizer
+  const BOUND_PROVIDER = 'test-anthropic'; // the one mock Workspace Agent mode permits
+  const CLI_PROVIDER = 'codex-mock'; // the only CLI-family provider needing no binary on PATH
+  const TRANSCRIPT_DIR = '.conversations'; // ConversationTranscriptWriter.TranscriptDirectory
+  const AGENTS_SUFFIX = '_agents'; // ConversationTranscriptWriter.AgentsDirectorySuffix
+  const STAMP = String(Date.now());
+  const TITLE = `Transcript Mirror E2E ${STAMP}`; // slugs to `transcript-mirror-e2e-<stamp>`
+  // Every `.jsonl` under `.conversations/` is NOT this run's transcript — the directory accumulates
+  // one per conversation, so on the SECOND run "the first .jsonl in the listing" is a previous run's
+  // file, whose MARKER1 carries a previous stamp. That reads as FAILED_TURN1_NOT_MIRRORED against a
+  // perfectly healthy mirror. Match this run's file by its own slug prefix instead.
+  const EXPECTED_PREFIX = `transcript-mirror-e2e-${STAMP}-`;
+  const isOurTranscript = (name) =>
+    String(name).startsWith(EXPECTED_PREFIX) && String(name).endsWith('.jsonl');
+  const MARKER1 = `MIRROR-TURN1-${STAMP}`;
+  const MARKER2 = `MIRROR-TURN2-CLI-${STAMP}`;
+  const ALPHA_TEXT = `mirror-alpha reporting ${MARKER1}`;
+  const BETA_TEXT = `mirror-beta reporting ${MARKER1}`;
+  // The flush is scheduled/debounced (TranscriptFlushScheduler), so the bytes trail the UI going
+  // idle. These are generous ceilings on a poll, not sleeps — every wait exits on first success.
+  const APPEAR_TIMEOUT_MS = 120000;
+  const POLL_MS = 2000;
+
+  const steps = [];
+  const record = (name, pass, detail) => {
+    steps.push({ name, pass, detail });
+    return pass;
+  };
+  const tid = (id) => page.locator(`[data-testid="${id}"]`);
+  const waitIdle = async () => {
+    await tid('stop-button').waitFor({ state: 'hidden', timeout: 120000 });
+    await tid('send-button').waitFor({ state: 'visible', timeout: 120000 });
+  };
+  const send = async (text) => {
+    await tid('chat-input-textarea').fill(text);
+    await tid('send-button').click();
+  };
+  const errorBanner = async () => {
+    const n = await tid('error-banner').count();
+    return n === 0 ? null : (await tid('error-banner').first().innerText().catch(() => '<unreadable>'));
+  };
+
+  let BASE = null;
+
+  // Same-origin, in-page fetch — the interactive path [InboundS2SAuth] deliberately lets through.
+  const api = (method, path, body) =>
+    page.evaluate(
+      async ({ method, path, body }) => {
+        const init = { method, headers: { 'content-type': 'application/json' } };
+        if (body !== undefined) init.body = JSON.stringify(body);
+        const res = await fetch(`${location.origin}${path}`, init);
+        const text = await res.text();
+        let json = null;
+        try {
+          json = text ? JSON.parse(text) : null;
+        } catch {
+          /* non-JSON body — the raw text is kept for the failure detail */
+        }
+        return { status: res.status, ok: res.ok, json, text: text.slice(0, 600) };
+      },
+      { method, path, body }
+    );
+
+  // Raw workspace bytes (transcripts are JSONL; `preview` refuses dot-directories by design).
+  const downloadWorkspaceFile = (threadId, wsPath) =>
+    page.evaluate(
+      async ({ threadId, wsPath }) => {
+        const url = `${location.origin}/api/conversations/${encodeURIComponent(threadId)}/files/download?path=${encodeURIComponent(wsPath)}`;
+        const res = await fetch(url);
+        return { status: res.status, ok: res.ok, text: res.ok ? await res.text() : null };
+      },
+      { threadId, wsPath }
+    );
+
+  // Classifies a listing into the four states that mean different things here. `no_session_yet` is
+  // NOT "empty": it means no sandbox has been established, so nothing could have been written.
+  const listWorkspaceDir = async (threadId, wsPath) => {
+    const res = await api('GET', `/api/conversations/${encodeURIComponent(threadId)}/files?path=${encodeURIComponent(wsPath)}`);
+    if (res.status === 404) return { state: 'not-found', entries: [] };
+    if (!res.ok) return { state: 'error', entries: [], status: res.status, body: res.text };
+    if (res.json && res.json.state === 'no_session_yet') return { state: 'no-session', entries: [] };
+    const entries = (res.json && res.json.entries) || [];
+    return { state: 'listed', entries, moreCount: (res.json && res.json.moreCount) || 0 };
+  };
+
+  const pollUntil = async (fn, timeoutMs) => {
+    const deadline = Date.now() + timeoutMs;
+    let last = null;
+    for (;;) {
+      last = await fn();
+      if (last && last.done) return last;
+      if (Date.now() >= deadline) return last;
+      await page.waitForTimeout(POLL_MS);
+    }
+  };
+
+  // UI-idle is NOT run-idle. When a turn spawns BACKGROUND sub-agents, the parent's stop-button
+  // hides as soon as the PARENT's own stream ends, but `agentPool.GetRunStateInfo(threadId)` stays
+  // IsInProgress until the whole run drains. The mode/provider switch guards
+  // (ConversationsController.cs:1263, :1380) read exactly that state, so switching on UI-idle earns
+  // a 409 `mode_switch_while_streaming` — observed as AgentIsRunning=True, RunTaskCompleted=False,
+  // IsStale=False, i.e. a genuinely live run, not a stale-state bug. Gate on the same signal the
+  // guard reads rather than on a proxy for it.
+  const waitRunIdle = (threadId, timeoutMs) =>
+    pollUntil(async () => {
+      const rs = await api('GET', `/api/conversations/${encodeURIComponent(threadId)}/run-state`);
+      return { done: !(rs.json && rs.json.isInProgress), runState: rs.json };
+    }, timeoutMs);
+
+  const result = {
+    pass: false,
+    verdict: 'NOT_REACHED',
+    failures: [],
+    steps,
+    base: null,
+    workspace: null,
+    conversation: null,
+    transcript: null,
+    markers: { turn1: MARKER1, turn2: MARKER2, alpha: ALPHA_TEXT, beta: BETA_TEXT },
+    subAgents: [],
+    diskChecks: null,
+  };
+
+  try {
+    // ---- 0. Find the running instance -------------------------------------------------------
+    for (const candidate of CANDIDATES) {
+      const res = await page.request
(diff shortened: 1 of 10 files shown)


gautam-achieveai:
[Revobot]

# Review: achieveai/LmDotnetTools#263

**Head:** `42a3771de42cd683a86ad6c2536dd7b6763b50de`  
**Verdict:** **Request changes**

The commit adds a manual Playwright transcript-mirror verifier and documentation. The new script is useful, but several assertions can produce false positives or false regressions, and its synchronization/error handling is unsafe for a test intended to validate asynchronous transcript behavior.

## Findings

### [HIGH] Failed run-state requests are interpreted as an idle run

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:172-176`

```js
const rs = await api('GET', `/api/conversations/${encodeURIComponent(threadId)}/run-state`);
return { done: !(rs.json && rs.json.isInProgress), runState: rs.json };
```

`api()` returns HTTP status and success information, but `waitRunIdle()` ignores them. A `404`, `500`, HTML fallback, or malformed JSON produces `rs.json === null`, which makes `done` true.

The script can consequently switch mode/provider while the parent or background sub-agents are still active. This can race with transcript flushing and turn an API failure into a misleading mirror regression.

Require a successful response and explicitly validate that `isInProgress === false`; otherwise continue polling or report an inconclusive environment failure.

---

### [MEDIUM] Turn 2 can be reported as a P1 regression before the run has drained

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:586-608`

The second server-side drain result is ignored:

```js
await waitRunIdle(threadId, APPEAR_TIMEOUT_MS);
```

The script proceeds to read and validate the transcript even if the run is still active or the run-state endpoint timed out. A partial transcript can then produce `P1_REGRESSED`, P2 ordering failures, or JSONL parsing failures that are only transient.

Capture the result and return an inconclusive result when `done` is false, matching the handling already used for the first provider switch.

---

### [MEDIUM] The P1 check can pass after the first turn was lost

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:610-619`

The P1 assertion only requires `hasM2`:

```js
const p1 = record(..., hasM2, {
  marker1Present: hasM1,
  marker2Present: hasM2,
});
```

It does not require `MARKER1` to remain in the final transcript. A broken implementation that truncates or replaces the transcript while retaining turn 2 would pass the P1 check.

Require both `hasM1` and `hasM2`, and verify that the first marker precedes the second.

---

### [MEDIUM] Child-agent verification checks counts, not names or content

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:432-436, 497-520`

The checks pass based only on counts:

```js
record('two named background sub-agents spawned', rows.length >= 2, ...)
record('sub-agent transcripts mirrored', agentFiles.length >= 2, ...)
```

The expected names and marker presence are included only in diagnostic output. Two unrelated child rows/files, stale files, or unreadable files could satisfy these assertions.

Require:

- child API rows named `mirror-alpha` and `mirror-beta`;
- filenames matching those names;
- successful downloads; and
- `ALPHA_TEXT` and `BETA_TEXT` present in the corresponding files.

This is particularly important because the accompanying documentation describes the names as load-bearing.

---

### [MEDIUM] Transcript discovery ignores the file-browser listing limit

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:145-151, 440-458`

`listWorkspaceDir()` preserves `moreCount`, but the script does not use it. The file-browser endpoint returns only a bounded number of entries, while the script searches only `listing.entries` for the current transcript.

After enough runs or other workspace activity, the current transcript may fall outside the returned entries. The script will then report `FAILED_NO_TRANSCRIPT_AT_ALL` even though the mirror worked.

Handle `moreCount` explicitly, add pagination if supported, or use a bounded direct file lookup that does not rely on an unpaged directory listing.

---

### [MEDIUM] In-page requests have no timeout or cancellation

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:114-140`

Both `api()` and `downloadWorkspaceFile()` use browser-side `fetch()` without an `AbortController`. Therefore, `pollUntil()` cannot enforce its timeout if a request or response-body read hangs.

A stalled gateway can retain the browser call indefinitely and prevent the script from reaching any cleanup logic. Downloads also materialize complete transcript responses in memory.

Add bounded request cancellation for all in-page calls and distinguish request timeouts from valid negative assertions.

---

### [MEDIUM] The raw reasoning check validates only a discriminator

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:646-666`

The check passes when any row has a `message_type` beginning with `Reasoning`:

```js
const reasoningRows = Object.entries(byType)
  .filter(([t]) => t.startsWith('Reasoning'))
  .reduce((n, [, c]) => n + c, 0);
```

It does not verify that the associated `message_json` is present, valid, non-empty, or contains the reasoning payload. A mirror that preserves the row type but drops the actual reasoning content would pass despite the script claiming to verify full raw fidelity.

Parse the reasoning rows and assert that their serialized payload contains the expected non-empty reasoning data.

---

### [MEDIUM] Each invocation permanently accumulates conversations and transcript artifacts

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:68-80, 238-309`

Every run creates a unique conversation title and leaves behind:

- the conversation metadata;
- the root transcript;
- the `_agents` directory;
- descendant transcripts; and
- potentially live state when the script exits early.

Repeated runs share the same workspace and grow the transcript directory, contributing directly to the listing-limit problem above. Concurrent runs also share the same workspace.

Add a teardown path, or create an isolated per-run workspace with an explicit retain-artifacts option for debugging.

---

### [LOW] Missing CLI prerequisite is detected only after the expensive first turn

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:217-235, 527-535`

The script records whether `codex-mock` is available but returns early only if the attaching provider is unavailable. If `codex-mock` is missing, it still provisions a conversation, sends turn 1, spawns sub-agents, polls files, and downloads transcripts before returning inconclusive.

Check all required providers before creating the conversation or sending any turn. This also avoids leaving artifacts when a prerequisite is known to be absent.

---

### [MEDIUM] The new end-to-end coverage is manual-only

**Files:**  
`samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:1-8`  
`samples/LmStreaming.Sample/playwright-scripts/README.md:1-5,44-55`

The new verifier is explicitly a manual `browser_run_code_unsafe` script. No CI registration or automated gateway-backed test is added in this commit.

The P1 scenario is a useful manual regression oracle, but the important deterministic assertions—marker continuity, child names/content, and reasoning payload preservation—will not protect future changes unless someone manually runs the script.

Keep the manual gateway test, but move deterministic portions into automated tests using the existing mock-provider and fake gateway/file-browser seams.

## Validation

- The reviewer sub-agents reported that `node --check samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs` passed.
- `git diff --check` passed.
- Targeted .NET tests could not be executed because `dotnet` is unavailable in the sandbox.
- No provider API calls or review comments were posted.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-78010db25a624c3ea57c3d28b098cffd&focus=1

gautam-achieveai (commented):
[Gautam's review bot] Deep review complete on `42a3771d`.

I found no additional blocking correctness, security, concurrency, or resource-lifetime defects in the base-to-head diff. The FIFO queue/set invariants, factory-wide mirror attachment and cleanup, and hard-link refusal behavior are coherent, and the affected build/tests completed cleanly.

I also independently checked the later Revobot Playwright feedback. Its code observations are mostly valid as hardening, but they do not justify a blocking verdict for this manual verifier: the run-state/listing/fetch cases fail toward inconclusive, false-negative, or a hung run, and none can manufacture the run-specific turn-2 marker required for `P1_HOLDS`. The highest-value non-blocking tightenings are requiring both markers in final order, asserting child names/content, treating truncated listings as inconclusive, and validating non-empty reasoning payloads.

Context traced through #251, #252, and ADR 0011. No duplicate inline comments added.

gautam-achieveai:
All eight actionable findings are fixed in `4367d0ef`, and the E2E was **re-run**, not just re-read: **`P1_HOLDS`, 23/23 steps, zero failures** (was 21 — two of the fixes add steps).

Thank you for these. Every one of them is the same shape, and it is the shape I spent this whole PR flagging in the *product*: **a property verified on a convenient surface and asserted of the real one.** Finding it eight times in my own test script is the useful part of the review.

## Fixed

| # | Finding | What it would have done |
|---|---|---|
| 1 | **`waitRunIdle` ignored HTTP status** *(HIGH — correct severity)* | `!(json && isInProgress)` is true for a 500, a dropped connection, and — most likely — a 404 falling through to the SPA fallback returning `index.html`, so `json` parses to `null`. Any of those **skipped the gate and handed the 409 back at the switch**: the exact failure the gate exists to prevent, wearing a green tick. Now requires an OK response that *explicitly* says `isInProgress === false`, and separates `INCONCLUSIVE_RUN_NEVER_DRAINED` from `INCONCLUSIVE_RUN_STATE_UNREADABLE` — "investigate the app" vs "investigate the route". |
| 2 | **Turn 2's drain result discarded** | A turn-2 run that never drained became a MARKER2 read taken mid-flight, and the resulting `P1_REGRESSED` blamed the mirror for a timing problem in the script. Recorded, and folded into the verdict as `INCONCLUSIVE_TURN2_NEVER_DRAINED`. |
| 3 | **P1 asserted MARKER2 alone** | MARKER1 was already **proven present in this same append-only file** two steps earlier, so if it is gone, previously-verified content was *destroyed* — strictly worse than the P1 hole the step is named for, and the old assertion scored it green. Now requires both, with a dedicated `FAILED_TURN1_LINES_DISAPPEARED`. |
| 4 | **Sub-agent checks counted instead of identified** | `rows.length >= 2` and "two `.jsonl`s exist" are both satisfied by the *wrong* pair, or by one agent mirrored twice — and the filename contract (`{slug(agentName)}-{shortAgentId}.jsonl`) is the thing under test. Now matched by name and by derived leaf, and the poll waits for *those two* rather than for any two. Worth calling out: `alphaPresent`/`betaPresent` were **already being collected** into the detail block while pass/fail rode on the file count. Evidence gathered and not asserted is the same bug in miniature. |
| 5 | **Listing cap ignored** | `FileBrowserLimits.MaxListingRows` is 500 and overflow returns as `moreCount`, not an error. `.conversations/` accumulates one file per conversation **forever**, so a long-lived workspace eventually pushes this run's file outside the window — and "not in the first 500" would read as `FAILED_NO_TRANSCRIPT_AT_ALL` against a mirror that wrote perfectly. Now reports `INCONCLUSIVE_LISTING_TRUNCATED` and says to prune. |
| 6 | **No fetch timeout** | `page.evaluate` imposes none, so a hung in-page request blocked the whole script with no diagnostic: the run died on the harness ceiling *after* `steps` stopped growing, so the report named a step that had already passed. Both helpers now abort (20 s API / 60 s download) and surface `{ok:false, status:0, timedOut:true}` — which flows into the same failure paths, and since #1 now demands an OK response, a timed-out poll **retries** instead of reading as idle. |
| 7 | **RAW fidelity checked the discriminator, not the payload** | A row carrying `message_type: "ReasoningMessage"` with an emptied `message_json` passed the step whose entire subject is *content* fidelity. Now also requires the mock's generated **lorem** text. That sentinel is deliberate: `reasoning:{length:30}` makes the mock emit lorem words, which the prompt never contains — unlike the word `"reasoning"`, which the User row echoes back from its own `"reasoning":{"length":30}` instruction chain and so would report hits on a broken mirror too. |
| 8 | **`codex-mock` prerequisite checked late** | It was checked *after* turn 1 had provisioned a workspace, run a full agent turn and spawned two sub-agents — all discarded for a verdict knowable in the first second, and leaving a half-finished conversation in the workspace every time. Gated up front; the now-unreachable late SKIP branch is gone. |

**#7 was verified rather than assumed.** The Anthropic mock also emits a `signature_delta` (base64 SHA-256 of the thinking text) simulating encrypted reasoning, so plaintext lorem reaching the transcript was *plausible but unproven* — asserting it could have turned a green E2E red. Run result:

```
reasoningRows: 3
reasoningRowsCarryingContent: 2      ← 0 here would now fail the step
payloadLengths: [382, 331, 478]
```

## Two I am handling differently — deliberately, with reasons

**Artifact accumulation / no teardown.** The finding is right about the *harm* and I fixed that, but a teardown would not have worked: **transcripts are deliberately retained on conversation delete** — a locked product decision for this feature, not an oversight. Deleting the conversation leaves the file. The real damage was that the script got *less trustworthy the longer the workspace lived*, via the 500-row cap, and that is exactly what #5 above closes: it now says `INCONCLUSIVE_LISTING_TRUNCATED` instead of accusing a healthy mirror. I would rather the script name the condition honestly than delete evidence to stay under a limit.

**Manual-only, no CI coverage.** Pushing back on this one. The deterministic seam is already covered in CI by `WorkspaceTranscriptMirrorAttachCompositionTests`. What this `.mjs` adds is the thing that suite **structurally cannot** give: a real provider switch on a real conversation writing through a real gateway — which is precisely what P1 broke and what the unit tests could not see. That is the top of the pyramid, and the sample's own guidance puts it there: assert deterministic browser-observable state in scripts; exact HTTP codes and timing-sensitive races belong in the C# suite. Building a parallel C# version of this would either mock away the gateway (removing the only thing it proves) or make CI depend on a live sandbox. If you think the balance is wrong I will file it as a follow-up rather than bolt it on here.

gautam-achieveai:
[Revobot]

# Re-Review Summary: PR #263

**Head:** `4367d0ef06dca49744c3f5ebd012347dbf38d0ba`  
**Verdict:** **Request changes**

## Findings

### [BLOCKER] HIGH — Raw diagnostic errors can disclose sensitive data

**File:** `samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs:146`  
**Related:** `:226`, with output consumed at `:636–641` and `:700–705`

The new timeout/error handling returns raw exception text through:

```js
text: aborted ? `request aborted after ${timeoutMs}ms` :
  `fetch failed: ${String(e && e.message ? e.message : e)}`
```

and forwards raw run-state response content through `rs.text`. These values become part of the Playwright diagnostic result and may be displayed or persisted in CI artifacts, logs, or reports. Gateway/proxy responses and browser exceptions can contain URLs, stack traces, credentials, or user-derived data.

**Suggested fix:** Return only allowlisted structured metadata, such as:

- error name
- HTTP status
- timeout flag
- response content type
- body-present and body-length indicators
- sanitized origin, if needed

Do not return raw exception messages or response bodies. Add regression coverage asserting that both exception and HTTP-error paths exclude raw text.

## Review notes

- The prior findings from the previous review were not re-posted; the current delta addresses them.
- Both dispatched sub-agents completed. They independently identified the same redaction issue; the grader confirmed **HIGH** severity and a blocking classification, without escalation to Critical.
- Node syntax validation passed. .NET tooling was unavailable for targeted test execution.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-94a32e4eddb0462ab5462995a5a0386c&focus=1

samples/LmStreaming.Sample/playwright-scripts/transcript-mirror.mjs
```diff
@@ -0,0 +1,855 @@
...
+      const dl = await downloadWorkspaceFile(threadId, `${agentsDir}/${f.name}`);
+      const text = (dl.ok && dl.text) || '';
+      agentTexts[f.name] = text.length;
+      // A failed download yields '' and would quietly drag the marker checks below to false, blaming
+      // the mirror for a reader problem. Name it instead.
+      if (!dl.ok) agentDownloadFailures.push({ file: f.name, status: dl.status });
+      agentsBlob += `${text}\n`;
+    }
+    const missingLeaves = EXPECTED_AGENTS.filter((n) => !agentFiles.some(leafFor(n)));
+    const alphaPresent = agentsBlob.includes(ALPHA_TEXT);
```
gautam-achieveai, line 607: [Gautam's review bot] **[MEDIUM] Verify each marker in its corresponding child transcript**

The filename checks and aggregate marker checks are still independent. `agentsBlob` combines every child file, so swapped contents (`BETA_TEXT` in `mirror-alpha-*`, `ALPHA_TEXT` in `mirror-beta-*`) satisfy `missingLeaves.length === 0`, `alphaPresent`, and `betaPresent`; one file containing both markers while the other is empty also passes. I reproduced the current predicate with swapped contents and it returns `true`.

Keep the downloaded text keyed by filename, select the matching `mirror-alpha-*` / `mirror-beta-*` leaf, and require each marker in that file. This completes the prior requirement to verify content in the corresponding transcript rather than merely somewhere under `_agents/`.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `42a3771d...4367d0ef` found one remaining Medium test-oracle gap, posted inline. The revised head otherwise addresses the earlier run-state, drain, marker-continuity, listing-cap, timeout, reasoning-payload, and prerequisite findings.

I do not confirm the new HIGH diagnostic-disclosure objection as blocking: this manually invoked verifier returns same-origin diagnostics to the already privileged operator, and no new trust-boundary or third-party disclosure path was identified.

Issue #237, opened 2026-08-01
P1: Identity, principals and authorization — federated + native login


gautam-achieveai:
> Part of #235. Runs **parallel to** #236 (different layer, no shared code). Blocks P2/P3/P4.
> Spec: `docs/specs/P1-identity-authorization.md`

## Problem

Authorization today is **per-app only**. `app.id` is authenticated (ADR 0029) and bound from the credential at create, `OwnedSession` gives uniform 404 cross-app, and per-app workspace rooting isolates filesystems. That is a genuine tenancy boundary and it works.

What does not exist: **an end-user principal**. Inside one tenant, two humans are indistinguishable. This blocks:

- "Which human claimed this assist?" (P2 requires exactly one)
- "May *this* human take over *this* sandbox?" (P3)
- Per-user audit — every action currently attributes to the app
- Per-user cost attribution
- `MultiTurnAgentPool` freezes a conversation to its creating `AppId`, so a second human in the same tenant is indistinguishable from the first

## Decision: BOTH federated and native

Federated is primary. Native login exists because the Developer Agent's defining moment is a human clicking an assist link from Slack **hours later** with no live client-service session — federation cannot serve that, there is no session to exchange.

**Both paths converge on one `EndUserSessionToken`** so nothing downstream bifurcates.

## Principal model

Four kinds — `App`, `EndUser`, `Agent`, `Service` — in one envelope with `Actor` and `OnBehalfOf`.

**Composition rule (normative): effective permissions are the INTERSECTION of `Actor` and `OnBehalfOf`, never the union.** An app acting for a user can never exceed that user; a user acting through an app can never exceed the app. A union would open a confused-deputy hole.

**Agent principals are real.** An agent acting autonomously is `Kind=Agent` with `OnBehalfOf` set to the initiating human, so "the agent did X on behalf of Alice" is expressible in audit without conflating it with "Alice did X" — a distinction that matters the first time an agent action is disputed.

## Federation

**Recommended: OAuth 2.0 Token Exchange (RFC 8693).** Rejected: bespoke signed assertions (reinvents JWT validation and key rotation badly); mTLS-only (identifies the service, not the user); passing raw client IdP tokens through (makes us a confused deputy holding foreign credentials).

**Layers on the existing app credential; does not replace it.** The app bearer still authenticates the calling service and still binds `app.id`. The exchanged token adds the *user* dimension. Requests carry both.

[...]

PR #266, opened 2026-08-05
docs: identify ADRs by slug, and record the missing signature-derived principal

gautam-achieveai:
Follow-up documentation from the **#227 / #231** lifecycle-and-approval work. Docs and XML comments only — no behaviour change. `LmAgentInfra` builds clean, **0 warnings, 0 errors**.

> **Revised after review.** An earlier version of this description claimed a subscriber "already proves who it is on every delivery it answers." That is false and has been corrected in both the code comment and this text — see [§3](#3-correction).

## 1. ADR 0010 — an ADR's identity is its slug

This directory contains **two ADRs numbered 0002 and two numbered 0003**. Both of the newer pair came from #231:

| Number | Records sharing it |
|---|---|
| `0002` | `0002-workflow-controller-transparency` (2026-07-23) · `0002-lifecycle-event-wire-contract` (2026-07-27) |
| `0003` | `0003-per-conversation-usage-collector` (2026-07-23) · `0003-fail-closed-tool-approval` (2026-07-27) |

Two work streams were in flight the same week, each drafting against the state of `docs/adrs/` visible from its own branch, where the highest merged number was `0001`. Nothing caught it at merge, because **a duplicate number is a duplicate filename *prefix*, not a duplicate filename** — git merges both without conflict.

**The damage is narrower than it looks.** Every in-file cross reference already links by filename, so every link in the repo resolves to exactly one record today. What is ambiguous is the *bare number in prose*, and the README index — which listed all four entries flat, reading as though 0002 and 0003 each decided two unrelated things.

### The decision

- **An ADR is identified by its slug.** A number is allocation order within the authoring branch — a sort key and a filename prefix, never a guaranteed-unique identifier.
- **The existing collisions stand.** Renumbering would rewrite four *accepted* records to fix a defect in a fifth file that is not a record at all, and would break PR #231's "ADRs 0002–0008" index.
- **The README index is corrected**, because it is a navigation aid describing current contents, not an immutable record.
- **A number is claimed at merge, not at drafting.** Whoever merges second checks the target branch and renames the file — a draft is not yet a record, so this is the one moment the correction is free.

### Why this needs to be committed

These files are not a speculative draft. They deliver decisions already taken on 2026-08-03 during the lifecycle-hooks survey — *"fix the collisions, but preserve ADR immutability"* and *"note the auth gap in a code comment, fix later."* That survey lives under `scratchPad/`, which is **gitignored** (`.gitignore:61`), so it never leaves one machine. Without this PR the decisions have no durable form anywhere in the repository, and the next person to hit the collision re-litigates it from scratch.

## 2. A known gap in the approval endpoint

`LifecycleApprovalController` refuses any request without a host-wired authentication scheme. That refusal is **correct** — a controller cannot tell a verified header from a merely typed one. But the burden is larger than the problem it solves for the common case: an integrator who wants only *"the holder of subscription X may decide X's approvals"* must stand up an entire separate scheme to say it.

The comment now records this as a **latent capability with three unmet prerequisites**, not as a settled fix:

1. **No subscriber-to-host signing convention exists.** The secret travels one way today.
2. **`ILifecycleSubscriptionRegistry` deliberately forbids resolving a subscription by id alone** — every lookup is owner-scoped, which *is* the authorization model.
3. **The HMAC binds only `{timestamp}.{deliveryId}.{body}`** — no direction, no route domain — so reusing it for caller authentication would let a captured outbound signature be replayed inbound.

That spans the registry and the wire format, so it is flagged as belonging in a **superseding ADR**. Registration is excluded on principle: it is the operation that *mints* the secret, so host authentication stays mandatory there regardless.

**No issue is filed for the handler** — related to #237 (*P1: Identity, principals and authorization*), but narrower.

## 3. Correction

The original commit asserted that a subscriber "already proves who it is on every delivery it answers." Verified false in review:

- `WebhookRequestSigner` has exactly one caller — `HttpLifecycleDeliverySender:105`, the **host**, signing **outbound**.
- `WebhookRequestVerifier`'s only consumer is `WebhookVerificationMiddleware`.
- **No subscriber signs anything, and no lifecycle route verifies a subscriber-produced signature.**

The underlying fact — the secret is host-minted and bound to one subscription and one owner — is true, but identity-proof does not follow from it. Direction was never checked. Corrected in `c774ec24`, along with four other review findings (all five valid, all confirmed against the code).

## Verification

| Claim | Result |
|---|---|
| The 0002/0003 collision is real and still present | both pairs on disk **and** on `origin/main` |
| Secret is host-minted, bound to one subscription + owner | `InMemoryLifecycleSubscriptionRegistry:143` → `LifecycleSubscription` carrying `SubscriptionId` + `Owner` |
| Subscriber does **not** sign inbound | `WebhookRequestSigner` single call site is host-outbound |
| Registry forbids id-alone lookup | every method owner-scoped; `ILifecycleSubscriptionRegistry:45` |
| README example filename is not a real number | `NNNN-`, after `0011` was found already taken |
| #227 carries no ADR reference | grep of body + comments returns nothing; citation dropped |
| Comments compile | `dotnet build` — 0 warnings, 0 errors |

Refs #227, #231.

### DIFF
--- a/docs/adrs/0010-adr-identity-is-the-slug.md
+++ b/docs/adrs/0010-adr-identity-is-the-slug.md
@@ -1,0 +1,80 @@
+# ADR 0010: An ADR's identity is its slug, and colliding numbers are left standing
+
+* Status: Accepted
+* Date: 2026-08-03
+* Related issues, PRs, or commits: [#227](https://github.com/achieveai/LmDotnetTools/issues/227),
+  [#231](https://github.com/achieveai/LmDotnetTools/pull/231); corrects the index for
+  [0002 — workflow controller transparency](0002-workflow-controller-transparency.md),
+  [0003 — per-conversation usage collector](0003-per-conversation-usage-collector.md),
+  [0002 — lifecycle event wire contract](0002-lifecycle-event-wire-contract.md),
+  [0003 — fail-closed tool approval](0003-fail-closed-tool-approval.md)
+
+## Context
+
+This directory contains two ADRs numbered 0002 and two numbered 0003.
+
+The cause is ordinary and will recur. Two work streams were in flight across the same
+week. The workflow-transparency and usage-collector records were written on 2026-07-23;
+the #227 lifecycle, approval, and delivery set was written on 2026-07-27. Each was drafted
+against the state of `docs/adrs/` visible from its own branch, where the highest merged
+number was 0001, and each therefore took 0002 and 0003 as the next free numbers. Nothing
+detected the collision at merge, because a duplicate number is a duplicate *filename
+prefix*, not a duplicate filename — git merges both without conflict.
+
+The damage is confined to one place, which is worth stating precisely before choosing a
+remedy. In-file cross references already link by filename — `[ADR 0002](0002-lifecycle-event-wire-contract.md)`
+in ADR 0005, `[Fail-closed tool approval](adrs/0003-fail-closed-tool-approval.md)` in the
+field matrix — so every link in the repository resolves to exactly one record today. What
+is ambiguous is the *bare number* in prose, and the README index, which lists all four
+entries and so reads as though 0002 and 0003 each decided two unrelated things.
+
+The obvious remedy is to renumber the newer set to 0010 and up. It is not available.
+This directory's own rule is that ADRs are append-only and that a later decision supersedes
+an earlier one rather than editing it, and a record's number is part of the record — it is
+in the filename, in the `# ADR NNNN` heading, and in every link and citation already made
+to it. Renumbering would rewrite four accepted records to fix a defect in a fifth file that
+is not a record at all. It would also break references that live outside this repository's
+control, in issue #227 and PR #231.
+
+## Decision
+
+**An ADR is identified by its slug, not by its number.** The canonical reference is the
+full filename — `0002-lifecycle-event-wire-contract` — and every cross reference, in prose
+or in code comments, links or names that slug. A number is allocation order within the
+branch that authored the record. It is a sort key and a filename prefix. It is not, and
+never was, a guaranteed-unique identifier.
+
+**The existing collisions stand.** 0002 and 0003 each name two records permanently. Bare
+"ADR 0002" and "ADR 0003" in already-merged prose are left as they are, because
+disambiguating them means editing accepted records to serve a convenience, which is the
+precise trade this directory has already refused.
+
+**A number is claimed at merge, not at drafting.** A branch drafts against whatever number
+looks free; whoever merges second checks `docs/adrs/` as it stands on the target branch and
+renames the file before merging if the number has since been taken. Renaming a draft is not
+an edit to a record — a draft is not yet a record — so this is the one moment at which the
+correction is free, and it is the only moment at which it is available.
+
+**The README index is corrected, because it is not a record.** It is a navigation aid
+describing the directory's current contents, and a navigation aid that misdescribes them is
+simply wrong rather than historically interesting. Colliding entries are listed with their
+slugs so the index distinguishes what the numbers cannot.
+
+## Consequences
+
+Two numbers in this directory are ambiguous forever, and a reader who cites one without a
+slug will be misunderstood roughly half the time. That is the accepted cost of the
+append-only rule, and it is cheaper than the alternative: a directory where numbers are
+authoritative is a directory where numbers get rewritten, and a record whose identity can
+be rewritten is not immutable in any sense that matters.
+
+Prevention is procedural rather than enforced. It relies on whoever merges an ADR looking
+at the directory first, which is one `ls` and is exactly the check that was skipped both
+times. A future collision is therefore possible; it is also harmless under this decision,
+because the slug still resolves it. If collisions become frequent enough to be irritating
+rather than rare enough to be tolerable, the answer is a merge check that fails on a
+duplicate prefix — not a renumbering pass.
+
+Finally, this ADR supersedes nothing. The four colliding records remain accepted, correct,
+and in force exactly as written. What changes is only how they are referred to and how the
+index presents them.
--- a/docs/adrs/0010-adr-identity-is-the-slug.md
+++ b/docs/adrs/0010-adr-identity-is-the-slug.md
@@ -33,8 +33,8 @@ This directory's own rule is that ADRs are append-only and that a later decision
 an earlier one rather than editing it, and a record's number is part of the record — it is
 in the filename, in the `# ADR NNNN` heading, and in every link and citation already made
 to it. Renumbering would rewrite four accepted records to fix a defect in a fifth file that
-is not a record at all. It would also break references that live outside this repository's
-control, in issue #227 and PR #231.
+is not a record at all. It would also break a reference that lives outside this repository's
+control, in PR #231, which indexes the set as "ADRs 0002–0008".
 
 ## Decision
 
--- a/docs/adrs/README.md
+++ b/docs/adrs/README.md
@@ -33,24 +33,43 @@ The architectural choice that was made.
 What improved, what became more complex, and what future work this implies.
 ```
 
+## Referring to an ADR
+
+Refer to an ADR by its **slug** — `0002-lifecycle-event-wire-contract` — never by its number
+alone. Numbers are allocation order within the branch that authored the record, not unique
+identifiers: `0002` and `0003` each name two different records here. See
+[ADR 0010](0010-adr-identity-is-the-slug.md) for why the collision was left standing rather
+than renumbered away.
+
 ## Creating future ADRs
 
 1. Copy [templates/adr-template.md](templates/adr-template.md) into this directory.
 2. Name it with the next four-digit number and a short kebab-case title, e.g.
-   `0002-use-example-backend.md`.
+   `0011-use-example-backend.md`.
 3. Open it as `Proposed`, and move it to `Accepted` once the decision is made.
+4. **Before merging, re-check this directory on the target branch.** A number that was free
+   when you drafted may have been taken by another branch since. Rename the file then —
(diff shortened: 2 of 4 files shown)


src/LmAgentInfra/Lifecycle/LifecycleApprovalController.cs
```diff
@@ -54,6 +54,22 @@ public sealed record ToolApprovalDecisionResponse
 /// misconfiguration rather than routine traffic.
 /// </para>
 /// <para>
+/// <b>Known gap, accepted for now: there is no supported way to turn a verified signature into a
+/// principal.</b> A subscriber already proves who it is on every delivery it answers — it holds a
```
gautam-achieveai: [Gautam's review bot] [P1] This describes an inbound authentication proof that does not exist. The subscription secret is currently used only by the host to sign outbound deliveries (`HttpLifecycleDeliverySender`); subscribers do not sign approval requests, and no lifecycle route verifies such a signature. Please rewrite this as a possible capability (the subscriber holds a secret that *could* authenticate an inbound protocol) and explicitly state that the subscriber-to-host signing convention still has to be defined.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found five actionable documentation issues: the proposed signature-derived identity path misstates the current protocol and omits security/registry prerequisites, the registration docs promise an authentication path that cannot bootstrap, and two ADR statements are factually incorrect.

src/LmAgentInfra/Lifecycle/LifecycleApprovalController.cs
```diff
@@ -54,6 +54,22 @@ public sealed record ToolApprovalDecisionResponse
...
+/// host-minted signing secret bound to one subscription and one owner — and yet that proof cannot be
+/// spent here, so an integrator who wants nothing more than "the holder of subscription X may decide
+/// X's approvals" must stand up a separate authentication scheme to express it. The refusal is
+/// correct (see the paragraph above: a controller cannot tell a verified header from a typed one, so
+/// trusting one would be trusting every deployment to have wired verification), but the burden it
+/// creates is larger than the problem it solves for the common case. The intended fix is a
+/// first-party <c>AuthenticationHandler</c> shipped alongside
+/// <see cref="Webhooks.WebhookRequestVerifier"/> that verifies the signature and, on success, emits a
```
gautam-achieveai: [Gautam's review bot] [P1] The intended fix is not implementable as “add an `AuthenticationHandler`” against the current contracts. Verification needs a way to resolve subscription → secret/owner before a principal exists, while `ILifecycleSubscriptionRegistry` deliberately forbids lookup by subscription id alone. The existing HMAC also binds only `{timestamp}.{deliveryId}.{body}`, with no inbound direction or route domain, so reusing it for caller authentication would allow cross-protocol signature reuse unless the format changes. Please record these prerequisites (or move this architectural change into a superseding ADR) rather than documenting a handler as the complete fix.

src/LmAgentInfra/Lifecycle/LifecycleSubscriptionsController.cs
```diff
@@ -92,7 +92,9 @@ public sealed record LifecycleSubscriptionResponse
 /// <see cref="HttpContext.User"/> — established by whatever authentication the host wired in front of
 /// it — and resolves an owner from that alone. No route, header, or body field names an owner, so a
 /// caller cannot register into, rotate, or revoke another tenant's subscriptions. A host that wires no
-/// authentication scheme gets an endpoint that refuses everything, which is the safe direction.
+/// authentication scheme gets an endpoint that refuses everything, which is the safe direction — and,
+/// as the known gap recorded on <see cref="LifecycleApprovalController"/> notes, a heavier burden on
+/// the integrator than it should be until a signature-derived principal is shipped.
```
gautam-achieveai: [Gautam's review bot] [P1] This promise cannot apply to the registration route covered by this controller: registration is the operation that mints and returns the subscription secret, so the caller cannot derive a principal from that secret beforehand. Host authentication remains mandatory for registration even if existing-subscription operations later gain signature authentication. Please scope this note to rotate/revoke (and any other existing-subscription routes) or state the registration exception explicitly.

docs/adrs/README.md
```diff
@@ -33,24 +33,43 @@ The architectural choice that was made.
...
+identifiers: `0002` and `0003` each name two different records here. See
+[ADR 0010](0010-adr-identity-is-the-slug.md) for why the collision was left standing rather
+than renumbered away.
+
 ## Creating future ADRs
 
 1. Copy [templates/adr-template.md](templates/adr-template.md) into this directory.
 2. Name it with the next four-digit number and a short kebab-case title, e.g.
-   `0002-use-example-backend.md`.
+   `0011-use-example-backend.md`.
```
gautam-achieveai: [Gautam's review bot] [P2] `0011` is already occupied by `0011-workspace-transcript-files.md`, so the new example immediately demonstrates the collision this section is meant to prevent. Use a stable placeholder such as `NNNN-use-example-backend.md` instead of a concrete number that is already taken and will become stale again.

docs/adrs/0010-adr-identity-is-the-slug.md
```diff
@@ -0,0 +1,80 @@
...
+is ambiguous is the *bare number* in prose, and the README index, which lists all four
+entries and so reads as though 0002 and 0003 each decided two unrelated things.
+
+The obvious remedy is to renumber the newer set to 0010 and up. It is not available.
+This directory's own rule is that ADRs are append-only and that a later decision supersedes
+an earlier one rather than editing it, and a record's number is part of the record — it is
+in the filename, in the `# ADR NNNN` heading, and in every link and citation already made
+to it. Renumbering would rewrite four accepted records to fix a defect in a fifth file that
+is not a record at all. It would also break references that live outside this repository's
+control, in issue #227 and PR #231.
```
gautam-achieveai: [Gautam's review bot] [P2] Issue #227 contains no ADR-number or ADR-filename reference in its body or comments; only the implementation PR mentions “ADRs 0002–0008”. Because this record becomes append-only once accepted, please remove #227 from this justification or replace it with an external reference that actually exists.

gautam-achieveai:
All five findings verified against the code and fixed in `c774ec24`. Every one was correct.

| # | Finding | Verification | Fix |
|---|---|---|---|
| P1 | Subscriber does not prove identity inbound | `WebhookRequestSigner` has exactly one caller — `HttpLifecycleDeliverySender:105`, the **host** signing **outbound**. `WebhookRequestVerifier`'s only consumer is `WebhookVerificationMiddleware`. No lifecycle route verifies a subscriber-produced signature. | Rewritten as a *latent capability*; states outright that no subscriber-to-host signing convention exists |
| P1 | `AuthenticationHandler` not implementable as described | `ILifecycleSubscriptionRegistry` is owner-scoped on every method (`Rotate`, `RevokePreviousKey`, `Unregister`, `TryGet`, `ForOwner`) — line 45: *"exist **or** belongs to another owner — the two are indistinguishable on purpose"*. HMAC binds `{timestamp}.{deliveryId}.{body}` only (`WebhookSigningSecret:86`) — no direction, no route domain | All three prerequisites now enumerated, and the change is flagged as belonging in a **superseding ADR**, not presented as a settled fix |
| P1 | Registration cannot bootstrap from the secret it mints | `InMemoryLifecycleSubscriptionRegistry:143` mints the secret during registration | Scoped to existing-subscription routes; registration exception stated explicitly |
| P2 | `0011` already taken | `0011-workspace-transcript-files.md` exists — the collision-prevention section was demonstrating a collision | Now `NNNN-use-example-backend.md` |
| P2 | #227 carries no ADR reference | Grep of #227 body + comments for `adr\|0002\|0003` returns nothing; only PR #231 says *"ADRs 0002–0008"* | #227 dropped from that justification; #231 kept and quoted |

Thanks — the first one is the one that mattered. I had "verified" that the secret is host-minted and bound to a subscription and owner, which is true, and then asserted from it that the subscriber *proves* something. I never checked which direction the signature travels. Container verified, contents asserted.

The last fix is also time-sensitive by this ADR's own rule: once 0010 is merged it becomes uneditable, so a wrong external citation inside it would have been permanent.

docs/adrs/README.md
gautam-achieveai: [Gautam's review bot] Addressed in `c774ec24`: the example now uses the stable `NNNN-use-example-backend.md` placeholder.

docs/adrs/0010-adr-identity-is-the-slug.md
gautam-achieveai: [Gautam's review bot] Addressed in `c774ec24`: the renumbering justification now cites only PR #231 for the external ADR-number reference; #227 remains only as accurate related context.

src/LmAgentInfra/Lifecycle/LifecycleSubscriptionsController.cs
gautam-achieveai: [Gautam's review bot] Addressed in `c774ec24`: the note now scopes signature-derived identity to existing-subscription routes and explicitly preserves host authentication for registration.

src/LmAgentInfra/Lifecycle/LifecycleApprovalController.cs
gautam-achieveai: [Gautam's review bot] Addressed in `c774ec24`: the documentation now records owner-scoped registry resolution and signature domain separation as prerequisites, and leaves the design to a superseding ADR.

gautam-achieveai (commented):
[Gautam's review bot] Incremental deep review of `c774ec24` found all five prior findings fully addressed. The corrected wording now matches the implemented signature direction, registry authorization boundary, HMAC binding, registration bootstrap constraint, ADR numbering state, and external references. No new actionable issues found.

PR #267, opened 2026-08-05
Add LmStreaming output-token policy and Copilot web search

gautam-achieveai:
## Summary

- add configurable output-token defaults for primary and delegated LmStreaming agents
- import Copilot's hosted `web_search` MCP tool while preserving Jina fallback composition
- apply delegated token policy to discovered, workflow, workspace, and nested sub-agent paths
- retain transcript-mirror resource ownership and cleanup from current `main`

## Key behavior

- primary/default work uses 24,576 output tokens unless the caller supplied a value
- delegated/fallback work uses 16,384 output tokens unless the caller supplied a value
- Copilot-backed conversations can register the hosted `web_search` tool from `/mcp/readonly`
- web tool policy prevents duplicate hosted/fallback search registration
- context-discovered templates inherit the delegated output-token policy

## Verification

- clean solution build: no warnings or errors
- LmStreaming.Sample.Tests: 1,402 passed, 1 platform skip
- CopilotAnthropicProxy.Tests: 365/365 passed
- transcript mirror / descendant scanner regression: 40/40 passed
- CopilotMcpLiveTests: 1/1 passed
- formatting pre-commit hook: passed
- `git diff --check`: passed

## Notes

- `.claude/scratchpad/**` is intentionally excluded
- the branch is based on current `origin/main` (`f518182c`)

### DIFF
--- a/samples/LmStreaming.Sample/Program.cs
+++ b/samples/LmStreaming.Sample/Program.cs
@@ -135,6 +135,13 @@ try
         options.WebSocketPath = "/ws";
         options.WriteIndentedJson = builder.Environment.IsDevelopment();
     });
+    _ = builder.Services
+        .AddOptions<AgentOutputTokenOptions>()
+        .Bind(builder.Configuration.GetSection(AgentOutputTokenOptions.SectionName))
+        .Validate(options => options.Validate().Succeeded, "Invalid AgentOutputTokens configuration.")
+        .ValidateOnStart();
+    _ = builder.Services.AddSingleton(sp =>
+        new AgentOutputTokenPolicy(sp.GetRequiredService<Microsoft.Extensions.Options.IOptions<AgentOutputTokenOptions>>().Value));
 
     // Service-to-service lifecycle observation and remote tool approval (ADR 0003 + ADR 0005). Both
     // are off unless the matching `Lifecycle:*:Enabled` flag is set, and this file carries no default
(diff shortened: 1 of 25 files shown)


samples/LmStreaming.Sample/Program.cs
```diff
@@ -3325,6 +3440,14 @@ out var parsedModelProbe
             : string.IsNullOrWhiteSpace(workingDirectory) ? null
             : workingDirectory;
 
+        IReadOnlyList<string>? enabledTools = mode.EnabledTools;
```
gautam-achieveai: [Gautam's review bot] The registration path expands `enabledBuiltInTools: [web_search]` through `ResolveEnabledTools`, so it registers the dynamic `WebSearch` and `WebFetch` contracts. This loop recomputes a narrower allow-list from only `mode.EnabledTools + mode.EnabledBuiltInTools`. For `workspace-agent`, that becomes just `web_search`; `CopilotToolPolicyEngine` then filters `WebSearch`/`WebFetch` out of `CopilotDynamicToolBridge`. As a result, `WebFetch` is unreachable even when hosted search succeeds, and if hosted MCP initialization fails the advertised Jina `WebSearch` fallback is also unreachable. Please pass the already-resolved list into this method (or call `ResolveEnabledTools` here) and cover the workspace-agent Copilot policy/bridge composition in a regression test.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found one blocking functional gap in the plain Copilot CLI web-tool composition.

gautam-achieveai:
[Revobot]

## Review — REQUEST CHANGES

### [BLOCKER] Plain Copilot CLI path drops the Jina tool allow-list

- **File:** `samples/LmStreaming.Sample/Program.cs:3443-3448`
- **Issue:** The plain `copilot` path constructs the CLI `enabledTools` list by unioning only `mode.EnabledTools` and `mode.EnabledBuiltInTools`. However, the web-tool composition path expands the built-in `web_search` declaration into the dynamic names `web_search`, `WebSearch`, and `WebFetch` via `WebToolRegistrationPolicy.ResolveEnabledTools`.

  As a result, for a mode that enables the built-in `web_search` capability, the registry can contain hosted `web_search` plus Jina `WebFetch`/`WebSearch`, but `CopilotToolPolicyEngine` receives only `web_search`. It will reject the Jina dynamic tools because their names are absent from the allow-list.

  This breaks the required behavior in both cases:

  - hosted search succeeds: Jina `WebFetch` is registered but cannot be called;
  - hosted search fails: Jina `WebSearch` and `WebFetch` are registered as fallback tools but cannot be called.

- **Suggested fix:** Use the same resolved web-tool list when calling `CreateCopilotAgentLoop`, or otherwise add the expanded `WebSearch` and `WebFetch` names to the CLI allow-list whenever the mode enables built-in `web_search`. Add a regression test that exercises the plain Copilot path and verifies that the effective dynamic-tool policy allows `web_search`, `WebSearch`, and `WebFetch`.

No sub-agent findings were available for this review. The focused build/tests could not be run in this environment because the `dotnet` executable was unavailable.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-20448c61cefd416ea8dda4b6b6eb4711&focus=1

samples/LmStreaming.Sample/Program.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in 9d70a1b4. The plain Copilot CLI path now uses WebToolRegistrationPolicy.ResolveEnabledTools, so Workspace Agent's built-in web_search intent expands to the model-facing WebSearch and WebFetch allow-list. The hosted MCP wire tool remains lowercase internally, is registered as WebSearch, and the new WorkspaceAgentMode regression asserts WebSearch + WebFetch are present and lowercase web_search is absent. Full LmStreaming.Sample.Tests passed 1,403 + 1 platform skip before push; fresh CI is running.

samples/LmStreaming.Sample/Services/CopilotWebSearchRegistration.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                Name = WebSearchTool.ToolName,
+                Description = descriptor.Contract.Description,
+                Namespace = descriptor.Contract.Namespace,
+                ClassName = descriptor.Contract.ClassName,
+                Parameters = descriptor.Contract.Parameters,
+                ReturnType = descriptor.Contract.ReturnType,
+                ReturnDescription = descriptor.Contract.ReturnDescription,
+            };
+            _ = registry.AddFunction(renamedContract, descriptor.Handler, providerName: ClientName);
```
gautam-achieveai, line 106: [Gautam's review bot] This registration runs after the conversation registry has already been populated, but unlike `WebToolRegistrationPolicy.Apply` it does not check for an existing `WebSearch`. If one exists, `FunctionRegistry.Build()` prefixes both colliding contracts (for example `SampleTools-WebSearch` and `copilot-web-search-WebSearch`). The Copilot policy allow-list contains only bare `WebSearch`, so neither prefixed contract is advertised; because this method still returns `Registered = true`, the Jina search fallback is suppressed too. Please perform the same case-insensitive collision preflight used by the Jina policy, preserve the existing tool, dispose the hosted resource, and add a regression test with a pre-registered `WebSearch`.

PR #268, opened 2026-08-05
feat(sample): professional markdown rendering in the chat client

gautam-achieveai:
## What

Markdown in the LmStreaming.Sample chat client rendered as an undifferentiated wall of text. This makes an assistant reply read like a document.

**The parser was never the problem** — all 18 GFM construct types rendered correctly from the start (h1, h2, h3, ul, ol, table, blockquote, pre, inline code, links, hr, task lists, strong, em…). The failure was purely styling.

`assets/styles.css` opens with a universal reset:

```css
*, *::before, *::after { margin: 0; padding: 0; box-sizing: border-box; }
```

Only `p`, `pre` and `code` had their spacing restored (in `TextMessage.vue`'s scoped `:deep()` rules). Everything else — headings, lists, tables, blockquotes, `hr` — was left flattened. Compounding it, `TextMessage.vue` used `.markdown-content` while `PendingMessage.vue` used `.markdown-body`, each with its own partial duplicate of the rules, which is how the two surfaces drifted apart.

## Changes

- **New shared `ClientApp/src/assets/markdown.css`**, imported from `main.ts`. Both components now render into `.markdown-content`; their scoped blocks keep only bubble chrome. Scoped `:deep()` is what *forced* the duplication in the first place.
- **Syntax highlighting** — `marked-highlight` + `highlight.js/lib/core` with 18 explicitly registered grammars. marked v17 dropped its built-in `highlight` option, so `marked-highlight` is the supported route; the core build avoids pulling ~190 unused grammars. Unknown fences fall back to `plaintext` with `ignoreIllegals`, because a model can label a fence with anything.
- **Tables** — outer border + radius, header fill, column dividers, zebra rows, `tabular-nums`.
- **Typography** — heading scale with rules under h1/h2, list indents, blockquote bar, task-list checkboxes without a stray bullet, `ui-monospace` font stack for code.

### Measured before → after

| Element | Before | After |
|---|---|---|
| heading scale | all three at body size, `margin: 0` | h1 25.6 / h2 20.8 / h3 18 / body 16px, rules under h1+h2 |
| `ul`/`ol` | `padding-left: 0px` | `24px` |
| table | no cell borders, no header fill | outer border + 8px radius, dividers, zebra, `tabular-nums` |
| column alignment | n/a | `["left","right","right","center"]` |
| syntax highlighting | 0 token spans (monochrome) | `hljs language-csharp`, 7 spans, distinct colours |
| blockquote | `border-left: 0` — rendered as a plain paragraph | 3px bar, padding, tinted background |
| `pre code` | chip background + padding (a chip inside a box) | transparent, `ui-monospace` |
| long code lines | **silently clipped** — 409/409, no scrollbar | 413/402 and 490/402, scrollbar 8px, scrolls |

## Two rendering bugs fixed along the way

**1. GFM column alignment was silently flattened.** My first pass shipped a blanket `text-align: left` on `th, td`. marked emits the *legacy presentational attribute* `<th align="right">` for `| ---: |`, not an inline style — and a presentational attribute has effectively zero specificity, so it loses to any CSS rule. One blanket `text-align` overrode every aligned column in the app. Aligned cells now opt back in via `[align='right'|'center'|'left']` selectors.

Generalisable: **if you style a construct the renderer expresses via HTML attributes, a blanket CSS rule on that element is a silent override, not a default.**

**2. Long code lines were silently CLIPPED, not scrolled.** `pre code` was `display: inline`. An inline box contributes no intrinsic width to its scroll container, so `pre` never grew its scroll area to the long line — the browser reported `scrollWidth === clientWidth`, painted no scrollbar, and cut the line off. Separately, `::-webkit-scrollbar` had a `width` but no `height`, which collapses the *horizontal* bar once that pseudo-element is styled at all. Both were live; fixing either alone leaves the symptom.

## Verification

- `playwright-scripts/markdown-render-audit.mjs` — **23 assertions green** against the running app. Everything is measured (computed styles, DOM state), not asserted from reading source.
- Client unit tests **840/840** (66 files).
- `npm run build` + `dotnet build` — 0 warnings, 0 errors. Asset hashes changed, confirming the CSS actually shipped (a stale `dist/` served through Static Web Assets is the classic false-green here).

The clipping assertion is worth calling out. Its first formulation — *"at least one panel overflows, and no overflowing panel is clipped"* — **stayed green when I injected the defect back in-page**. The bug's whole signature is that it makes `overflows` false, so the predicate skipped exactly the panels it existed to test. Re-anchored on a quantity the bug cannot suppress (the content's own rendered width, since an inline box's client rect is the union of its line boxes), it re-proved RED and then caught a *second* clipped panel the first formulation never touched.

## Not done

**HTML sanitization.** `marked` output still goes straight into `v-html`; there is no DOMPurify. This is a real exposure the moment a model or tool result echoes HTML. It adds a dependency and was out of scope here, so the audit reports it in `notes` rather than faking a passing assertion — a dependency-gated gap should be visible, not silently green.

## Also included

- `playwright-scripts/gen-md-showcase-prompt.mjs` — generates the escaped mock instruction-chain prompt now documented in `PromptExamples.md`. Generated rather than hand-edited because hand-inlining it ate a TS template literal.
- `playwright-scripts/markdown-showcase-screenshots.mjs` — captures the running app at 2× device scale.
- `PromptExamples.md` gains a "Markdown Rendering Showcase" entry. Note the determinism lever: `text_message: {"length": N}` only emits lorem ipsum; only `{"text": "..."}` pins exact literal output, which is what makes a rendering test possible at all.

### DIFF
--- a/samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
@@ -1,4 +1,4 @@
-import { marked } from 'marked';
+import { Marked } from 'marked';
 import { markedHighlight } from 'marked-highlight';
 import hljs from 'highlight.js/lib/core';
 
@@ -54,33 +54,70 @@ for (const [name, definition] of Object.entries(LANGUAGES)) {
   hljs.registerLanguage(name, definition);
 }
 
-marked.use({
+const MARKED_OPTIONS = {
   breaks: true, // Enable GFM line breaks
   gfm: true     // Enable GFM
-});
+};
+
+const ESCAPE_REPLACEMENTS: Record<string, string> = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  "'": '&#39;',
+};
+
+/*
+ * marked-highlight treats whatever the highlight callback returns as HTML, so the
+ * no-highlight path has to escape the fence body itself -- returning raw code here would
+ * put attacker-controlled markup straight into the `v-html` binding.
+ */
+function escapeHtml(code: string): string {
+  return code.replace(/[&<>"']/g, (ch) => ESCAPE_REPLACEMENTS[ch]);
+}
 
 /*
  * marked dropped its built-in `highlight` option; `marked-highlight` is the supported
  * replacement. The `hljs ` prefix on the class is what the highlight.js theme selects on --
  * without it the emitted token spans are styled but the block itself is not.
+ *
+ * TWO instances rather than one instance with a mutable flag: highlighting is skipped while a
+ * message is still streaming (see below), and a module-level flag toggled around `parse()` is
+ * hidden state that any concurrent parse would read. Both instances share MARKED_OPTIONS and
+ * the same `hljs language-x` class list, so finalizing a message only paints token colours in
+ * -- the block keeps its panel styling throughout and nothing shifts.
  */
-marked.use(
-  markedHighlight({
-    langPrefix: 'hljs language-',
-    emptyLangClass: 'hljs',
-    highlight(code: string, lang: string): string {
-      // An unknown or absent language must degrade to escaped plain text, never throw:
-      // a model can label a fence with anything at all (```mermaid, ```text, ```).
-      const language = lang && hljs.getLanguage(lang) ? lang : 'plaintext';
-      return hljs.highlight(code, { language, ignoreIllegals: true }).value;
-    },
-  })
-);
+function createMarked(highlightCode: (code: string, lang: string) => string): Marked {
+  return new Marked(
+    MARKED_OPTIONS,
+    markedHighlight({
+      langPrefix: 'hljs language-',
+      emptyLangClass: 'hljs',
+      highlight: highlightCode,
+    })
+  );
+}
+
+const markedHighlighted = createMarked((code, lang) => {
+  // An unknown or absent language must degrade to escaped plain text, never throw:
+  // a model can label a fence with anything at all (```mermaid, ```text, ```).
+  const language = lang && hljs.getLanguage(lang) ? lang : 'plaintext';
+  return hljs.highlight(code, { language, ignoreIllegals: true }).value;
+});
+
+const markedPlain = createMarked((code) => escapeHtml(code));
 
 /**
- * Parse markdown text to HTML
+ * Parse markdown text to HTML.
+ *
+ * `highlight: false` skips syntax highlighting. Highlighting a fence is O(fence length), and a
+ * streaming message re-parses its whole accumulated text on every delta, so highlighting a
+ * growing fence per chunk is quadratic on the main thread (measured: 500 incremental parses of
+ * a 10 KB JS fence took 2.02 s highlighted vs 8.5 ms plain). Callers rendering a message that
+ * is still streaming should pass `false` and re-render highlighted once it completes.
  */
-export function parseMarkdown(text: string): string {
+export function parseMarkdown(text: string, options?: { highlight?: boolean }): string {
   if (!text) return '';
-  return marked.parse(text) as string;
+  const instance = options?.highlight === false ? markedPlain : markedHighlighted;
+  return instance.parse(text) as string;
 }
--- a/samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
+++ b/samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
@@ -1,10 +1,81 @@
 import { marked } from 'marked';
+import { markedHighlight } from 'marked-highlight';
+import hljs from 'highlight.js/lib/core';
 
-// Configure marked options
+import bash from 'highlight.js/lib/languages/bash';
+import csharp from 'highlight.js/lib/languages/csharp';
+import css from 'highlight.js/lib/languages/css';
+import diff from 'highlight.js/lib/languages/diff';
+import go from 'highlight.js/lib/languages/go';
+import ini from 'highlight.js/lib/languages/ini';
+import java from 'highlight.js/lib/languages/java';
+import javascript from 'highlight.js/lib/languages/javascript';
+import json from 'highlight.js/lib/languages/json';
+import markdown from 'highlight.js/lib/languages/markdown';
+import plaintext from 'highlight.js/lib/languages/plaintext';
+import powershell from 'highlight.js/lib/languages/powershell';
+import python from 'highlight.js/lib/languages/python';
+import rust from 'highlight.js/lib/languages/rust';
+import sql from 'highlight.js/lib/languages/sql';
+import typescript from 'highlight.js/lib/languages/typescript';
+import xml from 'highlight.js/lib/languages/xml';
+import yaml from 'highlight.js/lib/languages/yaml';
+
+/*
+ * highlight.js is registered from `lib/core` with an explicit language list rather than the
+ * default bundle: the full bundle pulls in ~190 grammars (~1 MB) for the handful an assistant
+ * actually emits here. Adding a language is one import + one registerLanguage line.
+ *
+ * `registerLanguage` also registers each grammar's own aliases, so ```cs, ```ts, ```sh, ```yml
+ * and ```html all resolve without being listed.
+ */
+const LANGUAGES = {
+  bash,
+  csharp,
+  css,
+  diff,
+  go,
+  ini,
+  java,
+  javascript,
+  json,
+  markdown,
+  plaintext,
+  powershell,
+  python,
+  rust,
+  sql,
+  typescript,
+  xml,
+  yaml,
+};
+
+for (const [name, definition] of Object.entries(LANGUAGES)) {
+  hljs.registerLanguage(name, definition);
+}
+
 marked.use({
   breaks: true, // Enable GFM line breaks
   gfm: true     // Enable GFM
 });
+
+/*
+ * marked dropped its built-in `highlight` option; `marked-highlight` is the supported
+ * replacement. The `hljs ` prefix on the class is what the highlight.js theme selects on --
+ * without it the emitted token spans are styled but the block itself is not.
+ */
+marked.use(
+  markedHighlight({
+    langPrefix: 'hljs language-',
+    emptyLangClass: 'hljs',
+    highlight(code: string, lang: string): string {
+      // An unknown or absent language must degrade to escaped plain text, never throw:
+      // a model can label a fence with anything at all (```mermaid, ```text, ```).
+      const language = lang && hljs.getLanguage(lang) ? lang : 'plaintext';
+      return hljs.highlight(code, { language, ignoreIllegals: true }).value;
+    },
+  })
+);
 
 /**
  * Parse markdown text to HTML
--- a/samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
+++ b/samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
@@ -15,7 +15,16 @@
 async (page) => {
   const BASE = 'http://127.0.0.1:5000';
   const PROVIDER_ID = 'test-anthropic';
-  const OUT_DIR = 'B:/sources/LmDotnetTools/.worktrees/WT1/.logs';
+  // Output dir must follow the checkout that RUNS the script, never a hardcoded clone.
+  //
+  // Observed in the Playwright MCP runner: `process` is NOT defined there (no `process`, no
+  // `Buffer`, no `require`), so `process.cwd()` / `process.env` would throw a ReferenceError --
+  // hence the `typeof` guard. What DOES work, and is the route this script relies on, is a
+  // relative path: Playwright resolves `path` against the server process's cwd, which is the
+  // repo/worktree root (verified: '.logs/_probe/rel.png' landed in this worktree's .logs).
+  // The env override is for other runners that do expose `process`.
+  const OUT_DIR =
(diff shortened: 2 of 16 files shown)


samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
```diff
@@ -1,11 +1,82 @@
...
+ */
+marked.use(
+  markedHighlight({
+    langPrefix: 'hljs language-',
+    emptyLangClass: 'hljs',
+    highlight(code: string, lang: string): string {
+      // An unknown or absent language must degrade to escaped plain text, never throw:
+      // a model can label a fence with anything at all (```mermaid, ```text, ```).
+      const language = lang && hljs.getLanguage(lang) ? lang : 'plaintext';
+      return hljs.highlight(code, { language, ignoreIllegals: true }).value;
```
gautam-achieveai: [Gautam's review bot] **[P2] Avoid re-highlighting the complete response on every streamed chunk**

TextMessage's computed parser reruns whenever the accumulated message text changes, so this synchronous callback highlights the entire growing fenced block on the main thread for every delta. With the locked dependencies, 500 incremental parses of a 10 KB JavaScript fence took 2.02 s versus 8.5 ms before highlighting; a 50 KB fence took 8.90 s versus 37.8 ms (about 18 ms per update, before DOM replacement). Long code-heavy answers will therefore exceed the frame budget and visibly stall streaming. Please render fences as escaped/plain code while the message is active and highlight once when the final message arrives; if live highlighting is required, move it off the main thread rather than repeatedly lexing the full buffer.

samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
```diff
@@ -0,0 +1,360 @@
...
+// Run with:
+//   browser_run_code_unsafe({ filename: "samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs" })
+//
+// Prompt source: PromptExamples.md -> "Markdown Rendering Showcase". MARKDOWN below must stay
+// identical to the copy in gen-md-showcase-prompt.mjs (this file has to be a single self-contained
+// function expression, so it cannot import it).
+async (page) => {
+  const BASE = 'http://127.0.0.1:5000';
+  const PROVIDER_ID = 'test-anthropic';
+  const OUT_DIR = 'B:/sources/LmDotnetTools/.worktrees/WT1/.logs';
```
gautam-achieveai: [Gautam's review bot] **[P2] Keep the audit output inside the current workspace**

This absolute path points at the author's WT1 clone rather than the checkout running the script. On a Windows host without that exact B: drive/layout, page.screenshot() throws and the catch block reports the whole 23-check audit as failed after the DOM assertions; in another checkout it can silently write evidence into the wrong clone. The companion markdown-showcase-screenshots.mjs repeats the same path at line 15. Please use a runner-relative .logs path or accept the output directory from the invocation.

gautam-achieveai (commented):
[Gautam's review bot] Deep review found two introduced issues that should be addressed before merging.

samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
gautam-achieveai: Fixed in 70744a82.

`parseMarkdown(text, { highlight?: boolean })` now backs onto **two explicit `Marked` instances** — one with `markedHighlight` over hljs, one whose highlight callback only HTML-escapes — rather than a mutable module-level flag, so there's no hidden state a concurrent parse could read. Both share `{ breaks: true, gfm: true }` and both emit `hljs language-x`, so finalizing a message only paints token colours in: no class change, no restyle, no layout shift. The plain path escapes `& < > " '` itself, since `marked-highlight` treats the callback's return value as HTML and the output goes into `v-html`.

`TextMessage` gained an `isComplete` prop (defaults `true`). Deliberately **not** `isStreaming` — that one also renders the blinking `|`, which would land in `[data-testid="assistant-text"]` textContent and break browser-E2E assertions. Cursor behaviour is untouched.

`MessageList` computes the id of the last `assistant-message` in `splitGroups.current` while `isLoading`, and passes `:is-complete="item.id !== streamingItemId"` on the **current** section only. Scoped to that one bubble on purpose: blanket-disabling for the active group would turn already-finished code blocks monochrome mid-run.

Coverage — new `src/__tests__/utils/markdown.test.ts` (9 cases: token spans present/absent, escaping on both paths including no-language fences and an unknown language, identical `<code>` class list), 4 new `TextMessage` cases, and **5 new `MessageList` cases that assert the wiring, not just the helper** — per-bubble `isComplete` values (last-of-active `false`, earlier sibling and history `true`, all `true` once `isLoading` flips) plus rendered token-span counts before/after the run ends, which exercises `MessageList → TextMessage → parseMarkdown` end to end. RED-checked: removing the `:is-complete` binding fails 3 of them.

Suite: 858/858 across 67 files (was 840/840 across 66). `npm run build` (vue-tsc + vite) clean. `markdown-render-audit.mjs` re-run against the running app returns `{"pass":true,"failures":[]}` — `code-block-is-syntax-highlighted` still passes with 7 spans / 2 distinct colours, i.e. the finalized message is highlighted as before.

samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
gautam-achieveai: Fixed in 70744a82, in **both** scripts.

```js
const OUT_DIR =
  (typeof process !== 'undefined' && process.env && process.env.LMSTREAMING_OUT_DIR) || '.logs';
```
(the showcase script appends `/markdown-shots`).

One correction to the obvious fix worth recording: **`process` does not exist in the Playwright MCP script runner at all** — I probed it, and `process`, `Buffer` and `require` are all `undefined` there, so a bare `process.cwd()` would have thrown a `ReferenceError` and traded one breakage for another. Hence the `typeof` guard, and hence the *relative* path rather than a `process.cwd()`-derived absolute one: Playwright resolves a relative `path` against its server process's cwd, which is the checkout root. Verified directly — a probe screenshot at `.logs/_probe/rel.png` landed in this worktree's `.logs`, and the real audit run wrote `markdown-after-viewport.png` / `markdown-after-bubble.png` there. The env var stays as a documented override for runners that do expose `process`.

Separately, in `markdown-render-audit.mjs` only, the screenshot call is now in its own `try/catch` and its outcome is reported as `notes.screenshots` instead of a pass/fail step — that is exactly the failure mode you described, where a throwing screenshot escaped to the outer catch and reported the whole DOM audit as failed. Confirmed the catch is load-bearing: `page.screenshot({ path: 'Z:/no-such-drive/x.png' })` rejects with `ENOENT`, and it is now contained.

`markdown-showcase-screenshots.mjs` keeps its hard assertion that every PNG's IHDR width equals `viewport.width * DPR` — there the screenshot *is* the deliverable, and that assertion is what caught the silently-discarded device-scale override.

Audit re-run after the change: `{"pass":true,"failures":[]}`, 26 recorded steps, all green.

samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
gautam-achieveai: [Gautam's review bot] **[P2] This still re-highlights assistant-only sub-agent streams**

The main-chat case is fixed, but `streamingItemId` only searches `splitGroups.current`. A focused child transcript is a second `MessageList` consumer: `useSubAgentPanel` builds its live `TextUpdate` entries as assistant-only display items, and `SubAgentTranscript` passes `isStreaming=true`. With no user group, `splitGroups.current` is empty, `streamingItemId` stays `null`, and the growing child code fence defaults to `isComplete=true`, so highlight.js still reruns on every delta there. Please identify the live assistant item independently of the user/current split and add an assistant-only streaming transcript regression; otherwise the original frame-stall remains for code-heavy sub-agents.

samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
gautam-achieveai: [Gautam's review bot] **[P3] Report the bubble screenshot failure instead of claiming both files exist**

The portable output-directory fix is correct, but the new outcome note is inaccurate when only the second capture fails. Line 366 swallows the bubble locator screenshot rejection with `.catch(() => {})`, then `notes.screenshots` lists both paths as captured. This can send a reviewer looking for an artifact that was never written. Track the two captures separately and list only successful files (or include the individual non-fatal error).

samples/LmStreaming.Sample/ClientApp/src/utils/markdown.ts
gautam-achieveai: Fixed in b56d7de8 — and you were right that the plan itself was wrong, not just the implementation.

`streamingItemId` now scans the flat `displayItems` from the end instead of `splitGroups.current`, so it is consumer-agnostic. I did **not** widen it to all of `messageGroups`.

**On your explicit pill question — a trailing `pill` CAN follow a growing text bubble, and when it does the text is finished.** `buildDisplayItems` buffers reasoning/tool messages and `flushPill()` only emits the pill once something *follows* them (or at end of list). So a `pill` at the tail means the assistant already moved off the text and onto a tool call. Skipping past it, as the suggested rule did, would mark a *finished* bubble incomplete for the entire duration of the tool call — a visible monochrome regression. So the scan **stops** on `pill`, and skips only `notification` (out-of-band; can land mid-stream and says nothing about the text). Everything unrecognised also stops: losing the optimization is cheap, a monochrome flicker is not. The per-type reasoning is written into the comment on `streamingItemId`.

`:is-complete` is now bound in the `history` branch too, not just the active one — an assistant-only transcript has no user group, so *every* group lands in `history`. It cannot mis-fire for the main chat: an assistant item only reaches `history` when a later user message exists, and the scan stops at that user message first.

**Proof:**
- `MessageList` assistant-only + `isLoading: true` → last bubble `isComplete === false`, 0 hljs token spans.
- Monochrome guard `[user1, assistant1, user2]` + `isLoading: true` → assistant1 stays `isComplete === true` and *is* highlighted.
- Trailing-pill case → bubble stays complete and highlighted; `[user, a1, pill, a2]` → a2 incomplete, a1 not.
- Trailing-notification case → skipped, growing bubble still incomplete.
- **`SubAgentTranscript` end-to-end with the REAL `MessageList`** (not the `MessageListStub` every other test in that file uses) — this is the shape of false confidence you called out, so it is now covered directly.

**RED check** (restore the previous `MessageList.vue`, keep the new tests): 3 fail — the assistant-only case, the trailing-pill case, and the `SubAgentTranscript` end-to-end case. The `[user1, assistant1, user2]` guard passes in both states; it is a regression guard, not RED-provable, and I am calling that out rather than implying otherwise.

Client tests **865/865 across 67 files** (baseline 858/67), `npm run build` clean, audit re-run green with 26 steps.

samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs
gautam-achieveai: Fixed in b56d7de8. The two captures are now tracked independently:

```js
const captures = [ ['markdown-after-viewport.png', () => page.screenshot(...)],
                   ['markdown-after-bubble.png',   () => tid('assistant-text').last().screenshot(...)] ];
const captured = []; const captureErrors = {};
for (const [name, take] of captures) {
  try { await take(); captured.push(`${OUT_DIR}/${name}`); }
  catch (shotErr) { captureErrors[name] = String((shotErr && shotErr.message) || shotErr); }
}
notes.screenshots = { captured, failed: captureErrors };
```

`captured` lists only files actually written; `failed` carries the per-capture error. Screenshot failure stays non-fatal to the DOM assertions.

**Verified by forcing it, not by reasoning about it.** I pointed *only* the bubble capture at an unwritable path (`Z:/no-such-drive/...`) and re-ran the audit:

```json
{"htmlSanitizer":"none (marked output goes straight into v-html)",
 "screenshots":{"captured":[".logs/markdown-after-viewport.png"],
                "failed":{"markdown-after-bubble.png":"locator.screenshot: ENOENT: no such file or directory, open 'Z:\no-such-drive\markdown-after-bubble.png'"}}}
```

`pass: true`, `failures: []`, **26 steps** — so exactly one file is claimed, the other is named with its error, and the DOM assertions are unaffected. Reverted and re-ran clean: both files in `captured`, `failed: {}`.

One correction to the finding text: the audit records **26** steps, not 23. The 23 figure in the PR body is stale from an earlier revision.

gautam-achieveai:
[Revobot]

## Review: PR #268

### [P2] Redact per-capture exception details

**File:** `samples/LmStreaming.Sample/playwright-scripts/markdown-render-audit.mjs:379`

The new per-capture reporting stores the raw screenshot exception text:

```js
captureErrors[name] = String((shotErr && shotErr.message) || shotErr);
```

Playwright and filesystem errors can include absolute workspace paths, URLs, or other environment-derived details. Since this audit returns diagnostic output that may be persisted or posted, exposing the raw exception reintroduces a disclosure risk.

Record only a stable error name/type or a boolean failure indicator, and keep the raw exception text out of the returned notes.

### Verification

- Focused Vitest tests could not run because `vitest` is not installed in the checkout.
- Prior threads concerning output paths, per-capture success reporting, and sub-agent streaming highlighting were not repeated; those changes appear addressed in the current head.

**Verdict:** Request changes.

🔎 Full review conversation: http://localhost:5051/?threadId=thread-5936d3d6af9f431d8f6b1e4177c35c33&focus=1

